PR #3, opened 2020-02-28
Setup initial folder structure and build.

rghetia:
Makefile/tools are mostly copied from opentelemetry-go with some minor modifications.


CONTRIBUTING.md
```diff
@@ -0,0 +1,125 @@
...
+
+Any Collaborator/Maintainer can merge the PR once it is **ready to
+merge**.
+
+## Style Guide
+
+* Make sure to run `make precommit` - this will find and fix the code
+  formatting.
+
+## Adding a new Contrib package
```
paivagustavo: I think it is worth to define that every new contrib package should reside on its own module?

rghetia: done.

PR #9, opened 2020-04-10
Runtime metrics plugin

chrisleavoy:


plugins/metrics/runtime/runtime.go
```diff
@@ -0,0 +1,89 @@
...
+var (
+	initialisation time.Time
+)
+
+func init() {
+	initialisation = time.Now()
+}
+
+// various runtime metrics have a certain overhead, and as such, we only measure them on a defined interval
+func Runtime(ctx context.Context, meter metric.Meter, interval time.Duration) context.CancelFunc {
```
jmacd, line 23: I suggest naming this `Start()`.

plugins/metrics/runtime/runtime.go
```diff
@@ -0,0 +1,89 @@
...
+	ctx, cancel := context.WithCancel(ctx)
+
+	ctxTimeout, _ := context.WithTimeout(ctx, time.Second)
```
jmacd, line 26: This fragment of code is repeated twice--maybe add the call to `context.WithTimeout` into `metrics.Measure()`? Also, `time.Second` is used twice instead of `interval`, but ideally we'll be able to use Observers for all these metrics, leaving interval up to the metrics SDK.

plugins/metrics/runtime/runtime.go
```diff
@@ -0,0 +1,89 @@
...
+
+			if err != nil {
+				break
+			}
+
+			ctxTimeout, _ := context.WithTimeout(ctx, time.Second)
+			// store the previous measurement for reporting deltas
+			p = m
+			// we assume here we are replacing m and thus do ont require a concurrency lock
+			m, _ = metrics.Measure(ctxTimeout)
```
jmacd, line 47: This block of code is why we're working on batch observers. Until we get something done there, this approach is OK.  Eventually we should be able to use a single batch observer callback for all these instruments.

plugins/metrics/runtime/runtime.go
```diff
@@ -0,0 +1,89 @@
...
+
+			metric.Must(meter).NewFloat64Counter("runtime.go.cpu.user").Bind().Add(ctx,
```
jmacd, line 49: Creating new metric instruments is a heavyweight activity. Eventually these should be Observers, but for now the call to `meter.NewXXX` should be lifted outside of the `func()` body.

jmacd: The call to `Bind()` is can be simply removed. If you lifted the bound instrument creation outside the function body here, it would be a minor optimization (that would go away when we convert these instruments to Observers).  The instruments all support direct calls, so just remove `.Bind()` and it will compile with the same effect.

plugins/metrics/runtime/runtime.go
```diff
@@ -0,0 +1,89 @@
...
+				m.ProcessCPU.User-p.ProcessCPU.User,
+			)
+
+			metric.Must(meter).NewFloat64Counter("runtime.go.cpu.sys").Bind().Add(ctx,
```
jmacd, line 53: OTEPs 93/96 introduce a new metric instrument that will let us Observe a cumulative and not need to compute a delta.  This is fine since we don't have either the batch-observer or cumulative instrument here, yet. This will become a CumulativeObserver, deltas will be computed in the SDK because the exporter (in your case) will demand it.

plugins/metrics/runtime/runtime.go
```diff
@@ -0,0 +1,89 @@
...
+		for {
+			err := limit.Wait(ctx)
+
+			if err != nil {
+				break
+			}
+
+			ctxTimeout, _ := context.WithTimeout(ctx, time.Second)
+			// store the previous measurement for reporting deltas
+			p = m
```
jmacd, line 45: When we introduce the CumulativeObserver, storing the previous measurement won't be needed.

plugins/metrics/runtime/runtime.go
```diff
@@ -0,0 +1,89 @@
...
+				m.ProcessCPU.System-p.ProcessCPU.System,
+			)
+
+			metric.Must(meter).NewInt64Counter("runtime.go.gc.count",
+				metric.WithDescription("counter of completed GC cycles")).Bind().Add(ctx,
+				int64(m.Runtime.NumGC-p.Runtime.NumGC),
+			)
+		}
+	}()
+
```
jmacd, line 63: There's probably a potential race condition reading and writing `m`, which is not protected by any lock.

plugins/metrics/runtime/go.mod
```diff
@@ -0,0 +1,11 @@
+module github.com/open-telemetry/opentelemetry-go-contrib/plugins/metrics/runtime
+
+go 1.13
+
+require (
+	github.com/lightstep/lightstep-tracer-common/golang/gogo v0.0.0-20200310182322-adf4263e074b
+	github.com/lightstep/lightstep-tracer-go v0.20.0 // indirect
```
MrAlias, line 7: Why are these dependencies needed?

plugins/metrics/runtime/runtime.go
```diff
@@ -0,0 +1,89 @@
...
+var (
+	initialisation time.Time
+)
+
+func init() {
+	initialisation = time.Now()
+}
+
+// various runtime metrics have a certain overhead, and as such, we only measure them on a defined interval
```
MrAlias, line 22: Comment needs to be of the form `Runtime ...`

plugins/metrics/runtime/runtime.go
```diff
@@ -0,0 +1,89 @@
...
+func Runtime(ctx context.Context, meter metric.Meter, interval time.Duration) context.CancelFunc {
+	ctx, cancel := context.WithCancel(ctx)
+
+	ctxTimeout, _ := context.WithTimeout(ctx, time.Second)
+	var m, p metrics.Metrics
+	m, _ = metrics.Measure(ctxTimeout)
+
+	if interval == 0 {
+		interval = time.Second
```
MrAlias, line 31: Is this ever used?

plugins/metrics/runtime/runtime.go
MrAlias: Also, the cancel function should be called at a minimum on the function return. Meaning `defer`ing a call after it is returned would be appropriate.

plugins/metrics/runtime/runtime.go
```diff
@@ -0,0 +1,89 @@
...
+
+	"go.opentelemetry.io/otel/api/metric"
+	"go.opentelemetry.io/otel/api/unit"
+	"golang.org/x/time/rate"
+
+	"github.com/open-telemetry/opentelemetry-go-contrib/plugins/metrics/runtime/internal/metrics"
+)
+
+var (
+	initialisation time.Time
```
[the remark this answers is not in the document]
chrisleavoy: I wanted to report the true process start time. I'll look into something more idiomatic than using a global and init() to approximate process start time, versus when the runtime.Start() occurred.

Thanks for the review, I'll poke at this some more.

plugins/metrics/runtime/runtime.go
zlozano: > but ideally we'll be able to use Observers for all these metrics, leaving interval up to the metrics SDK.

@jmacd I am not super familiar with how the observer interval is implemented. If it's tied to the controller push interval, then I think we'd definitely want to decouple the two here. Assuming folks have one meter configured for their service, I think it would be valid for them to set a relatively frequent controller interval (e.g. 1 second), and want to limit the amount of times `runtime.ReadMemStats` to a less frequent interval (e.g. 10 seconds) is called since it is a STW process.

plugins/metrics/runtime/internal/procstats/procstats.go
```diff
@@ -0,0 +1,243 @@
...
+import (
+	"math"
+	"runtime"
+	"time"
+
+	// "github.com/segmentio/stats/v4"
+)
+
+func init() {
+	stats.Buckets.Set("go.memstats:gc_pause.seconds",
```
zlozano, line 12: Does this compile? It looks like the import is commented out?

plugins/metrics/runtime/internal/procstats/procstats.go
```diff
@@ -0,0 +1,243 @@
+package procstats
```
zlozano, line 1: It doesn't look like this internal package is used anywhere in here.

chrisleavoy: @zlozano, nah it isn’t used. Just example code (From segment.io stats) I copied for reference. Will remove. Didn’t notice as unused vendor code isn’t copied in by go mod

jmacd:
FYI this will help: https://github.com/open-telemetry/opentelemetry-go/pull/634

plugins/metrics/runtime/runtime.go
jmacd: Correction: This will become a SumObserver. But since this PR is ready now, and the new instruments are not through specification and haven't been implemented yet, this isfine.

plugins/metrics/runtime/runtime.go
```diff
@@ -0,0 +1,286 @@
...
+	"context"
+	goruntime "runtime"
+	"sync"
+	"time"
+
+	"go.opentelemetry.io/otel/api/metric"
+	"go.opentelemetry.io/otel/api/unit"
+)
+
+type Runtime struct {
```
jmacd, line 13: Please comment:
> Runtime reports the work-in-progress conventional runtime metrics specified by OpenTelemetry.

plugins/metrics/runtime/runtime.go
```diff
@@ -0,0 +1,286 @@
...
+		goLookups  metric.Int64Counter
+		goGcCount  metric.Int64Counter
+		gcPauseNs  metric.Int64Measure
+	}
+
+	memStats    goruntime.MemStats
+	numCgoCalls int64
+}
+
+func NewRuntime(meter metric.Meter, interval time.Duration) *Runtime {
```
jmacd, line 30: Let's name this "New", since it's going to be called `runtime.New()` there's less stutter. I noticed a concern. I think the ideal case here is that all the instruments are asynchronous ("observers"), meaning the SDK controls the interval. It won't be the runtime's decision how often to snapshot instruments, so at some point the `interval` parameter might go away. Is that ok with you?

plugins/metrics/runtime/runtime.go
```diff
@@ -0,0 +1,286 @@
...
+	r := &Runtime{
+		meter:    meter,
+		interval: interval,
+		done:     make(chan bool),
+	}
+
+	return r
+}
+
+func (r *Runtime) Start() error {
```
jmacd, line 40: Add a comment.

plugins/metrics/runtime/runtime.go
```diff
@@ -0,0 +1,286 @@
...
+	if err != nil {
+		return err
+	}
+
+	go r.ticker()
+
+	return nil
+}
+
+func (r *Runtime) Stop() {
```
jmacd, line 51: Add a comment.

plugins/metrics/runtime/runtime.go
```diff
@@ -0,0 +1,286 @@
+package runtime // import "go.opentelemetry.io/contrib/plugins/metrics/runtime"
```
jmacd, line 1: A `doc.go` comment file would be helpful listing all the metrics produced here. It's fine to make it a TODO, since we haven't actually created a conventional standard which it should be able to link to. How about:

> package runtime implements the work-in-progress conventional runtime metrics specified by OpenTelemetry. The metrics produced are:
> [list of metrics and brief descriptions]

plugins/metrics/runtime/runtime.go
chrisleavoy: Help me understand this. Once sdk is updated for new types and batch reporting is in, using async types for everything would be nice and clean. No mismatched ticker and the cache of the struct. 

However, since runtime.memstat metrics for go involve a stop-the-world action. It is not OK to set it arbitrarily frequent interval. So I would want either we allow an interval for how often to poll, or two metric.meter’s with different intervals.

plugins/metrics/runtime/runtime.go
chrisleavoy: Will do.

krnowak:
Does this package need to be in `plugins/metrics/runtime` directory? Can't it be `plugins/runtime`?

chrisleavoy:
Can anyone share what is required to get circleci working, if needed for this PR?

Aneurysm9:
The CircleCI config had been in the wrong place.  I merged master into your branch which should fix that.

plugins/metrics/runtime/runtime.go
MrAlias: > I would want either we allow an interval for how often to poll...

This interval is set when a Controller is created. Asynchronous instruments observe on this cadence.

I agree that this should be linked to that interval. If it is specified here it will be the instrumenter defining this interval instead of the operator or end-user of the instrumented code.

MrAlias: This function will also need a docstring.

jmacd: One of the considerations behind the asynchronous instrument API--the callback registration mechanism--is to support instruments with variable rate intervals.  So, this PR has done a fine job of demonstrating all the advanced things.

The approach here:
https://github.com/open-telemetry/opentelemetry-go/pull/634
uses a somewhat heavyweight API, and we're considering whether the semantics get confusing when we relax it a bit, please have a look.

jmacd: > we allow an interval for how often to poll

This is a bit complex, but understandable. I wonder if we should allow this code to set a minimum interval between polling, as a hard-stop on the rate of stop-the-world behavior. If an observer callback is called too frequently it can observe a cached result. 

Another option would be for the callback to return some kind of "not now" error code to say that the observer is not ready and would like to defer _without making an observation_.

plugins/runtime/runtime.go
```diff
@@ -0,0 +1,313 @@
...
+
+	r.metrics.goCgoCalls, err = r.meter.NewInt64Counter("runtime.go.cgo.calls",
+		metric.WithDescription("Number of cgo calls made by the current process"))
+	if err != nil {
+		return err
+	}
+
+	// poll now so that the first tick has a full delta
+	r.numCgoCalls = goruntime.NumCgoCall()
+	goruntime.ReadMemStats(&r.memStats)
```
chrisleavoy, line 130: @quentinmit, reading the meeting nodes of the metrics sig, you mentioned that this ReadMemStats function (GC) "is no longer expensive" in Go?   Its still labeled as stop-the-world however. I'm curious to hear from folks who have more expertise than me in this area.

jmacd: I found this: https://go-review.googlesource.com/c/go/+/34937/
(Not sure..)

quentinmit: Yeah, that's the change I was thinking of. Also, I believe [ReadGCStats](https://golang.org/pkg/runtime/debug/#ReadGCStats) does not STW at all.

PR #10, opened 2020-04-14
Restore Dogstatsd exporter

jmacd:
This imports the code that was removed in https://github.com/open-telemetry/opentelemetry-go/pull/591 without significant changes.

There was a report that the default MaxPacketSize is too large, which I was able to reproduce on my development machine. I've set it to 9216, which is Jumbo ethernet size.  Another good default would be 1500. (This is configurable.)


jmacd:
@chrisleavoy

exporters/metric/dogstatsd/internal/statsd/conn.go
```diff
@@ -0,0 +1,291 @@
...
+		if err := e.formatMetric(rec, buf); err != nil {
+			return err
+		}
+
+		if buf.Len() < e.config.MaxPacketSize {
+			return nil
+		}
+		if before == 0 {
+			// A single metric >= packet size
+			if err := e.send(buf.Bytes()); err != nil && sendErr == nil {
```
chrisleavoy, line 181: I'm still running into the same issue with this line of code....

my console is full of ```Metrics SDK error: write udp 127.0.0.1:62433->127.0.0.1:8125: write: message too long```

a breakpoint on this line shows buf has a len of 10791, and is essentially:

```
example.histogram:0|h|#climb:every_200ms,kube_pod:Chriss-MacBook-Pro.local,service:o11y-sensor-calibration
example.histogram:1|h|#climb:every_200ms,kube_pod:Chriss-MacBook-Pro.local,service:o11y-sensor-calibration
example.histogram:3|h|#climb:every_200ms,kube_pod:Chriss-MacBook-Pro.local,service:o11y-sensor-calibration
...
example.histogram:97|h|#climb:every_200ms,kube_pod:Chriss-MacBook-Pro.local,service:o11y-sensor-calibration
example.histogram:98|h|#climb:every_200ms,kube_pod:Chriss-MacBook-Pro.local,service:o11y-sensor-calibration
```

I tried tracing this code before and I think an assumption is off.

exporters/metric/dogstatsd/internal/statsd/conn.go
```diff
@@ -0,0 +1,291 @@
...
+	if sendErr != nil {
+		return sendErr
+	}
+	return aggErr
+}
+
+// send writes a complete buffer to the writer as a blocking call.
+func (e *Exporter) send(buf []byte) error {
+	for len(buf) != 0 {
+		n, err := e.writer.Write(buf)
```
chrisleavoy, line 212: specifically I think its this... writer.Write is assumed to accept a buf larger than MTU and write only first MTU bytes... but it doesn't.  it returns an error instead, writing out nothing. I was hacking the code locally before I opted to just wrap the datadog client instead.  I'd rather use this code instead but I think we'll need packet splitting first.

jmacd: I see it. The logic is trying to split packets between each metric record, but the Array-of-points record doesn't support splitting between values.

jmacd: (Fixed.)

jmacd:
I fixed the known problem, but there is still a defect where if the MaxPacketSize is too large, data won't be sent.  I tested on an OS X machine that the call to `Write()` returns 0 bytes written when the packet is too large. So the only way to avoid MTU-too-small errors is to dynamically adjust the packet size when we see the "message too long" error. I'll leave that up for discussion and/or another PR.

zlozano:
Does the default packet size take into account the UDP header size + max IP header size?

zlozano:
Also, it seems like the safer default would be the more conservative option of 1500 (minus room for header bytes). This seems to be the more widely accepted default configuration [1]

[1] https://docs.aws.amazon.com/AWSEC2/latest/UserGuide/network_mtu.html

exporters/metric/dogstatsd/internal/statsd/conn.go
```diff
@@ -0,0 +1,310 @@
...
+const (
+	formatCounter   = "c"
+	formatHistogram = "h"
+	formatGauge     = "g"
+	formatTiming    = "ms"
+
+	// MaxPacketSize defaults to Ethernet Jumbo frame size.  If
+	// this is too large, you will see "write: message too long"
+	// errors.
+	MaxPacketSize = 9216
```
kconwayinvision, line 83: Jumbo frames don't really exist in a any consistent form. AWS, for example, has a jumbo frame features that caps at 9001 bytes per packet. Most cloud providers stick to the 1500 MTU standard. I'd encourage a default size that is less than 1500 in order to account for both the required and optional vendor headers in the packets.

For reference, the official datadog client for Go uses 1432 (https://github.com/DataDog/datadog-go/blob/f137255bc8cca1abb54c897418b6bb752acd5839/statsd/statsd.go#L37).

jmacd:
(I had a default of 64kB because that used to work at my last employer.)

I take it 1432 is the value we want? Done.

exporters/metric/dogstatsd/internal/statsd/conn.go
jmacd: We'll use this value. Thanks.

PR #12, opened 2020-04-14
Support push options for dogstatsd exporter

jmacd:


jmacd:
@chrisleavoy

jmacd:
Maybe this is a bit lazy. We could create new options in the dogstatsd package, instead of referring to the push.Option type.

PR #15, opened 2020-04-16
Initial gin-gonic instrumentation with otel-go

krnowak:
This is some initial code that does the instrumentation. Still need to fix the tests, add docs and a Dockerfile with an example app. So it's work in progress.

There is the `common.go` file that is mostly about attributes for the span. I suppose it will end up being elsewhere in the long run.

PR #20, opened 2020-04-24
Add Macaron plugin

MikeGoldsmith:
- adds trace plugin for https://github.com/go-macaron/macaron routing
- has a TODO to get route template for span name (eg `/users/:id` instead of `/users/123`)

I'm working on a bug in docker example, running examples/server.go works but wget from one container to another with server.go fails with lost context.

Copies internal mock span, tracer, etc from #15 & #19


MikeGoldsmith:
Fixed docker-compose bug.

MikeGoldsmith:
Is someone available to review this PR please? All feedback welcome.

MrAlias:
> Is someone available to review this PR please? All feedback welcome.

Sorry about the delay. I was hoping to have #19 merged first to avoid having to review duplicate content. I'll see if I can't motivate more reviews on that to then prioritize this.

plugins/macaron/macaron.go
```diff
@@ -0,0 +1,79 @@
...
+
+		ctx := otelpropagation.ExtractHTTP(savedCtx, cfg.Propagators, c.Req.Header)
+		opts := []oteltrace.StartOption{
+			oteltrace.WithAttributes(trace.NetAttributesFromHTTPRequest("tcp", c.Req.Request)...),
+			oteltrace.WithAttributes(trace.EndUserAttributesFromHTTPRequest(c.Req.Request)...),
+			oteltrace.WithAttributes(trace.HTTPServerAttributesFromHTTPRequest(service, c.Req.RequestURI, c.Req.Request)...),
+			oteltrace.WithSpanKind(oteltrace.SpanKindServer),
+		}
+		// TODO: span name should be router template not the actual request path, eg /user/:id vs /user/123
+		spanName := c.Req.RequestURI
```
krnowak, line 60: There indeed seems to be no way to get the matched route. Boo. I think we should give a different name for the span then - the request URI is going to have a too high cardinality because of the parameters in the uri.

I see two ways out here:

1. convince macaron authors to provide the functionality for getting the matched route
    - that's a long way to resolve the issue - one thing is providing functionality, another thing is for the project to bump the macaron dependency to the new version
1. provide span name generator function as an option to `Middleware`, so `Middleware` could use it to get the span name.

The last option could be something like:

```go
// in config.go
type NameGenerator func(c *macaron.Context) string

func WithNameGenerator(g NameGenerator) Option {…}
```

So the example server could call `Middleware` like so:

```go
nameGenerator := func(c *macaron.Context) string {
	uri := c.Req.RequestURI
	if strings.HasPrefix(c.Req.RequestURI, "/users/") {
		return "/users/:id"
	}
	return "unknown route"
}
m.Use(macarontrace.Middleware("my-server", macarontrace.WithNameGenerator(nameGenerator)))
```

It's lame, but I guess you pay for what you use.

go.mod
```diff
@@ -1,3 +1,12 @@
 module go.opentelemetry.io/contrib
 
 go 1.14
+
+require go.opentelemetry.io/contrib/exporters/metric/dogstatsd v0.0.0-20200423201616-ec5be60591bf // indirect
```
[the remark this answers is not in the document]
MikeGoldsmith: Done :+1:

plugins/macaron/macaron.go
```diff
@@ -0,0 +1,79 @@
...
+		savedCtx := c.Req.Request.Context()
+		defer func() {
+			c.Req.Request = c.Req.Request.WithContext(savedCtx)
+		}()
+
+		ctx := otelpropagation.ExtractHTTP(savedCtx, cfg.Propagators, c.Req.Header)
+		opts := []oteltrace.StartOption{
+			oteltrace.WithAttributes(trace.NetAttributesFromHTTPRequest("tcp", c.Req.Request)...),
+			oteltrace.WithAttributes(trace.EndUserAttributesFromHTTPRequest(c.Req.Request)...),
+			oteltrace.WithAttributes(trace.HTTPServerAttributesFromHTTPRequest(service, c.Req.RequestURI, c.Req.Request)...),
```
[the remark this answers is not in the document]
MikeGoldsmith: Done 👍

MikeGoldsmith:
I've addressed some of the feedback & merged master into my  PR.

I'll take a look at @krnowak's suggestion of a span name generator next.

plugins/macaron/macaron.go
```diff
@@ -0,0 +1,79 @@
...
+	otelpropagation "go.opentelemetry.io/otel/api/propagation"
+	oteltrace "go.opentelemetry.io/otel/api/trace"
+)
+
+const (
+	tracerKey  = "otel-go-contrib-tracer"
+	tracerName = "go.opentelemetry.io/contrib/plugins/macaron"
+)
+
+// Middleware returns stuff
```
MrAlias, line 34:
```suggestion
// Middleware returns a macaron Handler to trace requests to the server.
```

plugins/macaron/macaron.go
jmacd: https://github.com/open-telemetry/opentelemetry-go-contrib/issues/37

jmacd:
@MikeGoldsmith This is looking great. I filed issue #37 so that we could merge it with a TODO. Would you mind updating for the 0.5.0 changes? Mostly it's that the "core" and "key" packages have been renamed/refactored. core-tracing things move into api/trace, and key/value stuff moves into api/kv. Thank you!

PR #22, opened 2020-04-24
add datadog metrics exporter

marwan-at-work:
Fixes https://github.com/marwan-at-work/otel-exporter-datadog/issues/3

This exporter uses the official DataDog client and is modeled after the OpenCensus exporter that they have written https://github.com/DataDog/opencensus-go-exporter-datadog


exporters/metric/datadog/datadog.go
```diff
@@ -0,0 +1,178 @@
...
+	return name
+}
+
+func (e *Exporter) Export(ctx context.Context, cs export.CheckpointSet) error {
+	return cs.ForEach(func(r export.Record) error {
+		agg := r.Aggregator()
+		name := e.sanitizeMetricName(r.Descriptor().LibraryName(), r.Descriptor().Name())
+		itr := r.Labels().Iter()
+		tags := append([]string{}, e.opts.Tags...)
+		for itr.Next() {
```
jmacd, line 73: This has me wondering whether the `label.Set.Encoded()` method could be improved for exporters that don't use strings.  If you could cache a `[]string` in the label set, it would be a win here, but at present we only support caching string encodings.

exporters/metric/datadog/datadog.go
```diff
@@ -0,0 +1,178 @@
+package datadog
```
jmacd, line 1: @lizthegrey can you comment on license and copyright requirements for packages in this repo?

jmacd:
FYI the OTel Resources will be coming through the export pipeline, soon. https://github.com/open-telemetry/opentelemetry-go/pull/659

exporters/metric/datadog/datadog.go
lizthegrey: Vendor code can keep its past licensing statement module-by-module, although we'd prefer it have future ongoing authorship attributed to the OpenTelemetry Contributors.

lizthegrey: Personally, I keep Honeycomb's exporter in its own repo so we don't have to wait for releases of -contrib.

marwan-at-work: Thank you both, just let me know if I need to add/change anything to this file or move the package to another repo 🙂 happy to do w/e!

marwan-at-work:
@jmacd thanks -- is there any documentation on Resources? What are they and how can this exporter use them? 

Thanks again for all the help!

exporters/metric/datadog/datadog.go
jmacd: DataDog is not here and hasn't responded in the past. I'm comfortable putting this here until DataDog would like us to remove it in favor of a repository of theirs.

exporters/metric/datadog/datadog.go
jmacd: (No action required.)

jmacd:
@lizthegrey @MrAlias I filed an issue in the DD repo: https://github.com/DataDog/datadog-go/issues/146

PR #25, opened 2020-04-27
Dogstatsd exporter resource support (for 0.4.3 release)

### LINKED ISSUE #24: Add resource support in the dogstatsd exporter
Like #23 but for dogstatsd exporter.

jmacd:
This updates the exporter interface for the new Resource argument, added in the base library's 0.4.3 release.  This removes the former workaround.

Resolves #24.

### DIFF
--- a/exporters/metric/dogstatsd/dogstatsd_test.go
+++ b/exporters/metric/dogstatsd/dogstatsd_test.go
@@ -31,10 +31,10 @@ import (
 )
 
 // TestDogstatsLabels that labels are formatted in the correct style,
-// whether or not the provided labels were encoded by a statsd label
-// encoder.
+// including Resources.
 func TestDogstatsLabels(t *testing.T) {
 	type testCase struct {
+		name      string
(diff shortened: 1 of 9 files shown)


exporters/metric/dogstatsd/dogstatsd_test.go
```diff
@@ -34,25 +34,62 @@ import (
...
+			resources: nil,
+			labels:    kvs(key.String("A", "B")),
+			expected:  "test.name:123|c|#A:B\n",
+		},
+		{
+			resources: kvs(key.String("R", "S")),
+			labels:    kvs(key.String("A", "B")),
+			expected:  "test.name:123|c|#R:S,A:B\n",
+		},
+		{
```
MikeGoldsmith, line 66: As we combine multiple test cases into a single test, I think a simple description for each test case would help here. There is only a subtle difference between the last two test cases.

jmacd: Done.

PR #26, opened 2020-04-30
Port mongodb driver instrumentation

rnburn:
Ported from Datadog's instrumentation.

Copies in key-value constants, though these should be reorganized to a location where they can be shared across instrumentations.


go.mongodb.org/mongo-driver/mongo/db.go
```diff
@@ -0,0 +1,52 @@
+// Unless explicitly stated otherwise all files in this repository are licensed
+// under the Apache License Version 2.0.
+// This product includes software developed at Datadog (https://www.datadoghq.com/).
+// Copyright 2016-2019 Datadog, Inc.
```
MrAlias, line 4: This needs to licensing needs to be resolved similarly to https://github.com/open-telemetry/opentelemetry-go-contrib/pull/15#pullrequestreview-398550926. The ask is to have DataDog sign off on this being solely licensed under `The OpenTelemetry Authors`.

go.mongodb.org/mongo-driver/mongo/mongo.go
```diff
@@ -0,0 +1,118 @@
...
+	m.Lock()
+	span, ok := m.spans[key]
+	if ok {
+		delete(m.spans, key)
+	}
+	m.Unlock()
+	if !ok {
+		return
+	}
+	span.SetAttributes(Error(err != nil))
```
krnowak, line 84: I think you shouldn't be setting this attribute at all if `err` is `nil`.

go.mongodb.org/mongo-driver/mongo/mongo_test.go
```diff
@@ -0,0 +1,72 @@
...
+	_, ok := os.LookupEnv("INTEGRATION")
+	if !ok {
+		fmt.Println("--- SKIP: to enable integration test, set the INTEGRATION environment variable")
+		os.Exit(0)
+	}
+	os.Exit(m.Run())
+}
+
+func Test(t *testing.T) {
+  mt := mocktracer.NewTracer("mongodb")
```
krnowak, line 35: Indentation is off. You could run `make` for the linter to do its work.

go.mongodb.org/mongo-driver/mongo/option.go
```diff
@@ -0,0 +1,49 @@
...
+func WithServiceName(name string) Option {
+	return func(cfg *config) {
+		cfg.serviceName = name
+	}
+}
+
+type config struct {
+	tracer      trace.Tracer
+	serviceName string
+}
```
krnowak, line 34: I think you could export the struct and its fields. It would be consistent with other plugins and there could be a possibility of writing your own `Option` like:

```go
func myOption(some stuff) Option {
	return func(cfg *Config) {
		cfg.ServiceName = some.FigureOutServiceName()
		cfg.Tracer = some.GetTracer()
	}
}
```

go.mongodb.org/mongo-driver/mongo/option.go
```diff
@@ -0,0 +1,49 @@
...
+
+func newConfig(opts ...Option) config {
+	var c config
+
+	defaultOpts := []Option{
+		WithTracer(trace.NoopTracer{}),
```
krnowak, line 40: I'd say that the default tracer shouldn't be a NoopTracer, but rather global.Tracer("go.opentelemetry.io/contrib/plugins/go.mongodb.org/mongo-driver")` or some such name.

go.mongodb.org/mongo-driver/mongo/option.go
```diff
@@ -0,0 +1,49 @@
...
+		WithServiceName("mongo"),
```
krnowak, line 41: I'm a bit on the fence with service name being an option. In case of web frameworks, the spec was saying (IIRC) that this is quite an important field, so I decided to have service name as a parameter to the function installing the instrumentation (`func Middleware(serviceName string, opts ...Option)`, so in mongodb case this could be `func NewMonitor(serviceName string, opts ...Option)`). What do you think?

go.mongodb.org/mongo-driver/mongo/mongo.go
```diff
@@ -0,0 +1,118 @@
+// Unless explicitly stated otherwise all files in this repository are licensed
+// under the Apache License Version 2.0.
+// This product includes software developed at Datadog (https://www.datadoghq.com/).
+// Copyright 2016-2019 Datadog, Inc.
+// Copyright 2020 The OpenTelemetry Authors
+
+// Package mongo provides functions to trace the mongodb/mongo-go-driver package (https://github.com/mongodb/mongo-go-driver).
```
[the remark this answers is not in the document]
krnowak: Also, the package documentation could be moved to `package.go`.

go.mongodb.org/mongo-driver/mongo/package.go
```diff
@@ -0,0 +1 @@
+package mongo // import "go.opentelemetry.io/contrib/go.mongodb.org/mongo-driver/mongo"
```
krnowak, line 1: Missing copyright. Also, this file could contain the package documentation, so the file wouldn't look superfluous. And maybe then you could rename it to `doc.go` (for consistency maybe with other plugins).

andrewhsu:
@rnburn looks like from the thread of conversation that lead to https://github.com/open-telemetry/opentelemetry-go-contrib/pull/15#issuecomment-628234917 indicates the copyright headers for datadog and the NOTICES file can be removed for this PR.

plugins/go.mongodb.org/mongo-driver/mongo_test.go
```diff
@@ -0,0 +1,80 @@
...
+	"go.mongodb.org/mongo-driver/bson"
+	"go.mongodb.org/mongo-driver/mongo"
+	"go.mongodb.org/mongo-driver/mongo/options"
+
+	"github.com/stretchr/testify/assert"
+)
+
+func TestMain(m *testing.M) {
+	_, ok := os.LookupEnv("INTEGRATION")
+	if !ok {
```
MrAlias, line 35: Create unit testing that doesn't need a running instance of Mongo to function.

MrAlias: https://github.com/open-telemetry/opentelemetry-go-contrib/issues/39

MrAlias:
@jmacd I think all the changes @krnowak requested have been addressed and this is ready to merge. It seems like I don't have the ability to dismiss his request for changes, so I think it's up to you if we merge this.

PR #32, opened 2020-05-09
WIP: Add instrumentation code for database/sql and database/sql/driver

krnowak:
This commit adds a dependency on a tool I wrote while working on this
instrumentation. The tools is wrappergen, it resides at
github.com/krnowak/wrappergen and it generates a lot of code. The code
is about having an exact wrapper of some interface. For example,
driver.Conn is a base interface which can be extended with 9 more
interfaces. This gives us 512 possible combinations of
implementations. And I wanted our tracing wrapper to implement the
same interfaces as the object it wraps. So instead of writing it by
hand, I wrote a generator.

Currently span names do not meet the requirements specified in
semantic conventions for database spans yet. The span names should be
SQL queries, likely without the values, but I'm not sure if it's
possible. In some cases the queries are separated from values, but in
some cases not. That's a thing to resolve before merging this commit.

Second thing is that this code needs tests, especially around figuring
out host, port, database name and so on from the data source name
(DSN) string. The tests should see if the tracing driver can figure
out the required attributes for various DSNs from different backends.


andrewhsu:
This one looks like it is generating quite a lot of code from a personal repo project code generator. Even though this is a WIP, I think this approach is a bit risky to continue pursuing and elect searching for other ways of getting the plugins in for database sql.

Issue #43, opened 2020-05-26
Metric test helpers needed


jmacd:
This repo has a copy of the trace testing helpers from the main repository. This repo does not have a copy of the metric testing helpers. We should add test helpers to support testing metrics in this repository, but first some code should be exposed from the main repository (e.g., AsyncInstrumentState).

matej-g:
Would take this toget with #385

MrAlias:
Closing, stale. A manual reader and the `metricdatatest` package are used to test expected metric exports.

PR #44, opened 2020-05-26
Update runtime package to use batch observer and new instruments

jmacd:
This updates the runtime plugin to the latest OTel release. It is not well tested, see issue #43.

The signature becomes `Start(meter, maxInterval)`. There is no `New` because there are no user-callable functions on this object.

See related issue: https://github.com/open-telemetry/opentelemetry-go/issues/767

CC: @chrisleavoy


plugins/runtime/example/main.go
```diff
@@ -32,11 +32,11 @@ func initMeter() *push.Controller {
 	pusher, err := metricstdout.NewExportPipeline(metricstdout.Config{
 		Quantiles:   []float64{0.5},
 		PrettyPrint: true,
-	}, 10*time.Second)
+	})
```
Aneurysm9, line 35:
```suggestion
	}, push.WithPeriod(10*time.Second)
```

Should this still specify a period?

chrisleavoy: I changed the default here because it was too long to wait for an example.  It made me think the example wasn't working because stdout was blank.

jmacd: The default has changed to 10s, so I removed the explicit parameter to keep an unnecessary detail out of the example.

PR #46, opened 2020-05-28
Upgrade Exporter to match new API

marwan-at-work:
This PR upgrades the datadog exporter and also uses the Resource:

 cc @jmacd I am combining both the Resource and the record's LabelSet(). I'm hoping that this the correct way to accumulate tags, but definitely let me know otherwise. 

Thanks!


exporters/metric/datadog/example_test.go
```diff
@@ -24,10 +24,10 @@ func ExampleExporter() {
 		panic(err)
 	}
 	defer exp.Close()
-	pusher := push.New(integrator, exp, time.Second*10)
+	pusher := push.New(integrator, exp, push.WithPeriod(time.Second*10))
```
jmacd, line 27: FWIW, the default is now 10s.

exporters/metric/datadog/datadog.go
```diff
@@ -66,12 +66,11 @@ func defaultFormatter(namespace, name string) string {
...
 }
 
-func (e *Exporter) Export(ctx context.Context, _ *resource.Resource, cs export.CheckpointSet) error {
-	// TODO: Use the Resource argument.
+func (e *Exporter) Export(ctx context.Context, cs export.CheckpointSet) error {
 	return cs.ForEach(func(r export.Record) error {
 		agg := r.Aggregator()
 		name := e.sanitizeMetricName(r.Descriptor().LibraryName(), r.Descriptor().Name())
-		itr := r.Labels().Iter()
+		itr := label.NewMergeIterator(r.Resource().LabelSet(), r.Labels())
```
jmacd, line 73: Oh, one thing: this call results in Resource labels overriding the metric labels, in case of conflict. I believe OTel recommends a policy of letting the metric label override the resource value, and likewise for Spans. I think we should reverse these arguments.

MrAlias:
I'm pretty sure this was superseded by #53

jmacd:
This contains a change to address a TODO about using the resource values.

PR #58, opened 2020-06-08
Fix module version

rnburn:


evantorrie:
I believe this will be fixed when #57 is merged, at which point in time you should be able to refer to the mongodb driver as v0.6.1.

Let's wait until #57 is merged and if you still have problems, we'll take this back up again.

MrAlias:
#57 is merged and tagged. PTAL

PR #60, opened 2020-06-08
Add tagging script for release process

MrAlias:
Also, rename `bump_and_tag.sh` to `bump_version.sh`.


tag.sh
```diff
@@ -0,0 +1,178 @@
+#!/bin/bash
```
evantorrie, line 1: Is there a reason to not have this as `#!/usr/bin/env bash`

Maybe I'm the only one running this on MacOSX, but by default, MacOS's bash at /bin/bash is 3.2.57 - released in 2007, and Apple clearly does not intend to upgrade the built-in version. I've often run into problems where a bash construct created sometime in the last 13 years has been used in a script, and the built-in /bin/bash chokes on it.

With `#!/usr/bin/env bash`, it will pick up the first bash from `$PATH`, which is easy to make point to an up-to-date bash with something like homebrew.

tag.sh
```diff
@@ -0,0 +1,178 @@
...
+    shift
+    commit="$1"
+    if [ -z "$commit" ]
+    then
+        echo "missing COMMIT_HASH"
+        usage
+        exit 1
+    fi
+    # Verify rev is for a commit and unify hashes into a complete SHA1.
+    readonly SHA="$( git rev-parse --quiet --verify "${commit}^{commit}" )"
```
evantorrie, line 100: Nice, I hadn't seen this trailing selector `^{commit}` before.

tag.sh
```diff
@@ -0,0 +1,178 @@
...
+        | sort \
+        | tail -1
+}
+
+print_changes() {
+    local tag="$1"
+    local previous
+
+    previous="$( previous_version "$tag" )"
+    if [ "$previous" ]
```
evantorrie, line 150:
```suggestion
    if [ -n "$previous" ]
```

tag.sh
```diff
@@ -0,0 +1,178 @@
...
+}
+
+previous_version() {
+    local current="$1"
+
+    git tag \
+        | grep -E "^${SEMVER_REGEX}$" \
+        | grep -v "$current" \
+        | sort \
+        | tail -1
```
evantorrie, line 142: Will this sort work for say, `v0.9.0` vs `v0.10.0`.

If we assume git 2.0 or later, this could be just `git tag -l --sort=v:refname` to guarantee sort by semantic version.

tag.sh
MrAlias: Yeah, that sounds good. More habit that I write `#!/bin/bash`.  I'll switch.

tag.sh
MrAlias: :smile:

tag.sh
MrAlias: Good point. I'll look into switching.

MrAlias: Yeah, 2.0 has been out for 6 years now. I think we can switch to this.

PR #117, opened 2020-06-30
Add link to go-redis/redis OpenTelemetry integration

### LINKED ISSUE #116: Update instrumentation readme to include links to instrumentation in external repos
As per discussion with @MrAlias, we agreed that the readme in the instrumentation directory should be updated to include links to integrations that are not in the opentelemetry-go-contrib repo.

@TigerHe7 and I found that [go-redis/go](https://github.com/go-redis/redis) already contains an integration for opentelemetry. 

This issue is to update the readme with a table including a link to that repo.

reggiemcdonald:
Fixes #116 

This PR adds a markdown table to the readme in the instrumentation directory to include a link to the go-redis/redis opentelemetry integration.


reggiemcdonald:
I signed it

Aneurysm9:
> Not sure whats up with the CLA. Is the linuxfoundation check needed?

No, the EasyCLA check is now required.  I believe the linuxfoundation check will be removed at some point.

instrumentation/README.md
```diff
@@ -35,3 +35,11 @@ Additionally the following guidelines for package composition need to be followe
...
    It SHOULD contain useful information like what the purpose of the instrumentation is, how to use it, and any compatibility restrictions that might exist. 
 - Examples of how to actually use the instrumentation SHOULD be included.
+
+## Additional Instrumentation Packages
+
+Below are additional instrumentation packages outside of the opentelemetry-go-contrib repo:
+
+|  Package Name  |                     Link                    |                                                          Notes                                                                 |
+| :------------: | :-----------------------------------------: | :----------------------------------------------------------------------------------------------------------------------------: |
+| go-redis/redis | [Github](https://github.com/go-redis/redis) | Trace only; add the hook found [here](https://github.com/go-redis/redis/blob/master/redisext/otel.go) to your go-redis client. |
```
MrAlias, line 45: We should probably use the fully qualified package name for the instrumentation: `github.com/go-redis/redis/v8/redisext`

It might also be helpful to structure the table with the package name linking to the source and the next column to the docs over at https://pkg.go.dev/. Maybe something like:

```suggestion
|  Package Name  | Documentation | Notes |
| :------------: | :-----------: | :---: |
| [`github.com/go-redis/redis/v8/redisext`](https://github.com/go-redis/redis/blob/v8.0.0-beta.5/redisext/otel.go) | [Go Docs](https://pkg.go.dev/github.com/go-redis/redis/v8@v8.0.0-beta.5.0.20200614113957-5b4d00c217b0/redisext?tab=doc) | Trace only; add the hook found [here](https://github.com/go-redis/redis/blob/master/redisext/otel.go) to your go-redis client. |
```

reggiemcdonald: This makes perfect sense. Should we maybe use the master branch for the links so that it doesn't fall out of date as quickly?

lizthegrey:
> > Not sure whats up with the CLA. Is the linuxfoundation check needed?
> 
> No, the EasyCLA check is now required. I believe the linuxfoundation check will be removed at some point.

Affirm, removal of the old lf check is in progress. https://github.com/open-telemetry/community/issues/379

instrumentation/README.md
MrAlias: Yeah, it is a bit of a trade off. Using the master branch has the downside of upstream changes causing the link to potentially break if the upstream directory changes. Linking to a specific hash or tag also helps give a clue as to what we see currently when making the doc, kind of like a timestamp.

I'm fine with either though, up to you.

reggiemcdonald: This is a great point - I stuck with the tagged links to avoid breaking changes

PR #118, opened 2020-07-01
Move all github.com instrumentation plugins down one level in directory hierarchy

evantorrie:
For

   1. consistency with other plugins placed under the `instrumentation/` directory
   2. consistency with [the documentation](https://github.com/open-telemetry/opentelemetry-go-contrib/tree/a24a7d579f5b3e345c4bb001f99a1fc29aec9114/instrumentation#packaging)

this moves all instrumentation plugins for `github.com/` packages underneath a `instrumentation/github.com/` directory.

In addition, also added a `doc.go` for `labstack/echo`. It was previously missing one.


lizthegrey:
This needed to also change `.github/dependabot.yml`

PR #136, opened 2020-07-16
Upgrade to new protocol

churanchen:
TODO: Support periods which aren't mutually divisible

Most of the change is upgrading to the new protocol.


churanchen:
Mistake

Issue #91, opened 2020-06-18
Add instrumentation for Cassandra: github.com/gocql/gocql


MrAlias:
https://github.com/gocql/gocql

## Tasks

- [x] code complete with tests passing
- [x] `Dockerfile` with example app showing instrumentation
- [x] documentation updated
- [x] instrumentation added to the opentelemetry [registry](https://opentelemetry.io/registry/)

## Prior Art

- https://github.com/DataDog/dd-trace-go/tree/v1/contrib/gocql/gocql

reggiemcdonald:
@MrAlias could I give this a try?

reggiemcdonald:
After a bit of research it appears that I can either: 
- Create an implementation of [`gocql.Tracer`](https://github.com/gocql/gocql/blob/c6852bb0463ade62d1234270419b7c948b45a7e1/session.go#L591-L598), which appears to be quite resource intensive with the multiple queries to the Database, but does not require the creation of a wrapper
- Create a wrapper around `gocql.Session` and avoid using the Tracer hook

Any suggestions which might be the better approach?

Aneurysm9:
The comment on the `gocql.Tracer` type says: 

> Gathering this information might be essential for debugging and optimizing queries, but this feature should not be used on production systems with very high load.

That would lead me to believe that creating a wrapper is the better approach.  Ideally we'd be able to leave the instrumentation enabled at all times.

reggiemcdonald:
> The comment on the `gocql.Tracer` type says:
> 
> > Gathering this information might be essential for debugging and optimizing queries, but this feature should not be used on production systems with very high load.
> 
> That would lead me to believe that creating a wrapper is the better approach. Ideally we'd be able to leave the instrumentation enabled at all times.

Perfect that's what I thought. Thanks!

reggiemcdonald:
> https://github.com/gocql/gocql
> 
> ## Tasks
> * [x]  code complete with tests passing
> * [x]  `Dockerfile` with example app showing instrumentation
> * [x]  documentation updated
> * [x]  instrumentation added to the opentelemetry [registry](https://opentelemetry.io/registry/)
> 
> ## Prior Art
> * https://github.com/DataDog/dd-trace-go/tree/v1/contrib/gocql/gocql

All tasks completed, ready to close

MrAlias:
Awesome :tada:! Thanks for the help @reggiemcdonald

PR #137, opened 2020-07-16
Add instrumentation for github.com/gocql/gocql

reggiemcdonald:
Addresses #91 

This PR adds an integration for the Cassandra DB golang client, gocql. 

Since the [previous implementation of this integration](https://github.com/DataDog/dd-trace-go/tree/v1/contrib/gocql/gocql) was created, the gocql community added Observer hooks that can be used to collect metrics for queries, batch  queries, and connection attempts. You can see the one for queries [here](https://github.com/gocql/gocql/pull/1023). This integration uses these observers, which differs from the prior art, which created a wrapper around the `gocql.Query`. 

Any feedback is appreciated!


instrumentation/github.com/gocql/gocql/example/go.mod
```diff
@@ -0,0 +1,14 @@
+module go.opentelemetry.io/contrib/github.com/gocql/gocql/example
+
+go 1.14
+
+require (
+	github.com/gocql/gocql v0.0.0-20200624222514-34081eda590e
+	go.opentelemetry.io/contrib/github.com/gocql/gocql v0.0.0
+	go.opentelemetry.io/otel v0.8.0
+	go.opentelemetry.io/otel/exporters/metric/prometheus v0.8.0
+	go.opentelemetry.io/otel/exporters/trace/zipkin v0.8.0
```
MrAlias, line 10: Should not include breaking changes.

```suggestion
	go.opentelemetry.io/otel v0.9.0
	go.opentelemetry.io/otel/exporters/metric/prometheus v0.9.0
	go.opentelemetry.io/otel/exporters/trace/zipkin v0.9.0
```

instrumentation/github.com/gocql/gocql/go.mod
```diff
@@ -0,0 +1,11 @@
+module go.opentelemetry.io/contrib/github.com/gocql/gocql
+
+go 1.14
+
+require (
+	github.com/gocql/gocql v0.0.0-20200624222514-34081eda590e
+	github.com/stretchr/testify v1.6.1
+	go.opentelemetry.io/contrib v0.7.0
+	go.opentelemetry.io/otel v0.7.0
```
MrAlias, line 9:
```suggestion
	go.opentelemetry.io/contrib v0.9.0
	go.opentelemetry.io/otel v0.9.0
```

instrumentation/github.com/gocql/gocql/gocql_test.go
```diff
@@ -0,0 +1,479 @@
...
+
+	ctx, parentSpan := tracer.Start(context.Background(), "gocql-test")
+
+	session, err := NewSessionWithTracing(
+		ctx,
+		cluster,
+		WithTracer(tracer),
+		WithConnectInstrumentation(false),
+	)
+	assert.NoError(t, err)
```
MrAlias, line 110:
```suggestion
	require.NoError(t, err)
```

instrumentation/github.com/gocql/gocql/gocql_test.go
```diff
@@ -0,0 +1,479 @@
...
+	defer session.Close()
+
+	id := gocql.TimeUUID()
+	title := "example-title"
+	insertStmt := fmt.Sprintf("insert into %s (id, title) values (?, ?)", tableName)
+	query := session.Query(insertStmt, id, title).WithContext(ctx)
+	assert.NotNil(t, query, "expected query to not be nil")
+	if err := query.Exec(); err != nil {
+		t.Fatal(err.Error())
+	}
```
MrAlias, line 120:
```suggestion
	require.NoError(t, query.Exec())
```

instrumentation/github.com/gocql/gocql/gocql_test.go
```diff
@@ -0,0 +1,479 @@
...
+	parentSpan.End()
+
+	// Get the spans and ensure that they are child spans to the local parent
+	spans := tracer.EndedSpans()
+
+	// Collect all the connection spans
+	// total spans:
+	// 1 span for the Query
+	// 1 span created in test
+	assert.Equal(t, 2, len(spans))
```
MrAlias, line 131:
```suggestion
	assert.Len(t, spans, 2)
```

instrumentation/github.com/gocql/gocql/gocql_test.go
```diff
@@ -0,0 +1,479 @@
...
+		assert.NotNil(t, span.Attributes[cassVersionKey].AsString())
+		assert.Equal(t, cluster.Hosts[0], span.Attributes[cassHostKey].AsString())
+		assert.Equal(t, int32(cluster.Port), span.Attributes[cassPortKey].AsInt32())
+		assert.Equal(t, "up", strings.ToLower(span.Attributes[cassHostStateKey].AsString()))
+	}
+
+	// Check metrics
+	controller.Stop()
+
+	assert.Equal(t, 3, len(exporter.records))
```
MrAlias, line 152:
```suggestion
	require.Len(t, exporter.records, 3)
```

instrumentation/github.com/gocql/gocql/gocql_test.go
```diff
@@ -0,0 +1,479 @@
...
+
+	ctx, parentSpan := tracer.Start(context.Background(), "gocql-test")
+
+	session, err := NewSessionWithTracing(
+		ctx,
+		cluster,
+		WithTracer(tracer),
+		WithConnectInstrumentation(false),
+	)
+	assert.NoError(t, err)
```
MrAlias, line 220:
```suggestion
	require.NoError(t, err)
```

instrumentation/github.com/gocql/gocql/gocql_test.go
```diff
@@ -0,0 +1,479 @@
...
+	batch := session.NewBatch(gocql.LoggedBatch).WithContext(ctx)
+	for i := 0; i < 10; i++ {
+		id := gocql.TimeUUID()
+		title := fmt.Sprintf("batch-title-%d", i)
+		stmt := fmt.Sprintf("insert into %s (id, title) values (?, ?)", tableName)
+		batch.Query(stmt, id, title)
+	}
+
+	err = session.ExecuteBatch(batch)
+	assert.NoError(t, err)
```
MrAlias, line 232:
```suggestion
	require.NoError(t, session.ExecuteBatch(batch))
```

instrumentation/github.com/gocql/gocql/gocql_test.go
```diff
@@ -0,0 +1,479 @@
...
+	assert.Equal(t, 2, len(spans))
+	span := spans[0]
+
+	assert.Equal(t, cassBatchQueryName, span.Name)
+	assert.Equal(t, parentSpan.SpanContext().SpanID, span.ParentSpanID)
+	assert.NotNil(t, span.Attributes[cassVersionKey].AsString())
+	assert.Equal(t, cluster.Hosts[0], span.Attributes[cassHostKey].AsString())
+	assert.Equal(t, int32(cluster.Port), span.Attributes[cassPortKey].AsInt32())
+	assert.Equal(t, "up", strings.ToLower(span.Attributes[cassHostStateKey].AsString()))
+	assert.Equal(t, int32(10), span.Attributes[cassBatchQueriesKey].AsInt32())
```
MrAlias, line 249:
```suggestion
	if assert.Len(t, spans, 2) {
		span := spans[0]

		assert.Equal(t, cassBatchQueryName, span.Name)
		assert.Equal(t, parentSpan.SpanContext().SpanID, span.ParentSpanID)
		assert.NotNil(t, span.Attributes[cassVersionKey].AsString())
		assert.Equal(t, cluster.Hosts[0], span.Attributes[cassHostKey].AsString())
		assert.Equal(t, int32(cluster.Port), span.Attributes[cassPortKey].AsInt32())
		assert.Equal(t, "up", strings.ToLower(span.Attributes[cassHostStateKey].AsString()))
		assert.Equal(t, int32(10), span.Attributes[cassBatchQueriesKey].AsInt32())
	}
```

instrumentation/github.com/gocql/gocql/gocql_test.go
```diff
@@ -0,0 +1,479 @@
...
+	spans := tracer.EndedSpans()
+	// total spans:
+	// 1 span for the query
+	// 1 span for the local span
+	assert.Equal(t, 2, len(spans))
+	span := spans[0]
+
+	assert.Equal(t, cassBatchQueryName, span.Name)
+	assert.Equal(t, parentSpan.SpanContext().SpanID, span.ParentSpanID)
+	assert.NotNil(t, span.Attributes[cassVersionKey].AsString())
```
MrAlias, line 245: IIRC `AsString` will return a string which is never `nil`. It might also be more useful for failures to test inclusion here directly.

```suggestion
	assert.Contains(t, span.Attributes, cassVersionKey)
```

instrumentation/github.com/gocql/gocql/gocql_test.go
```diff
@@ -0,0 +1,479 @@
...
+	assert.Equal(t, cluster.Hosts[0], span.Attributes[cassHostKey].AsString())
+	assert.Equal(t, int32(cluster.Port), span.Attributes[cassPortKey].AsInt32())
+	assert.Equal(t, "up", strings.ToLower(span.Attributes[cassHostStateKey].AsString()))
+	assert.Equal(t, int32(10), span.Attributes[cassBatchQueriesKey].AsInt32())
+
+	controller.Stop()
+
+	// Check metrics
+	assert.Equal(t, 2, len(exporter.records))
```
MrAlias, line 254:
```suggestion
	require.Len(t, exporter.records, 2)
```

instrumentation/github.com/gocql/gocql/gocql_test.go
```diff
@@ -0,0 +1,479 @@
...
+	tracer := mocktracer.NewTracer("gocql-test")
+	connectObserver := &mockConnectObserver{0}
+
+	session, err := NewSessionWithTracing(
+		context.Background(),
+		cluster,
+		WithTracer(tracer),
+		WithConnectObserver(connectObserver),
+	)
+	assert.NoError(t, err)
```
MrAlias, line 307:
```suggestion
	require.NoError(t, err)
```

instrumentation/github.com/gocql/gocql/gocql_test.go
```diff
@@ -0,0 +1,479 @@
...
+	spans := tracer.EndedSpans()
+
+	assert.Less(t, 0, connectObserver.callCount)
+
+	controller.Stop()
+
+	// Verify the span attributes
+	for _, span := range spans {
+		assert.Equal(t, cassConnectName, span.Name)
+		assert.NotNil(t, span.Attributes[cassVersionKey].AsString())
```
MrAlias, line 319: Same here w/r/t to testing inclusion.

instrumentation/github.com/gocql/gocql/observer.go
```diff
@@ -0,0 +1,263 @@
...
+// ------------------------------------------ Private Functions
+
+// getHost returns the hostname as a string.
+// gocql.HostInfo.HostnameAndPort() returns a string
+// formatted like host:port. This function returns the host.
+func getHost(hostPort string) string {
+	idx := strings.Index(hostPort, ":")
+	host := hostPort[0:idx]
+	return host
+}
```
MrAlias, line 228: Could [`net.SplitHostPort`](https://golang.org/pkg/net/#SplitHostPort) replace this or be used instead of `string.Index`?

instrumentation/github.com/gocql/gocql/gocql_test.go
```diff
@@ -0,0 +1,479 @@
...
+	if err = session.Query(fmt.Sprintf("truncate table %s", tableName)).Exec(); err != nil {
+		log.Fatalf("failed to truncate table, %v", err)
+	}
+}
+
+func TestMain(m *testing.M) {
+	if _, present := os.LookupEnv("INTEGRATION"); !present {
+		fmt.Println("--- SKIP: to enable integration test, set the INTEGRATION environment variable")
+		os.Exit(0)
+	}
```
MrAlias, line 476: Do we need to update our CI system to be able to run these?

instrumentation/github.com/gocql/gocql/instrument.go
```diff
@@ -0,0 +1,110 @@
...
+)
+
+// InstrumentWithProvider will recreate instruments using a meter
+// from the given provider p.
+func InstrumentWithProvider(p metric.Provider) {
+	defer func() {
+		if recovered := recover(); recovered != nil {
+			log.Print("failed to create meter. metrics are not being recorded")
+		}
+	}()
```
MrAlias, line 63: Instead of using panic as an error system, we should not use a `metric.Must` meter and directly handle the errors on each instrument creation.

instrumentation/github.com/gocql/gocql/instrument.go
```diff
@@ -0,0 +1,110 @@
...
+	"go.opentelemetry.io/otel/api/unit"
+)
+
+var (
+	// iQueryCount is the number of queries executed.
+	iQueryCount metric.Int64Counter
+
+	// iQueryErrors is the number of errors encountered when
+	// executing queries.
+	iQueryErrors metric.Int64Counter
```
MrAlias, line 32: Do we need two instruments for these? Couldn't we add error labels to the count of a failed query?

instrumentation/github.com/gocql/gocql/instrument.go
```diff
@@ -0,0 +1,110 @@
...
+
+	// iConnectionCount is the number of connections made
+	// with the traced session.
+	iConnectionCount metric.Int64Counter
+
+	// iConnectErrors is the number of errors encountered
+	// when making connections with the current traced session.
+	iConnectErrors metric.Int64Counter
+
+	// iLatency is the sum of attempt latencies.
```
MrAlias, line 52: Seems odd to say this is the sum here. Would it be acceptable to talk about the total instead?

```suggestion
	// iLatency is the total latency of a query to the Cassandra database.
```

instrumentation/github.com/gocql/gocql/instrument.go
```diff
@@ -0,0 +1,110 @@
...
+	}()
+	meter := metric.Must(p.Meter("github.com/gocql/gocql"))
+
+	iQueryCount = meter.NewInt64Counter(
+		"cassandra.queries",
+		metric.WithDescription("Number of queries executed"),
+	)
+
+	iQueryErrors = meter.NewInt64Counter(
+		"cassandra.query.errors",
```
MrAlias, line 72: For naming we should try to structure it hierarchically. So related instruments like this would share a naming pattern. Maybe something like:

```suggestion
		"cassandra.query",
		metric.WithDescription("Number of queries executed"),
	)

	iQueryErrors = meter.NewInt64Counter(
		"cassandra.query.errors",
```

or

```suggestion
		"cassandra.queries",
		metric.WithDescription("Number of queries executed"),
	)

	iQueryErrors = meter.NewInt64Counter(
		"cassandra.queries.errors",
```

instrumentation/github.com/gocql/gocql/observer.go
```diff
@@ -0,0 +1,263 @@
...
+	"github.com/gocql/gocql"
+
+	"go.opentelemetry.io/otel/api/kv"
+	"go.opentelemetry.io/otel/api/metric"
+	"go.opentelemetry.io/otel/api/trace"
+)
+
+// OtelQueryObserver implements the gocql.QueryObserver interface
+// to provide instrumentation to gocql queries.
+type OtelQueryObserver struct {
```
MrAlias, line 31: nit, OTel is an initialism and should maintain the capitalization pattern.

```suggestion
type OTelQueryObserver struct {
```

instrumentation/github.com/gocql/gocql/cass.go
```diff
@@ -0,0 +1,141 @@
...
+	cassQueryAttemptsKey = kv.Key("cassandra.attempts")
+
+	// cassQueryAttemptNumKey is the key for the span attribute describing
+	// which attempt the current query is as a 0-based index.
+	cassQueryAttemptNumKey = kv.Key("cassandra.attempt")
+
+	// Names of the spans for query, batch query, and connect respectively.
+	cassQueryName      = "cassandra.query"
+	cassBatchQueryName = "cassandra.batch.query"
+	cassConnectName    = "cassandra.connect"
```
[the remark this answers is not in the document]
reggiemcdonald: Ahh yes I missed that doc. I'll refactor to match the semantic conventions 🙂

instrumentation/github.com/gocql/gocql/instrument.go
reggiemcdonald: Thats a great idea!

instrumentation/github.com/gocql/gocql/gocql_test.go
reggiemcdonald: I used the same pattern here as in the [mongo integration](https://github.com/open-telemetry/opentelemetry-go-contrib/blob/2baa89862fad76c1ed79eb74f36c9978c2a7a512/instrumentation/go.mongodb.org/mongo-driver/mongo_test.go#L33-L40)

Just looking at the circle ci check, it doesn't appear to be running integration tests like this. This test relies on a running cassandra instance, and the mongo tests rely on mongo. Is there an interest in setting something like this up?

instrumentation/github.com/gocql/gocql/cass.go
```diff
@@ -0,0 +1,141 @@
...
+	// returned on a query to the database.
+	cassRowsReturnedKey = kv.Key("cassandra.rows.returned")
+
+	// cassQueryAttemptsKey is the key for the span attribute describing the number of attempts
+	// made for the query in question.
+	cassQueryAttemptsKey = kv.Key("cassandra.attempts")
+
+	// cassQueryAttemptNumKey is the key for the span attribute describing
+	// which attempt the current query is as a 0-based index.
+	cassQueryAttemptNumKey = kv.Key("cassandra.attempt")
```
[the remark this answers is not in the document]
reggiemcdonald: I kept them distinct because the gocql community has in the stats object passed to the observers. But I agree, I think it can be removed

instrumentation/github.com/gocql/gocql/observer.go
reggiemcdonald: Yep!
With my experimentation using `localhost`, `gocql.HostnameAndPort()` appears to return a resolved IP and port number (`127.0.0.1:9042`). So I thought that I'd do something along the lines of this:
```go
func hostOrIP(hostnameAndPort string) kv.KeyValue {
	hostname, _, err := net.SplitHostPort(hostnameAndPort)
	if err != nil {
		log.Printf("failed to parse hostname from port, %v", err)
	}
	if parse := net.ParseIP(hostname); parse != nil {
		return cassPeerIP(parse.String())
	}
	return cassPeerName(hostname)
}
```

instrumentation/github.com/gocql/gocql/instrument.go
reggiemcdonald: I think that language changes the meaning of the measurement. I tried to maintain consistency with the language used in the [gocql library](https://github.com/gocql/gocql/blob/4ee5414cce0b81345a411212c2cf8013bfab3ea9/session.go#L698-L705)

instrumentation/github.com/gocql/gocql/config.go
```diff
@@ -0,0 +1,130 @@
...
+// created with NewSessionWithTracing.
+type TracedSessionConfig struct {
+	otelConfig      *OtelConfig
+	queryObserver   gocql.QueryObserver
+	batchObserver   gocql.BatchObserver
+	connectObserver gocql.ConnectObserver
+}
+
+// OtelConfig provides OpenTelemetry configuration.
+type OtelConfig struct {
```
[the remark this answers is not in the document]
reggiemcdonald: I've been playing around with this for a while now. I think the issue here is that `NewSessionWithTracing` would need to accept the option functions for each of `NewQueryObserver`, `NewBatchObserver`, and `NewConnectObserver`, as well as generic option functions like `WithTracer()`; which seems to require quite a bit of code using the linked method.

I think what would be best to do, is remove all of the `New*` functions except for `NewSessionWithTracing`, and merge the configuration structs. Then we'd only need the one type of option function, and there'd be only one way to use the library - using `NewSessionWithTracing`. In retrospect, the additional constructor functions didn't really add any additional value to the library.

It looks like the following:

```suggestion
// TracedSessionConfig provides configuration for sessions
// created with NewSessionWithTracing.
type TracedSessionConfig struct {
	tracer            trace.Tracer
	instrumentQuery   bool
	instrumentBatch   bool
	instrumentConnect bool
	queryObserver     gocql.QueryObserver
	batchObserver     gocql.BatchObserver
	connectObserver   gocql.ConnectObserver
}

// TracedSessionOption is a function type that applies
// a particular configuration to the traced session in question.
type TracedSessionOption func(*TracedSessionConfig)
```

and for the `NewSessionWithTracing`: 

```go
// NewSessionWithTracing creates a new session using the given cluster
// configuration enabling tracing for queries, batch queries, and connection attempts.
// You may use additional observers and disable specific tracing using the provided `TracedSessionOption`s.
func NewSessionWithTracing(ctx context.Context, cluster *gocql.ClusterConfig, options ...TracedSessionOption) (*gocql.Session, error) {
	config := configure(options...)
	cluster.QueryObserver = &OTelQueryObserver{
		enabled:  config.instrumentQuery,
		observer: config.queryObserver,
		tracer:   config.tracer,
	}
	cluster.BatchObserver = &OTelBatchObserver{
		enabled:  config.instrumentBatch,
		observer: config.batchObserver,
		tracer:   config.tracer,
	}
	cluster.ConnectObserver = &OTelConnectObserver{
		ctx:      ctx,
		enabled:  config.instrumentConnect,
		observer: config.connectObserver,
		tracer:   config.tracer,
	}
	return cluster.CreateSession()
}
```

instrumentation/github.com/gocql/gocql/example/db.go
```diff
@@ -0,0 +1,54 @@
...
+package main
+
+import (
+	"fmt"
+	"log"
+
+	"github.com/gocql/gocql"
+)
+
+func initDb() {
```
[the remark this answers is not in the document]
reggiemcdonald: I originally had it in `client.go` but then thought that looked weird so I moved it. In retrospect, that was a bad move. I'll put it back! 🙂

instrumentation/github.com/gocql/gocql/example/client.go
```diff
@@ -0,0 +1,170 @@
...
+	traceExporter, err := zipkintrace.NewExporter(
+		"http://localhost:9411/api/v2/spans",
+		"zipkin-example",
+	)
+	if err != nil {
+		log.Fatalf("failed to create span traceExporter, %v", err)
+	}
+
+	provider, err := sdktrace.NewProvider(
+		sdktrace.WithBatcher(
```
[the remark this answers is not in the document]
reggiemcdonald: Since zipkin doesn't implement the `SpanSyncer` interface, would you recommend that I change the example to export to stdout?

instrumentation/github.com/gocql/gocql/config.go
```diff
@@ -0,0 +1,129 @@
...
+	return TracedSessionOptionFunc(func(cfg *TracedSessionConfig) {
+		cfg.instrumentConnect = enabled
+	})
+}
+
+// ------------------------------------------ Private Functions
+
+func configure(options ...TracedSessionOption) *TracedSessionConfig {
+	config := &TracedSessionConfig{
+		tracer:            global.Tracer("github.com/gocql/gocql"),
```
Aneurysm9, line 118: This needs to refer to the instrumentation library (`go.opentelemetry.io/contrib/github.com/gocql/gocql`) rather than the instrumented library.  See [the Trace spec](https://github.com/open-telemetry/opentelemetry-specification/blob/master/specification/trace/api.md#tracerprovider-operations).

instrumentation/github.com/gocql/gocql/instrument.go
```diff
@@ -0,0 +1,89 @@
...
+	iConnectionCount metric.Int64Counter
+
+	// iLatency is the sum of attempt latencies.
+	iLatency metric.Int64ValueRecorder
+)
+
+// InstrumentWithProvider will recreate instruments using a meter
+// from the given provider p.
+func InstrumentWithProvider(p metric.Provider) {
+	meter := p.Meter("github.com/gocql/gocql")
```
Aneurysm9, line 47: This needs to refer to the instrumentation library (go.opentelemetry.io/contrib/github.com/gocql/gocql) rather than the instrumented library. See the [Meter spec](https://github.com/open-telemetry/opentelemetry-specification/blob/master/specification/metrics/api.md#obtaining-a-meter).

instrumentation/github.com/gocql/gocql/config.go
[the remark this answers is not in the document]
MrAlias: I'm in favor of the merged constructor. :+1:

instrumentation/github.com/gocql/gocql/example/client.go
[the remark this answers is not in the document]
MrAlias: Didn't realize the Zipkin exporter didn't support that. I'll leave it up to you if you want to switch, but this should be fine for now.

instrumentation/github.com/gocql/gocql/gocql_test.go
MrAlias: I think we want to support this, but it is out of scope for this PR.

I opened https://github.com/open-telemetry/opentelemetry-go-contrib/issues/160 to track this work. If you're interested, we we would certainly be happy to have help on that :smile: .

instrumentation/github.com/gocql/gocql/gocql_test.go
```diff
@@ -0,0 +1,526 @@
...
+	"go.opentelemetry.io/otel/sdk/metric/selector/simple"
+
+	"github.com/gocql/gocql"
+	"github.com/stretchr/testify/assert"
+
+	mocktracer "go.opentelemetry.io/contrib/internal/trace"
+	"go.opentelemetry.io/otel/api/kv"
+	"go.opentelemetry.io/otel/api/metric"
+	export "go.opentelemetry.io/otel/sdk/export/metric"
+	"go.opentelemetry.io/otel/sdk/export/metric/aggregation"
```
MrAlias, line 40: We'll probably want to avoid this testing pattern where we depend on the default SDK. It will bring in a lot of dependencies for any user of this instrumentation (if they don't use the instrumentation (no SDK registered) it will still keep a copy of the SDK in the build because of this). Given the separation from the main repo, we cannot use the internal test structures like this https://github.com/open-telemetry/opentelemetry-go/blob/master/api/global/internal/meter_test.go#L28. We'll likely need to readdress when https://github.com/open-telemetry/opentelemetry-go/issues/872 is resolved.

This is a common issues we recently identified in other parts of the OTel API, so not uncommon.

MrAlias: Tracking with https://github.com/open-telemetry/opentelemetry-go-contrib/issues/161

instrumentation/github.com/gocql/gocql/gocql_test.go
reggiemcdonald: Would love to help!

instrumentation/github.com/gocql/gocql/README.md
```diff
@@ -0,0 +1,45 @@
...
+To enable tracing in your application: 
+
+```go
+package main
+
+import (
+	"context"
+
+	"github.com/gocql/gocql"
+	otelGocql "go.opentelemetry.io/contrib/github.com/gocql/gocql"
```
MrAlias, line 14:
```suggestion
	otelGocql "go.opentelemetry.io/contrib/instrumentation/github.com/gocql/gocql"
```

instrumentation/github.com/gocql/gocql/README.md
```diff
@@ -0,0 +1,45 @@
...
+```
+
+You can customize instrumentation by passing any of the following options to `NewSessionWithTracing`:
+
+| Function | Description |
+| -------- | ----------- |
+| `WithQueryObserver(gocql.QueryObserver)` | Specify an additional QueryObserver to be called. |
+| `WithBatchObserver(gocql.BatchObserver)` | Specify an additional BatchObserver to be called. |
+| `WithConnectObserver(gocql.ConnectObserver)` | Specify an additional ConnectObserver to be called. |
+| `WithTracer(trace.Tracer)` | The tracer to be used to create spans for the gocql session. If not specified, `global.Tracer("go.opentelemetry.io/contrib/github.com/gocql/gocql")` will be used. |
```
MrAlias, line 41:
```suggestion
| `WithTracer(trace.Tracer)` | The tracer to be used to create spans for the gocql session. If not specified, `global.Tracer("go.opentelemetry.io/contrib/instrumentation/github.com/gocql/gocql")` will be used. |
```

instrumentation/github.com/gocql/gocql/config.go
```diff
@@ -0,0 +1,129 @@
...
+// OpenTelemetry implementation, if it is not nil. Defaults to nil.
+func WithConnectObserver(observer gocql.ConnectObserver) TracedSessionOption {
+	return TracedSessionOptionFunc(func(cfg *TracedSessionConfig) {
+		cfg.connectObserver = observer
+	})
+}
+
+// WithTracer will set tracer to be the tracer used to create spans
+// for query, batch query, and connection instrumentation.
+// Defaults to global.Tracer("go.opentelemetry.io/contrib/github.com/gocql/gocql").
```
MrAlias, line 83:
```suggestion
// Defaults to global.Tracer("go.opentelemetry.io/contrib/instrumentation/github.com/gocql/gocql").
```

instrumentation/github.com/gocql/gocql/config.go
```diff
@@ -0,0 +1,129 @@
...
+	return TracedSessionOptionFunc(func(cfg *TracedSessionConfig) {
+		cfg.instrumentConnect = enabled
+	})
+}
+
+// ------------------------------------------ Private Functions
+
+func configure(options ...TracedSessionOption) *TracedSessionConfig {
+	config := &TracedSessionConfig{
+		tracer:            global.Tracer("go.opentelemetry.io/contrib/github.com/gocql/gocql"),
```
MrAlias, line 118: Might be worth making this into an unexported constant to share across the package (i.e. the `Meter` as well).

```suggestion
		tracer:            global.Tracer("go.opentelemetry.io/contrib/instrumentation/github.com/gocql/gocql"),
```

instrumentation/github.com/gocql/gocql/instrument.go
```diff
@@ -0,0 +1,89 @@
...
+	iConnectionCount metric.Int64Counter
+
+	// iLatency is the sum of attempt latencies.
+	iLatency metric.Int64ValueRecorder
+)
+
+// InstrumentWithProvider will recreate instruments using a meter
+// from the given provider p.
+func InstrumentWithProvider(p metric.Provider) {
+	meter := p.Meter("go.opentelemetry.io/contrib/github.com/gocql/gocql")
```
MrAlias, line 47:
```suggestion
	meter := p.Meter("go.opentelemetry.io/contrib/instrumentation/github.com/gocql/gocql")
```

PR #156, opened 2020-07-28
Move `google.golang.org/grpc` instrumentation from go.opentelemetry.io/otel

MrAlias:
- Includes a package rename: `go.opentelemetry.io/contrib/google.golang.org/grpc`
- Moves examples to actual example test.
- Packages it as its own module.

Part of https://github.com/open-telemetry/opentelemetry-go/issues/976


MrAlias:
Need to cut a release of go.opentelemetry.io/otel before I can fix the dependencies of the example here.

MrAlias:
Reattempt after the upcoming release of `go.opentelemetry.io/otel`

PR #188, opened 2020-08-04
Design doc

connorlindsey:
The attached design document outlines our plan to implement a Cortex Exporter for the Go SDK. This shouldn't be merged, but is instead a request for review before we begin implementing the Exporter. Please leave comments below or in the document. Thank you!

cc @ercl @alolita


MrAlias:
@connorlindsey I'd say this is "Accepted", but not sure if that means it can be closed? I'll leave it up to you.

Issue #158, opened 2020-07-28
Add instrumentation for net/http and net/httptrace


Aneurysm9:
See https://github.com/open-telemetry/opentelemetry-go/issues/974.

# Tasks

- [x]  code complete with tests passing
- [x]  Dockerfile with example app showing instrumentation
- [x]  documentation updated
- [x]  instrumentation added to the opentelemetry registry

MrAlias:
I think this might be complete.

PR #190, opened 2020-08-04
Add instrumentation for net/http and net/httptrace

Aneurysm9:
This PR moves the `othttp` and `httptrace` instrumentations from the main repo, renaming them to conform to this repo's naming conventions.  It also moves the HTTP example from the main repo and uses it to illustrate each instrumentation separately.

Part of #158 and https://github.com/open-telemetry/opentelemetry-go/issues/974


.gitignore
```diff
@@ -8,3 +8,4 @@ Thumbs.db
 *.so
 coverage.*
 
+instrumentation/github.com/gocql/gocql/example/example
```
lizthegrey, line 11: Inadvertent change?

internal/metric/mock.go
```diff
@@ -0,0 +1,191 @@
+// Copyright The OpenTelemetry Authors
```
lizthegrey, line 1: intentional to copy these over rather than leaving them in otel package?

.gitignore
Aneurysm9: `make precommit` summons that file into existence and I'm pretty sure it shouldn't be committed.

internal/metric/mock.go
Aneurysm9: They're internal to the otel package and thus can't be used here.  https://github.com/open-telemetry/opentelemetry-go/issues/872 could perhaps provide an exported solution, but one doesn't exist yet.

.gitignore
lizthegrey: odd. https://github.com/open-telemetry/opentelemetry-go-contrib/pull/137 doesn't seem to imply automatically building the example in that dir, wonder why it's doing that. it should write to $GOPATH/bin instead...

MrAlias: Yeah, I noticed this as well. Thanks for catching it here.

MrAlias:
@lizthegrey did your questions get answered or did we need to follow up on them?  Wondering if I can merge this.

instrumentation/net/http/common.go
```diff
@@ -0,0 +1,41 @@
...
+const (
+	ReadBytesKey  = kv.Key("http.read_bytes")  // if anything was read from the request body, the total number of bytes read
+	ReadErrorKey  = kv.Key("http.read_error")  // If an error occurred while reading a request, the string of the error (io.EOF is not recorded)
+	WroteBytesKey = kv.Key("http.wrote_bytes") // if anything was written to the response writer, the total number of bytes written
+	WriteErrorKey = kv.Key("http.write_error") // if an error occurred while writing a reply, the string of the error (io.EOF is not recorded)
+)
+
+// Server HTTP metrics
+const (
+	RequestCount          = "http.server.request_count"           // Incoming request count total
```
jmacd, line 33: (Note: there is a _belief_ that we can configure the `http.server.duration` metric, which has a count, to generate this `http.server.request_count` metric. I realize that we should not get hung up on views, so I think this looks good as a set of metrics.

MrAlias: FWIW, I noticed that this `const` isn't actually used in the code. It was one of the things I was alluding to needing clean up after the move.

Issue #82, opened 2020-06-18
Add instrumentation for github.com/astaxie/beego


MrAlias:
https://github.com/astaxie/beego

## Tasks

- [ ] code complete with tests passing
- [ ] `Dockerfile` with example app showing instrumentation
- [ ] documentation updated
- [ ] instrumentation added to the opentelemetry [registry](https://opentelemetry.io/registry/)

## Prior Art

- https://github.com/opentracing-contrib/beego

reggiemcdonald:
Could the beego integration leverage the `net/http` integration that was merged in this week?

MrAlias:
> Could the beego instrumentation leverage the `net/http` instrumentation that was merged in this week?

That seems to be what the OpenTracing project did: https://github.com/opentracing-contrib/beego/blob/master/middleware.go

Sounds like a good approach to me. :+1:

reggiemcdonald:
@MrAlias go ahead and close this one! `opentelemetry.io` has been updated with a link to the beego integration

MrAlias:
Awesome @reggiemcdonald!  Thanks for all the help :1st_place_medal:

PR #200, opened 2020-08-07
feat: Add integration for github.com/astaxie/beego

reggiemcdonald:
#### Description
This PR provides beego instrumentation.

#### Implementation
This integration makes use of the [`net/http` integration](https://github.com/open-telemetry/opentelemetry-go-contrib/tree/master/instrumentation/net/http) to provide tracing to HTTP requests.

HTTP requests to the beego app running this middleware will be traced. 

The following options are provided: 
- `WithTracer(trace.Tracer)`
- `WithMeter(metric.Meter)`
- `WithPropagators(propagation.Propagators)`
- `WithFilter(Filter)`
- `WithSpanNameFormatter(Formatter)`

which allows users to specify the tracer, meter and propagators to be used. In accordance with the [semantic conventions](https://github.com/open-telemetry/opentelemetry-specification/blob/master/specification/trace/semantic_conventions/http.md#name), span names are generated from the path of the HTTP request by default. Users are able to override this behaviour with the provided `WithSpanNameFormatter` option.
`WithFilter` allows users to filter out which requests should be traced.

Additionally, Beego supports HTML template rendering using calls to `beego.Controller.Render`. This instrumentation provides a wrapper function to add a child span to the rendering of the template files.

A dockerized example can be found [here](https://github.com/open-telemetry/opentelemetry-go-contrib/compare/master...reggiemcdonald:82-beego-integ-example). In an effort to keep this PR shorter, I've decided to move the example to a separate PR, which I'll open once this gets merged.

#### ~~TODO~~
- [x] Additional testing
- [x] Tracing render

Addresses: #82 

/cc @nlehrer @danherr


instrumentation/github.com/astaxie/beego/beego.go
```diff
@@ -0,0 +1,62 @@
...
+	if cfg.formatter != nil {
+		httpOptions = append(httpOptions, otelhttp.WithSpanNameFormatter(cfg.formatter))
+	}
+
+	return func(handler http.Handler) http.Handler {
+		return otelhttp.NewHandler(
+			handler,
+			name,
+			httpOptions...,
+		)
```
reggiemcdonald, line 60: @Aneurysm9 the `name` argument here is for the parameter `operation` for the `NewHandler` function in the http instrumentation package. I'm was wondering what exactly this variable is supposed to mean?

Aneurysm9: It is [provided to the `SpanNameFormatter`](https://github.com/open-telemetry/opentelemetry-go-contrib/blob/master/instrumentation/net/http/handler.go#L136) that is used to name the span created by the `Handler`.  The default formatter just uses it directly, custom formatters can do with it and the `http.Request` what they will.    I'm not familiar with Beego, but if it has a route name or template then that may be a good default choice that could reduce the required user input.

reggiemcdonald: Okay thanks. I set the default span name to be the url path, but I'll look into what a good default for this value would be

instrumentation/github.com/astaxie/beego/go.mod
```diff
@@ -0,0 +1,19 @@
...
+	go.opentelemetry.io/otel v0.10.0
+	golang.org/x/net v0.0.0-20200707034311-ab3426394381 // indirect
+	golang.org/x/sys v0.0.0-20200803210538-64077c9b5642 // indirect
+	golang.org/x/text v0.3.3 // indirect
+	google.golang.org/genproto v0.0.0-20200804131852-c06518451d9c // indirect
+	google.golang.org/grpc v1.31.0
+	google.golang.org/protobuf v1.25.0 // indirect
+)
+
+replace go.opentelemetry.io/contrib => ../../../..
```
reggiemcdonald, line 19: Both this package and `net/http` have this replace to be able to use the mock meter. Should I open an issue to change this following the next release?

Aneurysm9: Sure, I don't think it will be required once the next release is made.  Thanks.

MrAlias: https://github.com/open-telemetry/opentelemetry-go-contrib/issues/208

instrumentation/github.com/astaxie/beego/README.md
```diff
@@ -0,0 +1,82 @@
...
+}
+```
+
+You can customize instrumentation by passing any of the following options to `NewOtelBeegoMiddleWare`:
+
+| Function | Description |
+| -------- | ----------- |
+| `WithTracer(trace.Tracer)` | The tracer to be used to create spans for the beego server. If not specified, `global.Tracer("go.opentelemetry.io/contrib/instrumentation/github.com/astaxie/beego")` will be used. |
+| `WithMeter(metric.Meter)` | The meter to be used to create the instruments. If not specified, `global.Meter("go.opentelemery.io/contrib/instrumentation/github.com/astaxie/beego")` will be used. |
+| `WithPropagators(propagation.Propagators)` | The propagators to be used. If not specified, `global.Propagators()` will be used. |
```
MrAlias, line 35: This will need to be included in this cleanup: https://github.com/open-telemetry/opentelemetry-go-contrib/issues/213

instrumentation/github.com/astaxie/beego/README.md
```diff
@@ -0,0 +1,82 @@
...
+        
+        // Create the MiddleWare
+        mware := otelBeego.NewOTelBeegoMiddleWare("example-server")
+
+        // Run the server with the MiddleWare
+        beego.RunWithMiddleWares("localhost:8080", mware)
+}
+```
+
+You can customize instrumentation by passing any of the following options to `NewOtelBeegoMiddleWare`:
```
MrAlias, line 29:
```suggestion
You can customize instrumentation by passing any of the following options to `NewOTelBeegoMiddleWare`:
```

instrumentation/github.com/astaxie/beego/README.md
```diff
@@ -0,0 +1,82 @@
...
+        // Init the tracer and metric export pipelines
+
+        // Create your routes
+        
+        // Create the MiddleWare
+        mware := otelBeego.NewOTelBeegoMiddleWare("example-server")
+
+        // Run the server with the MiddleWare
+        beego.RunWithMiddleWares("localhost:8080", mware)
+}
```
MrAlias, line 26: `gofmt` this:

```suggestion
package main

import (
	"github.com/astaxie/beego"

	otelBeego "go.opentelemetry.io/contrib/instrumentation/github.com/astaxie/beego"
)

func main() {
	// Init the tracer and metric export pipelines

	// Create your routes

	// Create the MiddleWare
	mware := otelBeego.NewOTelBeegoMiddleWare("example-server")

	// Run the server with the MiddleWare
	beego.RunWithMiddleWares("localhost:8080", mware)
}
```

instrumentation/github.com/astaxie/beego/README.md
```diff
@@ -0,0 +1,82 @@
...
+
+        // Create your routes
+        beego.Router("/", &ExampleController{})
+
+        // Create the MiddleWare
+        mware := otelBeego.NewOTelBeegoMiddleWare("example-server")
+
+        // Run the server with the MiddleWare
+        beego.RunWithMiddleWares("localhost:8080", mware)
+}
```
MrAlias, line 75: `gofmt` this:

```suggestion
package main

import (
	"github.com/astaxie/beego"

	otelBeego "go.opentelemetry.io/contrib/instrumentation/github.com/astaxie/beego"
)

type ExampleController struct {
	beego.Controller
}

func (c *ExampleController) Get() {
	c.TplName = "index.ptl"

	// explicit call to traced Render function
	otelBeego.Render(&c.Controller)
}

func main() {
	// Init the tracer and metric export pipelines

	// Disable autorender
	beego.BConfig.WebConfig.AutoRender = false

	// Create your routes
	beego.Router("/", &ExampleController{})

	// Create the MiddleWare
	mware := otelBeego.NewOTelBeegoMiddleWare("example-server")

	// Run the server with the MiddleWare
	beego.RunWithMiddleWares("localhost:8080", mware)
}
```

MrAlias: Also this would work well in an [example](https://golang.org/pkg/testing/#hdr-Examples) for the go-docs.

instrumentation/github.com/astaxie/beego/README.md
```diff
@@ -0,0 +1,82 @@
...
+```
+
+| Function | Description |
+| -------- | ----------- |
+| `Render(*beego.Controller) error` | Provides tracing to `beego.Controller.Render`. |
+| `RenderString(*beego.Controller) (string, error)` | Provides tracing to `beego.Controller.RenderString`. |
+| `RenderBytes(*beego.Controller) ([]byte, error)` | Provides tracing to `beego.Controller.RenderBytes`. |
```
MrAlias, line 82: nit: this is likely not needed. The go-docs (where new users will be reading docs from) will render this info in the overview.

instrumentation/github.com/astaxie/beego/beego.go
```diff
@@ -0,0 +1,127 @@
...
+	"github.com/astaxie/beego"
+)
+
+// defaultSpanNameFormatter is the default formatter for spans created with
+// the beego integration. Returns the request URL path.
+func defaultSpanNameFormatter(operation string, req *http.Request) string {
+	if req.URL.Path == "" {
+		return req.Method
+	}
+	return req.URL.Path
```
MrAlias, line 35: Is there any way this can be made lower cardinality by using the route template instead of all the actual paths?

instrumentation/github.com/astaxie/beego/common.go
```diff
@@ -0,0 +1,50 @@
...
+const (
+	// packageName is the name of the this package, and is used as the default tracer
+	// and meter names.
+	packageName = "go.opentelemetry.io/contrib/instrumentation/github.com/astaxie/beego"
+
+	renderTemplateSpanName = "beego.render.template"
+	renderStringSpanName   = "beego.render.string"
+	renderBytesSpanName    = "beego.render.bytes"
+
+	templateKey = kv.Key("go.template")
```
MrAlias, line 34: Should this key be prefixed with `beego`?

```suggestion
	templateKey = kv.Key("beego.go.template")
```

instrumentation/github.com/astaxie/beego/README.md
reggiemcdonald: Yea definitely. Did you want this changed now? 
I could remove the options and replace them with `With*Provider()` options?

instrumentation/github.com/astaxie/beego/common.go
reggiemcdonald: I was following the example of the gin-gonic instrumentation which uses [`go.template`](https://github.com/open-telemetry/opentelemetry-go-contrib/blob/97f04f73f9f669f1d99492f6b914154149497037/instrumentation/github.com/gin-gonic/gin/gintrace.go#L105).

Beego uses the [html/template](https://golang.org/pkg/html/template/) package, so I thought that it made more sense to leave it as `go.template`. Happy to change it if you think otherwise!

instrumentation/github.com/astaxie/beego/README.md
MrAlias: Yeah, if you can that would be awesome! :heart:

instrumentation/github.com/astaxie/beego/common.go
MrAlias: Ah, makes sense. Leaving sounds good. :+1:

instrumentation/github.com/astaxie/beego/README.md
reggiemcdonald: Examples need to be runnable though. Wouldn't this cause problems for an HTTP server?

instrumentation/github.com/astaxie/beego/README.md
reggiemcdonald: Thoughts on getting rid of the readme entirely? All the functions are commented sufficiently

instrumentation/github.com/astaxie/beego/beego.go
reggiemcdonald: Okay so I've been working on this and I think I've come up with a solution. I would love to get your feedback on it, because its a little tricky. You can see what the diff looks like [here](https://github.com/reggiemcdonald/opentelemetry-go-contrib/compare/82-beego-integ...reggiemcdonald:82-beego-integ-span-name)

The problem is that the function beego provides to get the route template from a URL requires a `github.com/astaxie/beego/context.Contex`. You can get one of these contexts, but you need to set them with a `http.ResponseWriter` and `*http.Request` ([link to code](https://github.com/reggiemcdonald/opentelemetry-go-contrib/blob/950f7b2f90f670f244cb091071fcd48617d99ac8/instrumentation/github.com/astaxie/beego/beego.go#L36-L38)).

In the `SpanNameFormatter`, there is no `ResponseWriter` available to me. So I can't properly search up the route template from the `SpanNameFormatter`. 

So to solve this dilemma, what I've done is wrap the `ServeHTTP` function to lookup the route template and save it to the context before calling `otelhttp.ServeHTTP`. Then the `SpanNameFormatter` can retrieve the route template from the context if it exists, and return it as the span name. If the value isn't in the context (serving static files or using a router outside of the `beego.BeeApp`), then the HTTP method name is returned. In either case, cardinality is low.

instrumentation/github.com/astaxie/beego/README.md
MrAlias: > Thoughts on getting rid of the readme entirely? All the functions are commented sufficiently

SGTM.

instrumentation/github.com/astaxie/beego/README.md
MrAlias: I'm not sure what you mean by runnable, but the example needs to compile for sure. They don't necessarily need to output anything or actually serve web requests, but can look very close to, if not exactly, what you have here. An example of this in in the grpc instrumentation: https://github.com/open-telemetry/opentelemetry-go-contrib/blob/master/instrumentation/google.golang.org/grpc/example_interceptor_test.go

instrumentation/github.com/astaxie/beego/beego.go
MrAlias: This looks like a decent approach to me. :+1: 

I would say it would also be worth opening an issue in github.com/astaxie/beego to see if they can wrap this functionality themselves or maybe provide the template path in a more direct way to the `SpanNameFormatter`. Maybe by adding this to their context implementation? LMK if you would like help opening the issue.

reggiemcdonald: Awesome! I'll open an issue asking them to consider providing a function that can get the route template with just an http.Request, since looking through the code, thats all thats actually being used

reggiemcdonald: [#4162](https://github.com/astaxie/beego/issues/4162)

reggiemcdonald: The beego owners got back to me. An interceptor API will be released sometime soon that would make getting the template path more direct. They recently merged an example of using the new interface for opentracing [here](https://github.com/astaxie/beego/pull/4130).

Downside of the new approach is that it wouldn't let us use the existing net/http instrumentation package.

Did you want me to close this PR and start working on an implementation that uses the interceptor?

Aneurysm9: It looks like the `FilterChain` implementation described there is not yet available in a released version of beego.  Either approach is probably fine, but to rewrite using `FilterChain` we'd need to hold off shipping this instrumentation until after the next beego release.

instrumentation/github.com/astaxie/beego/go.mod
```diff
@@ -0,0 +1,19 @@
+module go.opentelemetry.io/contrib/instrumentation/github.com/astaxie/beego
+
+go 1.14
+
+require (
+	github.com/astaxie/beego v1.12.2
+	github.com/stretchr/testify v1.6.1
+	go.opentelemetry.io/contrib v0.10.0
+	go.opentelemetry.io/contrib/instrumentation/net/http v0.0.0-20200806162034-3fc65dc78f63
```
MrAlias, line 9:
```suggestion
	go.opentelemetry.io/contrib v0.10.1
	go.opentelemetry.io/contrib/instrumentation/net/http v0.10.1
```

instrumentation/github.com/astaxie/beego/beego.go
```diff
@@ -0,0 +1,152 @@
...
+	"github.com/astaxie/beego"
+)
+
+// OTelBeegoHandler implements the http.Handler interface and provides
+// trace and metrics to beego web apps.
+type OTelBeegoHandler struct {
+	http.Handler
+}
+
+func (o *OTelBeegoHandler) ServeHTTP(rr http.ResponseWriter, req *http.Request) {
```
MrAlias, line 35: Missing documentation. Possibly ...

```suggestion
// ServerHTTP calls the configured handler to serve HTTP for req to rr.
func (o *OTelBeegoHandler) ServeHTTP(rr http.ResponseWriter, req *http.Request) {
```

instrumentation/github.com/astaxie/beego/beego.go
MrAlias: > Either approach is probably fine, but to rewrite using FilterChain we'd need to hold off shipping this instrumentation until after the next beego release.

Let's go with the current approach for now and we can file a ticket that will track this as a future to possibly restructure. Thoughts?

reggiemcdonald: Yea I think thats fair. It doesn't seem like the `FilterChain` provides any functionality above the current implementation aside from making it easier to get the beego context

Aneurysm9:
@reggiemcdonald can you add a `CHANGELOG.md` entry for this?  I think this is otherwise ready to merge.

instrumentation/github.com/astaxie/beego/beego.go
MrAlias: #234

PR #211, opened 2020-08-12
Add convertToTimeseries for Sum, LastValue, and MinMaxSumCount 

connorlindsey:
This PR adds tests and code to convert Sum, LastValue, and MinMaxSumCount aggregated records to TimeSeries. It also finishes the Export function by creating and sending a request to the Cortex endpoint.

### DIFF
--- a/exporters/metric/cortex/cortex.go
+++ b/exporters/metric/cortex/cortex.go
@@ -18,6 +18,7 @@ import (
 	"bytes"
 	"context"
 	"fmt"
+	"log"
 	"net/http"
 
 	"github.com/gogo/protobuf/proto"
@@ -25,8 +26,10 @@ import (
 	"github.com/prometheus/prometheus/prompb"
 
 	"go.opentelemetry.io/otel/api/global"
+	"go.opentelemetry.io/otel/api/label"
 	apimetric "go.opentelemetry.io/otel/api/metric"
 	"go.opentelemetry.io/otel/sdk/export/metric"
+	export "go.opentelemetry.io/otel/sdk/export/metric"
 	"go.opentelemetry.io/otel/sdk/export/metric/aggregation"
 	"go.opentelemetry.io/otel/sdk/metric/controller/push"
 	"go.opentelemetry.io/otel/sdk/metric/selector/simple"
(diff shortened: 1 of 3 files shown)


exporters/metric/cortex/cortex.go
```diff
@@ -86,11 +109,228 @@ func InstallNewPipeline(config Config, options ...push.Option) (*push.Controller
...
+
+			timeSeries = append(timeSeries, tSeries...)
+		} else if lastValue, ok := agg.(aggregation.LastValue); ok {
+			tSeries, err := convertFromLastValue(record, lastValue)
+			if err != nil {
+				return err
+			}
+
+			timeSeries = append(timeSeries, tSeries)
+		}
```
jmacd, line 147: Should we catch this case where no match was found above?

exporters/metric/cortex/cortex.go
```diff
@@ -86,11 +109,228 @@ func InstallNewPipeline(config Config, options ...push.Option) (*push.Controller
...
+	name := sanitize(record.Descriptor().Name() + "_min")
+	labels := createLabelSet(record, "__name__", name)
+
+	// Create TimeSeries
+	minTimeSeries := &prompb.TimeSeries{
+		Samples: []prompb.Sample{minSample},
+		Labels:  labels,
+	}
+
+	// Convert Max
```
jmacd, line 235: I wonder if you'd consider refactoring this so the three blocks of nearly-identical code (for Min, Max, Count)?

exporters/metric/cortex/testutil_test.go
```diff
@@ -0,0 +1,81 @@
...
+
+	"go.opentelemetry.io/otel/api/metric"
+	export "go.opentelemetry.io/otel/sdk/export/metric"
+	"go.opentelemetry.io/otel/sdk/export/metric/metrictest"
+	"go.opentelemetry.io/otel/sdk/metric/aggregator/aggregatortest"
+	"go.opentelemetry.io/otel/sdk/metric/aggregator/lastvalue"
+	"go.opentelemetry.io/otel/sdk/metric/aggregator/minmaxsumcount"
+	"go.opentelemetry.io/otel/sdk/metric/aggregator/sum"
+)
+
```
jmacd, line 30: There are some improvements in the metric test helpers coming-- I'm glad to see you isolate this stuff.

exporters/metric/cortex/cortex.go
connorlindsey: What would the expected behavior be if no aggregations are matched? I assume that means the user provided their own and we should log that the record wasn't converted?

Also to note, I'll be adding in distribution and histogram in another PR soon.

exporters/metric/cortex/cortex.go
connorlindsey: Yeah, I think creating a helper function like createTimeSeries would be helpful. Maybe pass in the value I want and the labels. I'll make sure I can use it with histogram and distribution buckets/quantiles as well since that'll be pretty repetitive too.

Issue #212, opened 2020-08-13
Release v0.10.1


lizthegrey:
It'd be helpful to let -go-contrib evolve a little faster than -go, so that for instance people can migrate in advance to the moved othttp instrumentation instead of having things be temporarily broken between v0.11 of -go and -go-contrib being released. Thoughts?

MrAlias:
Seems reasonable to me :+1:

lizthegrey:
Not actually blocking for me, it turns out https://pkg.go.dev/mod/go.opentelemetry.io/contrib/instrumentation/net/http@v0.0.0-20200813041938-b948cd370862 will work fine, but ideally still it'd be tagged with v0.10.1

MrAlias:
Looking to merge PRs that are ready and then get this out.

PR #210, opened 2020-08-12
Cortex Exporter Send Pipeline

ercl:
This PR implements the pipeline for sending data to Cortex. It includes functions for converting a slice of TimeSeries to a Snappy-compressed protocol buffer message, building a HTTP request, and sending the request to Cortex. Authentication will be added in a separate PR.

The design doc for the project can be found here: https://github.com/open-telemetry/opentelemetry-go-contrib/pull/188.

cc @alolita @connorlindsey

PR #202, opened 2020-08-10
Cortex Exporter Project setup

connorlindsey:
This PR begins the implementation for a Cortex Exporter. It sets up the basic file structure, stubs out the Exporter interface, and copies the sanitize function and tests used in other Prometheus-based exporters. 

The design doc for this project can be found [here](https://github.com/open-telemetry/opentelemetry-go-contrib/pull/188)

cc @alolita @ercl

Issue #208, opened 2020-08-12
Remove replace on go.mod at next release


reggiemcdonald:
The `net/http` instrumentation, and the (not yet merged) `astaxie/beego` instrumentation both use a `replace` in the go.mod to use the mock meter for unit tests. This is because `internal/metric` is not in the most recent release of contrib. 

This ticket is to remove the replace from the go mods at the next release.

evantorrie:
As you observed, the `replace go.opentelemetry.io/contrib => ../../../..` directive in a submodule's `go.mod` is necessary only when the submodule's code depends on a new addition/change in the top-level `go.opentelemetry.io/contrib` module. 

Admittedly, that is likely to become more rare as the code base stabilizes, but https://github.com/open-telemetry/opentelemetry-go-contrib/issues/214#issuecomment-673747126 implies that this will indeed happen again soon with its suggestion of a constant `version` introduced into the "top-level repository".

Perhaps implementation of this ticket should be postponed to a GA release, and from then on, replaces directives introduced only when required for a particular module.

MrAlias:
Closing, stale.

PR #216, opened 2020-08-13
Release v0.10.1

MrAlias:
## 0.10.1 - 2020-08-13

This release includes instrumentation moved from the main `go.opentelemetry.io/otel` project for the `google.golang.org/grpc`, `net/http`, and `net/http/httptrace` packages.

### Added

- The `go.opentelemetry.io/contrib/instrumentation/google.golang.org/grpc` module has been added to replace the instrumentation that had previoiusly existed in the `go.opentelemetry.io/otel/instrumentation/grpctrace` package. (#189)
- Instrumentation for the stdlib `net/http` and `net/http/httptrace` packages. (#190)
- Initial Cortex exporter. (#202, #205, #210, #211, #215)

### Fixed

- Bump google.golang.org/grpc from 1.30.0 to 1.31.0. (#166)
- Bump go.mongodb.org/mongo-driver from 1.3.5 to 1.4.0 in /instrumentation/go.mongodb.org/mongo-driver. (#170)
- Bump google.golang.org/grpc in /instrumentation/github.com/gin-gonic/gin. (#173)
- Bump google.golang.org/grpc in /instrumentation/github.com/labstack/echo. (#176)
- Bump google.golang.org/grpc from 1.30.0 to 1.31.0 in /instrumentation/github.com/Shopify/sarama. (#179)
- Bump cloud.google.com/go from 0.61.0 to 0.63.0 in /detectors/gcp. (#181, #199)
- Bump github.com/aws/aws-sdk-go from 1.33.15 to 1.34.1 in /detectors/aws. (#184, #192, #193, #198, #201, #203)
- Bump github.com/golangci/golangci-lint from 1.29.0 to 1.30.0 in /tools. (#186)
- Setup CI to run tests that require external resources (Cassandra and MongoDB). (#191)
- Bump github.com/Shopify/sarama from 1.26.4 to 1.27.0 in /instrumentation/github.com/Shopify/sarama. (#206)

Resolves https://github.com/open-telemetry/opentelemetry-go-contrib/issues/212

### DIFF
--- a/instrumentation/net/http/go.mod
+++ b/instrumentation/net/http/go.mod
@@ -7,7 +7,7 @@ replace go.opentelemetry.io/contrib => ../../..
 require (
 	github.com/felixge/httpsnoop v1.0.1
 	github.com/stretchr/testify v1.6.1
-	go.opentelemetry.io/contrib v0.10.0
+	go.opentelemetry.io/contrib v0.10.1
 	go.opentelemetry.io/otel v0.10.0
 	google.golang.org/grpc v1.31.0
 )
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -8,10 +8,26 @@ This project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0.htm
 
 ## [Unreleased]
 
+## [0.10.1] - 2020-08-13
+
 ### Added
 
 - The `go.opentelemetry.io/contrib/instrumentation/google.golang.org/grpc` module has been added to replace the instrumentation that had previoiusly existed in the `go.opentelemetry.io/otel/instrumentation/grpctrace` package. (#189)
 - Instrumentation for the stdlib `net/http` and `net/http/httptrace` packages. (#190)
+- Initial Cortex exporter. (#202, #205, #210, #211)
+
+### Fixed
+
+- Bump google.golang.org/grpc from 1.30.0 to 1.31.0. (#166)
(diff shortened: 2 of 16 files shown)


instrumentation/net/http/go.mod
```diff
@@ -7,7 +7,7 @@ replace go.opentelemetry.io/contrib => ../../..
 require (
 	github.com/felixge/httpsnoop v1.0.1
 	github.com/stretchr/testify v1.6.1
-	go.opentelemetry.io/contrib v0.10.0
+	go.opentelemetry.io/contrib v0.10.1
```
Aneurysm9, line 10: Can you remove the `replace` directive in this file?  See #208.

MrAlias: Should this be done for all the `go.mod` files? It seems like they all have `replace` directives for the contrib repository.

IIRC this was something we did to bootstrap, but I also wondered if we should keep them if we want to change local dependencies in the repo and refer to them. I guess we could add them back if that is the case, right?

Also, will it not break things if I remove this `replace` before I actually do the tag and release? I.e. there won't actually be a `v0.10.1` release yet.

reggiemcdonald: Maybe this is a lapse in understanding how `go get` works, but wont the `replace` directive break things if a user calls `go get` on the particular integration?

But I agree that making the change before will temporarily break things

Aneurysm9: All good points.  Perhaps this is something we can discuss at the SIG meeting today.  We should probably  be safe and not change anything until after a release.

PR #223, opened 2020-08-14
Prototype implementation of metric config experimental specification

wtong98:
### Overview
This implementation relates to the experimental metric config specification, found [here](https://github.com/open-telemetry/opentelemetry-specification/blob/master/experimental/metrics/config-service.md).

An [example](https://github.com/vmingchen/opentelemetry-go-contrib/tree/master/sdk/dynamicconfig/example) is included to show how it works.

It's counterpart on the contrib-collector is available [here](https://github.com/open-telemetry/opentelemetry-collector-contrib/pull/712)

This PR seeks to present a functioning prototype of the experimental metric configuration spec, and is not intended to be finalized code ready for merging. We would love to get your thoughts on the large-scale architecture of the change, and if it seems reasonable, we will proceed to breaking it into smaller PRs that can be individually scrutinized and merged.

### About the code
This PR is a *monster*. Here's an overview of the code structure to make understanding our implementation easier:
- `sdk/dynamicconfig`: houses the relevant implementation
  - `example`: functioning example of per-metric configuration with our custom pusher
  - `internal`: various internal resources used by our implementation
    - `metricpattern`: bare-bones implementation of our pattern-matching "language" for specifying metric names
    - `proto`: generated proto files, copied wholesale with minimal change. This is a short-term solution. When OTel proto files are centralized again, the implementation should be updated to use the centralized proto's.
    - `transform`: internal files copied from the SDK used to transform an SDK-specific representation of proto objects into protobuf-ready structs. This is a short-term solution. When/if `transform` is exposed in a public API, we should of course switch to that.
  - `metric`: houses the implementation of the dynamic, per-metric push controller
    - `controller`: controller-specific files
      - `push`: implementation of the push controller
      - `remote`: utilities to communicate with a configuration service
    - `sdk.go`: identical to the SDK's `sdk.go`, with a small change made to `Collect()` so that it accepts a `rule`, deciding which metric should be collected based on their name

Note: much of this code overlaps with code from the SDK. The PR looks like an 8000+ line change, but perhaps 6000+ of those lines are copied code, either from generated proto files or internal code from the SDK.

Overall, per-metric configuration was achieved by using a filter, which we call a `rule`, that decides which metrics should be collected based on their name. The push controller organizes the creation of these `rule`s based on collection schedules received from the configuration service. For more information about this process, please refer to the specification and collector-contrib, linked above.


MrAlias:
Thanks for tackling this, it looks to have been a lot of work. I have tried to dig into this multiple times but due to the enormity of changes it includes I have not made any real headway.

@wtong98 I'm wondering if you can provide some context around what the goals of this PR are.

Is this something that can be split up into component changes with this acting as a prototype overview of what the ultimate goal is?  Is this something that should be prioritized after GA?

wtong98:
@MrAlias Thanks for taking a look! I agree it's a monster PR.

The overall goal is to present an example, working implementation of the experimental specification for dynamic, per-metric scheduling. The easiest way to get it up and running would be to clone the respective forks on [go-contrib](https://github.com/vmingchen/opentelemetry-go-contrib) and [collector-contrib](https://github.com/vmingchen/opentelemetry-collector-contrib), then follow the [example](https://github.com/vmingchen/opentelemetry-go-contrib/tree/master/sdk/dynamicconfig/example). 

In this way, it is indeed meant to be a full picture of what the system could look like. If it seems to look reasonable on a large scale, we can then proceed to break it down into more manageable PRs that can be scrutinized individually, then merged. Does that sound like a reasonable plan? I'll also expand the PR description to include a detailed breakdown of the code, to make digesting it easier.

This change is intended to be prioritized **after GA**.

Note: my internship has concluded and the school year is about to start, so I can't promise full attention to this project, but I will keep an eye on things. If something more immediate comes up, my former supervisors (@vmingchen, @jdmontana) are transitioning into a more active role in this work.

MrAlias:
Closing due to staleness and the complexity.

Issue #213, opened 2020-08-13
Unify instrumentation *Provider options


MrAlias:
Currently there are many instrumentation packages that are directly configurable with a `Tracer` or `Meter` to use in the telemetry they gather.

- https://github.com/open-telemetry/opentelemetry-go-contrib/blob/32a76fe5fd2bafc439463d61320152a2987cc881/instrumentation/go.mongodb.org/mongo-driver/config.go#L50
- https://github.com/open-telemetry/opentelemetry-go-contrib/blob/370078e3876afd8713a3e25076fe1e71f1f21587/instrumentation/github.com/gin-gonic/gin/option.go#L36
- https://github.com/open-telemetry/opentelemetry-go-contrib/blob/370078e3876afd8713a3e25076fe1e71f1f21587/instrumentation/github.com/gorilla/mux/config.go#L35
- https://github.com/open-telemetry/opentelemetry-go-contrib/blob/370078e3876afd8713a3e25076fe1e71f1f21587/instrumentation/github.com/labstack/echo/config.go#L35
- https://github.com/open-telemetry/opentelemetry-go-contrib/blob/19be05c53f11474ac0d1bf248b4a48a4c75100e0/instrumentation/github.com/emicklei/go-restful/config.go#L35
- https://github.com/open-telemetry/opentelemetry-go-contrib/blob/70957fc44e1fe809e2b961a43eeea390fa247038/instrumentation/github.com/Shopify/sarama/option.go#L55
- https://github.com/open-telemetry/opentelemetry-go-contrib/blob/7066c0829eab0b7992fe4b9f2ada1d8949d7c874/instrumentation/github.com/gocql/gocql/config.go#L84
- https://github.com/open-telemetry/opentelemetry-go-contrib/blob/3fc65dc78f63bed02e186a9c22b7eb32b565fc56/instrumentation/net/http/config.go#L62
- https://github.com/open-telemetry/opentelemetry-go-contrib/blob/3fc65dc78f63bed02e186a9c22b7eb32b565fc56/instrumentation/net/http/config.go#L70

They should not be configurable to do this. It is the instrumentations job to define the `Tracer` or `Meter` they use from a `*Provider`.

[...]

Issue #214, opened 2020-08-13
Instrumentation needs to include a version.


MrAlias:
`Tracer`s and `Meter`s both take instrumentation versions as an option. All instrumentation in this repo should included that. This will help in debugging and when a Views API is created could be used to disable or unify different versions of the same instrumentation running on the same server.

- A version should be defined for all instrumentation in this repository.
- All instrumentation needs to be updated to use this version as a `Meter` and `Tracer` instrumentation version.
- The [instrumentation guidelines](https://github.com/open-telemetry/opentelemetry-go-contrib/blob/master/instrumentation/README.md) need to be updated to specify that a version should be set for instrumentation.

### Instrumentation Punch List

- [x] instrumentation/host
- [x] instrumentation/google.golang.org/grpc/otelgrpc
- [x] instrumentation/go.mongodb.org/mongo-driver/mongo/otelmongo
- [x] instrumentation/net/http/otelhttp
- [x] instrumentation/net/http/httptrace/otelhttptrace
- [x] instrumentation/runtime
- [x] instrumentation/github.com/labstack/echo/otelecho
- [x] instrumentation/github.com/bradfitz/gomemcache/memcache/otelmemcache
- [x] instrumentation/github.com/gin-gonic/gin/otelgin
- [x] instrumentation/github.com/gocql/gocql/otelgocql
- [x] instrumentation/github.com/emicklei/go-restful/otelrestful
- [x] instrumentation/github.com/Shopify/sarama/otelsarama
- [x] instrumentation/github.com/gorilla/mux/otelmux
- [x] instrumentation/github.com/astaxie/beego/otelbeego
- [x] instrumentation/gopkg.in/macaron.v1/otelmacaron

MrAlias:
Reference example: https://github.com/open-telemetry/opentelemetry-go-contrib/pull/204/files#r470091488

MrAlias:
From the SIG meeting:

All instrumentation in this repo MUST share the same version and it needs to match repository version. This will unify this repos approach and avoid confusion of how things are versioned. It will also alleviate needing to track which versions are compatible with version of `otel`.

This also means that a top-level repository version `const` should be created so all instrumentation can reference it.

reggiemcdonald:
I could help out with adding versions to the existing instrumentation packages, if help is still needed on this?

MrAlias:
This is likely done, an audit is needed to verify this.

matej-g:
[...]

PR #224, opened 2020-08-14
Create runtime.Config struct with metric.Provider and WithMinimumReadMemStatsInterval() configuration options

jmacd:
Part of #213.


instrumentation/runtime/runtime.go
```diff
@@ -26,18 +26,42 @@ import (
...
 // Runtime reports the work-in-progress conventional runtime metrics specified by OpenTelemetry
 type runtime struct {
-	meter    metric.Meter
-	interval time.Duration
+	meter              metric.Meter
+	minGCStatsInterval time.Duration
+}
+
+// Option supports configuring optional settings for runtime metrics.
+type Option func(*runtime)
```
evantorrie, line 34: Should we standardize the `Option/Apply` interface approach in `opentelemetry-go-contrib` the same as in `opentelemetry-go`, i.e. the approach [described here](https://github.com/open-telemetry/opentelemetry-go/issues/536)?

instrumentation/runtime/runtime.go
```diff
@@ -26,18 +26,42 @@ import (
...
 // If the metric SDK attempts to observe MemStats-derived instruments more frequently than the
 // interval, a cached value will be used.
-func Start(meter metric.Meter, interval time.Duration) error {
+func Start(provider metric.Provider, opts ...Option) error {
 	r := &runtime{
-		meter:    meter,
-		interval: interval,
+		// TODO: How to set the instrumentation library version?
+		meter:              provider.Meter("runtime"),
+		minGCStatsInterval: DefaultMinimumGCStatsInterval,
```
evantorrie, line 61: Per https://github.com/open-telemetry/opentelemetry-go-contrib/issues/214 , I believe decision was to use a common instrumentation version across all implementations in this repo, but maybe that should be delayed to a later PR.

instrumentation/runtime/go.mod
```diff
@@ -6,6 +6,7 @@ replace go.opentelemetry.io/contrib => ../..
 
 require (
 	github.com/stretchr/testify v1.6.1
+	go.opentelemetry.io/contrib v0.0.0-00010101000000-000000000000
```
MrAlias, line 9:
```suggestion
	go.opentelemetry.io/contrib v0.10.1
```

instrumentation/runtime/runtime.go
```diff
@@ -26,18 +26,42 @@ import (
...
+		}
+	}
 }
 
 // New returns Runtime, a structure for reporting Go runtime metrics
 // interval is used to limit how often to invoke Go runtime.ReadMemStats() to obtain metric data.
 // If the metric SDK attempts to observe MemStats-derived instruments more frequently than the
 // interval, a cached value will be used.
-func Start(meter metric.Meter, interval time.Duration) error {
+func Start(provider metric.Provider, opts ...Option) error {
```
MrAlias, line 57: The `Provider` should be passed as an option with the default, if none are passed, being the `global.MeterProvider` (https://github.com/open-telemetry/opentelemetry-go-contrib/issues/213).

instrumentation/runtime/doc.go
```diff
@@ -1,3 +1,17 @@
+// Copyright The OpenTelemetry Authors
```
MrAlias, line 1: Yikes, we need to add a similar check to the build to catch these...

MrAlias: https://github.com/open-telemetry/opentelemetry-go-contrib/issues/227

instrumentation/runtime/go.mod
jmacd: Done.

instrumentation/runtime/runtime.go
jmacd: Fixed.

instrumentation/runtime/runtime.go
```diff
@@ -26,18 +26,42 @@ import (
...
+// frequently.  `d` values less than 0 will be disregarded silently.
+func WithMinimumGCStatsInterval(d time.Duration) Option {
+	return func(r *runtime) {
+		if d >= 0 {
+			r.minGCStatsInterval = d
+		}
+	}
 }
 
 // New returns Runtime, a structure for reporting Go runtime metrics
```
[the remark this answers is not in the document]
jmacd: Updated.

instrumentation/runtime/runtime.go
jmacd: Fixed.

instrumentation/runtime/runtime.go
jmacd: Thanks for #225.

instrumentation/runtime/runtime.go
```diff
@@ -20,48 +20,100 @@ import (
...
+	if c.MeterProvider == nil {
+		c.MeterProvider = global.MeterProvider()
+	}
+	r := &runtime{
+		meter: c.MeterProvider.Meter(
+			// TODO: should library names be qualified?
+			// e.g., contrib/runtime?
+			"runtime",
+			// TODO(#225): set the instrumentation library version
+			// metric.WithInstrumentationVersion(contrib.SemVersion())
```
evantorrie, line 114: #225 is merged to mainline, so should be able to uncomment this now.

jmacd: Done.

jmacd:
From a terminology perspective, there's one lingering concern.  We've written `WithMeterProvider()` for a class named `metric.Provider` over and over again. I wonder if we should rename `metric.Provider` to `metric.MeterProvider`, but such a rename might cause more trouble than it's worth.

MrAlias:
> From a terminology perspective, there's one lingering concern. We've written `WithMeterProvider()` for a class named `metric.Provider` over and over again. I wonder if we should rename `metric.Provider` to `metric.MeterProvider`, but such a rename might cause more trouble than it's worth.

I've been passively thinking of needing this kind of rename if we wanted to move the provider to the top level (alongside a `TraceProvider`). Sounds reasonable to me.

jmacd:
> I've been passively thinking of needing this kind of rename if we wanted to move the provider to the top level (alongside a TraceProvider).

 I like this idea!

instrumentation/runtime/runtime.go
```diff
@@ -20,24 +20,100 @@ import (
...
+	if c.MeterProvider == nil {
+		c.MeterProvider = global.MeterProvider()
+	}
 	r := &runtime{
-		meter:    meter,
-		interval: interval,
+		meter: c.MeterProvider.Meter(
+			// TODO: should library names be qualified?
+			// e.g., contrib/runtime?
+			"runtime",
```
MrAlias, line 113: Yeah, this should be fully qualified:

```suggestion
			// TODO: should library names be qualified?
			// e.g., contrib/runtime?
			"go.opentelemetry.io/contrib/instrumentation/runtime",
```

We can address this in a subsequent PR though.

MrAlias: https://github.com/open-telemetry/opentelemetry-go-contrib/issues/240

PR #226, opened 2020-08-15
Use top-level common SemVersion() in creation of versioned tracer for bradfitz/gomemcache

evantorrie:
This depends on #225, which would need to be merged prior to this.

Also noticed this module was missing from the `dependabot.yml` checks.

### DIFF
--- a/contrib.go
+++ b/contrib.go
@@ -1,0 +1,27 @@
+// Copyright The OpenTelemetry Authors
+//
+// Licensed under the Apache License, Version 2.0 (the "License");
+// you may not use this file except in compliance with the License.
+// You may obtain a copy of the License at
+//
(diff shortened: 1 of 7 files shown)


contrib.go
```diff
@@ -0,0 +1,27 @@
...
+// See the License for the specific language governing permissions and
+// limitations under the License.
+
+// Package contrib contains common values used across all
+// instrumentation, exporter, and detector contributions
+package contrib // import "go.opentelemetry.io/contrib"
+
+// Version is the current release version of OpenTelemetry Contrib in use.
+func Version() string {
+	return "0.10.1"
```
jmacd, line 21: Maybe add a comment that you're updating this via `sed` below. 😁

evantorrie: Done in #225

PR #231, opened 2020-08-18
Host metrics instrumentation

jmacd:
This instrumentation module implements metrics that the OTel Collector would also generate via its "hostmetrics" receiver. This covers only a minimum amount of information that is generally useful for monitoring, and leaves open questions about how to handle OS-specific metrics (esp. memory) and whether to encode per-CPU and per-NIC measurements for CPU and Network. See the new TODOs.

While working on this I filed several related specification questions:

https://github.com/open-telemetry/opentelemetry-specification/issues/818
https://github.com/open-telemetry/opentelemetry-specification/issues/819

This raises a very serious potential to share code with the OTel Collector's hostmetrics receiver, as this code is more thorough and handles OS-specific fields (with mock tests). On the other hand, factoring out that code module would be easier if we could migrate to using the real OTel-Go Metric SDK to generate these measurements. This is not out of the question, but too much work for this little PR, so I'll suggest that we start small with just a few metrics and evaluate this possibility later.


instrumentation/host/host.go
```diff
@@ -0,0 +1,279 @@
...
+		MeterProvider: global.MeterProvider(),
+	}
+	for _, opt := range opts {
+		opt.ApplyHost(&c)
+	}
+	return c
+}
+
+// Start initializes reporting of host metrics using the supplied Config.
+func Start(c Config) error {
```
MrAlias, line 99: nit, but I thought we creating by passing the options to the instantiation function: https://github.com/open-telemetry/opentelemetry-go/blob/master/CONTRIBUTING.md#instantiation

instrumentation/host/host.go
```diff
@@ -0,0 +1,279 @@
...
+	if c.MeterProvider == nil {
+		c.MeterProvider = global.MeterProvider()
+	}
```
MrAlias, line 102: This default configuration should live in the `Configure` function.

instrumentation/host/host.go
```diff
@@ -0,0 +1,279 @@
...
+	h := &host{
+		meter: c.MeterProvider.Meter(
+			"host",
```
MrAlias, line 105: Instrumentation needs to have the fully-qualified package name.

```suggestion
			"go.opentelemetry.io/contrib/instrumentation/host",
```

instrumentation/host/host.go
```diff
@@ -0,0 +1,279 @@
...
+		defer lock.Unlock()
+
+		// This follows the OpenTelemetry Collector's "hostmetrics"
+		// receiver/hostmetricsreceiver/internal/scraper/processscraper
+		// measures User and System IOwait time.
+		// TODO: the Collector has per-OS compilation modules to support
+		// specific metrics that are not universal.
+		processTimes, err := proc.TimesWithContext(ctx)
+		if err != nil {
+			global.Handler().Handle(err)
```
MrAlias, line 148: Same as [this](https://github.com/open-telemetry/opentelemetry-go/blob/fd61d2edece005195116e6b9218801a460a3868d/api/global/handler.go#L90)

```suggestion
			global.Handle(err)
```

instrumentation/host/host.go
```diff
@@ -0,0 +1,279 @@
...
+		hostTime := hostTimeSlice[0]
+		result.Observe(LabelCPUTimeUser, hostCPUTime.Observation(hostTime.User))
+		result.Observe(LabelCPUTimeSystem, hostCPUTime.Observation(hostTime.System))
+
+		// TODO: "other" is a placeholder for actually dealing
+		// with these states.  Do users actually want this
+		// (unconditionally)?  How should we handle "iowait"
+		// if not all systems expose it?  Should we break
+		// these down by CPU?  If so, are users could to want
+		// to aggregate in-process?
```
MrAlias, line 192: I think this is the correct initial approach :+1: 

If users want to have these exposed we can possibly add this to the `Config` as a something they can specify in the future.

```suggestion
		// TODO: "other" is a placeholder for actually dealing
		// with these states.  Do users actually want this
		// (unconditionally)?  How should we handle "iowait"
		// if not all systems expose it?  Should we break
		// these down by CPU?  If so, are users going to want
		// to aggregate in-process?
```

instrumentation/host/host_test.go
```diff
@@ -0,0 +1,256 @@
...
+	require.GreaterOrEqual(t, hostAfter[0].System, hostSystem)
+	// Ranges are not empty
+	require.NotEqual(t, hostAfter[0].System, hostBefore[0].System)
+	require.NotEqual(t, hostAfter[0].User, hostBefore[0].User)
+
+	// TODO: We are not testing host "Other" nor "Idle" and
+	// generally the specification hasn't been finalized, so
+	// there's more to do.  Moreover, "Other" is not portable and
+	// "Idle" may not advance on a fully loaded machine => both
+	// are difficult to test.
```
MrAlias, line 135: Do we have an issue to capture this?

jmacd: #244

instrumentation/host/host.go
jmacd: Also note: the recently-merged code in ../runtime has the same pattern.  I fixed this in both places.

instrumentation/host/host.go
MrAlias: Ah, gotcha. Makes sense :+1:

instrumentation/host/go.mod
```diff
@@ -0,0 +1,16 @@
+module go.opentelemetry.io/contrib/instrumentation/host
+
+go 1.14
+
+replace go.opentelemetry.io/contrib => ../..
+
+require (
+	github.com/StackExchange/wmi v0.0.0-20190523213315-cbe66965904d // indirect
```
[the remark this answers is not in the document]
MrAlias: Not sure if I'm bad at GitHub or maybe you forgot to push, or maybe it is actually needed, but I still see this requirement.

Going to move on as I don't think it's a bit deal.

jmacd:
It turns out that `gopsutil` depends on this `wmi` package for windows process state.

PR #236, opened 2020-08-18
Update build to use Go 1.15

### LINKED ISSUE #233: Update build to use Go 1.15
Similar to [this](https://github.com/open-telemetry/opentelemetry-go/pull/1064), update the build to stop building Go 1.13 and start using 1.15. This will likely require similar code fixes (see #232)

ilmanzo:
fix #233

### DIFF
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -2,7 +2,7 @@ version: 2.1
 jobs:
   build:
     docker:
-      - image: circleci/golang:1.13
+      - image: circleci/golang:1.15
 
     environment:
       TEST_RESULTS: /tmp/test-results # path to where test results will be saved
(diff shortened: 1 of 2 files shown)


.circleci/config.yml
```diff
@@ -2,7 +2,7 @@ version: 2.1
 jobs:
   build:
     docker:
-      - image: circleci/golang:1.13
+      - image: circleci/golang:1.15
```
MrAlias, line 5: We should update this to build the current and most recent version of Go similar to the `otel` repo. I'll open a ticket for this.

MrAlias: https://github.com/open-telemetry/opentelemetry-go-contrib/issues/239

ilmanzo: you mean `golang:latest` ? Don't know, maybe it's better to have always a precise reference on which go version you are using

MrAlias: No, I was referring to the fact that this should also be building for Go 1.14. I agree direct specifying of the version is preferred.

PR #237, opened 2020-08-18
Cortex: Add support for histogram and distribution

connorlindsey:
This PR adds support for histogram and distribution aggregations in the Cortex Exporter by adding functionality in `cortex.go`. I also refactored tests in `cortex_test.go`. To support buckets and quantiles, I also added two fields to `config.go`.

There are a few other minor changes like adding a link to Viper in the README, switching to a `NewWithHistogramDistribution` Processor, and modifying the conversion logic (control flow to select which aggregation needs to be converted).


exporters/metric/cortex/cortex.go
```diff
@@ -110,6 +111,8 @@ func InstallNewPipeline(config Config, options ...push.Option) (*push.Controller
 }
 
 // ConvertToTimeSeries converts a CheckpointSet to a slice of TimeSeries pointers
+// Based on the aggregation type, ConvertToTimeSeries will call helper function like
```
MrAlias, line 114:
```suggestion
// Based on the aggregation type, ConvertToTimeSeries will call helper functions like
```

exporters/metric/cortex/cortex.go
```diff
@@ -119,31 +122,52 @@ func (e *Exporter) ConvertToTimeSeries(checkpointSet export.CheckpointSet) ([]*p
...
-		// Check if aggregation has Sum value
-		if sum, ok := agg.(aggregation.Sum); ok {
+		// Check if aggregation has Histogram value
+		if histogram, ok := agg.(aggregation.Histogram); ok {
+			tSeries, err := convertFromHistogram(record, histogram)
+			if err != nil {
+				return err
+			}
+			timeSeries = append(timeSeries, tSeries...)
+			// Check if aggregation has sum value
```
MrAlias, line 132:
```suggestion
		// Check if aggregation has sum value
```

exporters/metric/cortex/cortex.go
```diff
@@ -119,31 +122,52 @@ func (e *Exporter) ConvertToTimeSeries(checkpointSet export.CheckpointSet) ([]*p
...
+				if distribution, ok := agg.(aggregation.Distribution); ok && len(e.config.Quantiles) != 0 {
+					tSeries, err := convertFromDistribution(record, distribution, e.config.Quantiles)
+					if err != nil {
+						return err
+					}
+
+					timeSeries = append(timeSeries, tSeries...)
+				}
+			}
+			// Check if aggregation has lastValue
```
MrAlias, line 160:
```suggestion
		// Check if aggregation has lastValue
```

exporters/metric/cortex/cortex.go
```diff
@@ -157,28 +181,34 @@ func (e *Exporter) ConvertToTimeSeries(checkpointSet export.CheckpointSet) ([]*p
...
+	// Create TimeSeries
 	name := sanitize(record.Descriptor().Name())
-	labels := createLabelSet(record, "__name__", name)
-
-	// Create TimeSeries and return
-	tSeries := &prompb.TimeSeries{
-		Samples: []prompb.Sample{sample},
-		Labels:  labels,
-	}
+	// Note: Cortex requires the name label to be in the format "__name__".
```
MrAlias, line 209: This duplicated pattern indicates that a refactor might be beneficial here. If `createLabelSet` accepted `label.KeyValue` pairs instead of runtime-guaranteed string pairs it would move the pair checking part of that code to be performed by the compiler and allow this to be templated with a common `label.Key`.

exporters/metric/cortex/cortex.go
```diff
@@ -264,21 +259,95 @@ func convertFromMinMaxSumCount(record metric.Record, minMaxSumCount aggregation.
...
+
+		// Create TimeSeries
+		tSeries := createTimeSeries(record, value, "__name__", metricName, "quantile", quantileStr)
+		timeSeries = append(timeSeries, tSeries)
+	}
+
+	return timeSeries, nil
+}
+
+// convertFromHistogram returns
```
MrAlias, line 301: partial comment.

exporters/metric/cortex/cortex.go
```diff
@@ -264,21 +259,95 @@ func convertFromMinMaxSumCount(record metric.Record, minMaxSumCount aggregation.
...
 
+	// Return all timeSeries
 	tSeries := []*prompb.TimeSeries{
 		minTimeSeries, maxTimeSeries, countTimeSeries,
 	}
 
 	return tSeries, nil
 }
 
+// convertFromDistribution returns n TimeSeries for n quantiles in a distributions
```
MrAlias, line 278:
```suggestion
// convertFromDistribution returns len(quantiles) number of TimeSeries in a distribution.
```

exporters/metric/cortex/cortex.go
```diff
@@ -264,21 +259,95 @@ func convertFromMinMaxSumCount(record metric.Record, minMaxSumCount aggregation.
...
+	}
+
+	var totalCount float64
+	// counts maps from the bucket upper-bound to the cumulative count.
+	// The bucket with upper-bound +inf is not included.
+	counts := make(map[float64]float64, len(buckets.Boundaries))
+	for i, boundary := range buckets.Boundaries {
+		// Add bucket count to totalCount and record in map
+		totalCount += buckets.Counts[i]
+		counts[boundary] = totalCount
```
MrAlias, line 327: Is this map used?

connorlindsey: It was, but I refactored it. I'm removing it.

exporters/metric/cortex/cortex.go
connorlindsey: To clarify, instead of strings, `createLabelSet` would take in `...label.KeyValue`? How would you recommend merging the three sets of labels (record labels, resource labels, and extra labels)? The desired behavior is to give precedence to labels the exporter adds (like `"__name__"`, `"le"`, and `"quantile"`), then record labels, then resource labels.

MrAlias: > To clarify, instead of strings, `createLabelSet` would take in `...label.KeyValue`?

Correct.

> How would you recommend merging the three sets of labels (record labels, resource labels, and extra labels)? The desired behavior is to give precedence to labels the exporter adds (like `"__name__"`, `"le"`, and `"quantile"`), then record labels, then resource labels.

Similar to what you are already doing in `createLabelSet` would be my suggestion. You would iterate through the labels in the order of precedence and overwrite any duplicate `Key` appropriately. So, instead of looping through the extra label strings last you would just loop through the extra label `label.KeyValue` last.

This can also be done in a separate PR. I didn't make that clear in my original comment, sorry. If we want to omit it from this PR we can create an issue to track it.

jmacd: There's an example of merging that leaves duplicate keys in place in the dogstatsd exporter, see `func (e *Exporter) AppendTags(rec export.Record, res *resource.Resource, buf *bytes.Buffer)`.

The `api/label` package has a MergeIterator to support merging label sets, which is another option, which is used to give precedence to labels over resources, it could be used here as well.

connorlindsey: That makes sense and would be a cleaner API. I'll file this in another PR.

lizthegrey:
for future, please prefix name of component in [] in PR description.

PR #238, opened 2020-08-18
Cortex example project

connorlindsey:
This PR adds an example project to the Cortex exporter demonstrating basic usage. 

It runs the following services with docker-compose:
- a go app instrumented by the OTel Go SDK
- Cortex
- Grafana

Note: Metrics will not display at the moment. I accidentally was sending metrics with the timestamp in seconds instead of milliseconds. We'll file a fix for that today.


exporters/metric/cortex/example/config.yml
```diff
@@ -0,0 +1,20 @@
...
+# WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+# See the License for the specific language governing permissions and
+# limitations under the License.
+
+url: http://host.docker.internal:9009/api/prom/push
+remote_timeout: 30s
+push_interval: 2s
+name: Test
+headers:
+  X-Scope-OrgID: 5
```
jmacd, line 20: I take it that this X-Scope-OrgID value is necessary and arbitrary, here?

connorlindsey: Yes, it's required in the exporter config but ignored by the "test" Cortex config we're running.

exporters/metric/cortex/example/docker-compose.yml
```diff
@@ -0,0 +1,35 @@
...
+
+services:
+  app:
+    image: golang:1.14-alpine
+    volumes:
+      - .:/app
+    working_dir: /app
+    command: go run main.go
+  cortex:
+    image: quay.io/cortexproject/cortex:v1.3.0-rc.1
```
MrAlias, line 25: Is there a need to use this release-candidate instead of a stable release?

exporters/metric/cortex/example/main.go
```diff
@@ -0,0 +1,68 @@
...
+	)
+
+	counter := metric.Must(meter).NewInt64Counter(
+		"pipeline.counter",
+		metric.WithDescription("Counts things"),
+	)
+	fmt.Println("Success: Created Int64ValueRecorder and Int64Counter instruments")
+
+	fmt.Println("Starting to write data to the instruments")
+	for i := 1; i <= 10; i++ {
```
MrAlias, line 60: Should this be switched to an infinite loop with random value writes? That way the demo could exist as long as the user wanted?

exporters/metric/cortex/example/docker-compose.yml
connorlindsey: There's no stable "latest" or comparable version. That was the latest versioned number, the rest are master-<commit-hash>.

exporters/metric/cortex/example/main.go
connorlindsey: Good point. I think I changed it when we were debugging and forgot to change it back.

exporters/metric/cortex/example/docker-compose.yml
MrAlias: Gotcha. Didn't mean to imply we should track `latest` here, but instead something like `1.2.0` (sorry cannot figure out how to link it):



It's not a huge deal. I was more so curious if that release candidate contained needed functionality.

connorlindsey: Makes sense, I'll double check but I don't think it contains any changes we need. If using 1.2.0 would be more stable, I'll switch to that.

connorlindsey: Actually, we do need 1.3 to use the most recent configuration.

MrAlias:
Looks like maintainer edits is turned off here. If you can sync with master this looks ready to merge.

PR #246, opened 2020-08-18
[cortex] Authentication Implementation and Timestamp Fix

ercl:
This PR implements BasicAuth and bearer token authentication. Additionally, it fixes an issue of timestamps being in seconds instead of milliseconds.


exporters/metric/cortex/cortex.go
```diff
@@ -185,7 +186,7 @@ func (e *Exporter) ConvertToTimeSeries(checkpointSet export.CheckpointSet) ([]*p
 func createTimeSeries(record metric.Record, value apimetric.Number, extraLabels ...string) *prompb.TimeSeries {
 	sample := prompb.Sample{
 		Value:     value.CoerceToFloat64(record.Descriptor().NumberKind()),
-		Timestamp: record.EndTime().Unix(),
+		Timestamp: record.EndTime().UnixNano() / int64(time.Millisecond),
```
MrAlias, line 189: To avoid a Y2K-esq bug and help my physics based brain can we change these to explicitly include the base time unit we are converting from? Almost certain the compile will optimize it away, but it is useful for the reader.

```suggestion
		Timestamp: int64(time.Nanosecond) * record.EndTime().UnixNano() / int64(time.Millisecond),
```

ercl: Definitely see how that's clearer!

exporters/metric/cortex/auth.go
```diff
@@ -0,0 +1,107 @@
...
+	// ErrFailedToReadFile occurs when a password / bearer token file exists, but could
+	// not be read.
+	ErrFailedToReadFile = fmt.Errorf("Failed to read password / bearer token file")
+)
+
+// addBasicAuth sets the Authorization header for basic authentication using a username
+// and a password / password file. To prevent the Exporter from potentially opening a
+// password file on every request by calling this method, the Authorization header is also
+// added to the Config header map.
+func (e *Exporter) addBasicAuth(req *http.Request) error {
```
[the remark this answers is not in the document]
ercl: That makes sense. I've moved it and some of the tests over in the latest commit.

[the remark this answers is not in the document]
ercl: That makes sense. I moved it and some of the tests over in the latest commit.

exporters/metric/cortex/auth.go
```diff
@@ -0,0 +1,86 @@
...
+	"io/ioutil"
+	"net/http"
+)
+
+// ErrFailedToReadFile occurs when a password / bearer token file exists, but could
+// not be read.
+var ErrFailedToReadFile = fmt.Errorf("failed to read password / bearer token file")
+
+// addBasicAuth sets the Authorization header for basic authentication using a username
+// and a password / password file.
```
MrAlias, line 28: We should probably add to this (and the `addBearerTokenAuth` documentation) that no action is taken if authorization already exists or is not configured for the Exporter.

Maybe something like:

```suggestion
// and a password / password file. The header value is not changed if it already exists and no action is taken
// if the Exporter is not configured with basic authorization credentials.
```

This isn't blocking.

MrAlias:
@ercl this looks ready to merge. Can you sync with master?

ercl:
Merged!

PR #255, opened 2020-08-21
[Prometheus Remote Write Exporter for Cortex] Add TLS Support and Default HTTP Client

ercl:
This PR implements TLS support for the Go SDK Prometheus Remote Write Exporter for Cortex. It includes functionality for creating a default HTTP client for the exporter, unit tests, and an integration test for mutual TLS.


exporters/metric/cortex/auth.go
```diff
@@ -88,3 +92,99 @@ func (e *Exporter) addBearerTokenAuth(req *http.Request) error {
...
+
+	// Convert proxy url to proxy function for use in the created Transport.
+	if e.config.ProxyURL != "" {
+		proxyURL, err := url.Parse(e.config.ProxyURL)
+		if err != nil {
+			return nil, err
+		}
+		proxy := http.ProxyURL(proxyURL)
+		transport.Proxy = proxy
+	}
```
MrAlias, line 117: Can we add a default environment variable handling here for proxy specification since we are overriding the default transport: https://golang.org/pkg/net/http/#ProxyFromEnvironment

exporters/metric/cortex/auth.go
```diff
@@ -88,3 +92,99 @@ func (e *Exporter) addBearerTokenAuth(req *http.Request) error {
...
+		certPool := x509.NewCertPool()
+		certPool.AppendCertsFromPEM(caFileData)
+		tlsConfig.RootCAs = certPool
+	}
+	return nil
+}
+
+// loadClientCertificate loads a x509 certificate from a certificate file and key file,
+// and stores it in a TLS Config struct.
+func (e *Exporter) loadClientCertificate(tlsConfig *tls.Config) error {
```
MrAlias, line 178: Is there a reason these need to be separate functions? It doesn't seem like they are called anywhere else and could just be an inline conditional. It would be my preference to do that instead and avoid mutating input on a function call.

exporters/metric/cortex/auth.go
```diff
@@ -88,3 +92,99 @@ func (e *Exporter) addBearerTokenAuth(req *http.Request) error {
...
+		return nil, err
+	}
+
+	transport := &http.Transport{
+		TLSClientConfig: tlsConfig,
+	}
+
+	// Convert proxy url to proxy function for use in the created Transport.
+	if e.config.ProxyURL != "" {
+		proxyURL, err := url.Parse(e.config.ProxyURL)
```
MrAlias, line 111: Can we switch the config to store the `ProxyURL` field as a `url.URL` to avoid this parse here? Seems like this should be part of the config validation.

exporters/metric/cortex/auth.go
```diff
@@ -88,3 +92,99 @@ func (e *Exporter) addBearerTokenAuth(req *http.Request) error {
...
+	if e.config.TLSConfig["server_name"] != "" {
+		tlsConfig.ServerName = e.config.TLSConfig["server_name"]
+	}
+
+	// Set InsecureSkipVerify. Viper reads the bool as a string since it is in a map.
+	parsedBool, err := strconv.ParseBool(e.config.TLSConfig["insecure_skip_verify"])
+	if err != nil {
+		return nil, err
+	}
+	tlsConfig.InsecureSkipVerify = parsedBool
```
MrAlias, line 144: `"insecure_skip_verify"` may not exist in the `TLSConfig` making this parse error unnecessarily. The map should be checked if it contains the key prior to trying to parse it. I.e.

```suggestion
	if isv, ok := e.config.TLSConfig["insecure_skip_verify"]; ok {
		var err error
		if tlsConfig.InsecureSkipVerify, err = strconv.ParseBool(isv); err != nil {
			return nil, err
		}
	}
```

exporters/metric/cortex/auth.go
ercl: No reason other than making the `buildClient` method smaller. I'll change them to inline conditionals.

exporters/metric/cortex/auth_test.go
```diff
@@ -153,3 +164,353 @@ func createFile(bytes []byte, filepath string) error {
...
+	if err != nil {
+		return nil, nil, err
+	}
+
+	// Write the certificate to the local directory.
+	certPEM := pem.EncodeToMemory(&pem.Block{
+		Type:  "CERTIFICATE",
+		Bytes: encodedCert,
+	})
+	err = createFile(certPEM, certFilepath)
```
MrAlias, line 359: definitely a nit, but having this seems pretty circuitous given the [`Encode`](https://golang.org/pkg/encoding/pem/#Encode) function from the `pem` module accepts an `io.Writer`. Instead of writing everything to memory it could be directly written to a file handle if this was restructured.

Pretty sure it's not worth spending time changing given this is for a testing file, but just wanted to encourage the change if you were already on the fence. :wink:

exporters/metric/cortex/auth_test.go
```diff
@@ -153,3 +164,353 @@ func createFile(bytes []byte, filepath string) error {
...
+
+			// Attempt to send the request and verify that the correct error occurred. If
+			// an error is expected, the test checks if the error string contains the
+			// expected error substring since the error can contain the server URL, which
+			// changes every test.
+			_, err = client.Get(server.URL)
+			if test.expectedErrorSubstring != "" {
+				require.Error(t, err)
+				errorSuffix := strings.Contains(err.Error(), test.expectedErrorSubstring)
+				require.True(t, errorSuffix)
```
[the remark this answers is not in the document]
ercl: I understand that assert will stop one failure from stopping other failures from showing. What do you mean by not make fixes serialized?

exporters/metric/cortex/auth_test.go
ercl: Did not notice that about the `Encode` function! I adjusted it in the recent commits.

ercl:
Thanks for all the feedback! The recent commits should address everything in the reviews.

exporters/metric/cortex/auth_test.go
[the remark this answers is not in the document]
MrAlias: > I understand that assert will stop one failure from stopping other failures from showing. What do you mean by not make fixes serialized?

So when you use `require.*` it will cause the test to end at that point, instead of recording a test error and continuing. This means that if you have multiple errors in a test you will need to fix each serially if you use `require.*` instead of `assert.*` which will output all the errors encountered in that run.

This doesn't mean that `require.*` should never be used though. It should be used when there is no way to continue based on the measured state. Otherwise, it's best to assert things and have the user notified of all failures at once.

MrAlias:
@ercl this looks ready to merge. If you can update with the master branch it should be good to go.

ercl:
Merged! I also updated the HTTP Transport's DialContext timeout since the CircleCI test was occasionally taking too much time on GOARCH 386.

MrAlias:
> Merged! I also updated the HTTP Transport's DialContext timeout since the mutual tls test was occasionally taking too much time on GOARCH 386.

Hmm, looks like this is still out of sync. I'll be sure to not merge anything before this if you can update again.

ercl:
Merged again! I think the reason why CircleCI tests fail is because the testing script sets the timeout at 30s, which the Mutual TLS test occasionally goes over.

PR #282, opened 2020-08-25
Prepare Release v0.11.0

Aneurysm9:
### Added

- Top-level `Version()` and `SemVersion()` functions defining the current version of the contrib package. (#225)
- Instrumentation for the `github.com/astaxie/beego` package. (#200)
- Instrumentation for the `github.com/bradfitz/gomemcache` package. (#204)
- Host metrics instrumentation. (#231)
- Cortex histogram and distribution support. (#237)
- Cortex example project. (#238)
- Cortex HTTP authentication. (#246)

### Changed

- Remove service name as a parameter of Sarama instrumentation. (#221)
- Replace `WithTracer` with `WithTracerProvider` in Sarama instrumentation. (#221)
- Switch to use common top-level module `SemVersion()` when creating versioned tracer in `bradfitz/gomemcache`. (#226)
- Use `IntegrationShouldRun` in `gomemcache_test`. (#254)
- Use Go 1.15 for CI builds. (#236)
- Improved configuration for `runtime` instrumentation. (#224)

### Fixed

- Update dependabot configuration to include newly added `bradfitz/gomemcache` package. (#226)
- Correct `runtime` instrumentation name. (#241)


instrumentation/google.golang.org/grpc/interceptor.go
```diff
@@ -269,7 +270,7 @@ func StreamClientInterceptor(tracer trace.Tracer, opts ...Option) grpc.StreamCli
 
 			if err != nil {
 				s, _ := status.FromError(err)
-				span.SetStatus(s.Code(), s.Message())
+				span.SetStatus(codes.Code(s.Code()), s.Message())
```
MrAlias, line 273: We might want to build out a similar translanation function to [the SDK one](https://github.com/open-telemetry/opentelemetry-go/blob/master/sdk/internal/codes.go) here. The problem with this conversion is that is assumes the underlying integer will remain the same. I don't think it is important to fix in this PR as it will only break when a change to the gRPC codes occurs.

MrAlias: https://github.com/open-telemetry/opentelemetry-go-contrib/issues/295

Aneurysm9: I felt slightly dirty doing this, but I'm also pretty sure that `span.SetStatus()` is [not long for this world](https://github.com/open-telemetry/oteps/pull/134), so it may be a moot point.

PR #306, opened 2020-08-27
Allow custom labels to be added to net/http metrics

### LINKED ISSUE #195: grpc/http: support custom KeyValues when reporting metrics
We've been running OpenCensus for ~2 years and are excited to move to OpenTelemetry. The standard API has served us well, but our biggest issue with some of the wrapper packages like ocgrpc and ochttp are that they don't support custom KeyValues (TagKeys) when recording metrics. I can register my own views that include them, but because of the way that they wrap my server, I'm not passing a context that contains said values. I need this on a per request basis, to report specific error types, latency per kv, etc. 

Is that in scope for OpenTelemetry? Without it, I'm forced to copy and fork the implementation.

Aneurysm9:
Fixes #195


instrumentation/net/http/labeler.go
```diff
@@ -0,0 +1,52 @@
...
+
+// Labeler is used to allow instrumented HTTP handlers to add custom labels to
+// the metrics recorded by the net/http instrumentation.
+type Labeler struct {
+	labels []label.KeyValue
+}
+
+// Add labels to a Labeler.
+func (l *Labeler) Add(ls ...label.KeyValue) {
+	l.labels = append(l.labels, ls...)
```
MrAlias, line 31: Does this need to be guarded by a mutex? Having this be used by handlers that will run in their own goroutine it would seem so.

instrumentation/net/http/labeler.go
```diff
@@ -0,0 +1,52 @@
...
+type labelerContextKeyType int
+
+const lablelerContextKey labelerContextKeyType = 0
+
+func injectLabeler(ctx context.Context, l *Labeler) context.Context {
+	return context.WithValue(ctx, lablelerContextKey, l)
+}
+
+// LabelerFromContext retrieves a Labeler instance from the provided context if
+// one is available.  If no labeler was found in the provided context a new, empty
```
MrAlias, line 43: nit

```suggestion
// one is available.  If no Labeler was found in the provided context a new, empty
```

instrumentation/net/http/labeler.go
```diff
@@ -0,0 +1,52 @@
...
+
+import (
+	"context"
+
+	"go.opentelemetry.io/otel/label"
+)
+
+// Labeler is used to allow instrumented HTTP handlers to add custom labels to
+// the metrics recorded by the net/http instrumentation.
+type Labeler struct {
```
MrAlias, line 25: Can we add an example test to show how to use this?

jmacd:
I wonder if this functionality _ought_ to be achieved using Correlation Context. The OpenCensus system had a TTL=0 concept that meant the correlations would only survive within a single process. Then we wouldn't need a new API here, I think.

MrAlias:
> I wonder if this functionality _ought_ to be achieved using Correlation Context. The OpenCensus system had a TTL=0 concept that meant the correlations would only survive within a single process. Then we wouldn't need a new API here, I think.

That's a really interesting idea. It could make this solution more universal for all instrumentation.

Aneurysm9:
> > I wonder if this functionality _ought_ to be achieved using Correlation Context. The OpenCensus system had a TTL=0 concept that meant the correlations would only survive within a single process. Then we wouldn't need a new API here, I think.
> 
> That's a really interesting idea. It could make this solution more universal for all instrumentation.

I wonder how this would work to get data from the handler up to the instrumentation.  It seems that context flow is one-way.  Or are you proposing that the instrumentation add the `Labeler` to the Correlation Context rather than directly to the `context.Context`?

instrumentation/net/http/labeler.go
Aneurysm9: I added use of the `Labeler` to the existing `Handler` example test since an independent example would look very similar.

MrAlias:
> I wonder how this would work to get data from the handler up to the instrumentation. It seems that context flow is one-way. Or are you proposing that the instrumentation add the `Labeler` to the Correlation Context rather than directly to the `context.Context`?

Good point. I think I was envisioning the use to add context prior to it getting to the middleware of the handler, but I don't think that is going to be easily achieved nor desirable to a user. It seems like they would rather an explicit solution like what you have here.

instrumentation/net/http/handler.go
```diff
@@ -170,13 +170,16 @@ func (h *Handler) ServeHTTP(w http.ResponseWriter, r *http.Request) {
...
+	ctx = injectLabeler(ctx, labeler)
+
 	h.handler.ServeHTTP(w, r.WithContext(ctx))
 
 	setAfterServeAttributes(span, bw.read, rww.written, rww.statusCode, bw.err, rww.err)
 
 	// Add request metrics
 
-	labels := semconv.HTTPServerMetricAttributesFromHTTPRequest(h.operation, r)
+	labels := append(labeler.labels, semconv.HTTPServerMetricAttributesFromHTTPRequest(h.operation, r)...)
```
MrAlias, line 182: With the addition of the mutex a getter method is needed to ensure reads like this are concurrency safe.

Aneurysm9: This would also need to return a copy of the labels then, wouldn't it?  I wonder whether it's sufficient to declare that the `Labeler` is not safe for use after the wrapped `http.Handler` has returned.  Maybe a getter that sets a flag that `Add()` checks before allowing the labels to be modified?

MrAlias: I think returning a copy and making it clear in the documentation that this is the behavior would be sufficient.

jmacd:
I reviewed this PR and thought about the underlying request.
This feature offers an equivalent to the act of calling `Span.SetAttribute()` during a request, with the intent of adding labels to metrics associated with the effective Span. This is one reason people like to (at least) think about automatically converting spans to metrics--because we already have APIs to modify the current Span, therefore we wouldn't need new APIs to support adding labels to correlated metrics mid-span. My proposal about Resource scope was intended to address a similar problem: https://github.com/open-telemetry/oteps/issues/78. That proposal stops short of adding a way to modify a current scope (equivalent to SetAttribute), it only gave us a way to create new scopes, but the same idea could be extended to support mutable scopes.

The use of Context in this PR, at a first glance, leads to confusion. We're only using context to pass a variable from the server to the handler. We would not expect the same labeler to be used in a child, that's why I prefer to think of this as "modify the current metric scope".

After studying this, Correlation Context is not a good way to propagate these labels--it would lead to an awkward API. Short of something radical like OTEP issue 78 (Resource scope), I don't see a way to avoid an interface like Labeler in this PR.

instrumentation/net/http/handler.go
jmacd: I don't think a copy is needed, as long as the labeler itself only appends and promises not to mutate the slice or the values it contains. It's not a big deal to copy, though.

MrAlias (approved):

derekperkins:
That's awesome, thanks for pushing that through! My only question on the PR is whether or not the labeler should be pulled out so that it can be shared between other implementations like grpc.

PR #311, opened 2020-08-27
Update dependabot to run once a week on Sundays.

evantorrie:
Include github-actions directory and exhaustive list of all current directories containing a `go.mod` file.

In addition, reordered items within `detectors`, `exporters` and `instrumentation` by alphabetical order.


evantorrie:
Never mind. See I was too late - made redundant by #311

Issue #5, opened 2020-03-03
Opentelemetry for database/sql


wingyplus:
It's good to see support for database/sql driver. We've some discussion here https://github.com/open-telemetry/opentelemetry-go/pull/490.

andrewhsu:
FYI initial attempt that looks like a different approach would be more desirable https://github.com/open-telemetry/opentelemetry-go-contrib/pull/32

seslattery:
I have an initial proof of concept I implemented at https://github.com/seslattery/otelsql.  If this seems like a useful direction to be explored more, I'd be happy to open a PR and work on polishing it up some more.

Aneurysm9:
@seslattery please do open a PR.  I'd like to see `Register()` use functional options for the tracer and attributes, to be more in line with the SDK and other instrumentations.  I'm also not sure about using `runtime.Caller()` to derive span names from callers rather than just putting static strings at the call site.  Overall, though, this is a good start.

MrAlias:
Prior work in OpenCensus might be useful to look at: https://github.com/opencensus-integrations/ocsql

I would not copy this, but might be interesting to look at.

MrAlias:
Also: https://github.com/DataDog/dd-trace-go/tree/v1.24.1/contrib/database/sql

MrAlias:
Prior attempt within this repo: https://github.com/open-telemetry/opentelemetry-go-contrib/pull/32

seanschade:
@seslattery are you still planning on submitting a PR?

> I have an initial proof of concept I implemented at https://github.com/seslattery/otelsql. If this seems like a useful direction to be explored more, I'd be happy to open a PR and work on polishing it up some more.

@MrAlias what are your thoughts on using something like https://github.com/ngrok/sqlmw as an interceptor to wrap drivers?

I think an approach similar to https://github.com/seslattery/otelsql is more desirable than https://github.com/open-telemetry/opentelemetry-go-contrib/pull/32. 

If we would like to forego the wrapper approach then we have both [OpenTracing](https://github.com/DataDog/dd-trace-go/tree/v1.24.1/contrib/database/sql) and [OpenCensus](https://github.com/opencensus-integrations/ocsql) examples to draw from. 

It would be nice if we could enumerate what we like about each, and what this should look like for OpenTelemetry.

seslattery:
[...]

PR #320, opened 2020-08-30
Initial instrumentation for database/sql

seslattery:
Here is an initial PR to add tracing support to database/sql as in https://github.com/open-telemetry/opentelemetry-go-contrib/issues/5.  This implementation is relying on https://github.com/ngrok/sqlmw to create the interceptor.  

I have a basic example case implemented.

This currently doesn't have any tests.


MrAlias (changes requested):
It seems like we need to find consensus on how to implement this instrumentation first: https://github.com/open-telemetry/opentelemetry-go-contrib/issues/5#issuecomment-684175108

I'm going to block this as we need to find this consensus before this can move forward. Though, it might be more prudent to close this until we have a more agreed upon approach.

jmacd:
I'm curious how this PR addresses the problems that led us to reject https://github.com/open-telemetry/opentelemetry-go-contrib/pull/32.

MrAlias:
Closing until consensus is reached in #5 on a solution.

PR #337, opened 2020-09-11
mongo: use db.operation instead of resource.name

vmihailenco:
[spec](https://github.com/open-telemetry/opentelemetry-specification/blob/master/specification/trace/semantic_conventions/database.md) says

>db.operation The name of the operation being executed, e.g. the MongoDB command name such as findAndModify


instrumentation/go.mongodb.org/mongo-driver/tags.go
```diff
@@ -17,16 +17,16 @@ package mongo
...
-	SpanTypeKey     = label.Key("span.type")
-	ServiceNameKey  = label.Key("service.name")
-	ResourceNameKey = label.Key("resource.name")
-	ErrorKey        = label.Key("error")
-	ErrorMsgKey     = label.Key("error.msg")
+	TargetHostKey  = label.Key("out.host")
+	TargetPortKey  = label.Key("out.port")
+	HTTPMethodKey  = label.Key("http.method")
+	HTTPCodeKey    = label.Key("http.code")
+	HTTPURLKey     = label.Key("http.url")
```
vmihailenco, line 24: It looks like `HTTP*` keys are unused and look out of place (and http.code should be http.status_code). Should I remove them?

Aneurysm9: Yes, if they're unused there's no point keeping them around.  There should be equivalents in the `semconv` package anyways.

Aneurysm9 (approved):

MrAlias (approved):
Looks good. I'll wait to merge after the HTTP keys are removed.

vmihailenco:
@MrAlias done

PR #378, opened 2020-10-01
fixed missing span.End() in otelhttp.Transport

ofw:


Aneurysm9:
This span should not necessarily end at the end of the `RoundTrip` function.  It is [passed to the `wrappedBody`](https://github.com/open-telemetry/opentelemetry-go-contrib/blob/master/instrumentation/net/http/otelhttp/transport.go#L101) which then closes it [at `io.EOF`](https://github.com/open-telemetry/opentelemetry-go-contrib/blob/master/instrumentation/net/http/otelhttp/transport.go#L121) or [when the `wrappedBody` is closed](https://github.com/open-telemetry/opentelemetry-go-contrib/blob/master/instrumentation/net/http/otelhttp/transport.go#L129).

PR #388, opened 2020-10-08
use httpsnoop to ensure http.ResponseWriter additional interfaces are preserved

toshok:
Noticed this problem when going through the websocket upgrade flow.  http.Hijacker is implemented by builtin response writers but not by the recordingResponseWriter.

Instead of doing all the leg work to ensure that interfaces are implemented correctly, convert things to use httpsnoop, which does codegen to generate the possible combinations of interface implementations so that things are preserved.

One question: given that the `recordingResponseWriter` initialization code now includes an additional set of allocations (from `httpsnoop`, but also the closures for the hooks), is it worth it to keep the sync.Pool?  My gut says no, but... ?


go.mod
```diff
@@ -4,7 +4,10 @@ go 1.14
 
 require (
 	github.com/davecgh/go-spew v1.1.1 // indirect
+	github.com/felixge/httpsnoop v1.0.1 // indirect
+	github.com/gorilla/mux v1.8.0 // indirect
 	github.com/kr/pretty v0.1.0 // indirect
+	go.opentelemetry.io/contrib/propagators v0.12.0 // indirect
```
MrAlias: These look like they were inadvertent. The `otelmux` module should be the only place these dependencies need to be added.

MrAlias (approved):

go.mod
toshok: yep, I failed to run `make precommit` 👍

toshok:
hm, seeing FD exhaustion in a service running with this now.  recommend holding off on accepting/landing

toshok:
phew, tracked down FD leak - not anything related to this change, so should be good to go

lizthegrey (approved):

MrAlias:
@toshok This looks ready to merge given your validation. It appears maintainer edits are off for this so I cannot sync with master. Can you sync so we can merge this?

toshok:
rebased off current master

Aneurysm9 (approved):

PR #390, opened 2020-10-09
Add convenience wrappers for http.Get/Post, etc.

johnbley:
Add convenience wrappers for http.{Get,Post,Head,PostForm}
using the otelhttp transport and taking a context.Context directly.
Makes instrumenting code using the original http convenience methods a
one-line change instead of a multiline one.


instrumentation/net/http/otelhttp/client.go
```diff
@@ -0,0 +1,58 @@
...
+	"context"
+	"io"
+	"net/http"
+	"net/url"
+	"strings"
+)
+
+var otelDefaultClient = &http.Client{Transport: NewTransport(http.DefaultTransport)}
+
+// Convenience replacement for http.Get
```
Aneurysm9, line 27:
```suggestion
// Get is a convenient replacement for http.Get that adds a span around the request.
```

instrumentation/net/http/otelhttp/client.go
```diff
@@ -0,0 +1,58 @@
...
+func Get(ctx context.Context, url string) (resp *http.Response, err error) {
+	req, err := http.NewRequestWithContext(ctx, "GET", url, nil)
+	if err != nil {
+		return nil, err
+	}
+	return otelDefaultClient.Do(req)
+}
+
+// Convenience replacement for http.Head
```
Aneurysm9, line 36:
```suggestion
// Head is a convenient replacement for http.Head that adds a span around the request.
```

instrumentation/net/http/otelhttp/client.go
```diff
@@ -0,0 +1,58 @@
...
+func Head(ctx context.Context, url string) (resp *http.Response, err error) {
+	req, err := http.NewRequestWithContext(ctx, "HEAD", url, nil)
+	if err != nil {
+		return nil, err
+	}
+	return otelDefaultClient.Do(req)
+}
+
+// Convenience replacement for http.Post
```
Aneurysm9, line 45:
```suggestion
// Post is a convenient replacement for http.Post that adds a span around the request.
```

instrumentation/net/http/otelhttp/client.go
```diff
@@ -0,0 +1,58 @@
...
+func Post(ctx context.Context, url, contentType string, body io.Reader) (resp *http.Response, err error) {
+	req, err := http.NewRequestWithContext(ctx, "POST", url, body)
+	if err != nil {
+		return nil, err
+	}
+	req.Header.Set("Content-Type", contentType)
+	return otelDefaultClient.Do(req)
+}
+
+// Convenience replacement for http.PostForm
```
Aneurysm9, line 55:
```suggestion
// PostForm is a convenient replacement for http.PostForm that adds a span around the request.
```

Aneurysm9 (commented):
I like the idea!  Can you please add some tests and a `CHANGELOG.md` entry?

johnbley:
> I like the idea! Can you please add some tests and a `CHANGELOG.md` entry?

Added to CHANGELOG.  Tests might need to wait a bit.

instrumentation/net/http/otelhttp/client.go
```diff
@@ -0,0 +1,58 @@
...
+
+import (
+	"context"
+	"io"
+	"net/http"
+	"net/url"
+	"strings"
+)
+
+var otelDefaultClient = &http.Client{Transport: NewTransport(http.DefaultTransport)}
```
XSAM, line 25: Maybe we should expose this client as a default client for these wrappers, just like Go `http` lib.

```go
// DefaultClient is the default Client and is used by Get, Head, and Post.
var DefaultClient = &Client{}
```

Therefore, end-users can use `otelhttp.DefaultClient.Do()` as well.

johnbley: I went the other direction, actually.  Because the wrapped transport captures (e.g.) propagators, this could lead to unwanted ordering issues, such as 
```
Http.Get("/usesOldPropagator")
global.SetPropagators(newProps)
Http.Get("/shouldUseNewButOnlyIfWeDidNotCacheTheOld")
```
I discovered this by writing the unit tests and being surprised by the behavior.  We could offer `otelhttp.NewDefaultClient()` as a substitute?

Aneurysm9: I'm wary of the construction of a new client for every request.  That seems like it could lead to performance issues as the client's connection pool cannot be re-used and the overhead of constructing a new client and transport must be incurred at every request rather than amortized over multiple requests.

XSAM: I agree with what @Aneurysm9 said about the performance. Adding a comment to the `DefaultClient` to notify users would be enough?

e.g., Some changes in `otel` might not affect the `DefaultClient`, such as setting a new global propagator.

johnbley: OK, I reverted back everything to an (exported) `DefaultClient`.  Thanks for the comments!

instrumentation/net/http/otelhttp/client.go
```diff
@@ -0,0 +1,58 @@
...
+
+import (
+	"context"
+	"io"
+	"net/http"
+	"net/url"
+	"strings"
+)
+
+var DefaultClient = &http.Client{Transport: NewTransport(http.DefaultTransport)}
```
XSAM, line 25:
```suggestion
// DefaultClient is the default Client and is used by Get, Head, Post and PostForm.
// Some changes in OTel might not affect the DefaultClient, such as setting a new global propagator.
// Setting up a new client to apply changes.
var DefaultClient = &http.Client{Transport: NewTransport(http.DefaultTransport)}
```

Don't forget the comment. ;)

CHANGELOG.md
```diff
@@ -14,6 +14,10 @@ This project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0.htm
    Instead, a new `WithTracerProvider` option is added to configure the `TracerProvider` used when creating the `Tracer` for the instrumentation. (#373)
 - Remove `go.opentelemetry.io/otel/sdk` dependency from instrumentation. (#381)
 
+### Added
+
+- `otelhttp.{Get,Head,Post,PostForm} convenience wrappers for their `http` counterparts. (#390)
```
MrAlias, line 19:
```suggestion
- `otelhttp.{Get,Head,Post,PostForm}` convenience wrappers for their `http` counterparts. (#390)
```

instrumentation/net/http/otelhttp/client.go
```diff
@@ -0,0 +1,61 @@
...
+	"context"
+	"io"
+	"net/http"
+	"net/url"
+	"strings"
+)
+
+// DefaultClient is the default Client and is used by Get, Head, Post and PostForm.
+// Some changes in OTel might not affect the DefaultClient, such as setting a new global propagator.
+// Setting up a new client to apply changes.
```
MrAlias, line 27: This is a bug with our implementation of the global propagators. I'm going to open an issue to address this.

instrumentation/net/http/otelhttp/client_test.go
```diff
@@ -0,0 +1,73 @@
...
+	"go.opentelemetry.io/otel/api/global"
+
+	mocktrace "go.opentelemetry.io/contrib/internal/trace"
+)
+
+func TestConvenienceWrappers(t *testing.T) {
+	provider, tracer := mocktrace.NewTracerProviderAndTracer(instrumentationName)
+	oldProvider := global.TracerProvider()
+	global.SetTracerProvider(provider)
+	defer global.SetTracerProvider(oldProvider)
```
MrAlias, line 34: This will have no effect, the TracerProvider can only be set once.

https://github.com/open-telemetry/opentelemetry-go/blob/8ed55f5980611f2e1effe471f9946aba263f255d/api/global/internal/state.go#L54-L68

Setting the `DefaultClient` here instead of relying on the global package would ensure the following tests use the intended Tracer and TextMapPropagator.

instrumentation/net/http/otelhttp/client_test.go
```diff
@@ -0,0 +1,73 @@
...
+		t.Fatal(err)
+	}
+
+	form := make(url.Values)
+	form.Set("foo", "bar")
+	_, err = PostForm(context, ts.URL, form)
+	if err != nil {
+		t.Fatal(err)
+	}
+
```
MrAlias, line 72: Can we validate that 4 spans were created on the client side?

instrumentation/net/http/otelhttp/client.go
MrAlias: https://github.com/open-telemetry/opentelemetry-go/issues/1257

instrumentation/net/http/otelhttp/client_test.go
johnbley: Cool, removed the attempt to re-set the global provider.

instrumentation/net/http/otelhttp/client_test.go
johnbley: 👍

MrAlias (approved):

instrumentation/net/http/otelhttp/client.go
Aneurysm9, line 27: What does this comment mean? It seems like it might be out of place or unneeded.

instrumentation/net/http/otelhttp/client_test.go
```diff
@@ -0,0 +1,79 @@
...
+	form := make(url.Values)
+	form.Set("foo", "bar")
+	res, err = PostForm(context, ts.URL, form)
+	if err != nil {
+		t.Fatal(err)
+	}
+	res.Body.Close()
+
+	if len(tracer.EndedSpans()) != 4 {
+		t.Fatal("expected 4 client spans")
```
Aneurysm9, line 76: I'm not sure that these tests are effective.  While they are now asserting that four spans were indeed created, it would be possible for them to pass if the `Get` method invoked itself four times and the other three were noops.  Perhaps this can be broken out into individual test cases for each method that can give higher confidence that the correct behavior is obtained?

Aneurysm9 (changes requested):

instrumentation/net/http/otelhttp/client.go
johnbley: I blindly approved a comment suggestion for this - I've now (hopefully) improved the comments here.

instrumentation/net/http/otelhttp/client_test.go
johnbley: I added a check that we got the expected ([GET,HEAD,POST,POST]) span names.

Aneurysm9 (approved):

Aneurysm9:
@johnbley one last thing, it looks like the `CHANGELOG.md` entry is still under the `v0.13.0` release section.  Can you move it up to `Unreleased`?

PR #381, opened 2020-10-02
Remove `go.opentelemetry.io/otel/sdk` dependency from instrumentation

matej-g:
In line with the aim stated in the #161, the instrumentation packages should be made more lightweight by removing the dependency on the main SDK package (go.opentelemetry.io/otel/sdk). In order achieve it, this changeset:

- Separates instrumentation examples (which usually depending on importing SDK in order to setup export pipeline) into individual modules, which removes dependency on SDK from the main instrumentation module.
- Adjusts `gocql` metric tests, which previously depended on SDK - this has been replaced by internal mock for metrics

Additionally, dependabot configuration has been updated to reflect his, as well as `.gitignore` to ignore binaries built by example modules.

Resolves #161

PR #377, opened 2020-09-30
fix(echo): return nil from middleware to handle error only once

stanleynguyen:
Resolves #364

Issue #364, opened 2020-09-25
Error handler triggered twice for echo


stanleynguyen:
Calling `c.Error`
https://github.com/open-telemetry/opentelemetry-go-contrib/blob/b313ddb2989e21338560e24b9e9478c563017d20/instrumentation/github.com/labstack/echo/otelecho/echo.go#L84
and returning err from a middleware 
https://github.com/open-telemetry/opentelemetry-go-contrib/blob/b313ddb2989e21338560e24b9e9478c563017d20/instrumentation/github.com/labstack/echo/otelecho/echo.go#L92
will trigger error handling twice.

I'm proposing to return nil instead

stanleynguyen:
I will submit  PR if there's no suggestion otherwise

MrAlias:
SGTM :+1:

PR #391, opened 2020-10-09
Pre release v0.13.0

Aneurysm9:
## Added

- A Jaeger propagator. (#375)

## Changed

- The `go.opentelemetry.io/contrib/instrumentation/google.golang.org/grpc/otelgrpc` package instrumentation no longer accepts a `Tracer` as an argument to the interceptor function.
   Instead, a new `WithTracerProvider` option is added to configure the `TracerProvider` used when creating the `Tracer` for the instrumentation. (#373)
- The `go.opentelemetry.io/contrib/instrumentation/gopkg.in/macaron.v1/otelmacaron` instrumentation now accepts a `TracerProvider` rather than a `Tracer`. (#374)
- Remove `go.opentelemetry.io/otel/sdk` dependency from instrumentation. (#381)

### Fixed

- The `go.opentelemetry.io/contrib/instrumentation/github.com/labstack/echo/otelecho.Middleware` no longer sends duplicate errors to the global `ErrorHandler`. (#377, #364)
- The import comment in `go.opentelemetry.io/contrib/instrumentation/net/http/otelhttp` is now correctly quoted. (#379)
- The B3 propagator sets the sample bitmask when the sampling decision is `debug`. (#369)


instrumentation/net/http/otelhttp/example/server/server.go
```diff
@@ -46,15 +49,19 @@ func initTracer() {
...
 func main() {
 	initTracer()
 
+	uk := label.Key("username")
+
 	helloHandler := func(w http.ResponseWriter, req *http.Request) {
 		ctx := req.Context()
 		span := trace.SpanFromContext(ctx)
-		span.AddEvent(ctx, "handling this...")
+		username := otel.BaggageValue(ctx, uk)
```
MrAlias, line 63: Nice :+1:

MrAlias (approved):
Awesome! This looks like it was a lot of work. Thanks for digging into it.  :tada:

XSAM (approved):

PR #401, opened 2020-10-14
[aws detector] Remove error when instance metadata is unavailable

### LINKED ISSUE #398: AWS Resource detector fails with an error if metadata not available
If the instance metadata is not available, no resource and an `unavailable EC2 client` error will be returned

https://github.com/open-telemetry/opentelemetry-go-contrib/blob/7e7d779/detectors/aws/aws.go#L42-L45

```
client := ec2metadata.New(session)
if !client.Available() {
	return nil, errors.New("unavailable EC2 client")
}
```

This is counter to the documented behaviour for resources which should be:

> Note the failure to detect any resource information MUST NOT be considered an error, whereas an error that occurs during an attempt to detect resource information SHOULD be considered an error.

https://github.com/open-telemetry/opentelemetry-specification/blob/master/specification/resource/sdk.md#detecting-resource-information-from-the-environment

A similar issue was raised (and fixed) with the ENV var detection in the main OTel library: https://github.com/open-telemetry/opentelemetry-go/issues/1114 / https://github.com/open-telemetry/opentelemetry-go/pull/1170

The GCP resource in this contrib library appears to implement detection correctly, returning no error:

https://github.com/open-telemetry/opentelemetry-go-contrib/blob/7e7d779/detectors/gcp/gce.go#L38-L40
https://github.com/open-telemetry/opentelemetry-go-contrib/blob/master/detectors/gcp/gce.go#L102-L104

```
if !metadata.OnGCE() {
	return nil, nil
}
```

```
if _, undefined := err.(metadata.NotDefinedError); undefined {
	return false
}
```

dackroyd:
> Note the failure to detect any resource information MUST NOT be considered an error

When instance metadata is not available, this should not be treated as an error.

Small refactoring required to allow tests to mock the client, which should be _mostly_ non-impacting for consumers.

Fixes #398

### DIFF
--- a/detectors/aws/aws_test.go
+++ b/detectors/aws/aws_test.go
@@ -1,0 +1,118 @@
+// Copyright The OpenTelemetry Authors
+//
+// Licensed under the Apache License, Version 2.0 (the "License");
+// you may not use this file except in compliance with the License.
+// You may obtain a copy of the License at
+//
+//     http://www.apache.org/licenses/LICENSE-2.0
+//
+// Unless required by applicable law or agreed to in writing, software
+// distributed under the License is distributed on an "AS IS" BASIS,
+// WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+// See the License for the specific language governing permissions and
+// limitations under the License.
+
+package aws
+
+import (
+	"context"
+	"errors"
+	"testing"
+	"time"
+
+	"github.com/aws/aws-sdk-go/aws/ec2metadata"
+	"github.com/stretchr/testify/assert"
+	"github.com/stretchr/testify/require"
+
+	"go.opentelemetry.io/otel/sdk/resource"
+	"go.opentelemetry.io/otel/semconv"
+)
+
+func TestAWS_Detect(t *testing.T) {
+	type fields struct {
+		Client client
+	}
+
+	type want struct {
+		Error    string
+		Resource *resource.Resource
+	}
+
+	testTable := map[string]struct {
+		Fields fields
+		Want   want
+	}{
+		"Unavailable": {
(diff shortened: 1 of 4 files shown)


detectors/aws/aws_test.go
```diff
@@ -0,0 +1,118 @@
...
+						PendingTime:             time.Date(2016, time.November, 19, 16, 32, 11, 0, time.UTC),
+						ImageID:                 "ami-5fb8c835",
+						Architecture:            "x86_64",
+					}
+
+					return doc, nil
+				}},
+			},
+			Want: want{Resource: resource.New(
+				semconv.CloudProviderAWS,
```
dackroyd, line 78: I noticed while making this change that there are several other `semconv` values that could be populated by AWS instance metadata, based on what is included in the example identity doc:

* `host.image.id` -> `ami-5fb8c835`
* `host.type` -> `t2.micro`
* `cloud.zone` -> `us-west-2b`

MrAlias (approved):
This is great! Thanks for the contribution.

detectors/aws/aws_test.go
MrAlias: @dackroyd can you open an issue to track this? These seem like good things for this detector to include.

Also, I can open an issue if you'd prefer.

dackroyd: Sure, I'll raise another issue for that one. Pretty simple change, I can probably push a PR for it as well, but might need confirmation that there isn't anything else that should also be captured before doing that work.

I've not worked in a GCP environment, so I'm not sure what else might be available to add to that detector.

dackroyd: Created #403

Aneurysm9 (approved):

PR #410, opened 2020-10-19
[aws detector] Additional Attributes

### LINKED ISSUE #403: Missing resource attributes from AWS detector
The AWS resource detector adds several of attributes following the semantic conventions for cloud provider/host, however several other values are available that could also be added by the detector. Currently the implementation captures:

```
labels := []label.KeyValue{
	semconv.CloudProviderAWS,
	semconv.CloudRegionKey.String(doc.Region),
	semconv.CloudAccountIDKey.String(doc.AccountID),
	semconv.HostIDKey.String(doc.InstanceID),
}
```

From the AWS EC2 instance metadata, we can additionally capture:

* `host.image.id` e.g. `ami-5fb8c835`
* `host.type` e.g. `t2.micro`
* `cloud.zone` e.g. `us-west-2b`

There are other fields available in the instance metadata that are not covered by the semantic conventions, some of which may be useful also: https://docs.aws.amazon.com/AWSEC2/latest/UserGuide/instance-identity-documents.html

I haven't explored the AWS APIs much deeper than that, but there are a number of attributes defined by the sementic conventions that _are not_ available from the instance metadata, but it's possible they might be discoverable in other ways.

dackroyd:
Adding extra host/cloud attributes from the semantic conventions that can be provided by AWS instance metadata:

 * `host.image.id`
 * `host.name`
 * `host.type`
 * `cloud.zone`

Metadata (currently only `hostname`) failures will still allow the resource to be returned, along with a wrapped ErrPartialResource (per the Detector interface design)

Fixes #403


detectors/aws/aws_test.go
```diff
@@ -53,32 +86,43 @@ func TestAWS_Detect(t *testing.T) {
...
+		},
+		"Hostname General Error": {
+			Fields: fields{
+				Client: &clientMock{available: true, idDoc: usWestInst, metadata: map[string]meta{"hostname": {err: errors.New("unknown error")}}},
+			},
+			Want: want{Error: `partial resource: ["hostname": unknown error]`, Partial: true, Resource: resource.New(usWestIDLabels...)},
+		},
+		"All Available": {
+			Fields: fields{
+				Client: &clientMock{available: true, idDoc: usWestInst, metadata: map[string]meta{"hostname": {value: "ip-12-34-56-78.us-west-2.compute.internal"}}},
```
MrAlias, line 115: Total nitpick, but I'd be in favor of breaking up some of these long lines into multiple lines.

MrAlias (approved):

detectors/aws/aws_test.go
dackroyd: I've split most cases of long lines now (struct fields), but I've left the AWS errors as single lines

Aneurysm9 (approved):

PR #412, opened 2020-10-19
Add sem version to remaining instrumentation packages

matej-g:
According to #214, every instrumentation should create its respective `Tracer` and / or `Meter` with a semantic version. This version should match the repository version, which is achieved by using the `SemVer` function in `contrib.go` from the top-level package.

Most of the instrumentation packages already include semantic version, this PR collects and adjusts the remaining ones, which are:
- instrumentation/github.com/Shopify/sarama/otelsarama
- instrumentation/github.com/emicklei/go-restful/otelrestful
- instrumentation/go.mongodb.org/mongo-driver/mongo/otelmongo
- instrumentation/net/http/otelhttp
- instrumentation/net/http/httptrace/otelhttptrace

It also adds this requirement to the `README.md` in instrumentation, as per request in the ticket description.

Should be last piece to resolving #214.


MrAlias (approved):
:1st_place_medal:

XSAM (approved):

jmacd (approved):

PR #414, opened 2020-10-19
Use tracer and meter test helpers from the OTEL repository instead of the internal ones

### LINKED ISSUE #31: Merge the mock tracer into the otel-go's mock tracer.
The mock tracer in `internal/trace` is a modified copy of the mock tracer in otel-go. When we decide that our copy of mock tracer meets the needs of the instrumentation testing code (so the mock tracer stabilized) then it would be good to reduce the code duplication and bring the changes back to otel-go.

### LINKED ISSUE #385: Consider using trace and meter test helpers from the OTEL repository
The `contrib` repository currently implements it's own set of trace / meter test helpers in the `internal` directory. However, these helper are functionally mostly identical to the helpers which are presently already available in the `otel` repository and can be exported for use in the `contrib` repository.

Specifically, this would mean replacing
- `go.opentelemetry.io/contrib/internal/metric`
- `go.opentelemetry.io/contrib/internal/trace`

by 
- `go.opentelemetry.io/otel/api/metric/metrictest`
- `go.opentelemetry.io/otel/api/trace/tracetest`

respectively.

This will ensure having unified approach to test helpers and deduplication of code between `contrib` and `otel` repositories.

matej-g:
For some time, the test helpers have existed in parallel in both `otel` and `contrib` repositories. Although helpers were functionally very similar in both repositories, not all helpers have been exported from the main `otel` repository, which resulted in some duplication of helper methods.

This is no longer the case, as helpers for both tracer and meter are now "stabilized" and exported in the main `otel` repository (`go.opentelemetry.io/otel/api/trace/tracetest` and `go.opentelemetry.io/otel/api/metric/metrictest` as of current version `0.13.0`). This makes it possible to replace internal helpers in `contrib` by the ones exported from `otel` repo and subsequently to remove internal helpers, resolving in cleaner and slimmer codebase.

Changes include:
- Replacing internal helpers by external ones in instrumentation modules + related test changes
- Replacing internal helpers by external ones in propagators modules + related test changes
- Removal of internal helper packages
- `go mod` related changes in affected modules

Resolves #385 and effectively resolves #31, #43 as well


matej-g:
FYI, looks like from next version, the tracer test package in `otel` repo has been moved by https://github.com/open-telemetry/opentelemetry-go/pull/1229, so the `go.opentelemetry.io/otel/api/trace/tracetest` import should be changed to `go.opentelemetry.io/otel/oteltest` before the next release.

instrumentation/github.com/astaxie/beego/otelbeego/common.go
```diff
@@ -23,10 +23,6 @@ import (
 // ------------------------------------------ Constants
 
 const (
-	// packageName is the name of the this package, and is used as the default tracer
-	// and meter names.
-	packageName = "go.opentelemetry.io/contrib/instrumentation/github.com/astaxie/beego/otelbeego"
```
matej-g, line 28: `packageName` was not used anywhere (apart from tests), and with reference to https://github.com/open-telemetry/opentelemetry-go-contrib/pull/303#issuecomment-680990488 this is desired behavior, so I removed it.

instrumentation/github.com/Shopify/sarama/otelsarama/producer_test.go
```diff
@@ -128,34 +120,34 @@ func TestWrapSyncProducer(t *testing.T) {
...
 func TestWrapAsyncProducer(t *testing.T) {
 	propagators := otelpropagators.TraceContext{}
 	// Create message with span context
-	createMessages := func(mt *mocktracer.Tracer) []*sarama.ProducerMessage {
+	createMessages := func(mt oteltrace.Tracer) []*sarama.ProducerMessage {
 		ctx, _ := mt.Start(context.Background(), "")
 		messageWithSpanContext := sarama.ProducerMessage{Topic: topic, Key: sarama.StringEncoder("foo")}
 		propagators.Inject(ctx, NewProducerMessageCarrier(&messageWithSpanContext))
-		mt.EndedSpans()
+		// mt.EndedSpans()
```
MrAlias, line 150: Should this just be removed?

MrAlias (approved):
Looks solid :+1: 

Thanks for digging into this!

XSAM (approved):
👍

instrumentation/github.com/Shopify/sarama/otelsarama/producer_test.go
matej-g: yes, this was leftover, removed :+1:

Aneurysm9 (approved):
I'm overall in favor of this, but [the `Event` addition](https://github.com/open-telemetry/opentelemetry-go-contrib/pull/413/files#diff-29598e8c63391fa1d4ffa748b3fcc107c5950485ca9f82893581c35b95249f33) in #413 makes me think perhaps we should ensure that we have fully-working helper implementations before relying on the main repo.  Otherwise we may find ourselves in a situation where multiple releases need to happen in order to properly test an addition in this repo.

matej-g:
> I'm overall in favor of this, but [the `Event` addition](https://github.com/open-telemetry/opentelemetry-go-contrib/pull/413/files#diff-29598e8c63391fa1d4ffa748b3fcc107c5950485ca9f82893581c35b95249f33) in #413 makes me think perhaps we should ensure that we have fully-working helper implementations before relying on the main repo. Otherwise we may find ourselves in a situation where multiple releases need to happen in order to properly test an addition in this repo.

I guess `Event`s are in the `oteltest` package already (https://github.com/MrAlias/opentelemetry-go/commits/master/oteltest/event.go). Are we missing something else in the main repo?

MrAlias:
@Aneurysm9 are you good with merging this and possibly opening an issue to address the concerns you raised?

Aneurysm9:
Yup, I'm good.  It looks like the main repo has more in the `oteltest` package than the subset here, so we should be good to use that package.

PR #427, opened 2020-10-30
Add metrics for otelhttp client

vchitai:
Add some metrics for the otelhttp transport for http client net/http

Includes those metrics:
- `http_client_duration`

Tested sample
```
$ curl localhost:10080/metrics
# HELP http_client_duration
# TYPE http_client_duration histogram
http_client_duration_bucket{http_client_host="localhost:10080",http_client_method="GET",http_client_path="/hello1",http_client_status="200",http_host="localhost:10080",http_method="GET",http_path="/hello1",http_status="200",le="+Inf"} 2
http_client_duration_sum{http_client_host="localhost:10080",http_client_method="GET",http_client_path="/hello1",http_client_status="200",http_host="localhost:10080",http_method="GET",http_path="/hello1",http_status="200"} 0.645514
http_client_duration_count{http_client_host="localhost:10080",http_client_method="GET",http_client_path="/hello1",http_client_status="200",http_host="localhost:10080",http_method="GET",http_path="/hello1",http_status="200"} 2
```

Referencing: https://github.com/census-instrumentation/opencensus-go/blob/master/plugin/ochttp/stats.go


MrAlias (changes requested):
The http metrics we implement need to conform to the OpenTelemetry metric semantic conventions: https://github.com/open-telemetry/opentelemetry-specification/blob/master/specification/metrics/semantic_conventions/http-metrics.md

vchitai:
> The http metrics we implement need to conform to the OpenTelemetry metric semantic conventions: https://github.com/open-telemetry/opentelemetry-specification/blob/master/specification/metrics/semantic_conventions/http-metrics.md

Thanks for the feedback, I will change it later.

vchitai:
@MrAlias I have updated this PR according to your document, can you review it again?

instrumentation/net/http/otelhttp/common.go
```diff
@@ -36,6 +36,11 @@ const (
 	ServerLatency         = "http.server.duration"                // Incoming end to end duration, microseconds
 )
 
+// Client HTTP metrics
```
MrAlias, line 39:
```suggestion
// Client HTTP metric instrument names
```

instrumentation/net/http/otelhttp/common.go
```diff
@@ -36,6 +36,11 @@ const (
...
+const (
+	ClientRequestDuration = "http.client.duration" // Incoming end to end duration, microseconds
```
MrAlias, line 41: Is there a reason this is exported? Also, this measures the duration of the outbound HTTP request.

```suggestion
	// clientRequestDuration is the name of the instrument that measures the duration of outbound HTTP requests.
	clientRequestDuration = "http.client.duration"
```

instrumentation/net/http/otelhttp/stats.go
```diff
@@ -0,0 +1,206 @@
+// Copyright The OpenTelemetry Authors
```
MrAlias, line 1: Naming this file `stats.go` seems like a hold over from OpenCensus. Can we rename it to `metrics.go` or something more idiomatic of OTel and the terminology of the project?

instrumentation/net/http/otelhttp/stats.go
```diff
@@ -0,0 +1,206 @@
...
+		// conform open-telemetry label definition
+		Method.String(req.Method),
+		Host.String(req.Host),
+		Scheme.String(req.URL.Scheme),
+		Path.String(req.URL.Path),
+		// for prometheus
+		KeyClientMethod.String(req.Method),
+		KeyClientHost.String(req.Host),
+		KeyClientScheme.String(req.URL.Scheme),
+		KeyClientPath.String(req.URL.Path),
```
MrAlias, line 108: These cannot be included here. If the prometheus exporter would like to translate the attributes that is fine, but this instrumentation needs to remain agnostic of downstream processing.

instrumentation/net/http/otelhttp/stats.go
```diff
@@ -0,0 +1,206 @@
...
+	// or "error" if a transport error occurred and no status code was read.
+	StatusCode = label.Key("http.status")
+
+	// Path is the URL path (not including query string) in the request.
+	//
+	// The value of this tag can be controlled by the HTTP client, so you need
+	// to watch out for potentially generating high-cardinality labels in your
+	// metrics backend if you use this tag in views.
+	Path = label.Key("http.path")
+)
```
MrAlias, line 73: The semconv package should be used instead of redefining these: https://github.com/open-telemetry/opentelemetry-go/blob/9ac3a08eef14d96687edb260e2c27114a658d6b7/semconv/trace.go#L77-L128

instrumentation/net/http/otelhttp/stats.go
```diff
@@ -0,0 +1,206 @@
...
+	"go.opentelemetry.io/otel/unit"
+
+	"go.opentelemetry.io/otel/api/metric"
+	"go.opentelemetry.io/otel/label"
+)
+
+type statTransport struct {
+	meter          metric.Meter
+	base           *Transport
+	valueRecorders map[string]metric.Float64ValueRecorder
```
MrAlias, line 33: Given there is only one metric instrument, it would make more sense to store that instrument as a field directly here.

instrumentation/net/http/otelhttp/stats.go
```diff
@@ -0,0 +1,206 @@
...
+	if err != nil {
+		track.statusCode = http.StatusInternalServerError
+		track.end()
+	} else {
+		track.statusCode = resp.StatusCode
+		if resp.Body == nil {
+			track.end()
+		} else {
+			track.body = resp.Body
+			resp.Body = wrappedBodyIO(track, resp.Body)
```
MrAlias, line 130: Reading of the response body should not be included in the measurement of a requests duration.

go.sum
```diff
@@ -1,3 +1,4 @@
+github.com/davecgh/go-spew v1.1.0 h1:ZDRjVQ15GmhC3fiQ8ni8+OwkZQO4DARzQgrnXU1Liz8=
```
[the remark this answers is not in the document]
vchitai: I will remove it

instrumentation/net/http/otelhttp/common.go
vchitai: I changed it

instrumentation/net/http/otelhttp/common.go
vchitai: I changed it

instrumentation/net/http/otelhttp/stats.go
vchitai: Renamed to `metrics.go`

instrumentation/net/http/otelhttp/stats.go
vchitai: Removed it then

instrumentation/net/http/otelhttp/stats.go
vchitai: I changed all the labels to `semconv` package. Thanks for the reference.

instrumentation/net/http/otelhttp/stats.go
vchitai: Changed it.

instrumentation/net/http/otelhttp/stats.go
vchitai: The request will be counted as ended after we finish reading all the data from the response body, isn't it?

Aneurysm9: There is probably utility in measurements of both time to first byte and time to the end of the body being read, though the latter is potentially less a measure of the client and more of whatever is processing the response data.

instrumentation/net/http/otelhttp/transport.go
```diff
@@ -42,9 +42,11 @@ var _ http.RoundTripper = &Transport{}
 
 // NewTransport wraps the provided http.RoundTripper with one that
 // starts a span and injects the span context into the outbound request headers.
-func NewTransport(base http.RoundTripper, opts ...Option) *Transport {
-	t := Transport{
-		rt: base,
+func NewTransport(base http.RoundTripper, opts ...Option) http.RoundTripper {
```
Aneurysm9, line 45: Since this is the primary mechanism by which I expect `otelhttp.Transport` instances to be created and this leaves no configuration option for enabling/disabling metric collection, does it make more sense to integrate the metrics collection directly in the `Transport.RoundTrip` method?  This would also be more in line with the approach taken by the `otelhttp.Handler`.

instrumentation/net/http/otelhttp/stats.go
vchitai: This only wrap and returns reader interface rather than actual read and parsing actual data itself. The actual parsing time could be useful, but should not be included in this package, dont you think ?

instrumentation/net/http/otelhttp/common.go
```diff
@@ -36,6 +36,12 @@ const (
 	ServerLatency         = "http.server.duration"                // Incoming end to end duration, microseconds
 )
 
+// Client HTTP metric instrument names
```
MrAlias, line 39:
```suggestion
// Client HTTP metric instrument names.
```

instrumentation/net/http/otelhttp/metrics.go
```diff
@@ -0,0 +1,143 @@
...
+	case io.EOF:
+		tracker.end()
+	}
+	return n, err
+}
+
+func (tracker *tracker) Close() error {
+	// Invoking endSpan on Close will help catch the cases
+	// in which a read returned a non-nil error, we set the
+	// span status but didn't end the span.
```
MrAlias, line 140: This comment seems unrelated.

instrumentation/net/http/otelhttp/metrics.go
```diff
@@ -0,0 +1,143 @@
...
+
+	"go.opentelemetry.io/otel/semconv"
+
+	"go.opentelemetry.io/otel/unit"
+
+	"go.opentelemetry.io/otel/api/metric"
+	"go.opentelemetry.io/otel/label"
+)
+
+type metricsTransport struct {
```
MrAlias, line 32: Can we rename this to `instrumentedTransport`? We try to avoid using `metrics` as a noun in OpenTelemetry.

instrumentation/net/http/otelhttp/metrics.go
```diff
@@ -0,0 +1,143 @@
...
+}
+
+type tracker struct {
+	ctx     context.Context
+	start   time.Time
+	body    io.ReadCloser
+	endOnce sync.Once
+	labels  []label.KeyValue
+
+	// recorders
```
MrAlias, line 45: Can we remove this comment or updated it to describe what clientDurationRecorder is?

instrumentation/net/http/otelhttp/metrics.go
```diff
@@ -0,0 +1,143 @@
...
+	labels := semconv.HTTPClientAttributesFromHTTPRequest(req)
+
+	ctx := req.Context()
+	tracker := &tracker{
+		start:                  time.Now(),
+		ctx:                    ctx,
+		clientDurationRecorder: trans.clientDurationRecorder,
+	}
+
+	// Perform request.;
```
MrAlias, line 67: Not sure this comment is needed.

instrumentation/net/http/otelhttp/metrics.go
```diff
@@ -0,0 +1,143 @@
...
+	resp, err := trans.base.RoundTrip(req)
+	if err != nil {
+		tracker.labels = append(labels, semconv.HTTPAttributesFromHTTPStatusCode(http.StatusInternalServerError)...)
```
MrAlias, line 70: This should include the response status not `http.StatusInternalServerError`

instrumentation/net/http/otelhttp/metrics.go
```diff
@@ -0,0 +1,143 @@
...
+		tracker.end()
+	} else {
+		tracker.labels = append(labels, semconv.HTTPAttributesFromHTTPStatusCode(http.StatusInternalServerError)...)
```
MrAlias, line 73: Same here.

instrumentation/net/http/otelhttp/metrics.go
```diff
@@ -0,0 +1,143 @@
...
+	case i0:
+		return struct {
+			io.ReadCloser
+			io.Writer
+		}{wrapper, wr}
+	default:
+		return struct {
+			io.ReadCloser
+		}{wrapper}
+	}
```
MrAlias, line 104: This seems like a long way of writing:

```
if wr, ok := body.(io.Writer); ok {
	return struct {
		io.ReadCloser
		io.Writer
	}{wrapper, wr}
}
return wrapper
```

Am I missing something?

instrumentation/net/http/otelhttp/metrics.go
```diff
@@ -0,0 +1,143 @@
...
+}
+
+func (trans *metricsTransport) createMeasures() {
+	var err error
+	trans.clientDurationRecorder, err = trans.meter.NewFloat64ValueRecorder(
+		clientRequestDuration,
+		metric.WithDescription("measure the duration of the outbound HTTP request"),
```
MrAlias, line 111:
```suggestion
		metric.WithDescription("measures the duration of the outbound HTTP request"),
```

instrumentation/net/http/otelhttp/metrics.go
vchitai: Updated. Thanks for the suggestion.

instrumentation/net/http/otelhttp/metrics.go
vchitai: Updated. Thanks for the suggestion.

instrumentation/net/http/otelhttp/metrics.go
vchitai: Updated. Thanks for the suggestion.

instrumentation/net/http/otelhttp/metrics.go
vchitai: Updated. Thanks for the suggestion.

instrumentation/net/http/otelhttp/metrics.go
vchitai: Updated. Thanks for the suggestion.

instrumentation/net/http/otelhttp/metrics.go
vchitai: Removed. Thanks for the suggestion.

instrumentation/net/http/otelhttp/metrics.go
vchitai: In case of HTTP connection error occurred, should we recorded nothing or 500 as the HTTP status code. What's your opinion?

instrumentation/net/http/otelhttp/transport.go
vchitai: I prefer to split up the implementation of metrics and tracing parts for ease of maintenance.

PR #428, opened 2020-10-30
Fix/cortex export metric example

redthor:
Just trying out https://aws.amazon.com/blogs/opensource/building-a-prometheus-remote-write-exporter-for-the-opentelemetry-go-sdk/
However, couldn't get docker-compose to run up successfully on Ubuntu.

One error:
```
/go.mod: no such file or directory
```
fixed with ab10ca0 - don't use Replace to a parent folder which is not mapped inside the container.

The other was with the use of `host.docker.internal` which is available on mac (and windows?) but not linux. This SO post to the rescue:
https://stackoverflow.com/a/49149020

I couldn't use the 172.x.x.x address because it keeps changing. The qoomon/docker-host solution seems better.

### DIFF
--- a/exporters/metric/cortex/example/go.mod
+++ b/exporters/metric/cortex/example/go.mod
@@ -2,11 +2,6 @@ module go.opentelemetry.io/contrib/exporters/metric/cortex/example
 
 go 1.14
 
-replace (
-	go.opentelemetry.io/contrib/exporters/metric/cortex => ../
-	go.opentelemetry.io/contrib/exporters/metric/cortex/utils => ../utils/
-)
-
 require (
 	go.opentelemetry.io/contrib/exporters/metric/cortex v0.13.0
 	go.opentelemetry.io/contrib/exporters/metric/cortex/utils v0.13.0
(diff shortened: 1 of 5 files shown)


exporters/metric/cortex/example/go.mod
```diff
@@ -2,11 +2,6 @@ module go.opentelemetry.io/contrib/exporters/metric/cortex/example
 
 go 1.14
 
-replace (
-	go.opentelemetry.io/contrib/exporters/metric/cortex => ../
-	go.opentelemetry.io/contrib/exporters/metric/cortex/utils => ../utils/
-)
```
MrAlias, line 8: These replace statements are critical to include for development. When there are unreleased changes to the `go.opentelemetry.io/contrib/exporters/metric/cortex` and `go.opentelemetry.io/contrib/exporters/metric/cortex/utils` packages and we want to update the example prior to a release they are needed to refer to the correct and local code. Additionally, when we are making a release we will bump the version of these packages prior to them actually existing so the build needs to rely on the local code instead of a non-existent tagged remote release.

Can we update the example in a way where these are preserved?

redthor: I guess we could map the parent directories into the container. I can give it a go.

MrAlias:
The cortex exporter and example were remove in https://github.com/open-telemetry/opentelemetry-go-contrib/releases/tag/v1.5.0

Issue #380, opened 2020-10-01
Ensure links to contributed plugins are updated in OpenTelemetry registry


evantorrie:
https://opentelemetry.io/registry/ acts as a registry of contributed plugins. For plugins that exist in this repo, we should ensure that the links to them are kept up to date when

* new plugins are added
* plugin names and/or locations are changed

This ticket is to track updating them as of the current repo snapshot, but we may also consider creating an automated "link checker" to create/update the content of the files in the [registry](https://github.com/open-telemetry/opentelemetry.io/tree/master/content/registry).

matej-g:
I tried to collect all currently missing entries

Instrumentation:
- [x] instrumentation/github.com/astaxie/beego/otelbeego
- [x] instrumentation/github.com/bradfitz/gomemcache/memcache/otelmemcache
- [x] instrumentation/github.com/emicklei/go-restful/otelrestful
- [x] instrumentation/github.com/gin-gonic/gin/otelgin
- [x] instrumentation/github.com/gocql/gocql/otelgocql
- [x] instrumentation/github.com/gorilla/mux/otelmux
- [x] instrumentation/github.com/labstack/echo/otelecho
- [x] instrumentation/github.com/Shopify/sarama/otelsarama
- [x] instrumentation/go.mongodb.org/mongo-driver/mongo/otelmongo
- [x] instrumentation/google.golang.org/grpc/otelgrpc
- [x] instrumentation/gopkg.in/macaron.v1/otelmacaron
- [ ] instrumentation/host
- [x] instrumentation/net/http/httptrace/otelhttptrace
- [x] instrumentation/net/http/otelhttp
- [x] instrumentation/runtime

Exporters:
- [ ] exporters/metric/cortex
- [ ] exporters/metric/datadog
- [ ] exporters/metric/dogstatsd

Detectors (I guess those are considered utilities in the registry):
- [ ] detectors/aws
- [ ] detectors/gcp

As to more automated check, I can imagine having a `make` target to check existing plugins against current entries in the `opentelemetry.io/content/en/registry/` (perhaps this could be part of the pre-release script, to ensure the links are up to date before a release?). However, I'm not sure about automated updating of content, that would probably require us to somehow automate opening PRs with updates in the `opentelemetry.io` repo. Is that the idea?

MrAlias:
[...]

PR #446, opened 2020-11-09
Add registry link check target to Makefile

matej-g:
This PR is a proposal which addresses #380, i.e. creation of mechanism to check whether all plugins included within `contrib` repository have a valid link in the [OpenTelemetry Registry](https://github.com/open-telemetry/opentelemetry.io/tree/master/content/en/registry).

The proposal adds a target to the `Makefile`, which checks the links by querying individual `.md` files representing entries in the registry and ensuring these are up to date in the following manner:

1. Each plugin should have a properly structured filename in the registry, which should have the format `<plugin_type>-go-<name>.md`, where:
- `plugin_type` should be one of `instrumentation`, `exporter` or `detector`
- `name` is the name of the actual plugin package (stripped of `otel` prefix, if there is one)

2. The links in the registry should point to `master` and lead to:
- In case of `instrumentation`, to directory which is identical with the import path of the instrumented package (e.g. for `github.com/gorilla/mux` it should be `https://github.com/open-telemetry/opentelemetry-go-contrib/tree/master/instrumentation/github.com/gorilla/mux`)
- In case of other plugins, to the directory of the module (where `go.mod` resides)

This will require also adjusting current entries in the registry, as not all file names adhere to this proposed format.

If the check it unsuccessful, it prints out a warning with the list of plugins, which do not have up to date links. The check should be run as a part of the pre-release script.

Still to decide:
~~- Should this target be run as a part of the pre-release script? Should it be left to the maintainers discretion to run it?~~


evantorrie:
> * `plugin_type` should be one of `instrumentation`, `exporter` or `tool` (plugins under `detector` or other plugins which might not fit into the first two categories)

 My one nit here is whether we should have a catchall `tool`, or instead, just have another type called `detector` for now, and add other "types" as and when we run into them.

> Still to decide:
> 
> * Should this target be run as a part of the pre-release script? Should it be left to the maintainers discretion to run it?

I don't think it should *block* the pre-release script, but it would be good for it to kick out a warning about "hey, some plugins are not registered".  To fix erroneous entries will require creating a PR against the registry repo and getting it merged.  While response to registry PRs has usually been quick in my experience, I don't think it should be a blocker to our releasing a new version.

matej-g:
> My one nit here is whether we should have a catchall `tool`, or instead, just have another type called `detector` for now, and add other "types" as and when we run into them.

I'm fine with having only `detector` for now. I was conflating this with the categorization they have in the registry, but looks like we can keep the filename independent of registry category and including tags in the `.md` should take care of this (should be `utilities` component).

> I don't think it should _block_ the pre-release script, but it would be good for it to kick out a warning about "hey, some plugins are not registered". To fix erroneous entries will require creating a PR against the registry repo and getting it merged. While response to registry PRs has usually been quick in my experience, I don't think it should be a blocker to our releasing a new version.

Warning in pre-release sounds good :+1:

matej-g:
Thanks for the suggestions @evantorrie, these have now been applied. Opening the PR for wider review.

Aneurysm9 (approved):

MrAlias (approved):

MrAlias:
@evantorrie can you take another look?

matej-g:
Any update on this one? Shall we proceed?

PR #455, opened 2020-11-17
Add detector for Cloud Run service.

yegle:


yegle:
Ping, this one is ready to review.

Aneurysm9 (approved):

Aneurysm9:
Please add a `CHANGELOG.md` entry for this addition.

yegle:
> Please add a `CHANGELOG.md` entry for this addition.

Done. Thanks for the review!

XSAM (commented):
Should we add some tests for this detector?

yegle:
> Should we add some tests for this detector?

Unit test added.

CHANGELOG.md
```diff
@@ -11,6 +11,8 @@ This project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0.htm
 ### Added
 
 - `otelhttp.{Get,Head,Post,PostForm}` convenience wrappers for their `http` counterparts. (#390)
+- `gcp.CloudRun` detector for detecting resource from a Cloud Run
+  instance.
```
MrAlias, line 15:
```suggestion
- `gcp.CloudRun` detector for detecting resource from a Cloud Run instance. (#455)
```

detectors/gcp/cloud-run.go
```diff
@@ -0,0 +1,114 @@
...
+	onGCE  func() bool
+	getenv func(string) string
+}
+
+// compile time assertion that CloudRun implements the resource.Detector
+// interface.
+var _ resource.Detector = (*CloudRun)(nil)
+
+// NewCloudRun creates a CloudRun detector
+// Specify nil to use the default metadata client.
```
MrAlias, line 47: This seems stale.

detectors/gcp/cloud-run.go
```diff
@@ -0,0 +1,114 @@
...
+	mc     metadataClient
+	onGCE  func() bool
+	getenv func(string) string
+}
+
+// compile time assertion that CloudRun implements the resource.Detector
+// interface.
+var _ resource.Detector = (*CloudRun)(nil)
+
+// NewCloudRun creates a CloudRun detector
```
MrAlias, line 46:
```suggestion
// NewCloudRun creates a CloudRun detector.
```

detectors/gcp/cloud-run.go
```diff
@@ -0,0 +1,114 @@
...
+	// Part of Cloud Run container runtime contract.
+	// See https://cloud.google.com/run/docs/reference/container-contract
+	// The same K_SERVICE value ultimately maps to both `namespace` and
+	// `job` label of `generic_task` metric type.
+	if service := c.getenv("K_SERVICE"); service == "" {
+		errInfo = append(errInfo, "envvar K_SERVICE contains empty string.")
+	} else {
+		labels = append(labels,
+			semconv.ServiceNamespaceKey.String(service),
+			semconv.ServiceNameKey.String(service),
```
MrAlias, line 104: It seems inappropriate for the namespace to be the same as the service name. The namespace is intended to encapsulate grouping of services that might be duplicated on another resource.

Would it make sense to instead use the project ID as the namespace? Can you run different jobs on CloudRun with the same name in the same project?

From the [linked docs](https://cloud.google.com/run/docs/reference/container-contract#env-vars), would it make sense to parse `K_REVISION` instead? For example, from the example, the namespace would be `hello-world` and the service name would be `1`.

detectors/gcp/cloud-run_test.go
```diff
@@ -0,0 +1,164 @@
...
+	return c.Get("instance/id")
+}
+
+func (c *client) ProjectID() (string, error) {
+	return c.Get("project/project-id")
+}
+
+var _ metadataClient = (*client)(nil)
+
+func TestCloudRunDetector_NotOnGCE(t *testing.T) {
```
MrAlias, line 66: Can we remove the underscores from the test function names?

detectors/gcp/cloud-run.go
```diff
@@ -0,0 +1,114 @@
...
+	getenv func(string) string
+}
+
+// compile time assertion that CloudRun implements the resource.Detector
+// interface.
+var _ resource.Detector = (*CloudRun)(nil)
+
+// NewCloudRun creates a CloudRun detector
+// Specify nil to use the default metadata client.
+func NewCloudRun() *CloudRun {
```
MrAlias, line 48: Should this accept a `metadata.Client` from the user?

detectors/gcp/cloud-run_test.go
```diff
@@ -0,0 +1,164 @@
...
+	}
+}
+
+func TestCloudRunDetector_ExpectSuccess(t *testing.T) {
+	ctx := context.Background()
+
+	metadata := map[string]string{
+		"project/project-id": "foo",
+		"instance/id":        "bar",
+		"instance/region":    "utopia",
```
MrAlias, line 82: :+1: :smile:

detectors/gcp/cloud-run.go
yegle: The Cloud Run is a Knative compatible service so the Knative concepts should be referenced.

https://github.com/knative/serving/blob/master/docs/spec/overview.md#service is the concepts in Knative. The "name of the service" (in the linked doc it's `hello-world`) naturally maps to `ServiceNameKey`. The `ServiceNamespaceKey` doesn't have a direct mapping in Knative world.

If we must, I think we can hardcode ServiceNamespace to "cloud-run-managed". The user can have other knative clusters (e.g. "cloud-run-anthos" or "on-prem") that can be considered different namespaces.

Also cc @steren in case he has some input on this.

detectors/gcp/cloud-run_test.go
yegle: Fixed.

detectors/gcp/cloud-run.go
yegle: I think in most cases the user don't need to customize the metadata client.

In rare cases when they do, one way to customize is to override the metadata server URL in environment variable.

detectors/gcp/cloud-run.go
yegle: Fixed.

detectors/gcp/cloud-run.go
jmacd: +1 to "cloud-run-managed"

jmacd (approved):

detectors/gcp/cloud-run.go
yegle: Fixed.

yegle:
I added a note to suggest not relying on the value of service.namespace just yet, in case we want to change it in the future.

XSAM (approved):

yegle:
Can I get this PR merged? Thank you!

CHANGELOG.md
```diff
@@ -16,7 +16,8 @@ This project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0.htm
 - The AWS detector now adds the cloud zone, host image ID, host type, and host name to the returned `Resource`. (#410)
 - Add Amazon ECS Resource Detector for AWS X-Ray. (#466)
 - Add propagator for AWS X-Ray (#462)
-  
+- A new `gcp.CloudRun` detector for detecting resource from a Cloud Run instance. (#455)
```
MrAlias, line 19: This needs to be moved the `Unreleased` section.

MrAlias (approved):

CHANGELOG.md
yegle: Moved to Unreleased.

yegle:
Merged latest master changes into this branch. PTAL.

detectors/gcp/cloud-run.go
```diff
@@ -0,0 +1,124 @@
...
+	return &CloudRun{
+		mc:     metadata.NewClient(nil),
+		onGCE:  metadata.OnGCE,
+		getenv: os.Getenv,
+	}
+}
+
+// for test only
+func (c *CloudRun) setupForTest(mc metadataClient, ongce func() bool, getenv func(string) string) {
```
dashpole, line 60: This probably belongs in the _test.go file. In addition to making the detector more readable, it would ensure test functions are not used in production code.

dashpole (approved):

detectors/gcp/cloud-run.go
yegle: Moved to _test.go and made it a function instead of a method.

yegle:
Can I get this PR merged? Thank you!

PR #456, opened 2020-11-18
Add go-kit instrumentation library

### LINKED ISSUE #434: Add instrumentation for github.com/go-kit/kit
Opening an issue for Go kit based on the discussion in #432
I intend to work on this one and submit a PR.

https://github.com/go-kit/kit

## Tasks

* [ ]  code complete with tests passing
* [ ]  `Dockerfile` with example app showing instrumentation (alternatively link to https://github.com/sagikazarmark/modern-go-application which serves as an example app for Go kit, showcasing OpenTelemetry as well)
* [ ]  documentation updated
* [ ]  instrumentation added to the opentelemetry [registry](https://opentelemetry.io/registry/)

## Prior Art

Go kit provides built-in support for OpenCensus, I'd like to use that as a template (together with existing framework integrations in this repo):

https://github.com/go-kit/kit/tree/master/tracing/opencensus

Apart from HTTP and gRPC integrations, it contains go kit specific logic (eg. business error code).

sagikazarmark:
Fixes #434

## Tasks

* [x]  code complete with tests passing
* [x]  `Dockerfile` with example app showing instrumentation (alternatively link to https://github.com/sagikazarmark/modern-go-application which serves as an example app for Go kit, showcasing OpenTelemetry as well)
* [x]  documentation updated
* ~~instrumentation added to the opentelemetry [registry](https://opentelemetry.io/registry/)~~


instrumentation/github.com/go-kit/kit/otelkit/endpoint.go
```diff
@@ -0,0 +1,135 @@
...
+
+				// test for business error
+				// business errors are often  successful requests carrying
+				// a business failure that the client can act upon
+				if res, ok := response.(endpoint.Failer); ok && res.Failed() != nil {
+					span.RecordError(res.Failed())
+
+					if cfg.IgnoreBusinessError {
+						span.SetStatus(codes.Unset, "")
+					}
```
matej-g, line 122: If the `IgnoreBusinessError` option is `true`, is it intended to still add error event via `span.RecordError` and to just "reset" the status to `Unset`? Or should the span's events also pretend as if there was no error? I was just wondering, as the option description did not make it 100% clear to me.

instrumentation/github.com/go-kit/kit/otelkit/endpoint.go
```diff
@@ -0,0 +1,135 @@
...
+			defer func() {
+				if err != nil {
+					if lberr, ok := err.(lb.RetryError); ok {
+						// handle errors originating from lb.Retry
+						for idx, rawErr := range lberr.RawErrors {
+							span.AddEvent(
+								"error",
+								trace.WithAttributes(
+									label.String("error.type", "gokit.lb.retry"),
+									label.String("gokit.lb.retry.count", strconv.Itoa(idx+1)),
```
matej-g, line 97: Could this be `label.Int` instead or is there a reason for converting to string?

instrumentation/github.com/go-kit/kit/otelkit/endpoint.go
sagikazarmark: Yeah, that's a little bit confusing concept and is the result of using the original OpenCensus integration as an example.

The idea is to add the error to the span, but not mark the span as a failed one. This in theory will make the request itself successful when carrying an error only meaningful to the user (eg. a validation error), but doesn't indicate an error in the system. Chances are an HTTP middleware will still mark the request as a failed one (I guess), so I'm not sure it's worth keeping.

There is a comment about this a few lines above:

```
				// business errors are often successful requests carrying
				// a business failure that the client can act upon
				// and therefore do not count as failed requests
```

I added another line. Let me know what you think.

sagikazarmark:
Thanks for the review @matej-g 

I think it's enough for me as a sanity check to finish the PR, so an approver can also review and merge it.

jmacd (approved):

matej-g (approved):
Nice work @sagikazarmark, this PR looks complete to me now.

> What is the procedure for adding it to the registry?

There is another repository for `opentelemetry.io`, there you can find the "registry" under https://github.com/open-telemetry/opentelemetry.io/tree/master/content/en/registry. Basically once this PR is merged, you can create a PR for that repository which will add a new entry, something like `instrumentation-go-kit.md`, to represent your instrumentation package. You can checkout the other registry entries as a reference to know what your `.md` file should include.

sagikazarmark:
Thanks @matej-g 

I guess adding it to the registry is not a task for this PR, so I crossed it off.

XSAM (approved):

sagikazarmark:
I've rebased the PR to use the latest version. Is there anything I can do to get this merged? I'd really like to avoid manually rebasing and updating the version. :)

sagikazarmark:
Happy new year everyone!

Friendly ping to see if we can get this merged before the next release. 🙂

sagikazarmark:
Hey folks,

Can someone merge this PR please? I'd really like to stop rebasing it, it's a waste of effort. It's been approved by three reviewers. What else do I have to do to get it merged?

Thanks!

MrAlias (changes requested):
Overall this looks good, thanks for the contribution. The blocking issues are the dependabot config change, upgrading to v0.18.0, and the `RecordError` question needs to be addressed.

sagikazarmark:
@MrAlias thanks for the review. I addressed most of it.

Regarding `RecordError`: it sets the status for the span and technically it's not correct for the intermediate errors. I guess we can use it since the status will be set by the final error anyway. I think the behavior of `RecordError` was somewhat different when I first submitted this PR, so there might have been some other reason as well.

sagikazarmark:
I changed it to use `RecordError`. Not fully convinced this is right though. I think semantically `RecordError` should only be called once for the terminal error (given it actually sets the status to error which might not always be desirable), but I could be wrong.

instrumentation/github.com/go-kit/kit/otelkit/endpoint.go
```diff
@@ -0,0 +1,134 @@
...
+						// handle errors originating from lb.Retry
+						for idx, rawErr := range lberr.RawErrors {
+							span.AddEvent(
+								"error",
+								trace.WithAttributes(
+									label.String("error.type", "gokit.lb.retry"),
+									label.Int("gokit.lb.retry.count", idx+1),
+									label.String("error.message", rawErr.Error()),
+								),
+							)
```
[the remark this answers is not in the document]
sagikazarmark: See my comments below.

MrAlias (approved):

MrAlias:
> I changed it to use `RecordError`. Not fully convinced this is right though. I think semantically `RecordError` should only be called once for the terminal error (given it actually sets the status to error which might not always be desirable), but I could be wrong.

This might be something that needs to be clarified better. You are correct that the semantics of `RecordError` are hard to distinguish if it should be used here. Ultimately, the user of this instrumentation will want to send this telemetry to a backend that will need to understand the events that it receives. Sending these errors as events that are semantically tied to errors will provide the backend a chance to highlight these in helpful manner.

That said, OTel in general wanted to provide a better errors interface in the future so this is likely going to be improved.

Thanks for the contribution and patience!

sagikazarmark:
Thanks for merging

PR #465, opened 2020-11-23
Add Amazon EKS Resource Detector

wilguo:
This PR adds an EKS Resource Detector.

#### Functionalities:
* If the EKS resource detector detects that it is running on EKS, then it will populate the [`resource`](https://github.com/open-telemetry/opentelemetry-specification/blob/master/specification/resource/sdk.md) object with metadata. The metadata will include the `containerId` and `clusterName`.
* If the EKS resource detector detects that the application is not running on EKS, then it will return an empty `Resource`

#### List of Changes:
* Add functionality for Amazon EKS resource detector defined above
* Add corresponding unit tests for Amazon EKS resource detector to detect when running in EKS and when running not in EKS


jmacd (approved):

XSAM (approved):

detectors/aws/eks/eks_resource_detector.go
```diff
@@ -0,0 +1,243 @@
...
+}
+
+// fetchString executes an HTTP request with a given HTTP Method and URL string
+// this function implements the detectorUtils interface
+func (eksUtils eksDetectorUtils) fetchString(httpMethod string, URL string) (string, error) {
+
+	// Create new HTTP request object
+	request, err := http.NewRequest(httpMethod, URL, nil)
+	if err != nil {
+		return "", fmt.Errorf("failed to create new HTTP request with method=%s, URL=%s", httpMethod, URL)
```
Aneurysm9, line 154:
```suggestion
		return "", fmt.Errorf("failed to create new HTTP request with method=%s, URL=%s: %w", httpMethod, URL, err)
```

detectors/aws/eks/eks_resource_detector.go
```diff
@@ -0,0 +1,243 @@
...
+
+const (
+	k8sSvcURL          = "https://kubernetes.default.svc"
+	k8sTokenPath       = "/var/run/secrets/kubernetes.io/serviceaccount/token"
+	k8sCertPath        = "/var/run/secrets/kubernetes.io/serviceaccount/ca.crt"
+	authConfigmapPath  = "/api/v1/namespaces/kube-system/configmaps/aws-auth"
+	cwConfigmapPath    = "/api/v1/namespaces/amazon-cloudwatch/configmaps/cluster-info"
+	defaultCgroupPath  = "/proc/self/cgroup"
+	containerIDLength  = 64
+	millisecondTimeOut = 2000
```
Aneurysm9, line 42: Perhaps this should be `timeoutMillis`?  When I see `millisecondTimeOut` used in the code I think it is set for 1ms.

detectors/aws/eks/eks_resource_detector.go
```diff
@@ -0,0 +1,243 @@
...
+			TLSClientConfig: &tls.Config{
+				RootCAs: caCertPool,
+			},
+		},
+	}
+
+	// Execute HTTP request
+	response, err := client.Do(request)
+	if err != nil {
+		return "", fmt.Errorf("failed to execute HTTP request with method=%s, URL=%s", httpMethod, URL)
```
Aneurysm9, line 185:
```suggestion
		return "", fmt.Errorf("failed to execute HTTP request with method=%s, URL=%s: %w", httpMethod, URL, err)
```

detectors/aws/eks/eks_resource_detector.go
```diff
@@ -0,0 +1,243 @@
...
+	}
+
+	// Retrieve response body from HTTP request
+	body, err := ioutil.ReadAll(response.Body)
+	if err != nil {
+		return "", fmt.Errorf("failed to read response from HTTP request with method=%s, URL=%s", httpMethod, URL)
```
Aneurysm9, line 191:
```suggestion
		return "", fmt.Errorf("failed to read response from HTTP request with method=%s, URL=%s: %w", httpMethod, URL, err)
```

detectors/aws/eks/eks_resource_detector.go
```diff
@@ -0,0 +1,243 @@
...
+			},
+		},
+	}
+
+	// Execute HTTP request
+	response, err := client.Do(request)
+	if err != nil {
+		return "", fmt.Errorf("failed to execute HTTP request with method=%s, URL=%s", httpMethod, URL)
+	}
+
```
Aneurysm9, line 187: Should there be a response status check here?  Is a `200` response expected, or would anything `2xx` be ok?

detectors/aws/eks/eks_resource_detector.go
```diff
@@ -0,0 +1,243 @@
...
+	return "Bearer " + string(content), nil
+}
+
+// getClusterName retrieves the clusterName resource attribute
+func getClusterName(utils detectorUtils) (string, error) {
+	resp, err := utils.fetchString("GET", k8sSvcURL+cwConfigmapPath)
+	if err != nil {
+		return "", fmt.Errorf("getClusterName() error: %s", err.Error())
```
Aneurysm9, line 211:
```suggestion
		return "", fmt.Errorf("getClusterName() error: %w", err)
```

detectors/aws/eks/eks_resource_detector.go
```diff
@@ -0,0 +1,243 @@
...
+	}
+
+	// parse JSON object returned from HTTP request
+	var parsedResp JSONResponse
+	err = json.Unmarshal([]byte(resp), &parsedResp)
+	if err != nil {
+		return "", fmt.Errorf("getClusterName() error: cannot parse JSON %s", resp)
```
Aneurysm9, line 218:
```suggestion
		return "", fmt.Errorf("getClusterName() error: cannot parse JSON: %w", w)
```

Putting the entire response into the error message might be a bit much.

detectors/aws/eks/eks_resource_detector.go
```diff
@@ -0,0 +1,243 @@
...
+}
+
+// getContainerID retrieves the containerID resource attribute
+// this function implements the detectorUtils interface
+func (eksUtils eksDetectorUtils) getContainerID() (string, error) {
+
+	// Read file
+	fileData, err := ioutil.ReadFile(defaultCgroupPath)
+	if err != nil {
+		return "", fmt.Errorf("getContainerID() error: cannot read file with path %s", defaultCgroupPath)
```
Aneurysm9, line 232:
```suggestion
		return "", fmt.Errorf("getContainerID() error: cannot read file with path %s: %w", defaultCgroupPath, err)
```

detectors/aws/eks/eks_resource_detector.go
```diff
@@ -0,0 +1,243 @@
...
+
+// getClusterName retrieves the clusterName resource attribute
+func getClusterName(utils detectorUtils) (string, error) {
+	resp, err := utils.fetchString("GET", k8sSvcURL+cwConfigmapPath)
+	if err != nil {
+		return "", fmt.Errorf("getClusterName() error: %s", err.Error())
+	}
+
+	// parse JSON object returned from HTTP request
+	var parsedResp JSONResponse
```
Aneurysm9, line 215: I'm not sure exporting the `JSONResponse` type is necessary or advisable here.  Could you instead use a `map[string]interface{}` to unmarshal the JSON and pull the expected values out of that?

wilguo: Yep! I have changed it to use `map[string]interface{}`

detectors/aws/eks/eks_resource_detector.go
```diff
@@ -0,0 +1,242 @@
...
+	if err != nil {
+		return "", fmt.Errorf("getK8sCredHeader() error: cannot read file with path %s", k8sTokenPath)
+	}
+
+	return "Bearer " + string(content), nil
+}
+
+// getClusterName retrieves the clusterName resource attribute
+func getClusterName(utils detectorUtils) (string, error) {
+	resp, err := utils.fetchString("GET", k8sSvcURL+cwConfigmapPath)
```
dashpole, line 202: Is there a reason you aren't using kubernetes' client-go library for this?

dashpole (commented):
Non-blocking, but I think it would be best to eventually have a generic "kubernetes" resource detector to grab things like pod name, namespace, container id, etc.  That way each cloud provider doesn't re-implement the same things.

Things like the cluster name are cloud-provider-specific, so this will still need to exist.  But hopefully the generic resource detector could add most attributes.

detectors/aws/eks/eks_resource_detector.go
wilguo: I used the same implementation that was found in [Java](https://github.com/open-telemetry/opentelemetry-java/blob/master/sdk-extensions/aws/src/main/java/io/opentelemetry/sdk/extension/aws/resource/EksResource.java) and [JS](https://github.com/open-o11y/opentelemetry-js/pull/2/files#diff-fc874f90e62b5fb67cbb9c77c66ac462f581291f75a9028d5e95441784e1cfa5).

alolita:
@MrAlias @Aneurysm9 Can you please merge this change? We're blocked on this merge. Ty!

detectors/aws/eks/eks_resource_detector.go
```diff
@@ -0,0 +1,242 @@
...
+		return "", fmt.Errorf("getContainerID() error: cannot read file with path %s: %w", defaultCgroupPath, err)
+	}
+
+	// Retrieve containerID from file
+	splitData := strings.Split(strings.TrimSpace(string(fileData)), "\n")
+	for _, str := range splitData {
+		if len(str) > containerIDLength {
+			return str[len(str)-containerIDLength:], nil
+		}
+	}
```
MrAlias, line 240: This seems like an error prone algorithm to determining the container ID. If I understand this correctly it is looking for a line longer than some length and then assuming that line ends with the container ID?

Based on this:

```
docker run --rm debian cat /proc/self/cgroup
12:cpu,cpuacct:/docker/422a3caf71aecd248eb39e2d6e2b4490992a7703e995745ca0ba4ae989f73e88
11:devices:/docker/422a3caf71aecd248eb39e2d6e2b4490992a7703e995745ca0ba4ae989f73e88
10:pids:/docker/422a3caf71aecd248eb39e2d6e2b4490992a7703e995745ca0ba4ae989f73e88
9:cpuset:/docker/422a3caf71aecd248eb39e2d6e2b4490992a7703e995745ca0ba4ae989f73e88
8:net_cls,net_prio:/docker/422a3caf71aecd248eb39e2d6e2b4490992a7703e995745ca0ba4ae989f73e88
7:memory:/docker/422a3caf71aecd248eb39e2d6e2b4490992a7703e995745ca0ba4ae989f73e88
6:hugetlb:/docker/422a3caf71aecd248eb39e2d6e2b4490992a7703e995745ca0ba4ae989f73e88
5:freezer:/docker/422a3caf71aecd248eb39e2d6e2b4490992a7703e995745ca0ba4ae989f73e88
4:rdma:/
3:blkio:/docker/422a3caf71aecd248eb39e2d6e2b4490992a7703e995745ca0ba4ae989f73e88
2:perf_event:/docker/422a3caf71aecd248eb39e2d6e2b4490992a7703e995745ca0ba4ae989f73e88
1:name=systemd:/docker/422a3caf71aecd248eb39e2d6e2b4490992a7703e995745ca0ba4ae989f73e88
0::/system.slice/containerd.service
```

Would it make more sense to look for lines that match `^.*/docker/(.+)$` instead? That or better document our assumptions as to why we think length is appropriate to match on.

detectors/aws/eks/eks_resource_detector_test.go
```diff
@@ -0,0 +1,98 @@
+// Copyright The OpenTelemetry Authors
```
MrAlias, line 1: Can we rename this to `detector_test.go`?

detectors/aws/eks/eks_resource_detector_test.go
```diff
@@ -0,0 +1,98 @@
...
+	// Expected resource object
+	eksResourceLabels := []label.KeyValue{
+		semconv.K8SClusterNameKey.String("my-cluster"),
+		semconv.ContainerIDKey.String("0123456789A"),
+	}
+	expectedResource := resource.New(eksResourceLabels...)
+
+	// Call EKS Resource detector to detect resources
+	eksResourceDetector := ResourceDetector{detectorUtils}
+	resourceObj, _ := eksResourceDetector.Detect(context.Background())
```
MrAlias, line 72:
```suggestion
	resourceObj, _ := eksResourceDetector.Detect(context.Background())
```
```suggestion
	resourceObj, err := eksResourceDetector.Detect(context.Background())
	require.NoError(t, err)
```

detectors/aws/eks/eks_resource_detector_test.go
```diff
@@ -0,0 +1,98 @@
...
+
+	assert.Equal(t, expectedResource, resourceObj, "Resource object returned is incorrect")
+	detectorUtils.AssertExpectations(t)
+}
+
+// Tests EKS resource detector not running in EKS environment
+func TestNotEks(t *testing.T) {
```
MrAlias, line 79:
```suggestion
func TestNotEKS(t *testing.T) {
```

detectors/aws/eks/eks_resource_detector_test.go
```diff
@@ -0,0 +1,98 @@
...
+
+	detectorUtils := new(MockDetectorUtils)
+
+	// Set EKS resource detector variables
```
MrAlias, line 83:
```suggestion
```

detectors/aws/eks/eks_resource_detector_test.go
```diff
@@ -0,0 +1,98 @@
...
+	detectorUtils.On("fileExists", k8sTokenPath).Return(false)
+
+	// Expected resource object
+	expectedResource := resource.Empty()
+
+	// Call EKS Resource detector to detect resources
+	eksResourceDetector := ResourceDetector{detectorUtils}
+	resourceObj, _ := eksResourceDetector.Detect(context.Background())
+
+	assert.Equal(t, expectedResource, resourceObj, "Resource object should be empty")
```
MrAlias, line 96:
```suggestion
	r, err := ResourceDetector{detectorUtils}.Detect(context.Background())
	require.NoError(t, err)
	assert.Equal(t, resource.Empty(), r, "Resource object should be empty")
```

MrAlias (changes requested):
The module path is a blocking change needed.

detectors/aws/eks/eks_resource_detector.go
wilguo: Yep that would make more sense! I have changed the implementation to use regex instead. Thanks!

detectors/aws/eks/eks_resource_detector.go
wilguo: I created a new feature request for this issue linked here https://github.com/open-telemetry/opentelemetry-go-contrib/issues/478.

detectors/aws/eks/detector.go
```diff
@@ -0,0 +1,234 @@
...
+}
+
+// getContainerID returns the containerID if currently running within a container.
+func (eksUtils eksDetectorUtils) getContainerID() (string, error) {
+	fileData, err := ioutil.ReadFile(defaultCgroupPath)
+	if err != nil {
+		return "", fmt.Errorf("getContainerID() error: cannot read file with path %s: %w", defaultCgroupPath, err)
+	}
+
+	r, _ := regexp.Compile(`^.*/docker/(.+)$`)
```
MrAlias, line 224:
```suggestion
	r, err := regexp.Compile(`^.*/docker/(.+)$`)
	if err != nil {
		return "", err
	}
```

detectors/aws/eks/detector.go
```diff
@@ -0,0 +1,234 @@
...
+
+	// Retrieve containerID from file
+	splitData := strings.Split(strings.TrimSpace(string(fileData)), "\n")
+	for _, str := range splitData {
+		if isMatch := r.MatchString(str); isMatch {
```
MrAlias, line 229:
```suggestion
		if r.MatchString(str) {
```

MrAlias (approved):

MrAlias:
cycling to see if it clears the checks.

wilguo:
Should I try and trigger a new commit for circleci to run again?

MrAlias:
> Should I try and trigger a new commit for circleci to run again?

Should be all set.

PR #467, opened 2020-11-24
Move opencensus binary propagation to contrib

dashpole:
Issue: https://github.com/open-telemetry/opentelemetry-go/issues/93

This was originally added to opentelemetry-go in https://github.com/open-telemetry/opentelemetry-go/pull/1334, but we decided it should be moved to contrib.

This copies the binary propagation, and adds a new go module so that opencensus dependencies aren't added to users of other propagation formats.


propagators/opencensus_binary/go.mod
```diff
@@ -0,0 +1,9 @@
+module go.opentelemetry.io/otel/bridge/opencensus/binary
```
MrAlias, line 1: Needs to be updated.

propagators/opencensus_binary/binary.go
```diff
@@ -0,0 +1,86 @@
+// Copyright The OpenTelemetry Authors
```
MrAlias, line 1: I would prefer to name the directory just `opencensus` and update the package name to match. That way the use of this propagator will be `opencensus.Binary`.

propagators/opencensus_binary/go.mod
dashpole: done

propagators/opencensus_binary/binary.go
dashpole: done

propagators/opencensus_binary/go.mod
dashpole: done

MrAlias (approved):

Aneurysm9 (approved):

XSAM (approved):

PR #477, opened 2020-12-03
otelmemcache: Simplify config and span status setting

### LINKED ISSUE #408: otelmemcache: Simplify and correct status code setting and attributes
There are some issues currently with the `otelmemcache` instrumentation:

1) The instrumentation should not be setting `service.name` attribute - this was never required by the spec and is redundant.

2) Since the specification now recognizes only three statuses - `Unset` / `Error` / `Ok` - the method `memcacheErrToStatusCode` to create quasi-error mapping is no longer needed (see https://github.com/open-telemetry/opentelemetry-specification/blob/master/specification/trace/api.md#set-status). Also, according to the specification:
> Generally, Instrumentation Libraries SHOULD NOT set the status code to Ok, unless explicitly configured to do so. Instrumention libraries SHOULD leave the status code as Unset unless there is an error, as described above.

I interpret this to mean that `memcache` instrumentation [should not be setting the status to `Ok`](https://github.com/open-telemetry/opentelemetry-go-contrib/blob/15f50e00ebfa4faa8721ad2a925e8328c8f003cd/instrumentation/github.com/bradfitz/gomemcache/memcache/otelmemcache/status.go#L26) even if no error is detected. This should be then fixed as well.

matej-g:
There are a couple of things with `otelmemcache` instrumentation that should be adjusted: to make it adhere to the specification on one hand, and to remove some unnecessary parts on the other. Specifically:
- Remove service name from configuration and from span attributes (it is redundant)
- Do not set span status to OK; remove unnecessary error-mapping method

Resolves #408


instrumentation/github.com/bradfitz/gomemcache/memcache/otelmemcache/config.go
```diff
@@ -18,25 +18,11 @@ import (
...
 )
 
-type config struct {
-	serviceName    string
-	tracerProvider oteltrace.TracerProvider
-}
-
 // Option is used to configure the client.
-type Option func(*config)
+type Option oteltrace.TracerProvider
```
dashpole, line 22: We should probably stick with a generic Option type like we had before.  Even though we only have a single option now, we may want to add others in the future.

matej-g: Yes, now I see this could potentially be a breaking change, good point :+1:. I'll keep the type as it was.

jmacd (approved):

XSAM (approved):

CHANGELOG.md
```diff
@@ -8,6 +8,13 @@ This project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0.htm
 
 ## [Unreleased]
 
+### Fixed
+
+- `otelmemcache` no longer sets span status to OK instead of leaving it unset. (#477)
+
+### Removed
+
+- Remove service name from `otelmemcache` configuration and span attributes. (#477)
```
XSAM, line 17:
```suggestion
- Remove service name from `otelmemcache` configuration and span attributes. (#477)

```
nit

PR #487, opened 2020-12-11
add ability to disable tracing ping requests in mongo

mdevilliers:
This PR allows the mongo instrumentor to ignore the traces for "ping" queries.

I've implemented it to ensure the default is to ignore 'ping' queries as we have found these traces to be of such low value as not to be of interest and the volume to be so great as to swamp out any generic algorithm to stop propagation.

I'm not too sure if that is the correct thing to do but figured I can change it quite easily.


jmacd:
This looks good. Please add a CHANGELOG entry.

mdevilliers:
@jmacd I've added the CHANGELOG entry.

XSAM (approved):

PR #500, opened 2020-12-21
Add instrumentation request issue templates

### LINKED ISSUE #433: Create instrumentation request documentation and Issue template
We expect in the post-RC phase to have a large number of these requests. We need to be sure to document this process.

What is needed...

* Documentation of when instrumentation belongs here. Should be OSS and should be a plug-able format (if it is native instrumentation it belongs in the library itself).
* How to make a suggestion. We should have an issue template for this.
* Document how to "vote" on these requests so we can gauge what to work on.

We will likely also need to sync our labels to match project standards so these types of requests can be found from the OpenTelemetry project level.

MrAlias:
Closes #433

### DIFF
--- a/.github/ISSUE_TEMPLATE/instrumentation-request.md
+++ b/.github/ISSUE_TEMPLATE/instrumentation-request.md
@@ -1,0 +1,58 @@
+---
+name: Instrumentation Request
+about: Suggest instrumentation to include in this project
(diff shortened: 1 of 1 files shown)


jmacd (approved):
:100:

XSAM (approved):

PR #512, opened 2021-01-06
Filter predicate to skip grpc trace

### LINKED ISSUE #355: Investigate GRPC Tracer to support filtering or exclude the Health Checking Protocol.
If my GRPC service implements the [GRPC Health Checking Protocol](https://github.com/grpc/grpc/blob/master/doc/health-checking.md) the GRPC Client library and my Kubernetes health probe call it regularly and generates a "noise" only. I'd like to filter out them the same way I could filter in HTTP the Health check. It could also be part of the library as bool option to ignore these calls.

Something like:

```golang
if stream := grpc.ServerTransportStreamFromContext(ctx); stream != nil {
  if stream.Method() == "/grpc.health.v1.Health/Check" || stream.Method() == "/grpc.health.v1.Health/Watch" {
    //Ignore the trace
  }
}
```

See HTTP [Filters](https://github.com/open-telemetry/opentelemetry-go-contrib/blob/master/instrumentation/net/http/config.go#L39)

```golang
// config represents the configuration options available for the http.Handler
// and http.Transport types.
type config struct {
	Filters           []Filter
}
```

rcrowe:
Fixes #355

A use-case that is common is to filter out traces before sending to the collector, for example...

```go
WithFilter(func(ctx context.Context, fullMethodName string) bool {
	return !strings.Contains(fullMethodName, "grpc.health.v1.Health/Check")
})
```

Opening this early to get feedback on the approach, before fixing up tests & changelog.

### DIFF
--- a/instrumentation/google.golang.org/grpc/otelgrpc/grpctrace.go
+++ b/instrumentation/google.golang.org/grpc/otelgrpc/grpctrace.go
@@ -35,6 +35,7 @@ const (
 
 // config is a group of options for this instrumentation.
 type config struct {
+	Filters        []Filter
 	Propagators    propagation.TextMapPropagator
 	TracerProvider trace.TracerProvider
 }
@@ -54,6 +55,21 @@ func newConfig(opts []Option) *config {
 		o.Apply(c)
 	}
 	return c
+}
+
+// Filter is a predicate used to determine whether a given gRPC request should
+// be traced. A Filter must return true if the request should be traced.
+type Filter func(ctx context.Context, fullMethodName string) bool
+
+type filtersOption struct{ f []Filter }
+
+func (o filtersOption) Apply(c *config) {
+	c.Filters = append(c.Filters, o.f...)
+}
+
+// WithFilter returns an Option that applies a filtering to intercepted requests
+func WithFilter(f ...Filter) Option {
+	return filtersOption{f: f}
 }
 
 type propagatorsOption struct{ p propagation.TextMapPropagator }
--- a/instrumentation/google.golang.org/grpc/otelgrpc/interceptor.go
+++ b/instrumentation/google.golang.org/grpc/otelgrpc/interceptor.go
@@ -75,10 +75,18 @@ func UnaryClientInterceptor(opts ...Option) grpc.UnaryClientInterceptor {
 		invoker grpc.UnaryInvoker,
 		callOpts ...grpc.CallOption,
 	) error {
+		cfg := newConfig(opts)
+
+		for _, f := range cfg.Filters {
+			if !f(ctx, method) {
+				return invoker(ctx, method, req, reply, cc, callOpts...)
+			}
+		}
+
 		requestMetadata, _ := metadata.FromOutgoingContext(ctx)
 		metadataCopy := requestMetadata.Copy()
 
-		tracer := newConfig(opts).TracerProvider.Tracer(
+		tracer := cfg.TracerProvider.Tracer(
 			instrumentationName,
 			trace.WithInstrumentationVersion(otelcontrib.SemVersion()),
 		)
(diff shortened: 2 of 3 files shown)


instrumentation/google.golang.org/grpc/otelgrpc/grpctrace.go
```diff
@@ -56,6 +57,21 @@ func newConfig(opts []Option) *config {
 	return c
 }
 
+// Filter is a predicate used to determine whether a given gRPC request should
+// be traced. A Filter must return true if the request should be traced.
+type Filter func(ctx context.Context, fullMethodName string) bool
```
Aneurysm9, line 62: Is there a more general interface that could be used here? See, e.g., the [`otelhttp.Filter`](https://github.com/open-telemetry/opentelemetry-go-contrib/blob/master/instrumentation/net/http/otelhttp/common.go#L41), which accepts an `http.Request`  and thus enables a [wide range of filters](https://github.com/open-telemetry/opentelemetry-go-contrib/blob/master/instrumentation/net/http/otelhttp/filters/filters.go) to be implemented.  Only being able to inspect the method name seems limiting.

rcrowe: For sure, I just wasn't sure given the differences between unary/stream across the client & server how to come to something consistent.

[OpenTracing](https://github.com/grpc-ecosystem/go-grpc-middleware/blob/48900393c7f37e2136684cf7b0657940f001048f/tracing/opentracing/options.go#L22) currently just uses method name.

gunturaf: I think `ctx context.Context, fullMethodName string` is sufficient considering that all UnaryServerInterceptor, UnaryClientInterceptor, StreamClientInterceptor and StreamServerInterceptor has both values in common.

Or to give more flexibility to users, we might go with approach like this:

```go
type Filter func(ctx context.Context, unaryInfo *grpc.UnaryServerInfo, steamInfo *grpc.StreamServerInfo) bool
```
then for UnaryServerInterceptor we can do like:
```go
for _, f := range cfg.Filters {
	if !f(ctx, info, &grpc.StreamServerInfo{}) {
		return handler(ctx, req)
	}
}
```
and for StreamServerInterceptor:
```go
for _, f := range cfg.Filters {
	if !f(ctx, &grpc.UnaryServerInfo{}, info) {
		return handler(srv, ss)
	}
}
```

With the above approach we only handle cases for server interceptors, because client interceptors did not have those values.
Should we have different interface for client interceptors? So there are `WithServerFilters` and `WithClientFilters`.
Considering that there are notable differences in `grpc.UnaryClientInterceptor` which is
```go
func(ctx context.Context,
	method string,
	req, reply interface{},
	cc *grpc.ClientConn,
	invoker grpc.UnaryInvoker,
	callOpts ...grpc.CallOption,
)
```
and `grpc.StreamClientInterceptor` which is
```go
func(ctx context.Context,
	desc *grpc.StreamDesc,
	cc *grpc.ClientConn,
	method string,
	streamer grpc.Streamer,
	callOpts ...grpc.CallOption,
)
```
now should we support all interesting values from both structs to our interface?

gunturaf: Or maybe this can be a simpler alternative?
```go
type FilterInfo struct {
	Context	context.Context
	Method	string
	IsServer	bool
	IsClient	bool
	// more fields we can support here later
}

type Filter func(ctx context.Context, info FilterInfo) bool
```
We can wait-and-see to expose more information in `FilterInfo` later on.

rcrowe:
I accomplished this using a sampler. For example `trace.WithSampler(ignoreNameSampler{})`

I'm closing as I didn't want to leave a draft open and keep it up to date. Was clear I hadn't settled on the right API and there's an already supported mechanism.

yuraxdrumz:
@rcrowe Hey, do you have an example of how you accomplished filtering our Health messages? They are adding a lot of "noise" to my datasource.
Thanks

PR #514, opened 2021-01-07
Merge code coverage output into single file for codecov.io

### LINKED ISSUE #300: Add codecov visualization/reporting to contrib repo
Similar to what was done for the main repo: https://github.com/open-telemetry/opentelemetry-go/pull/995

evantorrie:
Fixes #300

This fixes an issue where code coverage was not being reported.

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -56,7 +56,7 @@ test-with-coverage:
 	    go tool cover -html=coverage.out -o coverage.html); \
 	  [ -f "$${dir}/coverage.out" ] && cat "$${dir}/coverage.out" >> coverage.txt; \
 	done; \
-	sed -i.bak -e '2,$$ { /^mode: /d; }' coverage.txt && rm coverage.bak
+	sed -i.bak -e '2,$$ { /^mode: /d; }' coverage.txt && rm coverage.txt.bak
 
 .PHONY: ci
 ci: precommit check-clean-work-tree test-with-coverage test-386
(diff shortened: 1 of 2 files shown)


Aneurysm9 (approved):

Makefile
```diff
@@ -47,12 +47,17 @@ precommit: dependabot-check license-check generate build lint test
...
+	for dir in $(ALL_COVERAGE_MOD_DIRS); do \
 	  echo "go test ./... + coverage in $${dir}"; \
 	  (cd "$${dir}" && \
 	    $(GOTEST_WITH_COVERAGE) ./... && \
-	    go tool cover -html=coverage.txt -o coverage.html); \
-	done
+	    go tool cover -html=coverage.out -o coverage.html); \
+	  [ -f "$${dir}/coverage.out" ] && cat "$${dir}/coverage.out" >> coverage.txt; \
+	done; \
+	sed -i.bak -e '2,$$ { /^mode: /d; }' coverage.txt; \
```
MrAlias, line 59: Is the back up file needed here or can we remove that?

Makefile
```diff
@@ -47,12 +47,17 @@ precommit: dependabot-check license-check generate build lint test
...
+	ls -l
```
MrAlias, line 60: This seems like it will bloat the output a bit. Can we refine the listing to the coverage files or remove this?

Makefile
evantorrie: Removed by having a null extension backup.  

Yes, there's the warning about how this will cause a corrupted file if the disk is full, but more likely our build will have other problems if something like that happens -- an invalid coverage file is probably the least of our worries.

MrAlias (approved):

Makefile
MrAlias: Hmm, looks like that caused this error:

```
sed: can't read : No such file or directory
```

This isn't a critical ask if you just want to revert back to the working state. Sorry to cause this issue.

PR #569, opened 2021-02-07
Fix ECS resource detector bug

bhautikpip:
1. Fixes nil pointer bad reference panic when initialize resource detector as below.

```
ecsResourceDetector := new(ecs.ResourceDetector)
```

probably now with the fix we can initialize resource detector as below:
```
ecsResourceDetector := ecs.NewResourceDetector()
```

2. Minor changes
Reference Issue: https://github.com/aws-observability/aws-otel-go/issues/16


bonclay7:
Following the issue I've opened [here](aws-observability/aws-otel-go#16), I went ahead and did a test with this implementation and it working very great for me on ECS with the following modifications

```
replace go.opentelemetry.io/contrib/detectors/aws/ecs => github.com/bhautikpip/opentelemetry-go-contrib/detectors/aws/ecs v0.16.1-0.20210207012249-86c04d1b71f1
```

```go
ecsResourceDetector := ecs.ResourceDetector{Utils: new(ecs.DetectorUtils)}
ecsResource, err := ecsResourceDetector.Detect(ctx)
```

image

detectors/aws/ecs/ecs.go
```diff
@@ -43,25 +43,19 @@ var (
...
 	getContainerID() (string, error)
 }
 
-// struct implements detectorUtils interface
-type ecsDetectorUtils struct{}
+// struct implements detectorUtilsResources interface
+type DetectorUtils struct{}
 
 // resource detector collects resource information from Elastic Container Service environment
 type ResourceDetector struct {
```
XSAM, line 55: Is there a reason that we should export `ResourceDetector`? Should we provide an unexported struct `resourceDetector ` and provide a function called this `NewResourceDetector`?

detectors/aws/ecs/ecs.go
```diff
@@ -43,25 +43,19 @@ var (
...
-	utils detectorUtils
+	Utils detectorUtilsResources
 }
 
-// compile time assertion that ecsDetectorUtils implements detectorUtils interface
```
XSAM, line 59: Why remove these interface assertions?

XSAM (changes requested):

detectors/aws/ecs/ecs.go
MrAlias: :100:

bhautikpip: Good call. Added `NewResourceDetector` function.

detectors/aws/ecs/ecs.go
bhautikpip: Added assertions back as well.

detectors/aws/ecs/ecs.go
bhautikpip: Just the linter is failing with this error `exported func NewResourceDetector returns unexported type go.opentelemetry.io/contrib/detectors/aws/ecs.resourceDetector, which can be annoying to use (golint)`

bhautikpip: Probably either we can export this struct or maybe we can use `exclude-rules` for this. I think we should export the struct. Any suggestions?

MrAlias: @bhautikpip having `NewResourceDetector` return a [`resource.Detector`](https://github.com/open-telemetry/opentelemetry-go/blob/e50a1c8cecb8c584bdfada6cb21fe0b173a22646/sdk/resource/auto.go#L30-L38) instead of the unexported struct will solve this issue.

bhautikpip: Done @MrAlias

detectors/aws/ecs/ecs.go
```diff
@@ -43,27 +43,32 @@ var (
 )
 
 // Create interface for methods needing to be mocked
-type detectorUtils interface {
+type detectorUtilsResources interface {
```
MrAlias, line 46: This rename is not related to the nil reference bug fix. Please leave them the way they were.

detectors/aws/ecs/ecs.go
```diff
@@ -43,27 +43,32 @@ var (
...
 	getContainerName() (string, error)
 	getContainerID() (string, error)
 }
 
-// struct implements detectorUtils interface
-type ecsDetectorUtils struct{}
+// struct implements detectorUtilsResources interface
+type DetectorUtils struct{}
```
MrAlias, line 52: This should remain unexported and can we revert the rename to narrow the changes in this PR to solely address the bug.

detectors/aws/ecs/ecs.go
```diff
@@ -43,27 +43,32 @@ var (
...
-// compile time assertion that ecsDetectorUtils implements detectorUtils interface
-var _ detectorUtils = (*ecsDetectorUtils)(nil)
+// compile time assertion that ecsDetectorUtils implements detectorUtilsResources interface
+var _ detectorUtilsResources = (*DetectorUtils)(nil)
 
 // compile time assertion that resource detector implements the resource.Detector interface.
-var _ resource.Detector = (*ResourceDetector)(nil)
+var _ resource.Detector = (*resourceDetector)(nil)
+
+// returns resource detector struct
```
MrAlias, line 65: This comment needs to follow Go documentation guidelines (i.e. start with the function name)

```suggestion
// NewResourceDetector returns a resource detector that will detect AWS ECS resources.
```

detectors/aws/ecs/ecs.go
```diff
@@ -43,27 +43,32 @@ var (
...
+func NewResourceDetector(detectorUtils detectorUtilsResources) resource.Detector {
```
MrAlias, line 66: The detectorUtils of the resource interface is an abstraction introduced for testing purposes. We should not leak this implementation detail by requiring a value be passed here.

```suggestion
func NewResourceDetector() resource.Detector {
```

The testing for this detector, since it has internal access to the private fields and types of this package, can assert and then override the needed fields for testing still.

bhautikpip:
@MrAlias addressed your comments!

XSAM (approved):
LGTM

MrAlias (approved):

PR #579, opened 2021-02-15
Add tracing filter for otelgin based on request

clannadxr:
This PR allows otelgin to skip tracing the specific requests.

We add `WithFilter` API for the option to pass the user defined filter which returns false to skip tracing.

This is useful when you don't want to be annoyed by the robot requests like some security scans.

### DIFF
--- a/instrumentation/github.com/gin-gonic/gin/otelgin/gintrace.go
+++ b/instrumentation/github.com/gin-gonic/gin/otelgin/gintrace.go
@@ -38,10 +38,10 @@ const (
 	tracerName = "go.opentelemetry.io/contrib/instrumentation/github.com/gin-gonic/gin/otelgin"
 )
 
-// FilterFunc allows to skip tracing the incoming requests based on
-// request characteristics. Returns true means the filter rule is hit
-// and this request will not be recorded.
-type FilterFunc func(req *http.Request) bool
+// Filter allows to skip tracing the incoming requests based on
+// request characteristics. Returns false means this request will
+// not be recorded.
+type Filter func(req *http.Request) bool
 
 // Middleware returns middleware that will trace incoming requests.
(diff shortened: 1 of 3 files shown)


instrumentation/github.com/gin-gonic/gin/otelgin/gintrace.go
```diff
@@ -33,9 +34,15 @@ import (
...
 const (
 	tracerKey  = "otel-go-contrib-tracer"
+	skippedKey = "otel-go-contrib-skipped"
 	tracerName = "go.opentelemetry.io/contrib/instrumentation/github.com/gin-gonic/gin/otelgin"
 )
 
+// FilterFunc allows to skip tracing the incoming requests based on
+// request characteristics. Returns true means the filter rule is hit
+// and this request will not be recorded.
+type FilterFunc func(req *http.Request) bool
```
Aneurysm9, line 44: Can we get this to align with the [`otelhttp.Filter`](https://github.com/open-telemetry/opentelemetry-go-contrib/blob/main/instrumentation/net/http/otelhttp/common.go#L41) type?  Even if we don't reuse that type here, the signature is identical so having it be `true` to allow rather than reject would enable re-use of the numerous [pre-defined `Filter`s](https://github.com/open-telemetry/opentelemetry-go-contrib/blob/main/instrumentation/net/http/otelhttp/filters/filters.go).

Aneurysm9 (changes requested):

instrumentation/github.com/gin-gonic/gin/otelgin/gintrace.go
clannadxr: updated

XSAM (approved):

Issue #81, opened 2020-06-18
Add instrumentation for github.com/aws/aws-sdk-go


MrAlias:
https://github.com/aws/aws-sdk-go

## Tasks

- [ ] code complete with tests passing
- [ ] `Dockerfile` with example app showing instrumentation
- [ ] documentation updated
- [ ] instrumentation added to the opentelemetry [registry](https://opentelemetry.io/registry/)

## Prior Art

- https://github.com/opentracing-contrib/go-aws-sdk
- https://github.com/DataDog/dd-trace-go/tree/v1/contrib/aws/aws-sdk-go/aws

wangzlei:
Hi, 

AWS released new aws go sdk recent days, I am working on new [aws-sdk-go-v2](https://aws.github.io/aws-sdk-go-v2/) instrumentation but not v1. For v1 we can migrate aws-xray-sdk code to OTel, please see [xray aws-sdk-go instrumentation](https://github.com/aws/aws-xray-sdk-go/blob/master/xray/aws.go).

The sdk v2 instrumentation solution would be through [middleware](https://aws.github.io/aws-sdk-go-v2/docs/middleware/), the built-in hook.

danielwhite:
If I understand correctly, @MrAlias, #621 only addresses the AWS SDK v2. https://github.com/aws/aws-sdk-go is still in wide use and not yet deprecated, so there seems to still be a need to have instrumentation.

PR #599, opened 2021-03-01
support github.com/aws/aws-sdk-go-v2

wangzlei:
https://github.com/open-telemetry/opentelemetry-go-contrib/issues/81

[AWS sdk for go v2](https://aws.github.io/aws-sdk-go-v2/) instrumentation. See more details in https://aws.github.io/aws-sdk-go-v2/docs/middleware/


instrumentation/github.com/aws/aws-sdk-go-v2/otelaws/example/docker-compose.yml
```diff
@@ -0,0 +1,31 @@
...
+# Unless required by applicable law or agreed to in writing, software
+# distributed under the License is distributed on an "AS IS" BASIS,
+# WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+# See the License for the specific language governing permissions and
+# limitations under the License.
+version: "3.7"
+services:
+  aws-sdk-client:
+    build:
+      dockerfile: $PWD/Dockerfile
```
bhautikpip, line 18: Since in the instruction you mentioned use can run `docker-compose` command from `example` dir I think here we can remove `$PWD` from dockerfile path and use `.` in context as well.

instrumentation/github.com/aws/aws-sdk-go-v2/otelaws/example/go.mod
```diff
@@ -0,0 +1,21 @@
+module go.opentelemetry.io/contrib/instrumentation/github.com/aws/aws-sdk-go-v2/otelaws/example
+
+go 1.14
+
+replace (
+	go.opentelemetry.io/contrib => ../../../../../..
```
bhautikpip, line 6: since this is an example I think we should just use the released version right? If users want it they can always point this to local project using `replace`

bhautikpip (approved):
I have left some minor comments nothing major so approving.

instrumentation/github.com/aws/aws-sdk-go-v2/otelaws/aws.go
```diff
@@ -0,0 +1,132 @@
...
+
+	"go.opentelemetry.io/contrib"
+	"go.opentelemetry.io/otel"
+	"go.opentelemetry.io/otel/label"
+	"go.opentelemetry.io/otel/propagation"
+	"go.opentelemetry.io/otel/trace"
+)
+
+const (
+	instrumentationName = "go.opentelemetry.io/contrib/instrumentation/github.com/aws/aws-sdk-go-v2/otelaws"
```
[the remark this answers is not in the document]
wangzlei: yeah, tracerName sounds better.

instrumentation/github.com/aws/aws-sdk-go-v2/otelaws/aws.go
```diff
@@ -0,0 +1,132 @@
...
+
+	awsTracer := cfg.TracerProvider.Tracer(
+		instrumentationName,
+		trace.WithInstrumentationVersion(contrib.SemVersion()),
+	)
+
+	*apiOptions = append(*apiOptions,
+		func(stack *middleware.Stack) error {
+			return stack.Initialize.Add(middleware.InitializeMiddlewareFunc("OtelSpanCreator", func(
+				ctx context.Context, in middleware.InitializeInput, next middleware.InitializeHandler) (
```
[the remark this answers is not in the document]
wangzlei: No, here the ctx is passed from aws sdk, user just make sure initialize ctx when create the aws client.

instrumentation/github.com/aws/aws-sdk-go-v2/otelaws/aws_test.go
```diff
@@ -0,0 +1,158 @@
...
+func TestAppendOtelMiddlewares(t *testing.T) {
+	cases := map[string]struct {
+		responseStatus     int
+		responseBody       []byte
+		expectedError      string
+		expectedRequestID  string
+		expectedStatusCode int
+	}{
+		"invalidChangeBatchError": {
+			responseStatus: 500,
```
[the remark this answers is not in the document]
wangzlei: ok

instrumentation/github.com/aws/aws-sdk-go-v2/otelaws/example/docker-compose.yml
wangzlei: $PWD is more robust.

instrumentation/github.com/aws/aws-sdk-go-v2/otelaws/example/go.mod
wangzlei: Example can reflect the latest change in local by replacing path

instrumentation/github.com/aws/aws-sdk-go-v2/otelaws/aws.go
```diff
@@ -0,0 +1,132 @@
...
+		cfg.Propagators = otel.GetTextMapPropagator()
+	}
+
+	awsTracer := cfg.TracerProvider.Tracer(
+		instrumentationName,
+		trace.WithInstrumentationVersion(contrib.SemVersion()),
+	)
+
+	*apiOptions = append(*apiOptions,
+		func(stack *middleware.Stack) error {
```
Aneurysm9, line 82: Are these declared inline simply to close over `awsTracer`?  It looks like that's the only thing used from the containing scope, so I would suggest moving these to methods on a type that can hold on to a reference to the `Tracer` as it's a little hard to tease apart all the levels here.

```go
type mw struct { // this is probably a terrible type name, but it conflicts with the middleware package and I couldn't think of a better one :)
	tracer trace.Tracer
}

func (m mw) initialize(stack *middleware.Stack) error {...}
func (m mw) deserialize(stack *middleware.Stack) error {...}
```

Then registering the options becomes much cleaner: 

```go
m := mw{tracer: cfg.TracerProvider.Tracer(...)}
awsConfig.APIOptions = append(awsConfig.APIOptions, m.initialize, m.deserialize)
```

instrumentation/github.com/aws/aws-sdk-go-v2/otelaws/aws.go
```diff
@@ -0,0 +1,132 @@
...
+			return stack.Initialize.Add(middleware.InitializeMiddlewareFunc("OtelSpanCreator", func(
+				ctx context.Context, in middleware.InitializeInput, next middleware.InitializeHandler) (
+				out middleware.InitializeOutput, metadata middleware.Metadata, err error) {
+
+				commonLabels := []label.KeyValue{
+					label.String("aws.operation", v2Middleware.GetOperationName(ctx)),
+					label.String("aws.region", v2Middleware.GetRegion(ctx)),
+					label.String("aws.service", v2Middleware.GetServiceID(ctx)),
```
Aneurysm9, line 90: The `label` package has been renamed `attribute`.  This PR needs to be updated to the `main` branch and `v0.18.0` releases.

instrumentation/github.com/aws/aws-sdk-go-v2/otelaws/aws.go
```diff
@@ -0,0 +1,132 @@
...
+
+				return out, metadata, err
+			}),
+				middleware.After)
+		},
+		func(stack *middleware.Stack) error {
+			return stack.Deserialize.Add(middleware.DeserializeMiddlewareFunc("OtelSpanDecorator", func(
+				ctx context.Context, in middleware.DeserializeInput, next middleware.DeserializeHandler) (
+				out middleware.DeserializeOutput, metadata middleware.Metadata, err error) {
+				out, metadata, err = next.HandleDeserialize(ctx, in)
```
Aneurysm9, line 112: If this `err` is non-nil should the `Span` be pulled from `ctx` here to record the `error`?

instrumentation/github.com/aws/aws-sdk-go-v2/otelaws/aws.go
```diff
@@ -0,0 +1,132 @@
...
+				resp, ok := out.RawResponse.(*smithyhttp.Response)
+				if !ok {
+					// No raw response to wrap with.
+					return out, metadata, err
+				}
+
+				span := trace.SpanFromContext(ctx)
+				statusCode := resp.StatusCode
+				span.SetAttributes(label.Int("http.status_code", statusCode))
```
Aneurysm9, line 121:
```suggestion
				span.SetAttributes(semconv.HTTPStatusCodeKey.Int(statusCode))
```

Using the [semantic convention package](https://github.com/open-telemetry/opentelemetry-go/blob/main/semconv/trace.go#L96) here helps ensure this attribute key is aligned with the convention.

instrumentation/github.com/aws/aws-sdk-go-v2/otelaws/aws.go
```diff
@@ -0,0 +1,132 @@
...
+					label.String("aws.operation", v2Middleware.GetOperationName(ctx)),
+					label.String("aws.region", v2Middleware.GetRegion(ctx)),
+					label.String("aws.service", v2Middleware.GetServiceID(ctx)),
+				}
+				opts := []trace.SpanOption{
+					trace.WithAttributes(commonLabels...),
+					trace.WithSpanKind(trace.SpanKindClient),
+				}
+				ctx, span := awsTracer.Start(ctx, v2Middleware.GetServiceID(ctx), opts...)
+				defer span.End()
```
Aneurysm9, line 97: Does this function not return until after the `Deserialize` middleware has executed, or is each step executed independently?

instrumentation/github.com/aws/aws-sdk-go-v2/otelaws/aws.go
```diff
@@ -0,0 +1,132 @@
...
+
+				out, metadata, err = next.HandleInitialize(ctx, in)
+				if err != nil {
+					span.RecordError(err)
+				}
+
+				return out, metadata, err
+			}),
+				middleware.After)
```
Aneurysm9, line 106:
```suggestion
				middleware.Before)
```

Should this be at the front of the line to ensure that the created `Span` captures as much of the time the operation is active as possible?

instrumentation/github.com/aws/aws-sdk-go-v2/otelaws/aws.go
```diff
@@ -0,0 +1,132 @@
...
+func WithTracerProvider(provider trace.TracerProvider) Option {
+	return func(cfg *config) {
+		cfg.TracerProvider = provider
+	}
+}
+
+// AppendOtelMiddlewares attaches otel middlewares to aws go sdk v2 for instrumentation.
+// Otel middlewares can be appended to either all aws clients or a specific operation.
+// Please see more details in https://aws.github.io/aws-sdk-go-v2/docs/middleware/
+func AppendOtelMiddlewares(apiOptions *[]func(*middleware.Stack) error, opts ...Option) {
```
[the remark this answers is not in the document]
wangzlei: agree, duplicate `otel` here

instrumentation/github.com/aws/aws-sdk-go-v2/otelaws/aws.go
```diff
@@ -0,0 +1,132 @@
...
+	instrumentationName = "go.opentelemetry.io/contrib/instrumentation/github.com/aws/aws-sdk-go-v2/otelaws"
+)
+
+type config struct {
+	TracerProvider trace.TracerProvider
+	Propagators    propagation.TextMapPropagator
+}
+
+// Option specifies instrumentation configuration options.
+type Option func(*config)
```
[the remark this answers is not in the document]
wangzlei: Good to know the style guide!

instrumentation/github.com/aws/aws-sdk-go-v2/otelaws/aws.go
wangzlei: good point, I inline 2 middlewares just for the scope of tracer, will update code to type.

instrumentation/github.com/aws/aws-sdk-go-v2/otelaws/aws.go
MrAlias, line 63:
```suggestion
func AppendOTelMiddlewares(apiOptions *[]func(*middleware.Stack) error, opts ...Option) {
```

instrumentation/github.com/aws/aws-sdk-go-v2/otelaws/aws.go
wangzlei: Yes, it is like servlet, Deserialize is the step after Initialize in a pipeline. 
The request passes through Initialize then Deserialize, the response just the opposite.

instrumentation/github.com/aws/aws-sdk-go-v2/otelaws/aws.go
wangzlei: Good eyes. Here if adding Initialize middleware Before, Attributes like service name, region, etc are not initialized yet.
I will think about move there attributes to a middleware back.

instrumentation/github.com/aws/aws-sdk-go-v2/otelaws/aws.go
wangzlei: Middleware is like Servlet Filter, error will be back to Initialize.

wangzlei:
Close this PR due to EasyCLA failed

PR #612, opened 2021-03-09
Add instrumentation for github.com/streadway/amqp #522

previousdeveloper:


pellared (commented):
This instrumentation design differs a lot from the others. Usually, the instrumentation design is to provide decorators. Is this design based on some existing instrumentation? This PR is missing any description which would help to make a review 😉

Nit: you can update the docs as these would probably not change even if you would have to redesign the implementation. Here are the guidelines: https://github.com/open-telemetry/opentelemetry-go/blob/main/CONTRIBUTING.md#documentation

MrAlias:
Thanks for the contribution. According to our [new instrumentation policy](https://github.com/open-telemetry/opentelemetry-go-contrib/tree/main/instrumentation#new-instrumentation), we are no longer accepting new instrumentation to the project as we do not have the developer bandwidth to support it. If you are able, please consider maintaining this instrumentation in your own repository and listing it in the [the OpenTelemetry Registry](https://github.com/open-telemetry/opentelemetry.io#adding-a-project-to-the-opentelemetry-registry).

Issue #611, opened 2021-03-09
otelhttp.WithRouteTag does not set route tag for metrics


charleskorn:
I expected `otelhttp.WithRouteTag()` to set the route tag for both traces and metrics, but it appears to only set the route tag for traces. 

Happy to submit a PR to add this to `WithRouteTag()` using the functionality added in #306 if this would be reasonable default behaviour.

PR #615, opened 2021-03-14
WithRouteTag adds HTTP route attribute to metrics

charleskorn:
Resolves https://github.com/open-telemetry/opentelemetry-go-contrib/issues/611.


charleskorn:
Would it be possible to get a review on this? I'd love to see this functionality implemented.

RangelReale:
I'm also needing something like this.

MrAlias (commented):
Overall, these changes look good. Please sync with `main`.

instrumentation/net/http/otelhttp/handler_test.go
```diff
@@ -38,6 +38,16 @@ func assertMetricAttributes(t *testing.T, expectedAttributes []attribute.KeyValu
 	}
 }
 
+func assertTraceAttributes(t *testing.T, expectedAttributes []attribute.KeyValue, span *oteltest.Span) {
```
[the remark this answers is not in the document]
charleskorn: Done.

dmathieu (approved):

pellared:
@charleskorn Can you update the PR? Sorry for the delay.

charleskorn:
> @charleskorn Can you update the PR? Sorry for the delay.

No worries, done!

instrumentation/net/http/otelhttp/test/handler_test.go
```diff
@@ -468,3 +468,54 @@ func TestSpanStatus(t *testing.T) {
...
+				w.WriteHeader(http.StatusTeapot)
+			}),
+		),
+		"test_handler",
+		otelhttp.WithTracerProvider(tracerProvider),
+		otelhttp.WithMeterProvider(meterProvider),
+	)
+
+	h.ServeHTTP(httptest.NewRecorder(), httptest.NewRequest(http.MethodGet, "/", nil))
+	expectedAttribute := semconv.HTTPRouteKey.String(route)
```
pellared, line 495: Can you rename `expectedAttribute` to simply `want`?

instrumentation/net/http/otelhttp/test/handler_test.go
```diff
@@ -468,3 +468,54 @@ func TestSpanStatus(t *testing.T) {
...
+
+	require.Len(t, spanRecorder.Ended(), 1, "should emit a span")
+	span := spanRecorder.Ended()[0]
```
pellared, line 498: Can you rename `span` to `gotSpan`?

instrumentation/net/http/otelhttp/test/handler_test.go
```diff
@@ -468,3 +468,54 @@ func TestSpanStatus(t *testing.T) {
...
+	require.Contains(t, span.Attributes(), expectedAttribute, "should add route to span attributes")
+
+	rm := metricdata.ResourceMetrics{}
+	err := metricReader.Collect(context.Background(), &rm)
+	require.NoError(t, err)
+	require.Len(t, rm.ScopeMetrics, 1, "should emit metrics for one scope")
+	scopeMetrics := rm.ScopeMetrics[0]
```
pellared, line 505: How about

```suggestion
	gotMetrics := rm.ScopeMetrics[0].Metrics
```

instrumentation/net/http/otelhttp/test/handler_test.go
```diff
@@ -468,3 +468,54 @@ func TestSpanStatus(t *testing.T) {
...
+		case metricdata.Sum[int64]:
+			require.Len(t, d.DataPoints, 1, "metric '%v' should have exactly one data point", m.Name)
+			require.Contains(t, d.DataPoints[0].Attributes.ToSlice(), expectedAttribute, "should add route to attributes for metric '%v'", m.Name)
+
+		case metricdata.Histogram[float64]:
+			require.Len(t, d.DataPoints, 1, "metric '%v' should have exactly one data point", m.Name)
+			require.Contains(t, d.DataPoints[0].Attributes.ToSlice(), expectedAttribute, "should add route to attributes for metric '%v'", m.Name)
+
+		default:
+			require.Fail(t, "metric has unexpected data type", "metric '%v' has unexpected data type %T", m.Name, m.Data)
```
pellared, line 518: Do I understand correctly that this test would start failing if `otelhttp` will produce new metrics (of types other than `metricdata.Histogram[float64]` and `metricdata.Sum[int64]`)?

If I am correct then I think that this test would be too strict and I would propose to remove this `default` case.

pellared (commented):
Looks nice. I just left a few minor comments that I think are worth addressing.

instrumentation/net/http/otelhttp/test/handler_test.go
charleskorn: This is intentional - the expectation is that the route tag is applied to all metrics produced by `otelhttp`, so if a new metric is added that isn't an expected type, this test should fail and prompt the person adding the new type of metric to expand this test to cover that kind of metric.

pellared: >  so if a new metric is added that isn't an expected type, this test should fail and prompt the person adding the new type of metric to expand this test to cover that kind of metric.

I disagree that `TestWithRouteTag` should fail when one adds a new type of metric to be produced by `otelhttp`.
I find it as an "[overspecified](http://xunitpatterns.com/Fragile%20Test.html#Overspecified%20Software) test". This test should not test what types of metrics are produced by `otelhttp`. 

As an alternative, you could update the test to assert against all possible metrics types. If you extract such logic to a helper assertion function it could be used in other tests as well.

charleskorn: >As an alternative, you could update the test to assert against all possible metrics types. 

I've added this in fc1ec77.

pellared (approved):

pellared:
@charleskorn Thanks for your contribution 🎉

pellared:
I am planning to merge this PR. There are only some problems with EasyCLA integration which is blocking this PR. This problem affects other PRs as well.

dashpole (approved):

PR #713, opened 2021-03-26
otelhttp: handle nil base http transport

kjschnei001:
The purpose of this PR is to add a nil check in the `otelhttp.NewTransport` function.  I noticed this issue while I was using this library on a codebase where the transport was optionally nil.  When calling Do on the wrapped client, it was throwing a nil pointer.

Rather than put the onus on the user to make this check, it would be good to just transparently handle this case with a sane default (i.e. use the http.DefaultTransport).


instrumentation/net/http/otelhttp/transport.go
```diff
@@ -41,6 +41,10 @@ var _ http.RoundTripper = &Transport{}
 // NewTransport wraps the provided http.RoundTripper with one that
 // starts a span and injects the span context into the outbound request headers.
 func NewTransport(base http.RoundTripper, opts ...Option) *Transport {
+	if base == nil {
```
Aneurysm9, line 44: Can you add documentation of this behavior to the godoc for this function?

kjschnei001: updated

XSAM (approved):

MrAlias (approved):

Issue #435, opened 2020-11-03
Possible race condition in Sarama async producer


pkwarren:
https://github.com/open-telemetry/opentelemetry-go-contrib/blob/e7ba12942d4546d66b5853af7310c321888ff8de/instrumentation/github.com/Shopify/sarama/otelsarama/producer.go#L200-L203

This looks like a possible race condition during shutdown (we are returning from a goroutine as soon as the successes or errors channels are closed, potentially not reading any remaining messages in the other channel). I'm wondering if it would be safer to assign a local variable to both Successes() and Errors() channels and reassign it to nil once closed. Then keep track and only return from the goroutine once both channels are closed.

pellared:
@MrAlias Fix is here: https://github.com/open-telemetry/opentelemetry-go-contrib/pull/754

MrAlias:
Resolved in #754

PR #754, opened 2021-04-20
Fix issues in otelsarama.WrapAsyncProducer

pellared:
## Why

Usage of current Sarama instrumentation can lead to deadlocks and data races. These issues are not present if the instrumentation is not applied (created tests for deadlock scenarios).

The messages returned by the `Errors()` channel do not contain original `Metadata`. (I can make it as a separate PR if needed)

## What

1. Fix deadlocks by changing the concurrency design. 
   - Use dedicated goroutines for consuming messages and for producing messages. It prevents the deadlocks and also makes the implementation less dependent on Sarama's internal implementation.
   - Use dedicated channels for closing instead of reusing the `input` channel. 
   - Redesign the "spans cleanup" to make it NOT dependant on Sarma's implementation details (which may change). See:  https://github.com/open-telemetry/opentelemetry-go-contrib/issues/435
   - All changes were done in a TDD way, however still there may be some edge cases that I have not thought about.

2. Restore messages `Metadata` for the wrapped `Errors` channel.

## Important Notice

Kudos to @MadVikingGod for a lot of help and for having concerns with the initial fixes 👍 🥇 

Still very careful review is needed.

### Example deadlock

Reproducing a sample deadlock:

```go
package main

import (
	"fmt"
	"log"

	"github.com/Shopify/sarama"
	"go.opentelemetry.io/contrib/instrumentation/github.com/Shopify/sarama/otelsarama"
)

/*
	EXECUTE:

docker network create confluent

docker run --rm -d \
	--name zookeeper \
	--network confluent \
	-p 2181:2181 \
	-e ZOOKEEPER_CLIENT_PORT=2181 \
	confluentinc/cp-zookeeper:5.0.0

docker run --rm -d \
	--name kafka \
	--network confluent \
	-p 9092:9092 \
	-e KAFKA_ZOOKEEPER_CONNECT=zookeeper:2181 \
	-e KAFKA_ADVERTISED_LISTENERS=PLAINTEXT://localhost:9092 \
	-e KAFKA_LISTENERS=PLAINTEXT://0.0.0.0:9092 \
	-e KAFKA_CREATE_TOPICS=gotest:1:1 \
	-e KAFKA_OFFSETS_TOPIC_REPLICATION_FACTOR=1 \
	confluentinc/cp-kafka:5.0.0

go run -race .

*/

func main() {
	cfg := sarama.NewConfig()
	cfg.Version = sarama.V0_11_0_0 // minimum version that supports headers which are required for tracing

	producer, err := sarama.NewAsyncProducer([]string{"localhost:9092"}, cfg)
	if err != nil {
		panic(err)
	}
	defer producer.Close()

	producer = otelsarama.WrapAsyncProducer(cfg, producer)

	msg := &sarama.ProducerMessage{
		Topic: "some-topic",
		Value: sarama.StringEncoder("Hello World"),
	}

	// close the producer
	if err := producer.Close(); err != nil {
		log.Fatalln(err)
	}

	// esnure that WrapAsyncProducer has exited
	if _, ok := <-producer.Successes(); ok {
		log.Fatalln("should not receive anything")
	}

	log.Println("Sending message", msg)

	defer func() {
		if r := recover(); r != nil {
			fmt.Println("RECOVERED FROM MAIN:", r)
			return
		}
	}()

	sarama.PanicHandler = func(r interface{}) { // sarama has a data race for PanicHandler...
		fmt.Println("RECOVERED IN SARAMA:", r)
	}

	producer.Input() <- msg // hangs forever (should panic from HERE - not internaly in sarama)

	log.Println("Message sent", msg)
}
```


Aneurysm9 (approved):

instrumentation/github.com/Shopify/sarama/otelsarama/producer.go
```diff
@@ -155,6 +155,7 @@ func WrapAsyncProducer(saramaConfig *sarama.Config, p sarama.AsyncProducer, opts
 				finishProducerSpan(mc.span, 0, 0, nil)
 			}
 		}()
+		defer close(wrapped.input)
```
MadVikingGod, line 158: It looks like this is a consumer of input, not the producer.  If that's the case we shouldn't close it here, but where the producer is finished.

MadVikingGod (changes requested):

instrumentation/github.com/Shopify/sarama/otelsarama/producer.go
pellared: Meaning that `close(wrapped.input)` should be called in `AsyncClose` and `Close` methods after sending to message to `input` ?

MadVikingGod: I'm not very familiar with this portion of the code, but generally you want to close where you write to a channel, not where it's read.  And, this function only reads from `wrapped.input`.

If anything you might want to be closing `p.Input()`. And yes I would think that would be more appropriate in the `Close()` of the Producer.

pellared: @MadVikingGod I will try it next week and do my best to test it from different angles. After a quick look, I have a feeling that it will be a more "correct" code. Thanks a lot 👍

pellared: @MadVikingGod 
I think that the fix provided in this PR is the optimal one.
We have no other way to know when we can close the input channel other than waiting until `sarama` closes the `producer.Successes()` and `producer.Errors()` channels. It is caused mainly by sarama's `AsyncClose()` implementation which "closes at some point".

Here is the "problematic" code:
- https://github.com/Shopify/sarama/blob/master/async_producer.go#L320-L322
- https://github.com/Shopify/sarama/blob/master/async_producer.go#L1054-L1059

Still, I have managed at least to refactor the existing code a little... I hope it is better.

MadVikingGod: That is generally the accepted way of closing these async channels.  It's acceptable to panic if someone writes after `close()`, because that is the contract that channels have.  

What happens if you stick with the previous behavior, but just close the `p.input` in the `Close()` and `AsyncClose()`?

pellared: It panics in goroutines created by sarama... So it changes the behavior when instrumentation is used vs not applied. I think it is not acceptable. Let me know if you want stack traces (I can provide them in a few hours).

pellared: @MadVikingGod 
Thanks lot for your comments 👍 ❤️ 

I took another try. Probably I made some bug when closing the input channel.
Now I am closing the input channel when closing. It is a lot better because the previous " fix" was deadlocking in case of calling `p.Close()` more than once!

Is there anything that does not look good?

pellared: Getting race conditions in sarama when I do `p,Close()` and `p.Input() <- msg` in different goroutines. Makes sense...

pellared:
I do not think this fix is critical. Why? Because I think the only possibility to reproduce it is to send something to `p.Input()` after calling `p.Close()` or `p.AsyncClose()` on the same goroutine (that is strange isn't it?). If different goroutines are closing and sending then something is wrong and it probably also would result in race condition. 

This PR is more-less like the "defensive-programming" style allowing more freedom (e.g. calling `Close` twice). But also ensures similar behavior to one without instrumentation applied (at least it does not make it worse like the original codebase).

pellared:
Also solves: https://github.com/open-telemetry/opentelemetry-go-contrib/pull/755

pellared:
@Aneurysm9 @MadVikingGod Ready for review

lizthegrey (approved):

CHANGELOG.md
```diff
@@ -12,6 +12,10 @@ This project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0.htm
 
 - Remove service name from `otelmongodb` configuration and span attributes. (#763)
 
+### Fixed
+
+- Fix deadlocks and race conditions in Sarama instrumentation. (#754)
```
pellared, line 17: Should we mention that MessageID and KafkaPartitionID attributes are not set if a message was not processed?

pellared: Addressed https://github.com/open-telemetry/opentelemetry-go-contrib/pull/754/commits/341acf7b775884862866a492c182004232cf2d8d

instrumentation/github.com/Shopify/sarama/otelsarama/producer_test.go
```diff
@@ -335,6 +332,99 @@ func TestWrapAsyncProducerError(t *testing.T) {
 	assert.Equal(t, "test", span.StatusMessage())
 }
 
+func TestWrapAsyncProducer_ConcurrencyEdgeCases(t *testing.T) {
+	newAsyncProducer := func() sarama.AsyncProducer {
+		cfg := newSaramaConfig()
+		var ap sarama.AsyncProducer = mocks.NewAsyncProducer(t, cfg)
+		ap = WrapAsyncProducer(cfg, ap) // you can comment this line to ensure that sarama is working in the same way
```
Aneurysm9, line 339: Should we run these tests twice, once with a wrapped producer and once unwrapped?

pellared: I think it makes sense.

Side note: Do you have any idea of other edge cases? I want to be more confident that the rewritten implementation does not introduce any new bugs.

instrumentation/github.com/Shopify/sarama/otelsarama/producer.go
```diff
@@ -189,41 +174,67 @@ func WrapAsyncProducer(saramaConfig *sarama.Config, p sarama.AsyncProducer, opts
...
+		}()
+		defer close(wrapped.successes)
+		defer close(wrapped.errors)
+		for {
+			select {
 			case msg, ok := <-p.Successes():
 				if !ok {
-					// producer was closed, so exit
-					return
+					return // producer was closed, so exit
```
pellared, line 209: I do not remember why I am not ensuring that both `p.Successes()` and `p.Errors()` are closed before doing a `return`

I need to double-check it and leave a comment if I did it intentionally

instrumentation/github.com/Shopify/sarama/otelsarama/producer_test.go
pellared: Done

instrumentation/github.com/Shopify/sarama/otelsarama/producer.go
```diff
@@ -189,41 +173,73 @@ func WrapAsyncProducer(saramaConfig *sarama.Config, p sarama.AsyncProducer, opts
...
-				key := err.Msg.Metadata
-				if mc, ok := producerMessageContexts[key]; ok {
-					delete(producerMessageContexts, key)
-					finishProducerSpan(mc.span, err.Msg.Partition, err.Msg.Offset, err.Err)
-				}
-				wrapped.errors <- err
+	// Spawn spans cleanup goroutine.
+	// Sarama will consume all the successes and errors by itself while closing,
+	// so we need to end these these spans ourseleves.
+	cleanupCh := make(chan struct{}, 2)
```
Aneurysm9, line 192: I would use a `sync.WaitGroup` for this instead.  `wg.Add(1)`/`wg.Done()` in the cleanup goroutines and `wg.Wait()` before doing the work here.

pellared: I tried it initially. It produces a data race (sic!). It looks like Add and Wait have to be used in the same goroutine. I will just try it again. Maybe I made some silly error

Aneurysm9 (approved):

MadVikingGod (commented):
I haven't finished, but so far it looks good.  The producers of the channels are the ones closing them, There is overall good hygiene around the close semantics.  I will try and get to the tests later but consider this a soft +1.

instrumentation/github.com/Shopify/sarama/otelsarama/producer_test.go
```diff
@@ -331,6 +335,171 @@ func TestWrapAsyncProducerError(t *testing.T) {
...
+	gotErrors := 0
+	wg.Add(1)
+	go func() {
+		defer wg.Done()
+		for range ap.Errors() {
+			gotErrors++
+		}
+	}()
+
+	wg.Wait()
```
MadVikingGod, line 385: Not worth a change:
You should be able to do this synchronously correct?

MadVikingGod (approved):
LGTM

instrumentation/github.com/Shopify/sarama/otelsarama/producer_test.go
pellared: question if something like this is more readable:

```go

var gotSuccesses, gotErrors int
var successClosed, errorsClosed bool
for !successClosed && !errorsClosed  {
  select {
  case _, ok := <- ap.Successes():
    if !ok {
      successClosed = true
      continue
    }
    gotSuccesses++
  case _, ok := <- ap.Errors():
    if !ok {
      errorsClosed = true
      continue
    }
    gotErrors ++
  }
}
```

Issue #85, opened 2020-06-18
Add instrumentation for github.com/graph-gophers/graphql-go


MrAlias:
https://github.com/graph-gophers/graphql-go

## Tasks

- [ ] code complete with tests passing
- [ ] `Dockerfile` with example app showing instrumentation
- [ ] documentation updated
- [ ] instrumentation added to the opentelemetry [registry](https://opentelemetry.io/registry/)

## Prior Art

- https://github.com/DataDog/dd-trace-go/tree/v1/contrib/graph-gophers/graphql-go

gtourkas:
@MrAlias may I pick this up ?

pellared:
@ravilushqa @jay-manday @jjmengze @gtourkas 
Could you please try to collaborate and select one of your PRs as a preferred one (and make other PRs as a draft) to make the reviewers' life easier? 😉

ravilushqa:
@pellared
This issue how i see opened for graph-gophers/graphql-go lib, whereas mine and @jjmengze is for 99designs/gqlgen. I'm not sure that both lib should go in a one MR, and imo that would be harder for reviewers.

In situation with @jjmengze, our pr's are both for a one lib. I'm not experienced contributor, so i'm not pretty sure how to deal with that. I'm pretty long waiting for it, and I will be happy if we will collaborate around it.

My suggest is to contribute separately this one, and mine :)
https://github.com/open-telemetry/opentelemetry-go-contrib/pull/761

gtourkas:
@pellared 

@ravilushqa sufficiently described the case: this PR for for graph-gophers/graphql-go, which is also what the issue #85 is about, and the other two for 99designs/gqlgen . Apart from the registry item, which I guess is the last step, the PR is imo ready for review.

PR #761, opened 2021-04-26
Add 99designs/gqlgen instrumentation

ravilushqa:
This PR adds tracer for 99designs/gqlgen


ravilushqa:
hey @Aneurysm9 @dashpole, @evantorrie, @jmacd, @MrAlias, @paivagustavo @XSAM this is available for review now.
Please let me know if you have any comments. Waiting for your feedback

hirenvadalia:
Any plan to merge this?

ravilushqa:
@hirenvadalia im still waiting for review

jay-manday:
This is awesome @ravilushqa! we'd love to see this merged so we can start leveraging this ourselves

pellared:
We already have 3 PRs open addressing #85

instrumentation/github.com/99designs/gqlgen/otelgqlgen/config.go
```diff
@@ -0,0 +1,44 @@
...
+
+// config is used to configure the mongo tracer.
+type config struct {
+	TracerProvider          trace.TracerProvider
+	Tracer                  trace.Tracer
+	ComplexityExtensionName string
+}
+
+// Option specifies instrumentation configuration options.
+type Option func(*config)
```
pellared, line 29: The guidelines around the `Option` type have changed a little.

Please see: https://github.com/open-telemetry/opentelemetry-go/blob/main/CONTRIBUTING.md#option

Here is a sample PR how you can easily make it compliant: https://github.com/open-telemetry/opentelemetry-go-contrib/pull/860

.github/dependabot.yml
```diff
@@ -117,6 +117,24 @@ updates:
...
+    package-ecosystem: "gomod"
+    directory: "/instrumentation/github.com/99designs/gqlgen/otelgqlgen"
+    labels:
+      - dependencies
+      - go
+      - "Skip Changelog"
+    schedule:
+      interval: "weekly"
+      day: "sunday"
+  - package-ecosystem: "gomod"
```
pellared, line 130: please make the formatting consistent

```suggestion
  - 
    package-ecosystem: "gomod"
```

pellared (commented):
I just made a very quick few-minute review and looks nice so far 👍 

Can you please add a `README.md` file as recommended by https://github.com/open-telemetry/opentelemetry-go/blob/main/CONTRIBUTING.md#documentation?

ravilushqa:
@pellared thanks for review. Added more info and refactor Option. If something should be added feel free to say :)

instrumentation/github.com/99designs/gqlgen/otelgqlgen/gqlgen.go
```diff
@@ -0,0 +1,151 @@
...
+		// complexity extension is not used
+		complexityStats = &extension.ComplexityStats{}
+	}
+
+	if complexityStats.ComplexityLimit > 0 {
+		span.SetAttributes(
+			RequestComplexityLimit(int64(complexityStats.ComplexityLimit)),
+			RequestOperationComplexity(int64(complexityStats.Complexity)),
+		)
+	}
```
pellared, line 82: This feature is cool and I think it is worth adding a test for it to make sure it is always working.

instrumentation/github.com/99designs/gqlgen/otelgqlgen/gqlgen.go
```diff
@@ -0,0 +1,151 @@
...
+
+}
+
+func operationName(ctx context.Context) string {
+	requestContext := graphql.GetOperationContext(ctx)
+	requestName := "nameless-operation"
+	if requestContext.Doc != nil && len(requestContext.Doc.Operations) != 0 {
+		op := requestContext.Doc.Operations[0]
+		if op.Name != "" {
+			requestName = op.Name
```
pellared, line 146: how about adding a test to cover this line?

instrumentation/github.com/99designs/gqlgen/otelgqlgen/gqlgen.go
pellared, line 82: This feature is cool and I think it is worth adding a test for it to make sure it is always working.

pellared (approved):
I am just suggesting covering two code fragments. 

I do not have any `99designs/gqlgen` experience but this code looks very good and it can always be improved later.

pellared (changes requested):
I just spotted this uses an old version of OTel SDK and we should split the tests to another Go Module.

instrumentation/github.com/99designs/gqlgen/otelgqlgen/gqlgen.go
```diff
@@ -0,0 +1,151 @@
...
+func (a Tracer) ExtensionName() string {
+	return extensionName
+}
+
+func (a Tracer) Validate(_ graphql.ExecutableSchema) error {
+	return nil
+}
+
+func (a Tracer) InterceptResponse(ctx context.Context, next graphql.ResponseHandler) *graphql.Response {
+	ctx, span := a.tracer.Start(ctx, operationName(ctx))
```
pellared, line 56: how about setting the option `oteltrace.WithSpanKind(oteltrace.SpanKindServer)` ? 

probably same applies in `InterceptField` method

instrumentation/github.com/99designs/gqlgen/otelgqlgen/tags.go
```diff
@@ -0,0 +1,120 @@
...
+	ResolverFieldKey              = attribute.Key("resolver.field")
+	ResolverAliasKey              = attribute.Key("resolver.alias")
+	ResolverHasErrorKey           = attribute.Key("resolver.hasError")
+	ResolverErrorCountKey         = attribute.Key("resolver.errorCount")
+)
+
+// ServiceName defines the service name for this span.
+func ServiceName(serviceName string) attribute.KeyValue {
+	return ServiceNameKey.String(serviceName)
+}
```
pellared, line 45: I think this should be removed. The service name should be via  "resource"https://github.com/open-telemetry/opentelemetry-specification/blob/main/specification/resource/semantic_conventions/README.md by setting `OTEL_RESOURCE_ATTRIBUTES` or `OTEL_SERVICE_NAME` env var. It is possible to do it via code as well see:  [example](https://github.com/open-telemetry/opentelemetry-go/blob/main/example/fib/main.go)

instrumentation/github.com/99designs/gqlgen/otelgqlgen/tags.go
```diff
@@ -0,0 +1,120 @@
...
+	RequestQueryKey               = attribute.Key("request.query")
+	RequestComplexityLimitKey     = attribute.Key("request.complexityLimit")
+	RequestOperationComplexityKey = attribute.Key("request.operationComplexity")
+	ResolverPathKey               = attribute.Key("resolver.path")
+	ResolverObjectKey             = attribute.Key("resolver.object")
+	ResolverFieldKey              = attribute.Key("resolver.field")
+	ResolverAliasKey              = attribute.Key("resolver.alias")
+	ResolverHasErrorKey           = attribute.Key("resolver.hasError")
+	ResolverErrorCountKey         = attribute.Key("resolver.errorCount")
+)
```
pellared, line 40: what is the point of exporting these consts?

instrumentation/github.com/99designs/gqlgen/otelgqlgen/tags.go
```diff
@@ -0,0 +1,120 @@
...
+	ServiceNameKey                = attribute.Key("service.name")
+	RequestQueryKey               = attribute.Key("request.query")
+	RequestComplexityLimitKey     = attribute.Key("request.complexityLimit")
+	RequestOperationComplexityKey = attribute.Key("request.operationComplexity")
+	ResolverPathKey               = attribute.Key("resolver.path")
+	ResolverObjectKey             = attribute.Key("resolver.object")
+	ResolverFieldKey              = attribute.Key("resolver.field")
+	ResolverAliasKey              = attribute.Key("resolver.alias")
+	ResolverHasErrorKey           = attribute.Key("resolver.hasError")
+	ResolverErrorCountKey         = attribute.Key("resolver.errorCount")
```
pellared, line 39: I suggest adding a `gql.` prefix for all of these attributes

instrumentation/github.com/99designs/gqlgen/otelgqlgen/tags.go
hobbsh: I believe they are simple abstractions to make it easier to pull certain attributes out of the GraphQL request object

pellared: Correct, but it does not mean they have to be exported as the underlying values can change. Also if we would have a [semantic convention](https://github.com/open-telemetry/opentelemetry-specification/tree/main/specification/trace/semantic_conventions) for GraphQL then it would be good to remove these consts as they would live in a package like https://pkg.go.dev/go.opentelemetry.io/otel/semconv/v1.4.0

instrumentation/github.com/99designs/gqlgen/otelgqlgen/go.mod
```diff
@@ -0,0 +1,18 @@
...
+	go.opentelemetry.io/contrib => ../../../../../
+	go.opentelemetry.io/contrib/instrumentation/github.com/99designs/gqlgen/otelgqlgen => ../
+)
+
+require (
+	github.com/99designs/gqlgen v0.13.0
+	github.com/stretchr/testify v1.7.0
+	github.com/vektah/gqlparser/v2 v2.1.0
+	go.opentelemetry.io/contrib v0.0.0-00010101000000-000000000000
+	go.opentelemetry.io/otel v0.19.0
```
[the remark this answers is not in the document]
ravilushqa: I guess I did what you suggest

hobbsh:
First of all, awesome PR! I have a couple questions - possibly out of scope:

- How do you see users integrating this with their logging? We've only been able to retrieve the traceID/spanID from within the middleware itself and log that way, which means we would have to fork this to use it.
- How could we make use of exemplars (traceID in prometheus metrics) with this?

ravilushqa:
@hobbsh thanks!

> How do you see users integrating this with their logging? We've only been able to retrieve the traceID/spanID from within the middleware itself and log that way, which means we would have to fork this to use it.

If I understand correctly u talk about data from gql. In that situation i see using logging middleware like in my PR if it's needed. If u need data from my middleware, it's stored in spans, so u can catch it from it.

> How could we make use of exemplars (traceID in prometheus metrics) with this?
I didn't work with exemplars so it's hard to say for me.

hobbsh:
> If I understand correctly u talk about data from gql. In that situation i see using logging middleware like in my PR if it's needed. If u need data from my middleware, it's stored in spans, so u can catch it from it.
> 
> >

The use-case is to log the traceID with the GraphQL query so we can correlate logs to traces. I am not seeing an example of using logging middleware like `zap` in your PR to do this.

ravilushqa:
> The use-case is to log the traceID with the GraphQL query so we can correlate logs to traces. I am not seeing an example of using logging middleware like `zap` (or otherwise) in your PR to do this.

this pr has nothing to do with logging, so there is no such example.
In your use-case u can create trace in middleware. Trace will be stored in context.
For zap u can use  enrich logs with traceID by getting it from context.
for example:
`logger.With(zap.String("trace-id", trace.SpanContextFromContext(ctx).TraceID().String())`
in addition u can use zap.Hooks to enrich logs, and getting traceID same from context.

hobbsh:
> > The use-case is to log the traceID with the GraphQL query so we can correlate logs to traces. I am not seeing an example of using logging middleware like `zap` (or otherwise) in your PR to do this.
> 
> this pr has nothing to do with logging, so there is no such example.
> In your use-case u can create trace in middleware. Trace will be stored in context.
> For zap u can use  enrich logs with traceID by getting it from context.
> for example:
> `logger.With(zap.String("trace-id", trace.SpanContextFromContext(ctx).TraceID().String())`
> in addition u can use zap.Hooks to enrich spans by adding log info from spans that getting from ctx too.

I appreciate you taking the time to describe this. Thanks!

instrumentation/github.com/99designs/gqlgen/otelgqlgen/README.md
```diff
@@ -0,0 +1,39 @@
...
+  [`WithComplexityExtensionName`](https://pkg.go.dev/go.opentelemetry.io/contrib/instrumentation/github.com/99designs/gqlgen/otelgqlgen#WithComplexityExtensionName)
+  option.
+
+### Environment Variables
+
+The following environment variables can be used to override the default configuration.
+
+| Environment variable   | Option | Default value    |
+| ---------------------- | ------ | ---------------- |
+| `OTEL_SERVICE_NAME`    |        | `GraphQLService` |
```
pellared, line 34: The `Environment Variables` section should be removed. See further comment

instrumentation/github.com/99designs/gqlgen/otelgqlgen/gqlgen.go
```diff
@@ -0,0 +1,162 @@
...
+	}
+
+	oc := graphql.GetOperationContext(ctx)
+
+	serviceName := os.Getenv("OTEL_SERVICE_NAME")
+	if serviceName == "" {
+		serviceName = defaultServiceName
+	}
+	span.SetAttributes(
+		semconv.ServiceNameKey.String(serviceName),
```
pellared, line 72: this should be removed.
the resource detector is responsible for it
see:
- https://github.com/open-telemetry/opentelemetry-go/blob/b8561785c01cbcefa88d6fabe33d774869fcd637/sdk/resource/env.go
- https://opentelemetry.io/docs/go/exporting_data/#resources

ravilushqa: oh, sorry for misunderstood. fixed now :)

instrumentation/github.com/99designs/gqlgen/otelgqlgen/gqlgen.go
```diff
@@ -0,0 +1,162 @@
...
+	if complexityStats.ComplexityLimit > 0 {
+		span.SetAttributes(
+			RequestComplexityLimit(int64(complexityStats.ComplexityLimit)),
+			RequestOperationComplexity(int64(complexityStats.Complexity)),
+		)
+	}
+
+	span.SetAttributes(RequestVariables(oc.Variables)...)
+
+	return next(ctx)
```
pellared, line 86: shouldn't we do something more or less like this?

```go
resp := nex(ctx)
if len(resp.Errors) > 0 {
  span.RecordError(err)
}

File: instrumentation/github.com/99designs/gqlgen/otelgqlgen/gqlgen.go [hunk 13]
```diff
@@ -0,0 +1,162 @@
...
+	)
+	span.SetAttributes(ResolverArgs(fc.Field.Arguments)...)
+
+	resp, err := next(ctx)
+
+	errList := graphql.GetFieldErrors(ctx, fc)
+	if len(errList) != 0 {
+		span.SetStatus(codes.Error, errList.Error())
+		span.SetAttributes(ResolverErrors(errList)...)
+	}
```
pellared, line 114: shouldn't we set the span's status as error and record an error?

pellared:
PS. I do not have any more findings 😉

instrumentation/github.com/99designs/gqlgen/otelgqlgen/gqlgen.go
ravilushqa: i'm not sure that i catch the sence.

```
	errList := graphql.GetFieldErrors(ctx, fc)
	if len(errList) != 0 {
		span.SetStatus(codes.Error, errList.Error())
		span.RecordError(fmt.Errorf(errList.Error()))
		span.SetAttributes(ResolverErrors(errList)...)
	}
```

this will be correct in your opinion?

instrumentation/github.com/99designs/gqlgen/otelgqlgen/gqlgen.go
ravilushqa: same on this. this will be correct?

```
	resp := next(ctx)
	if len(resp.Errors) > 0 {
		span.RecordError(fmt.Errorf(resp.Errors.Error()))
		span.SetStatus(codes.Error, resp.Errors.Error())
		span.SetAttributes(ResolverErrors(resp.Errors)...)
	}

	return resp
```

pellared: Looks OK. And a unit test if possible 😉

instrumentation/github.com/99designs/gqlgen/otelgqlgen/gqlgen.go
pellared: Looks OK. And a unit test if possible 😉

pellared (approved):

instrumentation/github.com/99designs/gqlgen/otelgqlgen/example/go.mod
```diff
@@ -0,0 +1,18 @@
...
+
+replace (
+	go.opentelemetry.io/contrib => ../../../../../../
+	go.opentelemetry.io/contrib/instrumentation/github.com/99designs/gqlgen/otelgqlgen => ../
+)
+
+require (
+	github.com/99designs/gqlgen v0.14.0
+	github.com/vektah/gqlparser/v2 v2.2.0
+	go.opentelemetry.io/contrib/instrumentation/github.com/99designs/gqlgen/otelgqlgen v0.19.0
```
pellared, line 13: I guess it should be already `v0.23.0`

pellared:
@ravilushqa Nice work. I will try to ask others to review. In meantime, you can create a dedicated repository that will contain this instrumentation similarly to https://github.com/XSAM/otelsql

ravilushqa:
> @ravilushqa Nice work. I will try to ask others to review. In meantime, you can create a dedicated repository that will contain this instrumentation similarly to https://github.com/XSAM/otelsql

thanks! i'll work on package like that.

In addition i've found that we forget to remove serviceName from pr. i did it in the last commit

pellared:
I do not think it will be merged. See: https://github.com/open-telemetry/opentelemetry-go-contrib/issues/1100

However, I hope that my code reviews were helpful.

Issue #623, opened 2021-03-16
Examples build with golang:alpine'


matej-g:

matej-g:
Or maybe rather this might give the right context:
https://blog.golang.org/go116-module-changes
```
In Go 1.16, module-aware commands report an error after discovering a problem in go.mod or go.sum instead of attempting to fix the problem automatically.
```

It seems like building with Go < 1.16 resolves the issue 'automatically' and fixes `go.mod`:
```
go: finding module for package go.opentelemetry.io/otel/exporters/stdout
go: finding module for package go.opentelemetry.io/otel/sdk/trace
go: found go.opentelemetry.io/otel/exporters/stdout in go.opentelemetry.io/otel/exporters/stdout v0.18.0
go: found go.opentelemetry.io/otel/sdk/trace in go.opentelemetry.io/otel/sdk v0.18.0
```

Whereas with Go 1.16 we are getting an error by default.

MrAlias:
Thanks catching and documenting this!

> Otherwise it should be investigated what is the exact cause why Go fails to fetch the mentioned modules and how to mitigate it.

This sounds like the appropriate resolution. That way we ensure other users that use Go 1.16 will not encounter this error as well. This also motivates me to add 1.16 to the versions we validate in our CI system.

PR #767, opened 2021-05-03
Fixes Dockerfile based examples to run from correct directory

evantorrie:
With go-1.16, `go install` no longer modifies `go.mod` files to correct omissions of required module imports.  This means all `go.mod` files must provide a complete and correct set of `requires` module imports.

While the `go.mod` files inside the `examples` directories do include the correct requires, some of the `Dockerfile` examples run a `go install` command from the corresponding instrumentation directory (one level above the examples directory). This means `go install` looks at the `go.mod` file at the instrumentation directory level instead of the `go.mod` file included in the `examples/` directory.  

This changes all such cases to set the working directory to the `example` directory containing the correct `go.mod` before running `go install`.


evantorrie:
This fixes #623.

However it also indicates we may want to include checking that the `examples` for each contrib instrumentation build as part of the CI process, as we do in the  [opentelemetry-go main repo](https://github.com/open-telemetry/opentelemetry-go/blob/f40cad5ee20175a946763ef79a32741069e4492a/Makefile#L64-L69).

I've opened an issue #768 to track this

CHANGELOG.md
```diff
@@ -8,6 +8,10 @@ This project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0.htm
 
 ## [Unreleased]
 
+### Fixed
+
+- Dockerfile based examples of instrumentation to run with go-1.16 (#767)
```
Aneurysm9, line 13:
```suggestion
- Dockerfile based examples of instrumentation now run with go-1.16. (#767)
```

Aneurysm9 (approved):

MrAlias (commented):
Should this include upgrading the base image to 1.16, or have I missed something?

evantorrie:
> Should this include upgrading the base image to 1.16, or have I missed something?

It could, but I thought this could be done in a separate PR. For example, in `opentelemetry-go-contrib`, I believe we're still testing using only go-1.14 and go-1.15, so standardizing the base images on 1.16 should probably be done in conjunction with that.

In this case, the issue is that two examples were pulling from `golang:alpine" rather than a specific version of golang
1.  [`otelgin`](https://github.com/open-telemetry/opentelemetry-go-contrib/pull/767/files#diff-80b0b853ad2d3e677df5dc44fa80cb7c3d3aa1ae3bcafe0a03c75df0879a89fbL14)
2. [`otelmacaron`](https://github.com/open-telemetry/opentelemetry-go-contrib/pull/767/files#diff-28a97d3b9b156e7ba00a2b346028bb4709843d05ba42a035b0017037c4f1465bL14)

These were the only ones failing, but I took the opportunity to fix all the examples which were incorrectly trying to run `go install` from a parent directory with a *different* `go.mod` file than that inside the `example` directory, but just happened to work due to go-1.14 and go-1.15's behavior of silently fixing `go.mod` files that were incorrect.

In other words,  this PR just fixes the previously incorrect behavior - which means they now work with any version of go.

What do you think about creating a separate issue to go through and standardize the Dockerfile based images to all be a consistent version of `golang:alpine`?  Or perhaps use a `--build-arg` version which would allow building the docker-based examples using the same matrix of supported go versions as we use in the ci-build.

MrAlias (approved):

Issue #712, opened 2021-03-26
otelmongo: ability to hide Command attribute


ravilushqa:
Command attribute may contain sensitive data. would be great if it will be configurable.

davidgwcurve:
I have opened a MR for this option as we need this too for our applications :)

PR #769, opened 2021-05-03
Remove superfluous otelmongodb attribute names. Conform to specification and add more tests

### LINKED ISSUE #460: otelmongo: Various attributes improvements
Noticed that that there are couple of attributes-related areas in `otelmongo` that could be improved:
- Reconsider usage of `service.name` attribute - according to the specification this is a [resource attribute](https://github.com/open-telemetry/opentelemetry-specification/tree/master/specification/resource/semantic_conventions#service), not a trace attribute (see https://github.com/open-telemetry/opentelemetry-go-contrib/issues/454#issuecomment-729911547)
- There is a significant overlap between attribute keys in `db.go` / `tags.go` and `semconv` package - preferably, we should use the "standardized" ones from `go.opentelemetry.io/otel/semconv` package
- Go over https://github.com/open-telemetry/opentelemetry-specification/blob/master/specification/trace/semantic_conventions/database.md and look if any of the required / recommended attributes are missing

evantorrie:
fix #460 

* All custom attribute names previously used by `otelmongodb` instrumentation are removed and replaced with references to `semconv` conventions from the main `opentelemetry-go` repository.
* Set span name as per [OpenTelemetry specification](https://github.com/open-telemetry/opentelemetry-specification/blob/2a712f7626cb62b21ae210ef3175612e7744f285/specification/trace/semantic_conventions/database.md#mongodb): `<collection>.<operation>`, or just `<operation>` if no collection is involved.


instrumentation/go.mongodb.org/mongo-driver/mongo/otelmongo/config.go
```diff
@@ -57,11 +55,3 @@ func WithTracerProvider(provider trace.TracerProvider) Option {
 		cfg.TracerProvider = provider
 	}
 }
-
-// WithCommandAttributeDisabled specifies if the MongoDB command is added as an attribute to Spans or not.
-// The MongoDB command will be added as an attribute to Spans by default if this option is not provided.
-func WithCommandAttributeDisabled(disabled bool) Option {
```
MrAlias, line 63: Is there a reason this is being removed? Didn't this resolve https://github.com/open-telemetry/opentelemetry-go-contrib/issues/712?

evantorrie: This was being used only to determine whether to include the "full" command in the `db.statement` attribute.

But according to the [example MongoDB attributes](https://github.com/open-telemetry/opentelemetry-specification/blob/2a712f7626cb62b21ae210ef3175612e7744f285/specification/trace/semantic_conventions/database.md#mongodb) given in the DB semantics tags specification
   * `db.statement` is *unset*
   * `db.operation` is set to the name of the operation
There doesn't appear to be an appropriate attribute for the full "command".

Let me survey other implementations.

evantorrie: Looks like other implementations do set `db.statement`. Best implementation seems to be the [Java autoinstrumentation ](https://github.com/open-telemetry/opentelemetry-java-instrumentation/blob/0b0516bd1a0599c8f536cc6b2151782000681f39/instrumentation/mongo/mongo-3.1/library/src/main/java/io/opentelemetry/instrumentation/mongo/v3_1/MongoClientTracer.java)where it both sanitizes the command and constrains it to a maximum size.

I've restored setting `db.statement` and the `WithCommandAttributeDisabled()`, but with a TODO for better sanitizing and max size limit.

Also reorganized the tests a little including adding a check for `SpanKindClient`

MrAlias (approved):

evantorrie:
@MrAlias This now looks ready to re-review again. No changes really from last time other than being rebased and having the integration tests fixed so that it's tested during the PR run.

MrAlias (approved):

pellared (approved):

PR #770, opened 2021-05-04
otelsarama: Add WithAllowRootSpanStart option

keitwb:
This will allow the parent app to control whether a span will be
created based on whether the message contains injected context.

The default is 'true' which will preserve the original behavior.

### DIFF
--- a/instrumentation/github.com/Shopify/sarama/otelsarama/producer.go
+++ b/instrumentation/github.com/Shopify/sarama/otelsarama/producer.go
@@ -37,7 +37,7 @@ type syncProducer struct {
 func (p *syncProducer) SendMessage(msg *sarama.ProducerMessage) (partition int32, offset int64, err error) {
 	span := startProducerSpan(p.cfg, p.saramaConfig.Version, msg)
 	partition, offset, err = p.SyncProducer.SendMessage(msg)
-	finishProducerSpan(span, partition, offset, err)
+	span.Finish(partition, offset, err)
 	return partition, offset, err
 }
 
@@ -45,13 +45,13 @@ func (p *syncProducer) SendMessage(msg *sarama.ProducerMessage) (partition int32
 func (p *syncProducer) SendMessages(msgs []*sarama.ProducerMessage) error {
 	// Although there's only one call made to the SyncProducer, the messages are
 	// treated individually, so we create a span for each one
-	spans := make([]trace.Span, len(msgs))
+	spans := make([]producerSpan, len(msgs))
 	for i, msg := range msgs {
 		spans[i] = startProducerSpan(p.cfg, p.saramaConfig.Version, msg)
 	}
 	err := p.SyncProducer.SendMessages(msgs)
 	for i, span := range spans {
-		finishProducerSpan(span, msgs[i].Partition, msgs[i].Offset, err)
+		span.Finish(msgs[i].Partition, msgs[i].Offset, err)
 	}
 	return err
 }
@@ -121,7 +121,7 @@ func (p *asyncProducer) Close() error {
 }
 
 type producerMessageContext struct {
-	span           trace.Span
+	span           producerSpan
 	metadataBackup interface{}
 }
 
@@ -152,7 +152,7 @@ func WrapAsyncProducer(saramaConfig *sarama.Config, p sarama.AsyncProducer, opts
 		// cannot be closed.
 		defer func() {
 			for _, mc := range producerMessageContexts {
-				finishProducerSpan(mc.span, 0, 0, nil)
+				mc.span.Finish(0, 0, nil)
 			}
 		}()
 		defer close(wrapped.successes)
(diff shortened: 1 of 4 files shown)


instrumentation/github.com/Shopify/sarama/otelsarama/producer.go
```diff
@@ -176,25 +183,31 @@ func WrapAsyncProducer(saramaConfig *sarama.Config, p sarama.AsyncProducer, opts
...
-					finishProducerSpan(span, msg.Partition, msg.Offset, nil)
+
+				if ok {
+					if saramaConfig.Producer.Return.Successes {
+						producerMessageContexts[msg.Metadata] = mc
+					} else {
+						// If returning successes isn't enabled, we just finish the
+						// span right away because there's no way to know when it will
+						// be done.
+						finishProducerSpan(span, msg.Partition, msg.Offset, nil)
```
pellared, line 209: I think there is a bug in the conditional logic here.
should it not be something like:
```go
switch {
	case ok && saramaConfig.Producer.Return.Successes:
		producerMessageContexts[msg.Metadata] = mc
	case !saramaConfig.Producer.Return.Successes:
		span.End()
}
```

instrumentation/github.com/Shopify/sarama/otelsarama/producer.go
```diff
@@ -35,9 +35,11 @@ type syncProducer struct {
 
 // SendMessage calls sarama.SyncProducer.SendMessage and traces the request.
 func (p *syncProducer) SendMessage(msg *sarama.ProducerMessage) (partition int32, offset int64, err error) {
-	span := startProducerSpan(p.cfg, p.saramaConfig.Version, msg)
+	span, ok := startProducerSpan(p.cfg, p.saramaConfig.Version, msg)
 	partition, offset, err = p.SyncProducer.SendMessage(msg)
-	finishProducerSpan(span, partition, offset, err)
+	if ok {
+		finishProducerSpan(span, partition, offset, err)
+	}
```
pellared, line 42: this design is very error-prone. why not moving the `if` logic to `finishProducerSpan`

instrumentation/github.com/Shopify/sarama/otelsarama/producer.go
```diff
@@ -35,9 +35,11 @@ type syncProducer struct {
 
 // SendMessage calls sarama.SyncProducer.SendMessage and traces the request.
 func (p *syncProducer) SendMessage(msg *sarama.ProducerMessage) (partition int32, offset int64, err error) {
-	span := startProducerSpan(p.cfg, p.saramaConfig.Version, msg)
+	span, ok := startProducerSpan(p.cfg, p.saramaConfig.Version, msg)
```
pellared, line 38: I find the design with returning the `ok` is very error-prone and hard to read. 

Maybe it would be better if this if this `ok` would be somehow encapsulated in a type returned by `startProducerSpan` which then can be accepted by `finishProducerSpan` and used in `producerMessageContext` instead of `trace.Span`

Even this should be enough:
 
```
type producerSpan struct {
	span trace.Span
```

Then  the `finishProducerSpan` could even be implemented as `producerSpan`'s method

keitwb: What I originally wanted to do was just use a "noop" span and then avoid almost all conditionals, but I couldn't find anything like that in the Otel library.  I think what you are proposing is just a variation of a noop span, so I'll try that.

keitwb: ok @pellared take a look at my latest push and see what you think.

instrumentation/github.com/Shopify/sarama/otelsarama/producer.go
```diff
@@ -121,7 +121,7 @@ func (p *asyncProducer) Close() error {
 }
 
 type producerMessageContext struct {
-	span           trace.Span
+	span           *producerSpan
```
pellared, line 124: Aby reason why a pointer is used? It is especially strange as the method does not handle nil value.

keitwb: Not really, just standard. I made it not a pointer as it doesn't really hurt copying the whole struct since there is only one field in it.

pellared: BTW using values instead of pointers is often faster. E.g. take a look at https://medium.com/a-journey-with-go/go-should-i-use-a-pointer-instead-of-a-copy-of-my-struct-44b43b104963

pellared (approved):

pellared:
@kweitb Generally the code looks OK but I am not sure it would not be better to use `Sampler` which filter's out span's without a parent. Have you checked `Sampler` interface and existing samplers?

instrumentation/github.com/Shopify/sarama/otelsarama/producer.go
```diff
@@ -227,11 +228,21 @@ func WrapAsyncProducer(saramaConfig *sarama.Config, p sarama.AsyncProducer, opts
 	return wrapped
 }
 
-func startProducerSpan(cfg config, version sarama.KafkaVersion, msg *sarama.ProducerMessage) trace.Span {
+// producerSpan is a simple wrapper around trace.Span (which can be nil) that
+// simplifies the logic in the wrapper in the case that no span is really present.
+type producerSpan struct {
```
XSAM, line 233: I'm not sure about this wapper. startProducerSpan and startProducerSpan seem paired well, but now the wrapper breaks this.

XSAM (approved):
Looks good. Some comments and questions you might want to take a look.

instrumentation/github.com/Shopify/sarama/otelsarama/producer.go
pellared: The code without any additional type was not readable, contained more nesting, and was a lot more error-prone.

pellared:
@keitwb Do you still want/need this?

pellared:
`otelsarama` is deprecated. I am closing this PR. More info: https://github.com/open-telemetry/opentelemetry-go-contrib/issues/4091

PR #771, opened 2021-05-07
net/http: add requests count metric with status code label

### LINKED ISSUE #711: Add status code to net/http metrics
Hello! I have a minor request for improvement for one of the instrumentation packages.

My use-case:
There is a service, and I want to know the percentage of failed requests on it. Previously there was the combination of httpsnoop and Prometheus client_go to capture status code and add it to metrics as a label.

The net/http handler wrapper already extracts status code from the response writer but adds it only to span. [See here](https://github.com/open-telemetry/opentelemetry-go-contrib/blob/main/instrumentation/net/http/otelhttp/handler.go#L176). And there is a useful `labeler`, but it can only add some constant labels or information from a user's request, e.g., path, URL.

Maybe it is not a good idea to add status code to metric labels by default because of increasing metrics cardinality. I thought about adding an option to configure it somehow.

So, there are two questions left regarding all this:
1. Is it ok to add status code to metrics? I think it is if you want to use the RED method and differentiate `E`'s.
2. Looking through the code, it is not clear how to fulfill this request, but I hope that someone can give me a hint.

nabokihms:
Closes #711

This PR adds a counter for the total amount of requests with the http_status_code label. It will help us to be aware of the ratio of errors. 

P.S. I decided to avoid adding status code to other metrics because it looks excessive. Instead, counting the number of requests seems more sensible.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -10,6 +10,10 @@ This project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0.htm
 
 ## [0.21.0] - 2021-06-18
 
+### Added
+
+- Add total requests counter by response status code for the `net/http` instrumentation. (#771)
+
 ### Fixed
 
 - Dockerfile based examples for `otelgin` and `otelmacaron`. (#767)
--- a/instrumentation/net/http/otelhttp/handler_test.go
+++ b/instrumentation/net/http/otelhttp/handler_test.go
@@ -82,7 +82,19 @@ func TestHandlerBasics(t *testing.T) {
 		attribute.String("test", "attribute"),
 	}
 
-	assertMetricAttributes(t, attributesToVerify, meterimpl.MeasurementBatches)
+	var statusCodeBatch oteltest.Batch
+	var measurementBatches []oteltest.Batch
+	for _, batch := range meterimpl.MeasurementBatches {
+		batchName := batch.Measurements[0].Instrument.Descriptor().Name()
+		if batchName == RequestCount {
+			statusCodeBatch = batch
+			continue
+		}
+		measurementBatches = append(measurementBatches, batch)
+	}
+
+	assertMetricAttributes(t, attributesToVerify, measurementBatches)
+	assert.ElementsMatch(t, append(attributesToVerify, semconv.HTTPStatusCodeKey.Int(200)), statusCodeBatch.Labels)
 
 	if got, expected := rr.Result().StatusCode, http.StatusOK; got != expected {
 		t.Fatalf("got %d, expected %d", got, expected)
--- a/instrumentation/net/http/otelhttp/handler.go
+++ b/instrumentation/net/http/otelhttp/handler.go
@@ -102,11 +102,15 @@ func (h *Handler) createMeasures() {
 	responseBytesCounter, err := h.meter.NewInt64Counter(ResponseContentLength)
 	handleErr(err)
 
+	requestCount, err := h.meter.NewInt64Counter(RequestCount)
+	handleErr(err)
+
 	serverLatencyMeasure, err := h.meter.NewInt64ValueRecorder(ServerLatency)
 	handleErr(err)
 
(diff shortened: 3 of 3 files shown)


nabokihms:
I rebased the PR into the main branch and added a changelog entry. If it is possible, I would like to move this forward to merge. Maybe I have missed something from the contributing guide? Thanks in advance for any feedback.

CHANGELOG.md
```diff
@@ -10,6 +10,10 @@ This project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0.htm
 
 ## [0.21.0] - 2021-06-18
 
+### Added
+
+- Add total requests counter by response status code for the `net/http` instrumentation. (#771)
+
```
Aneurysm9, line 16: This would need to be moved up into the "Unreleased" section above.

Aneurysm9 (approved):
Thanks for the contribution and I apologize for taking so long to respond to it.  We have been deferring the addition of new features in contrib instrumentation while we worked through stabilizing the tracing API and SDK and separating it from the metrics API and SDK.  

Now that we have done that, we will need to review the instrumentation in contrib with a view toward making it stable (v1.0.0 with a stable public API and no unstable dependencies).  Unfortunately, I believe that this will require extracting the metrics capability from this instrumentation.  I'm not sure at this time what form a new HTTP metrics implementation will take or what metrics it will provide.  

I'm of two minds with regard to adding to the existing metrics instrumentation.  On the one hand I don't want to add to something that will be removed shortly, while on the other the fact that it will be removed shortly means it probably doesn't hurt to add it.  I will conditionally approve this PR and the other @open-telemetry/go-approvers can weigh in with their thoughts.

instrumentation/net/http/otelhttp/handler_test.go
```diff
@@ -82,7 +82,19 @@ func TestHandlerBasics(t *testing.T) {
...
 	}
 
-	assertMetricAttributes(t, attributesToVerify, meterimpl.MeasurementBatches)
+	var statusCodeBatch oteltest.Batch
+	var measurementBatches []oteltest.Batch
+	for _, batch := range meterimpl.MeasurementBatches {
+		batchName := batch.Measurements[0].Instrument.Descriptor().Name()
+		if batchName == RequestCount {
+			statusCodeBatch = batch
+			continue
```
pellared, line 91: is this continue needed? 

If I am not missing anything then I think that we could `assertMetricAttributes` for `RequestCount` metric as well

nabokihms: Thanks, I will check it. Maybe I missed something.

sagikazarmark:
> Unfortunately, I believe that this will require extracting the metrics capability from this instrumentation.

@Aneurysm9 can you elaborate why that is necessary and how things would look like from your perspective after that? I'm trying to understand how things would work, so we can build something around it while the reference implementation is finished.

merlinran:
unsolicited reply regards to why this is required: it's pretty natural for me to think about [the RED method](https://www.weave.works/blog/the-red-method-key-metrics-for-microservices-architecture) when it comes to monitoring services. While rate and duration are covered by otelhttp, error is not, and HTTP status code is a great way to observe if and why the service doesn't fulfill the request.

BTW it's also covered in [the open-telemetry HTTP metrics spec](https://github.com/open-telemetry/opentelemetry-specification/blob/main/specification/metrics/semantic_conventions/http-metrics.md), among many others, though it's not quite clear to me the relation between this one and the spec, and the spec only mentions gauges, but no counters, which is quite interesting.

nabokihms:
I had no chance to work on this one, bit I am still willing to add this feature. @Aneurysm9 I will rebase this PR if you don't mind.

Aneurysm9:
> I had no chance to work on this one, bit I am still willing to add this feature. @Aneurysm9 I will rebase this PR if you don't mind.

I would hold off until next week.  We just released a new metrics API and will be updating the instrumentation in this repo shortly to use it.

PR #777, opened 2021-05-13
Create .gitattributes

pellared:
## Why

Get rid of false negatives on Windows like: 
```
File is not `gofmt`-ed with `-s` (gofmt)
```

Also, it helps to work on Windows in native, WSL and Development Container modes on the same repository.

## What

Add .gitattributes so that most files (especially `*.sh` ,`*.go`, `Makefile`) use LF and only Windows-specific have CLRF line ending.

### DIFF
--- a/.gitattributes
+++ b/.gitattributes
@@ -1,0 +1,3 @@
+* text=auto eol=lf
+*.{cmd,[cC][mM][dD]} text eol=crlf
+*.{bat,[bB][aA][tT]} text eol=crl
--- a/.gitattributes
+++ b/.gitattributes
@@ -1,3 +1,3 @@
 * text=auto eol=lf
 *.{cmd,[cC][mM][dD]} text eol=crlf
-*.{bat,[bB][aA][tT]} text eol=crl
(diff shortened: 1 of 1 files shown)


pellared:
> Please add a CHANGELOG entry, or add the "Skip Changelog" label if not required

😉

.gitattributes
```diff
@@ -0,0 +1,3 @@
+* text=auto eol=lf
+*.{cmd,[cC][mM][dD]} text eol=crlf
+*.{bat,[bB][aA][tT]} text eol=crl
```
evantorrie, line 3: Is the last line supposed to end with `eol=crl`?

I don't really use Windows at all, but my understanding is the two possible attributes for `eol` are `crlf` or `lf`.

See https://www.git-scm.com/docs/gitattributes#_eol

evantorrie (commented):
Just checking whether the last character was dropped?

.gitattributes
pellared: typo!!! thanks!

evantorrie (approved):

MrAlias (approved):

PR #787, opened 2021-05-17
Update tested Go version

MrAlias:
Keep in line with the main repository and test version 1.15 and 1.16 of Go. Drop support for Go 1.14.


instrumentation/net/http/otelhttp/filters/header.go
```diff
@@ -12,8 +12,6 @@
 // See the License for the specific language governing permissions and
 // limitations under the License.
 
-// +build !go1.14
```
Aneurysm9, line 15: Shouldn't this be the file to be deleted and the other retained?  This was included to support <1.14 and the other file uses the `Values()` method that was added in 1.14 and should be available in all presently supported versions.

pre_release.sh
```diff
@@ -71,7 +71,7 @@ if [ -n "${OTEL_TAG}" ]; then
     TMPDIR=$(mktemp -d "/tmp/otel-contrib.XXXXXX") || exit 1
     trap "rm -fr ${TMPDIR}" EXIT
     (cd "${TMPDIR}" && go mod init tagtest)
-    # requires go 1.14 for support of '-modfile'
+    # requires go 1.15 for support of '-modfile'
```
Aneurysm9, line 74: Not sure that this version requirement changes, even if it's prior to the MSV.

MrAlias: Sure

```suggestion
    # requires go 1.14 for support of '-modfile'
```

instrumentation/net/http/otelhttp/filters/header.go
MrAlias: From my understanding of build constraints including a `!` means the matching constrains will not be built for. Thus this file is the one built for all non Go 1.14 build (i.e. this was the one supported for Go 1.15). This seems to be inline with what the file name is (i.e. `*_nongo14.go`). The other file deleted, `header_go14.go` had the build constraint (`+build go1.14`) that included it in builds for Go 1.14 AFAIK.

MadVikingGod: I think @Aneurysm9 is correct.  This file was the pre 1.14 version.  Now we can retrieve this slice via the `Values()` method.  I think we should switch to the other file, rename it to just header.go and remove the build constraint.

dashpole (approved):

evantorrie (approved):

PR #821, opened 2021-06-15
Switch from WithSyncer to WithBatcher in examples

dashpole:
Related to https://github.com/open-telemetry/opentelemetry-go/issues/1039, and similar to https://github.com/open-telemetry/opentelemetry-go/pull/2007, but for contrib.

A user accidentally used WithSyncer in their application (though not from this example), which added extra unexplainable latency in their request path. Use WithBatcher and ForceFlush in examples to make it explicit that we are waiting for spans to be sent to stdout at each point, and to discourage the use of WithSyncer.

### DIFF
--- a/instrumentation/github.com/Shopify/sarama/otelsarama/example/consumer/consumer.go
+++ b/instrumentation/github.com/Shopify/sarama/otelsarama/example/consumer/consumer.go
@@ -37,7 +37,12 @@ var (
 )
 
 func main() {
-	example.InitTracer()
+	tp := example.InitTracer()
+	defer func() {
+		if err := tp.Shutdown(context.Background()); err != nil {
+			log.Printf("Error shutting down tracer provider: %v", err)
+		}
+	}()
 	flag.Parse()
 
 	if *brokers == "" {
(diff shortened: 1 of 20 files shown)


instrumentation/github.com/Shopify/sarama/otelsarama/example/consumer/consumer.go
```diff
@@ -37,7 +37,8 @@ var (
 )
 
 func main() {
-	example.InitTracer()
+	tp := example.InitTracer()
+	defer tp.ForceFlush(context.Background())
```
MrAlias, line 41: Shutdown instead?

```suggestion
	defer tp.Shutdown(context.Background())
```

Aneurysm9, line 41: Should this `Shutdown()` instead of `ForceFlush()`?

dashpole: done

dashpole: done

Aneurysm9 (approved):

MrAlias (approved):

PR #829, opened 2021-06-24
Added option to using a skipper in the echo otel middleware

pcairns:
Allow for a skipper (https://github.com/labstack/echo/blob/master/middleware/middleware.go#L15) to be passed into the echo middleware. The use case for this is we would like the ability to bypass tracing for certain routes (eg. health check endpoints).


Aneurysm9:
@pcairns I think you may need to rebase this to eliminate commit `366d0ca`.  All commits must be from users that have signed the CLA.

pcairns:
@Aneurysm9 sorry I've been away on vacation, this should be rebased now

instrumentation/github.com/labstack/echo/otelecho/echo_test.go
```diff
@@ -225,3 +225,24 @@ func TestPropagationWithCustomPropagators(t *testing.T) {
...
+	router.Use(Middleware("foobar", WithSkipper(skipper)))
+	router.GET("/ping", func(c echo.Context) error {
+		span := oteltrace.SpanFromContext(c.Request().Context())
+		assert.False(t, span.SpanContext().HasSpanID())
+		assert.False(t, span.SpanContext().HasTraceID())
+		return c.NoContent(200)
+	})
+
+	router.ServeHTTP(w, r)
+
```
pellared, line 247: just to make sure that the "ping" handler is really invoked (eg. if you change the first line to `r := httptest.NewRequest("GET", "/bad", nil)` the test will pass)

```suggestion
	resp := w.Result()
	assert.Equal(t, http.StatusOK, resp .StatusCode)

```



would be good also to do the same in other tests

pellared (approved):
one little comment to make the test(s) more trustworthy

other than that looks great 👍

pcairns:
I've updated the tests to include a check for the status code

pellared (approved):

Aneurysm9 (approved):

instrumentation/github.com/labstack/echo/otelecho/config.go
```diff
@@ -52,3 +55,10 @@ func WithTracerProvider(provider oteltrace.TracerProvider) Option {
 		cfg.TracerProvider = provider
 	})
 }
+
+//WithSkipper specifies a skipper for allowing requests to skip generating spans.
+func WithSkipper(skipper middleware.Skipper) Option {
+	return func(cfg *config) {
+		cfg.Skipper = skipper
+	}
```
Aneurysm9, line 63:
```suggestion
	return optionFunc(func(cfg *config) {
		cfg.Skipper = skipper
	})
```

PR #835, opened 2021-06-28
add opentelemetry support for gorm2

wei840222:
Here is instrumentation for https://github.com/go-gorm/gorm.
A popular SQL database ORM framework.
Include README.md for how to used.
It seems need some test for this lib, it will be processed later.


vadimberezniker:
I am looking at adding GORM instrumentation to our service and came across this PR.
I see that this is derived from https://github.com/go-gorm/opentracing. Have you reached out the GORM maintainers to ask if they would prefer that this be maintained as part of GORM rather as part of the otel repo?
The other observation is that https://github.com/go-gorm/opentracing is MIT licensed and this PR is stripping the copyright & license.

Aneurysm9 (changes requested):
Thanks for the feedback @vadimberezniker.  I think our general preference is to have first-party instrumentation supported by library authors.  If the GORM maintainers wanted to support this instrumentation that would be great and I'd be happy to help them in any way possible.

It is concerning that this contribution appears to be derived from the existing GORM OpenTracing instrumentation without attribution or appropriate license provenance.  I'm not sure that we can accept this contribution or anything substantially similar to it.

MrAlias:
> Thanks for the feedback @vadimberezniker. I think our general preference is to have first-party instrumentation supported by library authors. If the GORM maintainers wanted to support this instrumentation that would be great and I'd be happy to help them in any way possible.
> 
> It is concerning that this contribution appears to be derived from the existing GORM OpenTracing instrumentation without attribution or appropriate license provenance. I'm not sure that we can accept this contribution or anything substantially similar to it.

Agreed. Thanks for the contribution, but this needs more strategy and community involvement before we are ready to accept this. Please open an "Instrumentation Request" type Issue where the solution can be discussed. I would, similar to others, prefer the instrumentation live natively in the library that is being instrumented. But am open to more discussion on the topic in the Issue if there is disagreement.

Issue #818, opened 2021-06-09
otelgprc: stream not closed on eof


tonistiigi:
Currently, when stream interceptor is used it closes only if EOF is received and the client also calls `CloseSend` https://github.com/open-telemetry/opentelemetry-go-contrib/blob/f2fb9c8d5524abba85c87ef341d707c7a8173932/instrumentation/google.golang.org/grpc/otelgrpc/interceptor.go#L225 . That behavior was a bit surprising to me as I've always considered `CloseSend` optional as grpc only allows one-sided close from one side. The docs in `CloseSend` are not very specific but the docs for `RecvMsg` do say that `EOF` indicates completion of the stream. https://pkg.go.dev/google.golang.org/grpc#ClientStream

My suggestion would be for the client stream to finish always if EOF is received. I think `CloseSend` should be completely ignored. It can log an event but it doesn't indicate that stream is complete nor is it required for the stream to finish.

I'm willing to implement the required changes if the above makes sense.

vadimberezniker:
Tõnis, I completely agree with you. We recently ran into massive goroutine leaks after adding gRPC instrumentation. 

I sent https://github.com/open-telemetry/opentelemetry-go-contrib/pull/836 to fix this. Let's see what the project approvers think.

PR #836, opened 2021-06-29
gRPC: fix goroutine leak if client stream CloseSend is not called.

vadimberezniker:
Outside of a unidirectional Client->Server stream there's no requirement
that CloseSend must be called (and it doesn't make much sense for a
Server->Client stream).

If CloseSend is not called, the RPC span will not be ended and a
goroutine will be leaked.

We saw a large number of goroutines leaked when we enabled gRPC
instrumentation in our service.

Resolves https://github.com/open-telemetry/opentelemetry-go-contrib/issues/818


Aneurysm9 (approved):

vadimberezniker:
Friendly ping :)

instrumentation/google.golang.org/grpc/otelgrpc/interceptor.go
```diff
@@ -208,24 +200,15 @@ func wrapClientStream(s grpc.ClientStream, desc *grpc.StreamDesc) *clientStream
...
-		// Both streams have to be closed
-		state := byte(0)
-
 		for event := range events {
 			switch event.Type {
-			case closeEvent:
-				state |= clientClosedState
 			case receiveEndEvent:
-				state |= receiveEndedState
+				finished <- nil
```
coryb, line 206: I was running into the same issues and after applying this PR I was still seeing some leaks.  I think this fixes some of the issues, but I think there is still an edge case where the client context is cancelled in which case we may not receive any more events.  I patched this loop to select on the `s.Context.Done()` along with `events` and all my leaks have gone away.  This is what my loop looks like:
```
		for {
			select {
			case <-s.Context().Done():
				finished <- s.Context().Err()
				return
			case event := <-events:
				switch event.Type {
				case receiveEndEvent:
					finished <- nil
					return
				case errorEvent:
					finished <- event.Err
					return
				}
			}
		}
```

vadimberezniker: Please see https://github.com/open-telemetry/opentelemetry-go-contrib/pull/840

tonistiigi (approved):

vadimberezniker:
Ping. This PR has been open for 2 weeks now.

XSAM (approved):
💯

vadimberezniker:
Thank you! Any blockers for merging this?

vadimberezniker:
What's left?

MrAlias (approved):

Issue #839, opened 2021-06-30
otelgrpc: goleaks if context canceled


glazychev-art:
Hello,
We found that `StreamClientInterceptor` doesn't process context cancellation. And we see goleaks in this case.
For example, let's take a look at the next test: 
```
func TestStreamClientInterceptorCancelContext(t *testing.T) {
	// enable goroutine leak detector 
	defer goleak.VerifyNone(t)

	clientConn, err := grpc.Dial("fake:connection", grpc.WithInsecure())
	if err != nil {
		t.Fatalf("failed to create client connection: %v", err)
	}
	defer clientConn.Close()

	// tracer
	sr := NewSpanRecorder()
	tp := oteltest.NewTracerProvider(oteltest.WithSpanRecorder(sr))
	streamCI := StreamClientInterceptor(WithTracerProvider(tp))

	var mockClStr mockClientStream
	method := "/github.com.serviceName/bar"
	name := "github.com.serviceName/bar"
	
	// create a context with cancel
	cancelCtx, cancel := context.WithCancel(context.Background())
	defer cancel()
	streamClient, err := streamCI(
		cancelCtx,
		&grpc.StreamDesc{ServerStreams: true},
		clientConn,
		method,
		func(ctx context.Context,
			desc *grpc.StreamDesc,
			cc *grpc.ClientConn,
			method string,
			opts ...grpc.CallOption) (grpc.ClientStream, error) {
			mockClStr = mockClientStream{Desc: desc, Ctx: ctx}
			return mockClStr, nil
		},
	)
	require.NoError(t, err, "initialize grpc stream client")
	_, ok := getSpanFromRecorder(sr, name)
	require.False(t, ok, "span should ended while stream is open")

	req := &mockProtoMessage{}
	reply := &mockProtoMessage{}

	// send and receive fake data
	for i := 0; i < 10; i++ {
		_ = streamClient.SendMsg(req)
		_ = streamClient.RecvMsg(reply)
	}

	// close client stream
	_ = streamClient.CloseSend()
}
```
Output:
```
=== RUN   TestStreamClientInterceptorCancelContext
    leaks.go:78: found unexpected goroutines:
        [Goroutine 9 in state chan receive, with go.opentelemetry.io/contrib/instrumentation/google.golang.org/grpc/otelgrpc.wrapClientStream.func1 on top of the stack:
        goroutine 9 [chan receive]:
[...]

PR #840, opened 2021-06-30
otelgrpc: fix goroutine leak after context cancellation

glazychev-art:
This PR is related to this issue: #839


glazychev-art:
@vadimberezniker @tonistiigi
Hello. As far as I know, you have caught a similar problem. Could you please take a look too?

denis-tingaikin (approved):

instrumentation/google.golang.org/grpc/otelgrpc/interceptor.go
```diff
@@ -211,19 +211,25 @@ func wrapClientStream(s grpc.ClientStream, desc *grpc.StreamDesc) *clientStream
...
+				case errorEvent:
+					finished <- event.Err
+					return
+				}
+				if state == clientClosedState|receiveEndedState {
+					finished <- nil
+					return
+				}
+
+			case <-s.Context().Done():
```
vadimberezniker, line 231: Disclaimer: I'm not a maintainer on this project but I have been looking at this code recently.

If we trust the ClientStream.Context() docs, it's not safe to use the Context() call here:
```
	// Context returns the context for this stream.
	//
	// It should not be called until after Header or RecvMsg has returned. Once
	// called, subsequent client-side retries are disabled.
	Context() context.Context
```

The alternative is to pass the parent context and check that.

Thanks for fixing this, I noticed this while fixing the other issue and this was next on my plate.

tonistiigi:
Does `<-s.Context().Done()` return then calling context is canceled or when stream is closed from the server-side and `recv` return eof? If former then I don't think this is correct and we should wait for server-side to close stream, maybe calling `recvmsg` is goroutine is an option if there isn't anything better.

vadimberezniker:
> Does `<-s.Context().Done()` return then calling context is canceled or when stream is closed from the server-side and `recv` return eof? If former then I don't think this is correct and we should wait for server-side to close stream, maybe calling `recvmsg` is goroutine is an option if there isn't anything better.

If the client properly handles the stream, the ctx.Done() case should not be reached. That should be handled by the event for loop.
The issue @glazychev-art is trying to address is the case where the parent context is cancelled before the client is done with the stream.

tonistiigi:
@vadimberezniker I understand that. The canceling flow should be 1) context client passed gets canceled 2) cancellation is passed to the server-side 3) server closes either cleanly or with an error 4) client receives the close and stream is finished. Just cancelling the context is just a signal and doesn't atomically close the stream by itself. I'm not sure about the internals, therefore it may be possible that `s.Context()` only closes after 4), in that case, the change looks correct to me. But if it closes after 1) then I don't think this approach is correct.

glazychev-art:
@vadimberezniker 
Thank you. Updated to parent context

glazychev-art:
@tonistiigi 
It is a good idea, but I'm not sure if this possible.. Because, let's take a look at:
https://github.com/grpc/grpc-go/blob/master/stream.go#L358-L372
```
		go func() {
			select {
			case <-cc.ctx.Done():
				cs.finish(ErrClientConnClosing)
			case <-ctx.Done():
				cs.finish(toRPCErr(ctx.Err()))
			}
		}()
```
ClientStream is keeping track context closing. And does not notify about it.
So, if we call `RecvMsg` it may be already finished.

vadimberezniker:
> @vadimberezniker I understand that. The canceling flow should be 1) context client passed gets canceled 2) cancellation is passed to the server-side 3) server closes either cleanly or with an error 4) client receives the close and stream is finished. Just cancelling the context is just a signal and doesn't atomically close the stream by itself. I'm not sure about the internals, therefore it may be possible that `s.Context()` only closes after 4), in that case, the change looks correct to me. But if it closes after 1) then I don't think this approach is correct.

Based on my understanding when the context is cancelled, gRPC automatically closes the underlying HTTP2 stream so there's no opportunity for the server to send anything after that.

tonistiigi:
I did some tests and indeed the server-side graceful cancel does not happen as I described before. I had forgotten that in the code where I use it I actually need to do two requests and only cancel one of them (that forces another to close cleanly) to achieve this behavior.

But it is still possible that `RecvMsg` returns cleanly after the context has been canceled if server-side sent the message before it received the cancel(after that sends will fail). Looking at the interceptor I think it probably just means that the events for the receive get dropped that isn't catastrophic. Although, I think I still kind of consider it a caller side issue if the client just closes the stream and doesn't recv until an error after that.

glazychev-art:
@Aneurysm9 @dashpole @evantorrie @jmacd @MadVikingGod @MrAlias @paivagustavo @XSAM
Any thoughts on this?

Aneurysm9:
The approach seems sane.  Please fix the linting issue.

glazychev-art:
@Aneurysm9
Thank you. Fixed

instrumentation/google.golang.org/grpc/otelgrpc/interceptor_test.go
```diff
@@ -372,6 +372,59 @@ func TestStreamClientInterceptor(t *testing.T) {
...
+			cc *grpc.ClientConn,
+			method string,
+			opts ...grpc.CallOption) (grpc.ClientStream, error) {
+			mockClStr = mockClientStream{Desc: desc, Ctx: ctx}
+			return mockClStr, nil
+		},
+	)
+	require.NoError(t, err, "initialize grpc stream client")
+	_, ok := getSpanFromRecorder(sr, name)
+	require.False(t, ok, "span should ended while stream is open")
```
XSAM, line 413:
```suggestion
	require.False(t, ok, "span should not ended while stream is open")
```

Seems more resonable.

XSAM (approved):
Looks good overall.

instrumentation/google.golang.org/grpc/otelgrpc/interceptor_test.go
glazychev-art: Ok, fixed

instrumentation/google.golang.org/grpc/otelgrpc/interceptor_test.go
```diff
@@ -372,6 +372,59 @@ func TestStreamClientInterceptor(t *testing.T) {
 	_ = streamClient.CloseSend()
 }
 
+// TestStreamClientInterceptorCancelContext tests a cancel context situation.
+// There should be no goleaks
+func TestStreamClientInterceptorCancelContext(t *testing.T) {
```
pellared, line 377: Cannot we simply modify the `TestStreamClientInterceptor` instead of creating a new test?

glazychev-art: I'm not sure.. because, as I understand, the `TestStreamClientInterceptor` demonstrates normal stream termination receiving `receiveEndEvent`. `TestStreamClientInterceptorCancelContext` ends the stream by closing the context.

pellared (approved):

edwarnicke:
@pellared Pardon my ignorance of the process... but what else has to be done before this fix can be merged?

pellared:
@Aneurysm9 @MrAlias PTAL

MrAlias (approved):

edwarnicke:
Yay!

Issue #851, opened 2021-07-06
Unify Configuration and Initialization


pellared:
## Why

This is a "clone" of https://github.com/open-telemetry/opentelemetry-go/issues/536 so that we can track adherence to https://github.com/open-telemetry/opentelemetry-go/blob/main/CONTRIBUTING.md#configuration

## What

Modules compliance table:

| Module | Compliant | Review date |
| --- | ---| --- |
| . | ✔️ | 2021-07-06 |
| ./detectors/aws | ✔️ | 2021-07-06 |
| ./detectors/aws/ecs | ✔️ | 2021-07-06 |
| ./detectors/aws/eks | ✔️ | 2021-07-06 |
| ./detectors/gcp | ✔️ | 2021-07-06 |
| ./exporters/metric/cortex | ❌ ❓   | 2021-07-06 |
| ./exporters/metric/cortex/utils | ❌ ❓ | 2021-07-06 |
| ./exporters/metric/datadog | ❌ ❓ | 2021-07-06 |
| ./exporters/metric/dogstatsd | ❌ ❓ | 2021-07-06 |
| ./instrumentation/github.com/Shopify/sarama/otelsarama |❌ https://github.com/open-telemetry/opentelemetry-go-contrib/pull/852 ⌛  | 2021-07-06 |
| ./instrumentation/github.com/astaxie/beego/otelbeego |❌ https://github.com/open-telemetry/opentelemetry-go-contrib/pull/853 ⌛ | 2021-07-06 |
| ./instrumentation/github.com/aws/aws-sdk-go-v2/otelaws |❌ https://github.com/open-telemetry/opentelemetry-go-contrib/pull/854 ⌛  | 2021-07-06 |
| ./instrumentation/github.com/bradfitz/gomemcache/memcache/otelmemcache |❌ https://github.com/open-telemetry/opentelemetry-go-contrib/pull/855 ⌛ | 2021-07-06 |
| ./instrumentation/github.com/emicklei/go-restful/otelrestful |❌ https://github.com/open-telemetry/opentelemetry-go-contrib/pull/856 ⌛ | 2021-07-06 |
| ./instrumentation/github.com/gin-gonic/gin/otelgin |❌ https://github.com/open-telemetry/opentelemetry-go-contrib/pull/857 ⌛ | 2021-07-06 |
[...]

PR #868, opened 2021-07-07
b3: Introduce New function and option type

pellared:
## Why

Per https://github.com/open-telemetry/opentelemetry-go-contrib/issues/851

## What

Adhere to https://github.com/open-telemetry/opentelemetry-go/blob/main/CONTRIBUTING.md#configuration

⚠️ It introduces breaking changes.

## Other remarks

During the last SIG meeting, we agreed that we should have a single Go Module per propagator. This will be done in a dedicated PR.


propagators/b3/b3_propagator.go
```diff
@@ -89,22 +80,21 @@ const (
 //    x-b3-spanid: {SpanId}
 //    x-b3-sampled: {SamplingState}
 //    x-b3-flags: {DebugFlag}
-type B3 struct {
-	// InjectEncoding are the B3 encodings used when injecting trace
-	// information. If no encoding is specified (i.e. `B3Unspecified`)
-	// `B3SingleHeader` will be used as the default.
-	InjectEncoding Encoding
+// The Single Header propagator is used by default.
+func New(opts ...Option) propagation.TextMapPropagator {
```
pellared, line 84: Should this function return a `propagation.TextMapPropagator` or `B3` type?

If we return  `propagation.TextMapPropagator` then we can also unexport `B3` type and have more flexibility in the future (e.g. pointer vs struct usage depending on benchmarks).

MrAlias: +1 returning `propagation.TextMapPropagator` and unexporting `B3`.

pellared: Unexported `B3`. Done here: 3119ced

MrAlias (approved):

Aneurysm9 (approved):

PR #870, opened 2021-07-08
Fix host and runtime instrumentation examples

Frefreak:
I believe current host example stops the meter [too early](https://github.com/open-telemetry/opentelemetry-go-contrib/blob/main/instrumentation/host/example/main.go#L52). When testing there is no output.

This PR also set the collect period to 3 seconds instead of the default 10 seconds to make it more "responsive".


Frefreak:
A quick grep reveals runtime has the same issue. Also updated.

MrAlias (approved):

instrumentation/host/example/main.go
```diff
@@ -42,14 +43,18 @@ func initMeter() *controller.Controller {
...
 	global.SetMeterProvider(pusher.MeterProvider())
 	return pusher
 }
 
 func main() {
-	defer handleErr(initMeter().Stop(context.Background()))
+	pusher := initMeter()
+	defer func(pusher *controller.Controller, ctx context.Context) {
+		handleErr(pusher.Stop(ctx))
+	}(pusher, context.Background())
```
pellared, line 57: I find this example code a little too hard to read.
What about something like:

```go
func main() {
	exporter, err := stdout.New(stdout.WithPrettyPrint())
	if err != nil {
		log.Fatalln("failed to initialize metric stdout exporter:", err)
	}
	pusher := controller.New(
		processor.New(
			simple.NewWithInexpensiveDistribution(),
			exporter,
		),
		controller.WithExporter(exporter),
		controller.WithCollectPeriod(3*time.Second),
	)
	if err := pusher.Start(context.Background()); err != nil {
		log.Fatalln("failed to start the metric controller:", err)
	}
	global.SetMeterProvider(pusher.MeterProvider())

	if err := host.Start(); err != nil {
		log.Fatalln("failed to start host instrumentation:", err)
	}

	ctx, cancel := newOSSignalContext()
	defer cancel()
	<-ctx.Done()

	if err := pusher.Stop(context.Background()); err != nil {
		log.Fatalln("failed to stop the metric controller:", err)
	}
}

func newOSSignalContext() (context.Context, func()) {
	// trap Ctrl+C and call cancel on the context
	ctx, cancel := context.WithCancel(context.Background())
	c := make(chan os.Signal, 1)
	signal.Notify(c, os.Interrupt)
	go func() {
		select {
		case <-c:
			cancel()
		case <-ctx.Done():
		}
	}()

	return ctx, func() {
		signal.Stop(c)
		cancel()
	}
}
```

pellared (commented):
I propose simplifying the example. The current one is (at least for me) quite hard to follow.

The especially funny thing is that the current example uses the following packages for logging errors:
- `fmt`
- `panic`
- `log`

instrumentation/host/example/main.go
jmacd: ^^ This looks good.

Frefreak:
Modified _host_ and _runtime_ using code given by @pellared (As a relatively new golang user I feels the `newOSSignalContext` thing is a bit complex though, but for the main purpose of the example this can be ignored if read by users)

pellared (approved):

PR #873, opened 2021-07-10
otelhttp: get tracer from current context if not set in constructor

tonistiigi:
When `otelhttp` is used as a `Transport` the traces should go to the parent span tracer by default. If custom `TracerProvider` is set then it is still used and if none is set then global one is used like it was before.

@Aneurysm9 

Signed-off-by: Tonis Tiigi <tonistiigi@gmail.com>


instrumentation/net/http/otelhttp/handler.go
```diff
@@ -127,8 +127,18 @@ func (h *Handler) ServeHTTP(w http.ResponseWriter, r *http.Request) {
...
 
+	tracer := h.tracer
+
+	if tracer == nil {
+		if span := trace.SpanFromContext(r.Context()); span.SpanContext().IsValid() {
+			tracer = newTracer(span.TracerProvider())
+		} else {
+			tracer = newTracer(otel.GetTracerProvider())
+		}
+	}
```
pellared, line 138: Maybe this code should be moved to some function in the `otel` or package like `otel.TracerProviderFromContext(ctx)`?

instrumentation/net/http/otelhttp/config.go
```diff
@@ -59,18 +59,17 @@ func (o optionFunc) apply(c *config) {
...
 	}
 	for _, opt := range opts {
 		opt.apply(c)
 	}
 
-	c.Tracer = c.TracerProvider.Tracer(
-		instrumentationName,
-		trace.WithInstrumentationVersion(contrib.SemVersion()),
-	)
+	// Tracer is only initialized if manually specified. Otherwise, can be passed with the tracing context.
```
pellared, line 69: a comment in code like this is not enough for documenting the behavior
it should be at least noted in "entry" point functions/types e.g. `NewHandler` and `NewTransport`

Moreover the new feature  should be noted in `CHANGELOG.md`

pellared (commented):
I am not sure if there was any issue or discussion before this PR was created, please link it if there is some. I imagine that if we would like to support such behavior then we should probably create some helper in SDK and then use it in all applicable instrumentation packages (you already repeated the functionality in https://github.com/open-telemetry/opentelemetry-go-contrib/pull/874). This would deserve an issue at least for tracking purposes.

instrumentation/net/http/otelhttp/config.go
tonistiigi: This is not documenting behavior but explaining why the `.Meter` and `.Tracer` are handled differently to the reader. This switch could be in `RoundTrip()` as well.

instrumentation/net/http/otelhttp/handler.go
tonistiigi: Not really related to this change as API already supports this behavior. I think I'm personally hesitant of adding new public functions for every case two other public functions are used together. In my codebase, this same check is used a lot of times but there I always return early if `IsValid` is not true as I don't want anything to rely on the global tracer. I guess if this was discussed in the `otel` repo as a shorthand I would find `otel.StartSpan(ctx, name, ..)` more useful for the general case.

instrumentation/net/http/otelhttp/config.go
tonistiigi: Added changelog item to the "Added" section. Could have been in "Fixed" as well as if the global tracer and the parent span tracer would not have matched previously I think it would have lead to some bad behavior. Lmk if you want to change that.

instrumentation/net/http/otelhttp/handler.go
Aneurysm9: I'd prefer to avoid adding API surface to the `otel` package if at all possible, particularly while the question of even getting a `TracerProvider` from a `Span` is still outstanding in the spec.  It doesn't seem like there are any particular subtleties to this invocation that would lead me to believe that providing an abstraction is highly valuable.

instrumentation/net/http/otelhttp/config.go
pellared: For sure it should not be in "fixed":

> particularly while the question of even getting a `TracerProvider` from a `Span` is still outstanding in the spec.

Therefore if we accept such behavior as acceptable, then, in my opinion, it should be explicitly described in the godoc.

Aneurysm9 (approved):

thaJeztah:
https://github.com/open-telemetry/opentelemetry-go-contrib/pull/879 was merged, so this needs another rebase @tonistiigi

tonistiigi:
The base of this PR was completely refactored while it was stuck in review 😞 . The refactoring was merged in a matter of hours.

thaJeztah:
Looks like this needs a re-approval to get CI running; @pellared @Aneurysm9 are you able to help getting this one moving?

pellared (approved):

thaJeztah:
@tonistiigi needs a rebase (again) 😞

crazy-max:
Can someone approve CI run please? Thanks!

thaJeztah:
Looks to be green ✅

PR #874, opened 2021-07-10
otelhttptrace: get tracer from request if available

tonistiigi:
Similar change as #873 for `otelhttptrace`. Picks up the correct `TracingProvider` based on parent span in the config.

@Aneurysm9 

Signed-off-by: Tonis Tiigi <tonistiigi@gmail.com>

PR #875, opened 2021-07-10
otelhttp: add WithClientTrace to connect with otelhttptrace 

tonistiigi:
I couldn't quite figure out how `otelhttp` and `otelhttptrace` are supposed to be used together. From they being in separate packages I assume maintainers don't want them to depend on each other.

This adds a new option so caller can initialize `Transport` with `otelhttp.NewTransport(base, otelhttp.WithClientTrace(otelhttptrace.NewClientTrace))` (or any other `httptrace` implementation) so that `otelhttp` spans will have `otelhttptrace` spans as children.

Depends on #873


thaJeztah:
@tonistiigi https://github.com/open-telemetry/opentelemetry-go-contrib/pull/873 was merged; guess this needs to be rebased and moved out of draft

thaJeztah:
https://github.com/open-telemetry/opentelemetry-go-contrib/pull/874 was merged as well; this ready to be moved out of draft?

thaJeztah:
@pellared  @MadVikingGod @Aneurysm9 ptal; this is the remaining one for us to get rid of our fork ❤️

instrumentation/net/http/otelhttp/transport.go
```diff
@@ -102,6 +105,10 @@ func (t *Transport) RoundTrip(r *http.Request) (*http.Response, error) {
 
 	ctx, span := tracer.Start(r.Context(), t.spanNameFormatter("", r), opts...)
 
+	if t.clientTrace != nil {
+		ctx = httptrace.WithClientTrace(ctx, t.clientTrace(ctx))
+	}
```
errordeveloper, line 110: Just out of curiosity, would having a default client trace be helpful? would that be considered an API breakage at this point?

Aneurysm9 (approved):

instrumentation/net/http/otelhttp/transport.go
Aneurysm9: I don't think it would be API breaking, but perhaps a surprising change in behavior.  I'm not sure a default client trace is a good idea, either, as it adds overhead and verbosity that is not necessary in many cases.

errordeveloper: Makes sense, thanks. Very helpful, as always!

thaJeztah:
Looks like it may need another rebase

Aneurysm9:
This still requires another approval.  @open-telemetry/go-approvers can anyone else weigh in on this?

evantorrie:
LGTM 👍

thaJeztah:
Thank you!

PR #879, opened 2021-07-11
Add WithoutSubSpans option to NewClientTrace

### LINKED ISSUE #876: otelhttptrace: allow tracing by adding events
In OpenTracing the equivalent httptrace implementation worked by adding logs(events in opentelemetry) to the parent span https://github.com/opentracing-contrib/go-stdlib/blob/00fa856341a83c0ee331ef758856ad41fdabbea7/nethttp/client.go#L286-L292 while `otelhttptrace` creates new spans.

In an app that creates a lot of traces, this creates a lot of noise and in my case, I would prefer the timestamped logs. Assuming some users prefer spans, can we make this configurable. For example, variadic `Option` field to `NewClientTrace`.

coryb:
This adds an optional behavior to NewClientTrace so that it only
adds events and annotations to the span found in SpanFromContext.
The default behavior remains unchanged, several sub-spans will be
created for each request.  This optional behavior is useful when
tracing "complex" processes that have many http calls.  In this case
the added sub-spans become "noise" and may distract from the overall
trace.

This also adds several useful attributes from data available
in the httptrace callbacks:
* http.conn.reused - bool if connection was reused
* http.conn.wasidle - bool if connection was idle
* http.conn.idletime - if wasidle, duration of idletime
* http.conn.start.network - start connection network type [tcp/udp]
* http.conn.done.network - end connection network type [tcp/udp]
* http.conn.done.addr - connection address when done
* http.dns.addrs - list of addrs returned from dns lookup

Fixes #876


instrumentation/net/http/httptrace/otelhttptrace/clienttrace.go
```diff
@@ -199,14 +288,28 @@ func (ct *clientTracer) tlsHandshakeDone(_ tls.ConnectionState, err error) {
...
 	}
-	ct.root.SetAttributes(attribute.String("http."+strings.ToLower(k), sliceToString(v)))
+	if !ct.addHeaders {
+		return
+	}
+	k = strings.ToLower(k)
+	value := sliceToString(v)
+	if _, ok := ct.redactedHeaders[k]; ok {
+		// redact all non-space runes to 'x'
+		value = strings.Map(func(r rune) rune {
```
tonistiigi, line 301: I'd prefer if this wouldn't leak the length/spaces or waste space with long tokens. Just constant `***` sgtm

pellared: I have even a stronger point of view. I think in such a case, the attribute should NOT be included at all. This is to ensure we do not leak any security-related internals. 

Reasoning: https://github.com/open-telemetry/opentelemetry-specification/pull/1502#discussion_r587860542

 This would be also more in line with other specification recommendations such us:
- https://github.com/open-telemetry/opentelemetry-specification/blame/main/specification/trace/semantic_conventions/http.md#L71
- https://github.com/open-telemetry/opentelemetry-specification/blame/main/specification/trace/semantic_conventions/database.md#L160

coryb: For now I have updated to use fixed `****` value.  We can instead drop those annotations completely, I dont really have a preference.  If we drop them, I guess I would change the `WithRedactedHeaders` to `WithIgnoredHeaders` to alllow appending to the fixed list.  We can also swap the default to not collect any headers unless a `WithHeaders` option is specified.  For the edge case of trying to debug authentication-related issues, we could add `WithRedactgnoredHeaders` (or some such name) to keep headers in the spans but always hide the values.

tonistiigi: I prefer the redacted value instead of ignoring one. Seeing which requests were authenticated and which were not is an important common feature in tracing HTTP traffic and can point to issues.

instrumentation/net/http/httptrace/otelhttptrace/clienttrace.go
```diff
@@ -115,6 +179,15 @@ func (ct *clientTracer) start(hook, spanName string, attrs ...attribute.KeyValue
 }
 
 func (ct *clientTracer) end(hook string, err error, attrs ...attribute.KeyValue) {
+	if !ct.useSpans {
+		ct.root.AddEvent(hook + ".done")
+		if err != nil {
+			attrs = append(attrs, attribute.String(hook+".error", err.Error()))
+		}
+		ct.root.SetAttributes(attrs...)
```
Aneurysm9, line 187: Should these attributes be added to the event rather than the span?

coryb: whoops, fixed that in `start`, but forgot to also fix in `end`, I have updated the PR.  Thanks

pellared:
> The default behavior remains unchanged, 

I think this is not true b/c of the default `redactedHeaders`.

Also is it possible to remove the default `redactedHeaders` if someone really wants to trace them (e.g. for some local testing)? I think that security should be "enabled" by default but there should always be an option to explicitly configure to some insecure settings.

coryb:
> > The default behavior remains unchanged,
> 
> I think this is not true b/c of the default `redactedHeaders`.

True, there is a behaviour change due to the redacted headers.  I didn't know about the leaking headers when I opened the issue and made that comment.

> Also is it possible to remove the default `redactedHeaders` if someone really wants to trace them (e.g. for some local testing)? I think that security should be "enabled" by default but there should always be an option to explicitly configure to some insecure settings.

We could add a `WithInsecureHeaders` to disable the redaction and reset the current list.  Motivated users could always just add an http.RoundTrip wrapper to their client that adds more attributes to the spans.  Either option seems fine to me, curious if others have opinions...

coryb:
> curious if others have opinions...

No response, so I went ahead and added `WithInsecureHeaders` to facilitate the "debugging auth issues" use-case.  Also extended the tests to verify WithoutHeaders, WithRedactedHeaders, and the new WithInsecureHeaders so there should not be a drop in test coverage.

CHANGELOG.md
```diff
@@ -14,6 +14,10 @@ This project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0.htm
   The `messaging.message_id` and `messaging.kafka.partition` attributes are now not set if a message was not processed. (#754) (#755)
 - Fix `otelsarama.WrapAsyncProducer` so that the messages from the `Errors` channel contain the original `Metadata`. (#754)
 
+### Changed
+
+- Added `WithoutSubSpans`, `WithRedactedHeaders`, and `WithoutHeaders` options to `NewClientTrace` in the `instrumentation/net/http/httptrace/otelhttptrace` package (#879)
```
pellared, line 19: The changelog does not mention `WithInsecureHeaders` and changes in the default behavior.

coryb: Added an `### Added` section for the new options and modified the `### Changed` section to mention the redacted headers by default.  Also added the `.`

coryb:
Updated lint issue that broke the tests, I forgot to run the precommit check but it is working now.

Aneurysm9 (approved):

instrumentation/net/http/httptrace/otelhttptrace/clienttrace.go
```diff
@@ -53,20 +60,81 @@ func parentHook(hook string) string {
 	return hookMap[hook]
 }
 
+// ClientTraceOption allows customizations to how the httptrace.Client
+// collects information.
+type ClientTraceOption func(*clientTracer)
```
pellared, line 65: It does not follow https://github.com/open-telemetry/opentelemetry-go/blob/main/CONTRIBUTING.md#option

The option type has to be an interface. But you can make a functional option implementation: https://github.com/open-telemetry/opentelemetry-go/blob/main/CONTRIBUTING.md#functional-options

Most of the other Go Modules in this repository should already follow this pattern.

instrumentation/net/http/httptrace/otelhttptrace/clienttrace.go
```diff
@@ -53,20 +60,81 @@ func parentHook(hook string) string {
...
+	activeHooks     map[string]context.Context
+	root            trace.Span
+	mtx             sync.Mutex
+	redactedHeaders map[string]struct{}
+	addHeaders      bool
+	useSpans        bool
 }
 
-func NewClientTrace(ctx context.Context) *httptrace.ClientTrace {
+func NewClientTrace(ctx context.Context, opts ...ClientTraceOption) *httptrace.ClientTrace {
```
pellared, line 121: Can you document this function, please? The defaults should be described.

instrumentation/net/http/httptrace/otelhttptrace/clienttrace.go
coryb: updated

instrumentation/net/http/httptrace/otelhttptrace/clienttrace.go
coryb: updated

instrumentation/net/http/httptrace/otelhttptrace/clienttrace_test.go
```diff
@@ -243,3 +254,192 @@ func TestEndBeforeStartCreatesSpan(t *testing.T) {
 	spans := getSpansFromRecorder(sr, name)
 	require.Len(t, spans, 1)
 }
+
+func TestClientTraceOptions(t *testing.T) {
```
pellared, line 258: please split it into several unit tests.

you can also create a helper like this to make the tests more self-contained:

```go
type clientTraceTestFixutre struct {
  Address string
  SpanRecorder *oteltest.SpanRecorder
}

func prepareClientTraceTest(t *testing.T) clientTraceTestFixutre {
  fixture := clientTraceTestFixutre {}
  fixture.SpanRecorder =&oteltest.SpanRecorder{}
  otel.SetTracerProvider(
		oteltest.NewTracerProvider(oteltest.WithSpanRecorder(fixture.SpanRecorder)),
	)
  
  ts := httptest.NewServer(
		http.HandlerFunc(func(w http.ResponseWriter, r *http.Request) {
		}),
	)
  t.Cleanup(ts.Close))
  fixture.Address = ts.Listener.Addr().String()
  return fixture
}
```

coryb: updated

pellared (approved):

CHANGELOG.md
```diff
@@ -29,6 +29,14 @@ This project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0.htm
...
 - Fix `otelsarama.WrapAsyncProducer` so that the messages from the `Errors` channel contain the original `Metadata`. (#754)
 
+### Added
+
+- Add `WithoutSubSpans`, `WithRedactedHeaders`, `WithoutHeaders`, and `WithInsecureHeaders` options for `otelhttptrace.NewClientTrace`. (#879)
+
+### Changed
+
+- `otelhttptrace.NewClientTrace` now redacts known sensitive headers by default. (#879)
+
```
Aneurysm9, line 39: Please move this section up under `Unreleased` since this change was not included in `v0.22.0`.

coryb: Updated, that was caused by the git automatic merge since I had not rebased since the latest release.

thaJeztah:
@coryb needs a rebase

coryb:
Thanks for the ping, rebased, again.  Hopefully we can get this merged in, please.

coryb:
Could someone please poke the workflow approval button so the test will run?

coryb:
Can we please get this merged in?  We have two approvals and tests are passing ... what more is left?

coryb:
Hi folks, I have rebased again to address merge conflicts on the CHANGELOG.    It has now been a month since the last change other than the rebases, so this is pretty frustrating.

I need someone to poke the workflow approval button again.

thaJeztah:
@pellared @Aneurysm9 are you able to help out to this PR moving? 🤗

thaJeztah:
Thanks!

MrAlias:
@dependabot rebase

Issue #843, opened 2021-07-02
Add Instrumentation for `gofiber/fiber`


WLun001:
## Background

**Package Link**: https://github.com/gofiber/fiber

<describe how this package is commonly used>

## Proposed Solution

middlewares

### Tracing

- attributes:
  - semconv.NetAttributesFromHTTPRequest
  - semconv.EndUserAttributesFromHTTPRequest
  - semconv.HTTPServerAttributesFromHTTPRequest
  - semconv.HTTPAttributesFromHTTPStatusCode
  - semconv.SpanStatusFromHTTPStatusCode
- events:
  - Incoming HTTP request

### Prior Art

- https://github.com/open-telemetry/opentelemetry-go-contrib/tree/main/instrumentation/github.com/gin-gonic/gin/otelgin

## Tasks

- Code complete:
  - [x] Comprehensive unit tests.
  - [x] End-to-end integration tests.
  - [x] Tests all passing.
  - [x] Instrumentation functionality verified.
- Documented
  - [x] Added to the [OpenTelemetry Registry](https://opentelemetry.io/registry/)
  - [x] README included for the module describing high-level purpose.
  - [x] Complete documentation of all public API including package documentation.
  - [x] [Instrumentation documentation](https://github.com/open-telemetry/opentelemetry-go-contrib/blob/main/instrumentation/README.md#instrumentation-packages) updated.
- Examples
  - [x] `Dockerfile` file to build example application.
  - [x] `docker-compose.yml` to run example in a docker environment to demonstrate instrumentation.

WLun001:
working on it

MrAlias:
In accordance with the [new instrumentation policy hosting guidelines](https://github.com/open-telemetry/opentelemetry-go-contrib/tree/main/instrumentation#new-instrumentation) I'm closing this.

PR #880, opened 2021-07-12
add Fiber

WLun001:
close https://github.com/open-telemetry/opentelemetry-go-contrib/issues/843


instrumentation/github.com/gofiber/fiber/otelfiber/fiber_test.go
```diff
@@ -0,0 +1,211 @@
...
+
+	app := fiber.New()
+	app.Use(Middleware("foobar", WithTracerProvider(provider)))
+	app.Get("/user/:id", func(ctx *fiber.Ctx) error {
+		span := oteltrace.SpanFromContext(ctx.UserContext())
+		mspan, ok := span.(*oteltest.Span)
+		require.True(t, ok)
+		assert.Equal(t, pspan.SpanContext().TraceID(), mspan.SpanContext().TraceID())
+		assert.Equal(t, pspan.SpanContext().SpanID(), mspan.ParentSpanID())
+		return ctx.SendStatus(http.StatusNoContent)
```
pellared, line 180: there is nothing that ensures that this code is executed

instrumentation/github.com/gofiber/fiber/README.md
```diff
@@ -0,0 +1,41 @@
+### OpenTelemetry for [Fiber](https://gofiber.io/)
```
pellared, line 1: This markdown is not properly formatted. E.g. this line should have `#` instead of `###`

I would also add a hyperlink to the `example` directory instead of putting the code here as it may easily get out-of-sync.

I think it is very good that you want to provide `README.md` 👍 

However, because no other instrumentation has a `README.md` and also I want to propose some documentation guidelines as part of https://github.com/open-telemetry/opentelemetry-go/issues/2046, I suggest to simply remove this file 😉

instrumentation/github.com/gofiber/fiber/otelfiber/config.go
```diff
@@ -0,0 +1,46 @@
...
+)
+
+// config is used to configure the mux middleware.
+type config struct {
+	TracerProvider oteltrace.TracerProvider
+	Propagators    propagation.TextMapPropagator
+}
+
+// Option specifies instrumentation configuration options.
+type Option func(*config)
```
pellared, line 29: This is not adhering to https://github.com/open-telemetry/opentelemetry-go/blob/main/CONTRIBUTING.md#option

See https://github.com/open-telemetry/opentelemetry-go-contrib/pull/857 how to easily address it

pellared:
Could you please also add a line to `CHANGELOG.md`?

instrumentation/github.com/gofiber/fiber/otelfiber/example/docker-compose.yml
```diff
@@ -0,0 +1,39 @@
...
+version: "3.7"
+services:
+  fiber-client:
+    image: golang:alpine
+    networks:
+      - example
+    command:
+      - "/bin/sh"
+      - "-c"
+      - "wget http://echo-server:3000/users/123 && cat 123"
```
Aneurysm9, line 23:
```suggestion
      - "wget http://fiber-server:3000/users/123 && cat 123"
```

instrumentation/github.com/gofiber/fiber/otelfiber/config.go
```diff
@@ -0,0 +1,46 @@
...
+// limitations under the License.
+
+package otelfiber
+
+import (
+	"go.opentelemetry.io/otel/propagation"
+	oteltrace "go.opentelemetry.io/otel/trace"
+)
+
+// config is used to configure the mux middleware.
```
Aneurysm9, line 22:
```suggestion
// config is used to configure the Fiber middleware.
```

instrumentation/github.com/gofiber/fiber/otelfiber/fiber.go
```diff
@@ -0,0 +1,124 @@
...
+			oteltrace.WithAttributes(semconv.HTTPClientIPKey.String(c.IP())),
+			oteltrace.WithAttributes(semconv.NetTransportTCP),
+			oteltrace.WithSpanKind(oteltrace.SpanKindServer),
+		}
+		// temporary set to c.Path() first
+		// update with c.Route().Path after c.Next() is called
+		// to get pathRaw
+		spanName := c.Path()
+		if spanName == "" {
+			spanName = fmt.Sprintf("HTTP %s route not found", c.Path())
```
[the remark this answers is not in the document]
WLun001: `c.Path()` is not empty at this point, the value will be `users/123`, and later on update the spanName to `users/:id`

instrumentation/github.com/gofiber/fiber/otelfiber/fiber_test.go
WLun001: what do you mean? It does executed

pellared: now it does get executed but this code does not ensure it is really executed.

what I mean is that I suggest changing the test code to something more-or-less like this:

```go
	var gotSpan *oteltest.Span
	app.Get("/user/:id", func(ctx *fiber.Ctx) error {
		gotSpan = oteltrace.SpanFromContext(ctx.UserContext()).(*oteltest.Span)
		mspan, ok := span
		require.True(t, ok)
		
		return ctx.SendStatus(http.StatusNoContent)
	})

	_, _ = app.Test(r)

	assert.Equal(t, pspan.SpanContext().TraceID(), gotSpan.SpanContext().TraceID())
	assert.Equal(t, pspan.SpanContext().SpanID(), gotSpan.ParentSpanID())
```

this will panic if `gotSpan` is not assigned instead of a silent pass
you can of course make type assertions and nil-checks to avoid panics in tests

the same comment applies to other tests

instrumentation/github.com/gofiber/fiber/otelfiber/fiber_test.go
```diff
@@ -0,0 +1,211 @@
...
+	})
+
+	_, _ = app.Test(r)
+
+	otel.SetTextMapPropagator(propagation.NewCompositeTextMapPropagator())
```
pellared, line 185: it is would be better to add it as:
```
defer otel.SetTextMapPropagator(propagation.NewCompositeTextMapPropagator())
```
just after:
```
otel.SetTextMapPropagator(propagation.TraceContext{})
```

b/c if the test stops sooner (e.g. b/c of `require` or `panic`) then the "noop" propagator would not be restored

pellared (commented):
Was instrumentation/github.com/gofiber/fiber/README.md not supposed to be removed?

Reference https://github.com/open-telemetry/opentelemetry-go-contrib/pull/880/files#r667995107

WLun001:
@pellared sorry I mistakenly removed the wrong README

instrumentation/github.com/gofiber/fiber/otelfiber/example/server.go
```diff
@@ -0,0 +1,79 @@
...
+	"go.opentelemetry.io/contrib/instrumentation/github.com/gofiber/fiber/otelfiber"
+	"go.opentelemetry.io/otel"
+	"go.opentelemetry.io/otel/attribute"
+	stdout "go.opentelemetry.io/otel/exporters/stdout/stdouttrace"
+	"go.opentelemetry.io/otel/propagation"
+	sdktrace "go.opentelemetry.io/otel/sdk/trace"
+	oteltrace "go.opentelemetry.io/otel/trace"
+)
+
+var tracer = otel.Tracer("fiber-server")
```
pellared, line 32: @open-telemetry/go-approvers Would it be better to recommend:
- adding `sdktrace.WithResource(resource.NewWithAttributes(semconv.SchemaURL, semconv.ServiceNameKey.String("fiber-server"))),` to `sdktrace.NewTracerProvider` call
- creating the new span inside `getUser` using `oteltrace.SpanFromContext()`
?

Reference: https://github.com/open-telemetry/opentelemetry-go-contrib/blob/main/instrumentation/net/http/otelhttp/example/server/server.go

instrumentation/github.com/gofiber/fiber/otelfiber/fiber.go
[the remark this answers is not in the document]
pellared: Like... Why it is not empty?

```
panName := c.Path()
if spanName == "" {
```

if you say that it is never empty then this code does nothing (which is probably not what you intended to do)

CHANGELOG.md
```diff
@@ -7,6 +7,9 @@ The format is based on [Keep a Changelog](https://keepachangelog.com/en/1.0.0/).
 This project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0.html).
 
 ## [Unreleased]
+### Added
+
+- `gofiber/fiber` trace instrumentation (#880)
```
pellared, line 12: nit (just to be a little more consistent in the changelog)
```suggestion
- Trace instrumentation for the `github.com/gofiber/fiber` package. (#880)
```

instrumentation/github.com/gofiber/fiber/otelfiber/fiber.go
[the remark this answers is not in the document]
pellared: I would remove the `if spanName == "" ` block

instrumentation/github.com/gofiber/fiber/otelfiber/fiber.go
```diff
@@ -0,0 +1,123 @@
...
+	oteltrace "go.opentelemetry.io/otel/trace"
+)
+
+const (
+	tracerKey  = "otel-go-contrib-tracer-gofiber-fiber"
+	tracerName = "go.opentelemetry.io/contrib/instrumentation/github.com/gofiber/fiber/otelfiber"
+)
+
+// Middleware returns fiber handler which will trace incoming requests.
+func Middleware(service string, opts ...Option) fiber.Handler {
```
pellared, line 38: @open-telemetry/go-approvers 
side comment: do we have an example of documentation that tells what attributes are being set for given instrumentation?

CHANGELOG.md
```diff
@@ -7,6 +7,9 @@ The format is based on [Keep a Changelog](https://keepachangelog.com/en/1.0.0/).
 This project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0.html).
 
 ## [Unreleased]
+### Added
```
pellared, line 10: nit:

```suggestion
## [Unreleased]

### Added
```

otherwise `markdownlint` complains 😉

instrumentation/github.com/gofiber/fiber/otelfiber/fiber.go
```diff
@@ -0,0 +1,123 @@
...
+		// serve the request to the next middleware
+		err := c.Next()
+
+		span.SetName(c.Route().Path)
+		span.SetAttributes(semconv.HTTPRouteKey.String(c.Route().Path))
+
+		if err != nil {
+			span.SetAttributes(attribute.String("fiber.error", err.Error()))
+			// invokes the registered HTTP error handler
+			_ = c.App().Config().ErrorHandler(c, err)
```
pellared, line 113: should it be the responsibility of the instrumentation middleware to do it?
if so then maybe it would be good to explain why in comments as it is not obvious (at least for me)

instrumentation/github.com/gofiber/fiber/otelfiber/fiber.go
```diff
@@ -0,0 +1,123 @@
...
+		reqHeader := make(http.Header)
+		c.Request().Header.VisitAll(func(k, v []byte) {
+			sk := string(k)
+			sv := string(v)
+
+			switch sk {
+			case fiber.HeaderTransferEncoding:
+				reqHeader[fiber.HeaderTransferEncoding] = append(reqHeader[fiber.HeaderTransferEncoding], sv)
+			default:
+				reqHeader[sk] = []string{sv}
```
pellared, line 72: I am not sure that this handling is correct. Here is an example why: https://www.w3.org/TR/trace-context/#tracestate-header-field-values

> Tracestate MAY be sent or received as multiple header fields. Multiple tracestate header fields MUST be handled as specified by RFC7230 Section 3.2.2 Field Order. The tracestate header SHOULD be sent as a single field when possible, but MAY be split into multiple header fields.

instrumentation/github.com/gofiber/fiber/otelfiber/fiber.go
WLun001: in order to get the correct response status code to add to span attributes, `ErrorHandler` has to be called.

instrumentation/github.com/gofiber/fiber/otelfiber/fiber.go
WLun001: hmmm i borrow the code from fasthttp

https://github.com/valyala/fasthttp/blob/b3ece39dcdd40ab814988d7a322b2f1a050e13f4/fasthttpadaptor/request.go#L46-L56

instrumentation/github.com/gofiber/fiber/otelfiber/fiber.go
pellared: can you change the comment to:
```
// invoke the registered HTTP error handler
// to get the correct response status code
```

CHANGELOG.md
WLun001: nice catch!

instrumentation/github.com/gofiber/fiber/otelfiber/fiber.go
WLun001: something like this instead? 

```go
c.Request().Header.VisitAll(func(k, v []byte) {
	sk := string(k)
	sv := string(v)
	reqHeader.Add(sk, sv)
})
```

pellared (approved):

pellared:
@WLun001 You need to add the new modules to `/.github/dependabot.yml`

instrumentation/github.com/gofiber/fiber/otelfiber/example/go.mod
```diff
@@ -0,0 +1,20 @@
+module go.opentelemetry.io/opentelemetry-go-contrib/instrumentation/github.com/gofiber/fiber/otelfiber/example
+
+go 1.16
```
pellared, line 3:
```suggestion
go 1.15
```

instrumentation/github.com/gofiber/fiber/otelfiber/go.mod
```diff
@@ -0,0 +1,18 @@
+module go.opentelemetry.io/contrib/instrumentation/github.com/gofiber/fiber/otelfiber
+
+go 1.16
```
pellared, line 3:
```suggestion
go 1.15
```

pellared:
@WLun001 Looks like you need to `go mod tidy` and `go fmt`.

WLun001:
just ran `go mod tidy`, `go fmt` and `make precommit`  locally, should be ok now

instrumentation/github.com/gofiber/fiber/otelfiber/example/server.go
Aneurysm9: A service name should definitely be provided in the resource given to the tracer provider initialization.

instrumentation/github.com/gofiber/fiber/otelfiber/fiber.go
```diff
@@ -0,0 +1,116 @@
...
+		c.SetUserContext(ctx)
+
+		// serve the request to the next middleware
+		err := c.Next()
+
+		span.SetName(c.Route().Path)
+		span.SetAttributes(semconv.HTTPRouteKey.String(c.Route().Path))
+
+		if err != nil {
+			span.SetAttributes(attribute.String("fiber.error", err.Error()))
```
Aneurysm9, line 103: Should this use `span.RecordError()` instead?

instrumentation/github.com/gofiber/fiber/otelfiber/fiber.go
```diff
@@ -0,0 +1,116 @@
...
+			// to get the correct response status code
+			_ = c.App().Config().ErrorHandler(c, err)
+		}
+
+		attrs := semconv.HTTPAttributesFromHTTPStatusCode(c.Response().StatusCode())
+		spanStatus, spanMessage := semconv.SpanStatusFromHTTPStatusCode(c.Response().StatusCode())
+		span.SetAttributes(attrs...)
+		span.SetStatus(spanStatus, spanMessage)
+
+		return nil
```
Aneurysm9, line 114: does this need to propagate the error value?

```suggestion
		return err
```

instrumentation/github.com/gofiber/fiber/otelfiber/fiber.go
WLun001: should we? Looking at other example, it doesn't. Not really sure

https://github.com/open-telemetry/opentelemetry-go-contrib/blob/0548aff5c36f20d9da23a633a4468b853f8a0dab/instrumentation/github.com/labstack/echo/otelecho/echo.go#L92

instrumentation/github.com/gofiber/fiber/otelfiber/fiber.go
WLun001: not need, errorHandler already called

WLun001:
any updates on this PR? good to merge this?

instrumentation/github.com/gofiber/fiber/otelfiber/fiber.go
pellared: I agree with @Aneurysm9

efectn:
Do you have any plan to add offical Fiber support?

pellared:
I do not think it will be merged. See: https://github.com/open-telemetry/opentelemetry-go-contrib/issues/1100

However, I hope that all the code reviews were helpful.

PR #882, opened 2021-07-13
Add support for end-to-end tracing on AWS Lambda

garrettwegan:
See for example use case: https://github.com/garrettwegan/opentelemetry-lambda/blob/main/go/sample-apps/function/function.go


instrumentation/github.com/aws/aws-lambda-go/otellambda/lambda.go
```diff
@@ -0,0 +1,351 @@
...
+
+// Logic to wrap up OTel Tracing
+func tracingEnd(ctx context.Context, span trace.Span) {
+	if tp != nil {
+		// span will be valid if tp is not nil
+		span.End()
+
+		// yield processor to attempt to attempt to ensure
+		// all spans have been consumed and are ready to be
+		// flushed - see https://github.com/open-telemetry/opentelemetry-go/issues/2068
```
wangzlei, line 341: 2080

instrumentation/github.com/aws/aws-lambda-go/otellambda/lambda.go
```diff
@@ -0,0 +1,351 @@
...
+	// logic with no tracing
+	if tp != nil {
+		// Get a named tracer with package path as its name.
+		tracer := tp.Tracer(tracerName)
+
+		var span trace.Span
+		spanName := os.Getenv("AWS_LAMBDA_FUNCTION_NAME")
+
+		var attributes []attribute.KeyValue
+		lc, ok := lambdacontext.FromContext(ctx)
```
wangzlei, line 308: don't find this function

instrumentation/github.com/aws/aws-lambda-go/otellambda/lambda.go
```diff
@@ -0,0 +1,351 @@
...
+func tracingEnd(ctx context.Context, span trace.Span) {
+	if tp != nil {
+		// span will be valid if tp is not nil
+		span.End()
+
+		// yield processor to attempt to attempt to ensure
+		// all spans have been consumed and are ready to be
+		// flushed - see https://github.com/open-telemetry/opentelemetry-go/issues/2068
+		// to be removed upon resolution of above issue
+		runtime.Gosched()
```
rakyll, line 343: This would only yield to another goroutine with no guarantees that it will be the flusher. I think we need to think a bit more what we should do here. Ideally, a graceful shutdown event would be ideal place to force flush but this doesn't exist at the moment.

Can we write a custom batcher for Lambda that flushes in the same goroutine and document the limitation?

rakyll: I talked to the Lambda team and they said the extensions API is supported in the runtime/functions: https://docs.aws.amazon.com/lambda/latest/dg/runtimes-extensions-api.html#runtimes-extensions-api-lifecycle

We should be able to listen to the "shutdown" event and force flush there instead of force flushing after each handler. The shutdown happens after the Lambda VM is recycled and this may add up to 500ms of latency but I think we are fine with this.

The Lambda collector is implementing something to listen to these events: https://github.com/open-telemetry/opentelemetry-lambda/blob/main/collector/extension/client.go You can use that as a reference.

rakyll (changes requested):

instrumentation/github.com/aws/aws-lambda-go/otellambda/lambda.go
wangzlei: @rakyll Refer [design](https://github.com/open-telemetry/opentelemetry-lambda/blob/main/docs/design_proposal.md#21-lambda-freeze)
1. adding force flush at the end of each Lambda invocation but not in `shutdown` is for no data delay.
2. Collector monitor Lambda event does not make sense. Because Collector flush data does not mean SDK has flushed data to Collector.

wangzlei: > This would only yield to another goroutine with no guarantees that it will be the flusher. I think we need to think a bit more what we should do here. Ideally, a graceful shutdown event would be ideal place to force flush but this doesn't exist at the moment.
> 
> Can we write a custom batcher for Lambda that flushes in the same goroutine and document the limitation?

The issue should be addressed by https://github.com/open-telemetry/opentelemetry-go/issues/2080. We list some workarounds before fixing ForceFlush():
1. As your suggestion, create a specific batch processor for Lambda only // difficulty? more maintenance; useless after fixing ForceFlush;
2. Using SimpleProcessor instead of BatchProcessor  // bad performance, ie: create 100 spans in one Lambda function.
3. Add a yield, give up CPU to other gorountine, and hopefully Enqueue can clean the data on the way.

If in actual testing the `yield` can solve problem, it maybe the best option.

instrumentation/github.com/aws/aws-lambda-go/otellambda/lambda.go
```diff
@@ -0,0 +1,351 @@
...
+	}
+
+	if err := json.Unmarshal(remarshalledPayload, event.Interface()); err != nil {
+		return reflect.Value{}, err
+	}
+	return event, nil
+}
+
+// LambdaHandlerWrapper Provides a lambda handler which wraps customer lambda handler with OTel Tracing
+func LambdaHandlerWrapper(handlerFunc interface{}) interface{} {
```
Aneurysm9, line 152:
```suggestion
func WrapHandler(handlerFunc interface{}) interface{} {
```

`Lambda` can be dropped since it will be in the package name when referenced externally, and using a verb phrase rather than a noun for naming a function makes more sense to me.  I'd expect a `HandlerWrapper` to be a struct or interface rather than a function.

instrumentation/github.com/aws/aws-lambda-go/otellambda/lambda.go
```diff
@@ -0,0 +1,351 @@
...
+	}
+}
+
+// Ensure handler takes 0-2 values, with context
+// as its first value if two arguments exist
+func validateArguments(handler reflect.Type) (bool, error) {
+	handlerTakesContext := false
+	if handler.NumIn() > 2 {
+		return false, fmt.Errorf("handlers may not take more than two arguments, but handler takes %d", handler.NumIn())
+	} else if handler.NumIn() > 0 {
```
Aneurysm9, line 76: Perhaps use a `switch handler.NumIn()` here, and add a `case 1` that avoids doing reflection on the only argument when if a context can only be provided with two arguments.

instrumentation/github.com/aws/aws-lambda-go/otellambda/lambda.go
```diff
@@ -0,0 +1,351 @@
...
+
+// Ensure handler returns 0-2 values, with an error
+// as its first value if any exist
+func validateReturns(handler reflect.Type) error {
+	errorType := reflect.TypeOf((*error)(nil)).Elem()
+
+	switch n := handler.NumOut(); {
+	case n > 2:
+		return fmt.Errorf("handler may not return more than two values")
+	case n > 1:
```
Aneurysm9, line 96:
```suggestion
	case n == 2:
```

instrumentation/github.com/aws/aws-lambda-go/otellambda/lambda.go
```diff
@@ -0,0 +1,351 @@
...
+		if !handler.Out(0).Implements(errorType) {
+			return fmt.Errorf("handler returns a single value, but it does not implement error")
+		}
+	}
+
+	return nil
+}
+
+// Wraps and calls customer lambda handler then unpacks response as necessary
+func wrapperInternals(handlerFunc interface{}, event reflect.Value, ctx context.Context, takesContext bool) (interface{}, error) {
```
Aneurysm9, line 110:
```suggestion
func wrapperInternals(ctx context.Context, handlerFunc interface{}, event reflect.Value, takesContext bool) (interface{}, error) {
```

The context should always be the first argument.

instrumentation/github.com/aws/aws-lambda-go/otellambda/lambda.go
```diff
@@ -0,0 +1,351 @@
...
+	}
+	return true
+}
+
+type wrappedHandler struct {
+	handler lambda.Handler
+}
+
+// Compile time check our Handler implements lambda.Handler
+var _ lambda.Handler = wrappedHandler{}
```
Aneurysm9, line 261: There seem to be two different interfaces being implemented here.  They should probably be split into separate files so that it is clear what code goes with which implementation.  The difference between the interfaces should be well documented in the `README`.

instrumentation/github.com/aws/aws-lambda-go/otellambda/lambda.go
```diff
@@ -0,0 +1,351 @@
...
+	if tp == nil {
+		initTracerProvider()
+	}
+
+	// if tracer provider successfully initializes then
+	// we add tracing, otherwise do customer business
+	// logic with no tracing
+	if tp != nil {
+		// Get a named tracer with package path as its name.
+		tracer := tp.Tracer(tracerName)
```
Aneurysm9, line 302: It would be preferable to have this function be a method hanging off a type like `wrappedHandler` that could retain a reference to the `Tracer`.  The value of `tracerName` doesn't change and there is no need to obtain a new `Tracer` with every invocation.

instrumentation/github.com/aws/aws-lambda-go/otellambda/lambda.go
```diff
@@ -0,0 +1,351 @@
...
+		spanName := os.Getenv("AWS_LAMBDA_FUNCTION_NAME")
+
+		var attributes []attribute.KeyValue
+		lc, ok := lambdacontext.FromContext(ctx)
+		if !ok {
+			errorLogger.Println("failed to load lambda context from context, ensure tracing enabled in Lambda")
+		}
+		if lc != nil {
+			ctxRequestID := lc.AwsRequestID
+			attributes = append(attributes, attribute.KeyValue{Key: semconv.FaaSExecutionKey, Value: attribute.StringValue(ctxRequestID)})
```
Aneurysm9, line 314:
```suggestion
			attributes = append(attributes, semconv.FaaSExecutionKey.String(ctxRequestID))
```

instrumentation/github.com/aws/aws-lambda-go/otellambda/lambda.go
```diff
@@ -0,0 +1,351 @@
...
+
+			// Resource attrs added as span attr due to static tp
+			// being created without meaningful context
+			ctxFunctionArn := lc.InvokedFunctionArn
+			attributes = append(attributes, attribute.KeyValue{Key: semconv.FaaSIDKey, Value: attribute.StringValue(ctxFunctionArn)})
```
Aneurysm9, line 319:
```suggestion
			attributes = append(attributes, semconv.FaaSIDKey.String(ctxFunctionArn))
```

instrumentation/github.com/aws/aws-lambda-go/otellambda/lambda.go
```diff
@@ -0,0 +1,351 @@
...
+			arnParts := strings.Split(ctxFunctionArn, ":")
+			if len(arnParts) >= 5 {
+				attributes = append(attributes, attribute.KeyValue{Key: semconv.CloudAccountIDKey, Value: attribute.StringValue(arnParts[4])})
```
Aneurysm9, line 322:
```suggestion
				attributes = append(attributes, semconv.CloudAccountIDKeyString(arnParts[4]))
```

instrumentation/github.com/aws/aws-lambda-go/otellambda/lambda.go
```diff
@@ -0,0 +1,351 @@
...
+		lc, ok := lambdacontext.FromContext(ctx)
+		if !ok {
+			errorLogger.Println("failed to load lambda context from context, ensure tracing enabled in Lambda")
+		}
+		if lc != nil {
+			ctxRequestID := lc.AwsRequestID
+			attributes = append(attributes, attribute.KeyValue{Key: semconv.FaaSExecutionKey, Value: attribute.StringValue(ctxRequestID)})
+
+			// Resource attrs added as span attr due to static tp
+			// being created without meaningful context
```
Aneurysm9, line 317: Is there no mechanism for retrieving these values prior to the first invocation so that they can be added to the `Resource` where they belong?  If not, can they at least be persisted somewhere to avoid reconstructing them on every request?

instrumentation/github.com/aws/aws-lambda-go/otellambda/lambda.go
```diff
@@ -0,0 +1,351 @@
...
+			if len(arnParts) >= 5 {
+				attributes = append(attributes, attribute.KeyValue{Key: semconv.CloudAccountIDKey, Value: attribute.StringValue(arnParts[4])})
+			}
+		}
+
+		ctx, span = tracer.Start(ctx, spanName, trace.WithSpanKind(trace.SpanKindServer), trace.WithAttributes(attributes...))
+
+		return ctx, span
+	}
+	return ctx, nil
```
Aneurysm9, line 330: Return a no-op span here if necessary.  There should never be a nil span value.  This would be unnecessary if the instrumentation received the `TracerProvider` as an option and used `otel.GetTracerProvider()` in the absence of an explicit option.  It would have a valid tracer provider at all times, though it may be a no-op provider if no SDK is configured.

Aneurysm9 (changes requested):
Both modules require `README.md` describing their use and the instrumentation should have an example.  See other instrumentation for samples.

instrumentation/github.com/aws/aws-lambda-go/otellambda/lambda.go
```diff
@@ -0,0 +1,351 @@
...
+	sdktrace "go.opentelemetry.io/otel/sdk/trace"
+	semconv "go.opentelemetry.io/otel/semconv/v1.4.0"
+	"go.opentelemetry.io/otel/trace"
+)
+
+const (
+	tracerName = "go.opentelemetry.io/contrib/instrumentation/github.com/aws/aws-lambda-go/otellambda"
+)
+
+var tp *sdktrace.TracerProvider
```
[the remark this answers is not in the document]
garrettwegan: TracerProvider is now set externally via `WithTracerProvider` option and defaults to the global Otel tracerprovider if none provided. 

More research/conversation needed to be determine if decoupling `Propagator` is currently possible do to issues with Lambda itself.

instrumentation/github.com/aws/aws-lambda-go/otellambda/lambda.go
garrettwegan: Context can be the only argument and when we are only given one argument we still need to know whether it is context or an event in order to correctly handle an event in the LambdaHandlerWrapper so I believe the reflection is necessary either way.

instrumentation/github.com/aws/aws-lambda-go/otellambda/lambda.go
garrettwegan: Unfortunately these resources are not available until an invocation. I have now stored them to avoid computing them every time.

instrumentation/github.com/aws/aws-lambda-go/otellambda/lambda.go
garrettwegan: Now stores a Tracer to use each invocation

instrumentation/github.com/aws/aws-lambda-go/otellambda/lambda.go
garrettwegan: Moved the two wrappers into separate files. Usage is now shown in doc.go to indicate that the two wrappers are to be used with their respective lambda entry points.

instrumentation/github.com/aws/aws-lambda-go/otellambda/lambda.go
garrettwegan: Renamed wrappers to `WrapHandlerFunction` and `WrapHandler`.

instrumentation/github.com/aws/aws-lambda-go/otellambda/example/main.go
```diff
@@ -0,0 +1,114 @@
...
+	// HTTP
+	orig := otelhttp.DefaultClient
+	otelhttp.DefaultClient = &http.Client{
+		Transport: otelhttp.NewTransport(
+			http.DefaultTransport,
+			otelhttp.WithTracerProvider(otel.GetTracerProvider()),
+		),
+	}
+	defer func() { otelhttp.DefaultClient = orig }()
+	res, err := otelhttp.Get(ctx, "https://api.github.com/repos/open-telemetry/opentelemetry-go/releases/latest")
```
Aneurysm9, line 68: Rather than replacing `otelhttp.DefaultClient` and restoring it later (which I'm not sure is doing much, since it should already have a handle to a delegate to the global tracer provider), perhaps just use `NewTransport()` to create a local `http.Client` instance.

```suggestion
	client := &http.Client{
		Transport: otelhttp.NewTransport(
			http.DefaultTransport,
			otelhttp.WithTracerProvider(otel.GetTracerProvider()),
		),
	}
	req, err := http.NewRequestWithContext(ctx, http.MethodGet, "https://api.github.com/repos/open-telemetry/opentelemetry-go/releases/latest", nil)
	if err != nil {
		//handle err
	}
	res, err := client.Do(req)
```

instrumentation/github.com/aws/aws-lambda-go/otellambda/go.mod
```diff
@@ -0,0 +1,17 @@
...
+replace (
+	go.opentelemetry.io/contrib => ../../../../..
+	go.opentelemetry.io/contrib/propagators/aws => ../../../../../propagators/aws
+)
+
+require (
+	github.com/aws/aws-lambda-go v1.24.0
+	github.com/stretchr/testify v1.7.0
+	go.opentelemetry.io/contrib v0.21.0
+	go.opentelemetry.io/contrib/propagators/aws v0.21.0
```
Aneurysm9, line 14:
```suggestion
	go.opentelemetry.io/contrib v0.22.0
	go.opentelemetry.io/contrib/propagators/aws v0.22.0
```

instrumentation/github.com/aws/aws-lambda-go/otellambda/go.mod
```diff
@@ -0,0 +1,17 @@
...
+	go.opentelemetry.io/otel v1.0.0-RC1
+	go.opentelemetry.io/otel/trace v1.0.0-RC1
```
Aneurysm9, line 16:
```suggestion
	go.opentelemetry.io/otel v1.0.0-RC2
	go.opentelemetry.io/otel/trace v1.0.0-RC2
```

instrumentation/github.com/aws/aws-lambda-go/otellambda/lambda.go
```diff
@@ -0,0 +1,158 @@
...
+type noopFlusher struct{}
+
+func (*noopFlusher) ForceFlush(context.Context) error { return nil }
+
+// Compile time check our noopFlusher implements FLusher
+var _ Flusher = &noopFlusher{}
+
+type InstrumentationOption func(o *InstrumentationOptions)
+
+type InstrumentationOptions struct {
```
Aneurysm9, line 51: Please see [the contributing guidelines related to configuration](https://github.com/open-telemetry/opentelemetry-go/blob/main/CONTRIBUTING.md#configuration).  The struct should probably be called simply `config` and not exported.  The option type should be an interface and acan be called `Option` since it will be scoped with the package name if it is used elsewhere.

```suggestion
type Option interface {
	apply(*config)
}

type config struct {
```

A helper `Option` implementation can be added for options that are simple:

```go
type optionFunc func(*config)

func (o optionFunc) apply(c *config) {
	o(c)
}
```

instrumentation/github.com/aws/aws-lambda-go/otellambda/lambda.go
```diff
@@ -0,0 +1,158 @@
...
+	// each Lambda Invocation to avoid spans being unexported for long
+	// when periods of time if Lambda freezes the execution environment
+	// The default value of Flusher is a noop Flusher, using this
+	// default can result in long data delays in asynchronous settings
+	Flusher Flusher
+}
+
+var configuration InstrumentationOptions
+var resourceAttributesToAddAsSpanAttributes []attribute.KeyValue
+var tracer trace.Tracer
```
Aneurysm9, line 68: These should be encapsulated in a struct and not left as package global variables.  The free functions in this package that operate on them should probably then be methods with the new struct that contains them as a receiver.

instrumentation/github.com/aws/aws-lambda-go/otellambda/lambda.go
```diff
@@ -0,0 +1,158 @@
...
+	err := configuration.Flusher.ForceFlush(ctx)
+	if err != nil {
+		errorLogger.Println("failed to force a flush, lambda may freeze before instrumentation exported: ", err)
+	}
+}
+
+func WithTracerProvider(tracerProvider trace.TracerProvider) InstrumentationOption {
+	return func(o *InstrumentationOptions) {
+		o.TracerProvider = tracerProvider
+	}
```
Aneurysm9, line 151: If an `optionFunc` wrapper is created, these can easily be adapted to the new `Option` interface:

```suggestion
	return optionFunc(func(c *config) {
		c.TracerProvider = tracerProvider
	})
```

instrumentation/github.com/aws/aws-lambda-go/otellambda/wrapHandler.go
```diff
@@ -0,0 +1,48 @@
...
+	o := InstrumentationOptions{
+		TracerProvider: otel.GetTracerProvider(),
+		Flusher:        &noopFlusher{},
+	}
+	for _, opt := range options {
+		opt(&o)
+	}
+	configuration = o
+	// Get a named tracer with package path as its name.
+	tracer = configuration.TracerProvider.Tracer(tracerName, trace.WithInstrumentationVersion(contrib.SemVersion()))
```
Aneurysm9, line 45: These should both be added to the `wrappedHandler` struct rather than changing package-global values.

Aneurysm9:
```
 license header checking failed:
./instrumentation/github.com/aws/aws-lambda-go/otellambda/wrapLambdaHandler.go
./instrumentation/github.com/aws/aws-lambda-go/otellambda/wrapHandler.go
make: *** [Makefile:165: license-check] Error 1
```

Please ensure appropriate license headers are in place.  `make precommit` can help catch these before the CI (though I'm bad myself and often forget to run it).

instrumentation/github.com/aws/aws-lambda-go/otellambda/lambda.go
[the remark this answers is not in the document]
garrettwegan: Propagator is now decoupled as well as discussed

instrumentation/github.com/aws/aws-lambda-go/otellambda/README.md
```diff
@@ -0,0 +1,117 @@
...
+| `WithFlusher` | `otellambda.Flusher`  | This instrumentation will call the `ForceFlush` method of its `Flusher` at the end of each invocation. Should you be using asynchronous logic (such as `sddktrace's BatchSpanProcessor`) it is very import for spans to be `ForceFlush`'ed before [Lambda freezes](https://docs.aws.amazon.com/lambda/latest/dg/runtimes-context.html) to avoid data delays. | `Flusher` with noop `ForceFlush`
+| `WithEventToTextMapCarrierConverter` | custom function `func xxx(eventJSON []byte) propagation.TextMapCarrier{}` | Function for providing custom logic to support retrieving trace header from different event types that are handled by AWS Lambda (e.g., SQS, CloudWatch, Kinesis, API Gateway) and returning them in a `propagation.TextMapCarrier` which a Propagator can use to extract the trace header into the context. | Function which returns an empty `TextMapCarrier` - new spans will be part of a new Trace and have no parent past Lambda instrumentation span
```
Aneurysm9, line 60:
```suggestion
| `WithEventToTextMapCarrierConverter` | `func(eventJSON []byte) propagation.TextMapCarrier{}` | Function for providing custom logic to support retrieving trace header from different event types that are handled by AWS Lambda (e.g., SQS, CloudWatch, Kinesis, API Gateway) and returning them in a `propagation.TextMapCarrier` which a Propagator can use to extract the trace header into the context. | Function which returns an empty `TextMapCarrier` - new spans will be part of a new Trace and have no parent past Lambda instrumentation span
```

instrumentation/github.com/aws/aws-lambda-go/otellambda/config.go
```diff
@@ -0,0 +1,116 @@
...
+// Compile time check our noopFlusher implements Flusher
+var _ Flusher = &noopFlusher{}
+
+// An EventToTextMapCarrierConverter defines how the instrumentation should
+// prepare a TextMapCarrier for the configured propagator to read from. This
+// extra step is necessary because Lambda does not have HTTP headers to read
+// from and instead stores the headers it was invoked with (including TraceID, etc.)
+// as part of the invocation event. If using the AWS XRay tracing then the
+// trace information is instead stored in the Lambda environment.
+type EventToTextMapCarrierConverter func(eventJSON []byte) propagation.TextMapCarrier
```
Aneurysm9, line 47: This name is a mouthful!  Is there something shorter that would work just as well?

instrumentation/github.com/aws/aws-lambda-go/otellambda/lambda.go
```diff
@@ -0,0 +1,89 @@
...
+	if lc != nil {
+		ctxRequestID := lc.AwsRequestID
+		attributes = append(attributes, semconv.FaaSExecutionKey.String(ctxRequestID))
+
+		// Some resource attrs added as span attrs because lambda
+		// resource detectors are created before a lambda
+		// invocation and therefore lack lambdacontext.
+		// Create these attrs upon first invocation
+		if i.resourceAttributesToAddAsSpanAttributes == nil {
```
Aneurysm9, line 64: Can the same lambda instance handle multiple concurrent invocations?  If so, this should be guarded with a `sync.Once`.  Otherwise, multiple concurrent invocations may both hit this block and double up on the added resource attributes.

Also, it doesn't look like `i.resourceAttributesToAddAsSpanAttributes` is ever set.

instrumentation/github.com/aws/aws-lambda-go/otellambda/lambda.go
```diff
@@ -0,0 +1,89 @@
...
+
+const (
+	tracerName = "go.opentelemetry.io/contrib/instrumentation/github.com/aws/aws-lambda-go/otellambda"
+)
+
+var errorLogger = log.New(log.Writer(), "OTel Lambda Error: ", 0)
+
+type instrumentor struct {
+	configuration                           config
+	resourceAttributesToAddAsSpanAttributes []attribute.KeyValue
```
Aneurysm9, line 38:
```suggestion
	resAttrs []attribute.KeyValue
```

Sometimes, less is more.

instrumentation/github.com/aws/aws-lambda-go/otellambda/wrapLambdaHandler.go
```diff
@@ -0,0 +1,207 @@
...
+		Propagator:                     otel.GetTextMapPropagator(),
+	}
+	for _, opt := range options {
+		opt.apply(&cfg)
+	}
+
+	i := instrumentor{}
+	i.configuration = cfg
+	// Get a named tracer with package path as its name.
+	i.tracer = i.configuration.TracerProvider.Tracer(tracerName, trace.WithInstrumentationVersion(contrib.SemVersion()))
```
Aneurysm9, line 116: This code block seems to be repeated.  Perhaps extract it as `newInstrumentor(opts ...Option) instrumentor`.

instrumentation/github.com/aws/aws-lambda-go/otellambda/lambda.go
garrettwegan: We shouldn't have to worry about concurrency here, lambda will create new instances to handle concurrent requests: https://docs.aws.amazon.com/lambda/latest/dg/configuration-concurrency.html

instrumentation/github.com/aws/aws-lambda-go/otellambda/config.go
garrettwegan: Going to leave some brainstorming ideas here and come back to them later:

- TraceDetector: **Pros**: detects whether there is a trace in the event/environ, short and intuitive name **Cons**: don't want to create confusion with ResourceDetectors, sounds like a struct not function
- EventToCarrier: **Pros** describes datatypes in/out, short **Cons** desired outcome is not intuitive
- GetTraceHeader: **Pros** describes purpose of function, intuitive output **Cons** Implies trace is in a Header obj somewhere (not always true eg xray)

pellared: `EventToCarrier` LGTM

Aneurysm9 (approved):

instrumentation/github.com/aws/aws-lambda-go/otellambda/wrapLambdaHandler.go
```diff
@@ -0,0 +1,188 @@
...
+	} else if handler.NumIn() > 0 {
+		contextType := reflect.TypeOf((*context.Context)(nil)).Elem()
+		argumentType := handler.In(0)
+		handlerTakesContext = argumentType.Implements(contextType)
+		if handler.NumIn() > 1 && !handlerTakesContext {
+			return false, fmt.Errorf("handler takes two arguments, but the first is not Context. got %s", argumentType.Kind())
+		}
+	}
+
+	return handlerTakesContext, nil
```
willarmiros, line 51: Is it intentional that this returns `false` in the event that there are 0 arguments? If so, should we also provide an error message?

willarmiros (approved):
Generally lgtm, just a couple questions!

instrumentation/github.com/aws/aws-lambda-go/otellambda/wrapLambdaHandler.go
garrettwegan: Yes this is intentional. This return is used to determine whether to pass context to the user's function. The user's function only needs to take context if they plan to instrument downstream operations. Not taking context is a valid choice such as in a case where the Lambda function is the final operation in a trace. Such cases are admittedly unlikely though if they are bothering to instrument lambda.

willarmiros:
@bhautikpip would you mind taking a quick look from AWS side as well?

@open-telemetry/go-approvers would anyone else also be able to give this a look and get it merged? Many of the files changed are examples/documentation, and we would love to release this to OTel Go users!

instrumentation/github.com/aws/aws-lambda-go/otellambda/wrapHandler.go
```diff
@@ -0,0 +1,50 @@
...
+// wrappedHandler is a struct which holds an instrumentor
+// as well as the user's original lambda.Handler and is
+// able to instrument invocations of the user's lambda.Handler
+type wrappedHandler struct {
+	instrumentor instrumentor
+	handler      lambda.Handler
+}
+
+// Compile time check our Handler implements lambda.Handler
+var _ lambda.Handler = wrappedHandler{}
```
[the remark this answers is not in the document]
garrettwegan: Yes this is true. Same as if they change handler function signature requirements for Start() entrypoint.

instrumentation/github.com/aws/aws-lambda-go/otellambda/wrapLambdaHandler.go
willarmiros: Ok I see, so it's a silent failure by design. That makes sense.

instrumentation/github.com/aws/aws-lambda-go/otellambda/wrapHandler.go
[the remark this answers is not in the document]
willarmiros: Ok should be infrequent, just something to lookout for.

garrettwegan:
@jmacd and @MrAlias I think both of y'all checked out the links to my PR's in the SIG last week, if either of y'all have had a chance to look them over a bit I'd love some feedback

bhautikpip (approved):
LGTM :)

garrettwegan:
Closing in favor of #983, @bhautikpip will own related PRs going forward

PR #897, opened 2021-07-23
Remove oteltest dependency from otelmux

MrAlias:
Replace the use of the `oteltest` package with the `sdk/trace/tracetest` package for the `otelmux` instrumentation. Do this in its own `test` module to not impose a dependency on the default SDK to the instrumentation user.

This uses a pre-release version of the `sdk/trace/tracetest` to prepare for the next release of otel-go.

Part of https://github.com/open-telemetry/opentelemetry-go/issues/2121


Aneurysm9 (approved):

pellared (approved):
nice refactoring 😉

MrAlias:
> nice refactoring 
> 
> PS. Do we want to play with sth like https://github.com/ory/go-acc to more accurate code coverage results?
> 
> More: https://www.ory.sh/golang-go-code-coverage-accurate/

Yes! I was looking for something exactly like this. I think we should try to incorporate this.

PR #476, opened 2020-12-02
Migrate CircleCI ci.yml workflow to GitHub Actions

AzfaarQureshi:
### What does this do

This  addresses issue https://github.com/open-telemetry/opentelemetry-go/issues/880 by migrating CI/CD from CircleCI to GitHub Actions

### Changes

* Migrated `current-go`, `prior-go` and `integration-test` CircleCI jobs to GitHub Actions. 
* Changelog updated

### Migration Plan
I suggest having CircleCI and GitHub Action jobs run in parallel for a few weeks. After the GitHub Actions jobs are running fine for a week or so, remove the CircleCI workflows from `config.yml`

cc- @alolita

PR #935, opened 2021-08-03
Ensure Integration Tests run with Github Actions

evantorrie:
These have been "skipped" since  #476 was merged due to the lack of setting the `INTEGRATION` environment variable prior to running the test.


MrAlias (approved):
:facepalm:

pellared (approved):

PR #950, opened 2021-08-10
Mark the span as error when the HTTP transporter fails

slok:
Hi! :wave: 

Currently, the HTTP Transporter is recording the errors, however, is not marking the span as an error, and this is misleading. For example, if the Request fails because there is nobody listening to our client calls we would not see that the request has failed easily (unless an upper layer records the error): 

Before:

Screenshot_20210810_070512

After the fix:

Screenshot_20210810_071420


dashpole (approved):
Nice!  I ran into this yesterday when instrumenting a kubernetes webhook myself...

instrumentation/net/http/otelhttp/transport.go
```diff
@@ -97,6 +98,7 @@ func (t *Transport) RoundTrip(r *http.Request) (*http.Response, error) {
 	res, err := t.rt.RoundTrip(r)
 	if err != nil {
 		span.RecordError(err)
+		span.SetStatus(codes.Error, err.Error())
```
MrAlias, line 101: Can we add tests to prevent regressions. This behavior used to exist in old versions of `RecordError`'s predecessor and I'm guessing it was intended by the original author. It would have been nice to have tests fail when it was removed upgrading to the latest `RecordError`. Adding tests now will help preserve this behavior in the future.

slok: I didn't add in the first place because apparently the coverage was good enough, and 100% coverage doesn't always mean better due to the complexity of use cases and the tests. Knowing that this is important enough, I added for both cases!

MrAlias (approved):

MrAlias:
This looks ready to merge outside of fixing the lint issues:

```
transport_test.go:313:2: ineffectual assignment to ctx (ineffassign)
	ctx, span := tracer.Start(ctx, "test")
	^
```

Aneurysm9 (approved):

slok:
Done!

instrumentation/net/http/otelhttp/transport_test.go
```diff
@@ -243,3 +246,101 @@ func TestTransportUsesFormatter(t *testing.T) {
 	}
 
 }
+
+func TestTransportErrorStatus(t *testing.T) {
+	// Prepare tracing stuff.
+	prop := propagation.TraceContext{}
```
pellared, line 252: I guess this can be removed as it is not used

instrumentation/net/http/otelhttp/transport_test.go
```diff
@@ -243,3 +246,101 @@ func TestTransportUsesFormatter(t *testing.T) {
...
+
+	spanStatusCode := gotSpans[0].StatusCode()
+	if spanStatusCode != codes.Error {
+		t.Fatalf("expected error status code on span; got: %q", spanStatusCode)
+	}
+
+	spanStatusMessage := gotSpans[0].StatusMessage()
+	if !strings.Contains(spanStatusMessage, "connect: connection refused") {
+		t.Fatalf("expected error status message on span; got: %q", spanStatusMessage)
+	}
```
pellared, line 297: I would use `t.Errorf` in this section instead of `t.Fatalf`

instrumentation/net/http/otelhttp/transport_test.go
```diff
@@ -243,3 +246,101 @@ func TestTransportUsesFormatter(t *testing.T) {
...
+
+	spanStatusCode := gotSpans[0].StatusCode()
+	if spanStatusCode != codes.Error {
+		t.Fatalf("expected error status code on span; got: %q", spanStatusCode)
+	}
+
+	spanStatusMessage := gotSpans[0].StatusMessage()
+	if !strings.Contains(spanStatusMessage, "something") {
+		t.Fatalf("expected error status message on span; got: %q", spanStatusMessage)
+	}
```
pellared, line 345: I would use `t.Errorf` in this section instead of `t.Fatalf`

pellared (approved):
Just a few minor comments to the test code. Nice 👍

instrumentation/net/http/otelhttp/transport_test.go
slok: Right! done in [`ed15ce8` (#950)](https://github.com/open-telemetry/opentelemetry-go-contrib/pull/950/commits/ed15ce8d81b07ac9aedd98d29aa8455741a1464d)

instrumentation/net/http/otelhttp/transport_test.go
slok: Done in [`ed15ce8` (#950)](https://github.com/open-telemetry/opentelemetry-go-contrib/pull/950/commits/ed15ce8d81b07ac9aedd98d29aa8455741a1464d)

instrumentation/net/http/otelhttp/transport_test.go
slok: Done in [`ed15ce8` (#950)](https://github.com/open-telemetry/opentelemetry-go-contrib/pull/950/commits/ed15ce8d81b07ac9aedd98d29aa8455741a1464d)

PR #983, opened 2021-08-12
End to end tracing support - Lambda

bhautikpip:
See for example use case: https://github.com/garrettwegan/opentelemetry-lambda/blob/main/go/sample-apps/function/function.go


CHANGELOG.md
```diff
@@ -8,6 +8,11 @@ This project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0.htm
 
 ## [Unreleased]
 
+### Added
+
+- Add instrumentation for the `github.com/aws/aws-lambda-go` package. (#882)
+- Add resource detector for AWS Lambda. (#882)
```
garrettwegan, line 14: Update PR number
```suggestion
- Add instrumentation for the `github.com/aws/aws-lambda-go` package. (#983)
- Add resource detector for AWS Lambda. (#983)
```

bhautikpip:
@Aneurysm9 @MrAlias Please take a look whenever you get a chance. This PR is derived from https://github.com/open-telemetry/opentelemetry-go-contrib/pull/882

Aneurysm9 (approved):

detectors/aws/lambda/README.md
```diff
@@ -0,0 +1,67 @@
...
+Now your `TracerProvider` will have the following resource attributes and attach them to new spans:
+
+| Resource Attribute | Example Value |
+| --- | --- |
+| `cloud.provider` | aws
+|`cloud.region` | us-east-1 
+|`faas.name` | MyLambdaFunction 
+|`faas.version` | $LATEST
+
+Of note, `faas.id` and `cloud.account.id` are not set by the Lambda resource detector because they are not available outside a Lambda invocation. For this reason, when using the AWS Lambda Instrumentation these attributes are set as additional span attributes.
```
punya, line 50: Could this be solved through lazy initialization?

willarmiros: Not without asking the user to add more code. The `faas.id` and `cloud.account.id` attributes are only available inside of the handler, and our solution is meant to be strictly a wrapper of the handler such that we can't actually do anything inside the wrapper. So we could ask the user to lazy-initialize the tracer provider in their handler, but that's a much worse experience.

detectors/aws/lambda/README.md
```diff
@@ -0,0 +1,67 @@
...
+	detector := lambdadetector.NewResourceDetector()
+	res, err := detector.Detect(context.Background())
+	if err != nil {
+		fmt.Printf("failed to detect lambda resources: %v\n", err)
+	}
+
+	tp := sdktrace.NewTracerProvider(
+		sdktrace.WithResource(res),
+	)
+	lambda.Start(<some lambda handler>)
```
willarmiros, line 37:
```suggestion
	lambda.Start(myHandler, otellambda.WithTracerProvider(tp))
	
```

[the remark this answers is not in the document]
willarmiros: Good catch, made a suggestion

detectors/aws/lambda/README.md
punya: Why would the *user* have to do that lazy-initialization, rather than the framework? E.g., at [this point](https://github.com/open-telemetry/opentelemetry-go-contrib/pull/983/files#diff-76cf3df45b0cb4100c653ba5800e2123291f47369b5c96e68a6df167c2403c9dR122) it seems like you'd have access to those invocation-scoped variables. It's true that you'd have to do separately for each possible reflective type of the handler function.

willarmiros: I think the problem is that we will then couple the Lambda resource detector and the Lambda instrumentation, which ought to be two separate components. The Lambda instrumentation does indeed have access to those invocation-scoped attributes, and is able to initialize the tracer provider. But the Lambda resource detector must operate only with what is available in the static environment, so if a user wants to initialize a custom tracer provider outside their handler as seen in the example it will work the same.

I agree that it's unfortunate the lambda resource detector doesn't capture these attributes, but I believe this is a limitation of Lambda itself more than any individual instrumentation, which is why the span attribute is called out [in the spec](https://github.com/open-telemetry/opentelemetry-specification/blob/main/specification/resource/semantic_conventions/faas.md) and there's talks of [not even marking faas.id as a resource attribute](https://github.com/open-telemetry/opentelemetry-specification/issues/1261) at all.

punya: Thanks for explaining. I was commenting based on "what would an operator find most useful" vs "what is a clean way to factor the code." Given the ambiguity / ongoing discussion around faas.id, I think this code makes a reasonable choice.

instrumentation/github.com/aws/aws-lambda-go/otellambda/test/go.mod
```diff
@@ -0,0 +1,22 @@
...
+)
+
+require (
+	github.com/aws/aws-lambda-go v1.25.0
+	github.com/stretchr/testify v1.7.0
+	go.opentelemetry.io/contrib v0.22.0
+	go.opentelemetry.io/contrib/detectors/aws/lambda v0.22.0
+	go.opentelemetry.io/contrib/instrumentation/github.com/aws/aws-lambda-go/otellambda v0.22.0
+	go.opentelemetry.io/contrib/propagators/aws v0.22.0
+	go.opentelemetry.io/otel v1.0.0-RC2
```
codeboten, line 19: should these be updated to 1.0.0 or 1.0.1?

instrumentation/github.com/aws/aws-lambda-go/otellambda/go.mod
```diff
@@ -0,0 +1,13 @@
...
+
+go 1.15
+
+replace go.opentelemetry.io/contrib => ../../../../..
+
+require (
+	github.com/aws/aws-lambda-go v1.24.0
+	github.com/stretchr/testify v1.7.0
+	go.opentelemetry.io/contrib v0.22.0
+	go.opentelemetry.io/otel v1.0.0-RC2
```
codeboten, line 11: should these be updated to 1.0.0 or 1.0.1?

detectors/aws/lambda/detector.go
```diff
@@ -0,0 +1,68 @@
...
+	"errors"
+	"os"
+
+	"go.opentelemetry.io/otel/attribute"
+	"go.opentelemetry.io/otel/sdk/resource"
+	semconv "go.opentelemetry.io/otel/semconv/v1.4.0"
+)
+
+const (
+	lambdaFunctionNameEnvVar    = "AWS_LAMBDA_FUNCTION_NAME"
```
codeboten, line 28: are any of the other variables valuable here? if not maybe a note that points a user to where they can find the list would be useful: https://docs.aws.amazon.com/lambda/latest/dg/configuration-envvars.html

detectors/aws/lambda/go.mod
```diff
@@ -0,0 +1,11 @@
+module go.opentelemetry.io/contrib/detectors/aws/lambda
+
+go 1.15
+
+require (
+	github.com/davecgh/go-spew v1.1.1 // indirect
+	github.com/stretchr/testify v1.7.0
+	go.opentelemetry.io/otel v1.0.0-RC2
```
codeboten, line 8: Please update the versions

detectors/aws/lambda/detector.go
willarmiros: None of the other environment variables store data that should be recorded per the [faas spec](https://github.com/open-telemetry/opentelemetry-specification/blob/main/specification/resource/semantic_conventions/faas.md), but I'll make a suggestion to include that link in a comment

detectors/aws/lambda/detector.go
```diff
@@ -0,0 +1,68 @@
...
+	"context"
+	"errors"
+	"os"
+
+	"go.opentelemetry.io/otel/attribute"
+	"go.opentelemetry.io/otel/sdk/resource"
+	semconv "go.opentelemetry.io/otel/semconv/v1.4.0"
+)
+
+const (
```
willarmiros, line 27:
```suggestion
// For a complete list of reserved environment variables in Lambda, see:
// https://docs.aws.amazon.com/lambda/latest/dg/configuration-envvars.html
const (
```

instrumentation/github.com/aws/aws-lambda-go/otellambda/go.mod
bhautikpip: Updated!

detectors/aws/lambda/go.mod
willarmiros: Thanks!

instrumentation/github.com/aws/aws-lambda-go/otellambda/README.md
```diff
@@ -0,0 +1,117 @@
...
+	lambda.Start(otellambda.WrapHandlerFunction(HandleRequest))
+}
+```
+
+## AWS Lambda Instrumentation Options
+
+| Options | Input Type  | Description | Default |
+| --- | --- | --- | --- |
+| `WithTracerProvider` | `trace.TracerProvider` | Provide a custom `TracerProvider` for creating spans. Consider using the [AWS Lambda Resource Detector][lambda-detector-url] with your tracer provider to improve tracing information. | `otel.GetTracerProvider()`
+| `WithFlusher` | `otellambda.Flusher`  | This instrumentation will call the `ForceFlush` method of its `Flusher` at the end of each invocation. Should you be using asynchronous logic (such as `sddktrace's BatchSpanProcessor`) it is very import for spans to be `ForceFlush`'ed before [Lambda freezes](https://docs.aws.amazon.com/lambda/latest/dg/runtimes-context.html) to avoid data delays. | `Flusher` with noop `ForceFlush`
```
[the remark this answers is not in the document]
willarmiros: IIRC we didn't want to force Lambda customers using async patterns to incur additional execution time from this instrumentation, but I'm not 100% sure what the reasoning behind making the default no-op was. Perhaps @Aneurysm9 remembers better.

instrumentation/github.com/aws/aws-lambda-go/otellambda/lambda.go
```diff
@@ -0,0 +1,107 @@
...
+}
+
+// Logic to start OTel Tracing
+func (i *instrumentor) tracingBegin(ctx context.Context, eventJSON []byte) (context.Context, trace.Span) {
+	// Add trace id to context
+	mc := i.configuration.EventToCarrier(eventJSON)
+	ctx = i.configuration.Propagator.Extract(ctx, mc)
+
+	var span trace.Span
+	spanName := os.Getenv("AWS_LAMBDA_FUNCTION_NAME")
```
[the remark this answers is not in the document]
willarmiros: It's guaranteed to be set: https://docs.aws.amazon.com/lambda/latest/dg/configuration-envvars.html#configuration-envvars-runtime

instrumentation/github.com/aws/aws-lambda-go/otellambda/README.md
```diff
@@ -0,0 +1,117 @@
...
+```
+
+Now use the provided wrapper to instrument your basic Lambda function:
+```go
+// Add import
+import "go.opentelemetry.io/contrib/instrumentation/github.com/aws/aws-lambda-go/otellambda"
+
+// wrap lambda handler function
+func main() {
+	lambda.Start(otellambda.WrapHandlerFunction(HandleRequest))
```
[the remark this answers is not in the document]
willarmiros: Not sure about `NewHandler` since we're not instantiating a new function handler, we are indeed just wrapping it. I think `InstrumentHandler()` makes sense but would like to have @bhautikpip weigh in.

bhautikpip:
@dashpole addressed your comments! Feel free to take a look again.

dashpole (approved):

PR #1008, opened 2021-08-17
Fix test-with-coverage target

MrAlias:
Currently the coverage tests (from #985) will show the coverage for the entirety of the `go.opentelemetry.io/contrib/...` packages. This means that any package from the repository that uses another, unrelated, package will cover that package. For example any HTTP framework instrumentation that uses the underlying `go.opentelemetry.io/contrib/instrumentation/net/http/otelhttp` instrumentation will also be considered in the coverage of that underlying instrumentation.

This overlap is unintentional and can lead to [test failures](https://github.com/open-telemetry/opentelemetry-go-contrib/runs/3345557976).


Aneurysm9 (approved):

evantorrie (approved):
Only quibble I would have is that we alternate between using `$$variable` and `$${variable}` within the same code block, e.g. `$$dir` and `$${dir}`, `$$CMD` and `$${CMD}`.

I'd pick one and keep it consistent.

Makefile
```diff
@@ -55,9 +55,13 @@ test-with-coverage: $(TOOLS_DIR)/gocovmerge
...
 	for dir in $(ALL_COVERAGE_MOD_DIRS); do \
-	  echo "go test ./... + coverage in $${dir}"; \
+	  CMD="$(GOTEST_WITH_COVERAGE)"; \
+	  echo "$$dir" | \
+	    grep -q 'test$$' && \
+		CMD="$${CMD} -coverpkg=go.opentelemetry.io/contrib/$$( dirname "$$dir" | sed -e "s/^\.\///g" )/..."; \
+	  echo "$$CMD $${dir}/..."; \
 	  (cd "$${dir}" && \
-	    $(GOTEST_WITH_COVERAGE) ./... && \
+	    eval "$$CMD ./..." && \
```
evantorrie, line 64: Do you need the `eval` here?  Or is `$$CMD ./...` sufficient?

PR #1017, opened 2021-08-20
added new dir for EC2 resource detector 

bhautikpip:
This PR only changes the location of ec2 detectors and created a dir for ec2 so that ec2 resource detector has parity with other resource detectors directory structure like ecs and eks. Also, importing ec2 resource detector would be more intuitive with this change.


bhautikpip:
@Aneurysm9 @MrAlias Can you give it a review? It's a very small change :)

MrAlias (changes requested):
This is more than a directory change, it is a package rename. Please be sure to include an entry in the CHANGELOG to track this break change and be sure to check this is a desired enough change for users of the AWS instrumentation to warrant the change.

bhautikpip:
> This is more than a directory change, it is a package rename. Please be sure to include an entry in the CHANGELOG to track this break change and be sure to check this is a desired enough change for users of the AWS instrumentation to warrant the change.

I submitted this PR because I myself got confused that we didn't have EC2 resource detector support since I was only looking at ecs and eks modules and not checking code at root of the package. So, I thought of adding separate ec2 module for EC2 resource detector as well to kind of have a parity (with ECS and EKS) and that will probably help causing less confusion. I think this is a good change specifically because this will enable customer to only import EC2 module if they just want to use EC2 resource detector. Added a CHANGELOG entry for this as well and since it's a breaking change it would be good if we can add this before GA release.

bhautikpip:
Checks are failing due to some internal error feel free to re-run again :)

MrAlias (approved):

Aneurysm9 (approved):

PR #1050, opened 2021-09-04
Add Inject Response Header in gin middleware

drinks5:
* Impletment the traceresponse in [W3C Trace-Context HTTP Propagator](https://w3c.github.io/trace-context/#traceresponse-header) and [b3 response header](https://github.com/openzipkin/b3-propagation/blob/master/RATIONALE.md)


drinks5:
@XSAM Hello, could u help me to review this PR?

pellared (changes requested):
I think that by default the HTTP responses should not contain the propagation data. It would leak the internals.

The W3C specs is an "Editor's Draft" version which does not seem to be stable. The current Trace Context specs (recommended by W3C) do not contain the `Traceresponse` header.

At Splunk, we have coined a `TraceResponseHeader` option to add to this info via `Server-Timing` header.

References: 
- https://github.com/open-telemetry/opentelemetry-specification/blob/main/specification/context/api-propagators.md
- https://www.w3.org/TR/trace-context/
- https://github.com/signalfx/splunk-otel-go/tree/main/instrumentation/net/http/splunkhttp
- https://www.w3.org/TR/server-timing/

MrAlias:
Closing as there seems consensus from the community that this is not something we want to support until it is stabilized in the W3C. Please feel free to fork this and codebase and use as you see fit @drinks5.

Issue #1104, opened 2021-09-16
Options should not allow setting `nil` providers


Aneurysm9:
All instrumentations accept functional options to set the `TracerProvider` and/or `MeterProvider`, as appropriate for the instrumentation.  These providers are used to create the `Tracer` and `Meter` used by the instrumentation.  Most (all?) instrumentations follow a pattern whereby they set a default provider in their configuration struct and then apply user-provided options.  The instrumentations assume that at the end of this operation the configuration struct will still hold a valid provider.  Unfortunately, a `nil` value can be provided to an option which results in a panic when the instrumentation attempts to use the provider.

Most provider options look something like this:

```go
func WithTracerProvider(provider trace.TracerProvider) Option {
	return optionFunc(func(cfg *config) {
		cfg.TracerProvider = provider
	})
}
```

A simple guard can be added around setting the provider in the `config` struct:

```go
func WithTracerProvider(provider trace.TracerProvider) Option {
	return optionFunc(func(cfg *config) {
		if provider != nil {
			cfg.TracerProvider = provider
		}
	})
}
```

PR #1107, opened 2021-09-17
Added guard to prevent Options from setting nil providers

erichsueh3:
**Description:**
This PR adds a guard to prevent options from setting nil providers, in the form of a if provider != nil check.

**Link to tracking Issue:**
#1104


Aneurysm9 (approved):

pellared (approved):
Should we do the same for `WithPropagators` as a followup?

erichsueh3:
> Should we do the same for `WithPropagators` as a followup?

If yes please open up an issue and assign it to me if possible, thanks!

PR #1144, opened 2021-09-30
OTel-Go-Contrib 0.24 release

jmacd:
This updates the exporter interfaces and test changes mainly due to
https://github.com/open-telemetry/opentelemetry-go/pull/2197

This integrates the branch from @Aneurysm9 here: https://github.com/open-telemetry/opentelemetry-go-contrib/compare/main...Aneurysm9:release_prep?expand=1

PR #1330, opened 2021-10-04
Upgrade gopsutil package used in instrumentation/host

evantorrie:
gopsutil-v2.x was failing during tests on my Mac.  Took the opportunity to upgrade it to the next major version `gopsutil-v3.x` which no longer fails.

Also updated this particular instrumentation to use the new API from `go.opentelemetry.io/otel/metric-v0.24.0`


jmacd (approved):

jmacd:
Note I'd like to see https://github.com/open-telemetry/opentelemetry-go-contrib/pull/1144 merge first, conflicts either way.

pellared (approved):
👍

Aneurysm9:
I've incorporated this into #1144 as I also had these test errors.

PR #1342, opened 2021-10-12
update deps as per dependabot suggestions

codeboten:


MrAlias:
It looks like https://github.com/Shopify/sarama introduced function use that was added in Go 1.16. We may need to skip that upgrade.

codeboten:
> It looks like https://github.com/Shopify/sarama introduced function use that was added in Go 1.16. We may need to skip that upgrade.

Happy to revert that portion of the change. What's the process on updating the minimum version of go required for the repo?

PR #1353, opened 2021-10-13
Auto-scaling base-2 histogram aggregator

jmacd:
**Not ready for review.**

This implementation of the OTLP v0.11 exponential histogram provides an aggregator for calculating ExponentialHistogramDataPoint values from individual measurements. The only configurable parameter of this implementation presently is the maximum size: it implements auto-scaling to fit all observed values into the available buckets.

This behavior will not work for all consumers, thus more development will be needed.

This PR is a placeholder, linked with prerequisites in the collector and collector-contrib repositories.


jmacd:
I moved this work into the OTel-Go repo because it was difficult to develop an end-to-end demonstration. We need support for a new aggregator type and a new OTLP export code path; I may have to move it back to this repo, in which case I'll re-open this PR.

PR #1358, opened 2021-10-20
use true/false for ot-tracer-sampled header

codeboten:
Other implementations of the ot trace propagator are setting the values to `true`/`false`, this change is making the go implementation consistent.

See other implementations:
- [Java](https://github.com/open-telemetry/opentelemetry-java/blob/9cea4ef1f92d3186b1bd8296e9daac4281c0f759/extensions/trace-propagators/src/main/java/io/opentelemetry/extension/trace/propagation/OtTracePropagator.java#L73)
- [Python](https://github.com/open-telemetry/opentelemetry-python-contrib/blob/c24c77dd3adcbfe9406b139acaa08d68155d6edb/propagator/opentelemetry-propagator-ot-trace/src/opentelemetry/propagators/ot_trace/__init__.py#L130-L133)
- [Javascript](https://github.com/open-telemetry/opentelemetry-js-contrib/blob/f5851e72512117dbce571a42930a90c560dbf63d/propagators/opentelemetry-propagator-ot-trace/src/OTTracePropagator.ts#L75)

### DIFF
--- a/propagators/ot/ot_data_test.go
+++ b/propagators/ot/ot_data_test.go
@@ -198,7 +198,7 @@ var injectHeaders = []injectTest{
 		wantHeaders: map[string]string{
 			traceIDHeader: traceID16Str,
 			spanIDHeader:  spanIDStr,
-			sampledHeader: "1",
+			sampledHeader: "true",
 		},
 	},
 	{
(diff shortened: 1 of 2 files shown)


MrAlias (approved):
For those also interested, this is the behavior in the original OpenTracing implementation as well: https://github.com/opentracing/basictracer-go/blob/6bcd2fc91c40602d5f6a14991fa933cb9c90df71/propagation_ot.go#L45

PR #1385, opened 2021-11-02
Update semantic conventions to v1.7.0

### LINKED ISSUE #1384: Cannot merge resource due to conflicting Schema URL in OTEL v1.1.0 / OTEL-contrib v1.1.0
If we attempt to upgrade to the latest OTEL and OTEL contrib, we're hitting the error `cannot merge resource due to conflicting Schema URL` attempting to create a resource detector on startup. This appears to be due to the fact that the OTEL libraries have updated to semconv v1.7.0 while several of the OTEL-contrib packages are still on semconv v1.4.0.

Aneurysm9:
Fixes #1384


MrAlias (approved):

odsod:
@MrAlias @Aneurysm9 Thanks for the quick fix! Would you be open to cut a patch release for this?

PR #1413, opened 2021-11-08
Go lambda instrumentation enhancements

bhautikpip:
1. This PR introduces Go Lambda instrumentation enhancements to make `xrayconfig` package more useful.


instrumentation/github.com/aws/aws-lambda-go/otellambda/xrayconfig/xrayconfig.go
```diff
@@ -35,48 +34,39 @@ func xrayEventToCarrier([]byte) propagation.TextMapCarrier {
...
 		sdktrace.WithIDGenerator(xray.NewIDGenerator()),
-		sdktrace.WithResource(res),
-	)
+		sdktrace.WithResource(resource),
+	), nil
+}
+
+// TracerProvider returns an otellambda.Option(s) to
+// enable using a TracerProvider configured for AWS XRay via a collector
+func TracerProvider(tp *sdktrace.TracerProvider) otellambda.Option {
```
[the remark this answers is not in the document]
bhautikpip: We have followed that pattern in `config.go` file in `otellambda` package. I think for `xrayconfig.go` we have followed naming without `With` keyword. See [here](https://github.com/open-telemetry/opentelemetry-go-contrib/blob/main/instrumentation/github.com/aws/aws-lambda-go/otellambda/xrayconfig/xrayconfig.go#L91) for `Propagator` and `EventToCarrier` method name for reference.

instrumentation/github.com/aws/aws-lambda-go/otellambda/xrayconfig/README.md
```diff
@@ -46,17 +46,17 @@ import "go.opentelemetry.io/contrib/instrumentation/github.com/aws/aws-lambda-go
 
 // add options to WrapHandlerFunction call
 func main() {
-	lambda.Start(otellambda.WrapHandlerFunction(HandleRequest, xrayconfig.AllRecommendedOptions()...))
+	lambda.Start(otellambda.InstrumentHandler(HandleRequest, xrayconfig.RecommendedOptions()...))
```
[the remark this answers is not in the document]
bhautikpip: Looks like `InstrumentHandler` is the correct name. We have changed the name in original merge of lambda instrumentation to address reviewer feedback. Ref: https://github.com/open-telemetry/opentelemetry-go-contrib/blob/main/instrumentation/github.com/aws/aws-lambda-go/otellambda/wrapLambdaHandler.go#L98

instrumentation/github.com/aws/aws-lambda-go/otellambda/xrayconfig/README.md
```diff
@@ -34,7 +34,7 @@ func HandleRequest(ctx context.Context, name MyEvent) (string, error) {
 }
 
 func main() {
-	lambda.Start(otellambda.WrapHandlerFunction(HandleRequest))
+	lambda.Start(otellambda.InstrumentHandler(HandleRequest))
```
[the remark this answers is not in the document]
bhautikpip: yep. see this [comment](https://github.com/open-telemetry/opentelemetry-go-contrib/pull/1413#discussion_r745193392)

instrumentation/github.com/aws/aws-lambda-go/otellambda/xrayconfig/xrayconfig.go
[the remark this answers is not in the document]
Aneurysm9: I understand that other functions in the `xrayconfig` package were not adhering to the convention.  I now think that to be a mistake that we should fix at the earliest opportunity, a.k.a., now.

instrumentation/github.com/aws/aws-lambda-go/otellambda/lambda.go
```diff
@@ -24,6 +24,10 @@ import (
 
 	"go.opentelemetry.io/otel"
 	"go.opentelemetry.io/otel/attribute"
+
+	// import to call ForceFlush method of batch_span_processor.go
+	//https://github.com/open-telemetry/opentelemetry-go/blob/main/sdk/trace/batch_span_processor.go#L167
+	_ "go.opentelemetry.io/otel/sdk/trace"
```
Aneurysm9, line 30:
```suggestion
```

This is unnecessary and must be removed.  Instrumentation packages must never depend on an SDK implementation.  The instrumentation defines its own `Flusher` interface so that it does not need to explicitly depend on any concrete type.

bhautikpip: If we remove this then we might need to add this method (https://github.com/open-telemetry/opentelemetry-go-contrib/pull/1413#discussion_r745082701) again right? I tested the code without it and we're not seeing instrumentation nodes coming up properly.

instrumentation/github.com/aws/aws-lambda-go/otellambda/xrayconfig/xrayconfig.go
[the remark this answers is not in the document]
bhautikpip: I think it would be confusing because then public methods in `otellambda` packages would have the same name as public methods in `xrayconfig` packages and would confuse users but I'm open do it like the way you mentioned but just posting this as a concern.

instrumentation/github.com/aws/aws-lambda-go/otellambda/lambda.go
Aneurysm9: No, there is no need for that.  This won't be released until after the next SDK release.  The `go.mod` files that need to import the SDK will be updated as appropriate at that time.  We do not make any guarantees regarding the ability to use unreleased commits of the SDK or instrumentation.

instrumentation/github.com/aws/aws-lambda-go/otellambda/xrayconfig/xrayconfig.go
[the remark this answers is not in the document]
Aneurysm9: We should simply remove the superfluous functions.  The `TracerProvider()` and `Flusher()` functions provide no value as they merely pass through to the functions in the instrumentation package.  `WithRecommendedOptions()` can call the `Option`-generating functions in the instrumentation package directly.  `Propagator()` and `EventToCarrier()` should be renamed with the `With` prefix.

Aneurysm9 (changes requested):
Instrumentation must not depend on an SDK.

instrumentation/github.com/aws/aws-lambda-go/otellambda/xrayconfig/README.md
```diff
@@ -44,19 +44,19 @@ Now configure the instrumentation with the provided options to export traces to
...
-| `WithPropagator` | An `xray.propagator` | Individually exported as `EventToCarrier()`, also included in `AllRecommendedOptions()`
+| `WithTracerProvider` | An `sdktrace.TracerProvider` configured to export in batches to an OTel Collector running locally in Lambda | Not individually exported. Can only be used via `WithRecommendedOptions()`
+| `WithFlusher` | An `otellambda.Flusher` which yields before calling ForceFlush on the configured `sdktrace.TracerProvider`. Yielding mitigates data delays caused by asynchronous nature of batching TracerProvider when in Lambda | Not individually exported. Can only be used via `WithRecommendedOptions()`
+| `WithEventToCarrier` | Function which reads X-Ray TraceID from Lambda environment and inserts it into a `propagtation.TextMapCarrier` | Individually exported as `EventToCarrier()`, also included in `WithRecommendedOptions()`
+| `WithPropagator` | An `xray.propagator` | Individually exported as `Propagator()`, also included in `WithRecommendedOptions()`
```
Aneurysm9, line 59:
```suggestion
| `WithEventToCarrier` | Function which reads X-Ray TraceID from Lambda environment and inserts it into a `propagtation.TextMapCarrier` | Individually exported as `WithEventToCarrier()`, also included in `WithRecommendedOptions()`
| `WithPropagator` | An `xray.propagator` | Individually exported as `WithPropagator()`, also included in `WithRecommendedOptions()`
```

instrumentation/github.com/aws/aws-lambda-go/otellambda/xrayconfig/xrayconfig.go
```diff
@@ -35,72 +34,45 @@ func xrayEventToCarrier([]byte) propagation.TextMapCarrier {
...
 		sdktrace.WithIDGenerator(xray.NewIDGenerator()),
-		sdktrace.WithResource(res),
-	)
-
-	return []otellambda.Option{otellambda.WithTracerProvider(tp), otellambda.WithFlusher(asyncSafeFlusher{tp: tp})}, nil
+		sdktrace.WithResource(resource),
+	), nil
 }
 
 // EventToCarrier returns an otellambda.Option to enable
```
Aneurysm9, line 61:
```suggestion
// WithEventToCarrier returns an otellambda.Option to enable
```

instrumentation/github.com/aws/aws-lambda-go/otellambda/xrayconfig/xrayconfig.go
```diff
@@ -35,72 +34,45 @@ func xrayEventToCarrier([]byte) propagation.TextMapCarrier {
...
 // an otellambda.EventToCarrier function which reads the XRay trace
 // information from the environment and returns this information in
 // a propagation.HeaderCarrier
-func EventToCarrier() otellambda.Option {
+func WithEventToCarrier() otellambda.Option {
 	return otellambda.WithEventToCarrier(xrayEventToCarrier)
 }
 
 // Propagator returns an otellambda.Option to enable the xray.Propagator
```
Aneurysm9, line 69:
```suggestion
// WithPropagator returns an otellambda.Option to enable the xray.Propagator
```

Aneurysm9:
The failing unit test is a result of no longer trying to yield to the go runtime before invoking `ForceFlush()`.  Once https://github.com/open-telemetry/opentelemetry-go/pull/2335 is available in a release then the test should start passing and this PR can proceed.

bhautikpip:
@MrAlias @dashpole Can we review this change? So that when we release the core this should be ready to merge.

Aneurysm9 (approved):

alolita:
@open-telemetry/proto-go-approvers @dashpole can you please provide a second review so that this PR can be merged. We have a downstream dependency on this. Many thanks in advance.

instrumentation/github.com/aws/aws-lambda-go/otellambda/xrayconfig/xrayconfig.go
```diff
@@ -35,72 +34,45 @@ func xrayEventToCarrier([]byte) propagation.TextMapCarrier {
...
-	// Do not need transport security in Lambda because collector
-	// runs locally in Lambda execution environment
+// NewTracerProvider returns a TracerProvider configured with an exporter,
+// ID generator, and lambda resource detector to send trace data to AWS X-Ray
+// via a Collector instance listening on localhost
+func NewTracerProvider(ctx context.Context) (*sdktrace.TracerProvider, error) {
 	exp, err := otlptracegrpc.New(ctx, otlptracegrpc.WithInsecure())
 	if err != nil {
-		return []otellambda.Option{}, err
+		errorLogger.Println("failed to create exporter: ", err)
```
dashpole, line 43: We probably don't need to log here.  Since we return the error, we expect the caller to handle it (and log it if they would like).  Same applies below and elsewhere.

instrumentation/github.com/aws/aws-lambda-go/otellambda/xrayconfig/xrayconfig_test.go
```diff
@@ -154,6 +154,9 @@ func assertSpanEqualsIgnoreTimeAndSpanID(t *testing.T, expected *v1trace.Resourc
 func TestWrapEndToEnd(t *testing.T) {
 	setEnvVars()
 
+	ctx := context.Background()
+	tp, _ := NewTracerProvider(ctx)
```
dashpole, line 158: Probably best to check that the err is non-nil here.

dashpole (approved):
Just nits.  Looks good overall

instrumentation/github.com/aws/aws-lambda-go/otellambda/xrayconfig/xrayconfig_test.go
```diff
@@ -154,6 +154,12 @@ func assertSpanEqualsIgnoreTimeAndSpanID(t *testing.T, expected *v1trace.Resourc
 func TestWrapEndToEnd(t *testing.T) {
 	setEnvVars()
 
+	ctx := context.Background()
+	tp, err := NewTracerProvider(ctx)
+	if err != nil {
+		return
+	}
```
Aneurysm9, line 161:
```suggestion
	assert.NoError(t, err)
```

PR #1536, opened 2021-12-21
Remote Sampling - XRay (Part 1) (Fetching Sampling Rules)

bhautikpip:
This is the intial PR for X-Ray remote sampling. The scope of this PR is to introduce data model for the business logic of remote sampling, adding support for fetching sampling rules and stroring.

**Upcoming PRs just a heads up**

1. Support for getting sampling targets from AWS X-Ray + Business logic for centralized reservoir + fallback sampling (traceIDRationBasedSampler)
2. Validate rules are being matched using matching logic


bhautikpip:
@Aneurysm9 @dashpole @MrAlias @jmacd @XSAM and other maintainers feel free to review this PR! The main reason behind breaking up the PRs to kind of make review easier. :)

bhautikpip:
Also, heads up I will be out of the office until Jan 28 but will be around to address comments so feel free to post any comments.

samplers/aws/go.mod
```diff
@@ -0,0 +1,8 @@
+module go.opentelemetry.io/contrib/samplers/aws
+
+go 1.16
+
+require (
+	github.com/stretchr/testify v1.7.0
+	go.opentelemetry.io/otel/sdk v1.2.0
```
Aneurysm9, line 7:
```suggestion
	go.opentelemetry.io/otel/sdk v1.3.0
```

Aneurysm9 (commented):
This looks reasonable overall, but needs significantly better test coverage.  Much of this is not at all tested.

bhautikpip:
> This looks reasonable overall, but needs significantly better test coverage. Much of this is not at all tested.

Okay! will add more tests to make coverage better!

samplers/aws/xray/client.go
```diff
@@ -0,0 +1,76 @@
...
+	"encoding/json"
+	"fmt"
+	"net/http"
+)
+
+type xrayClient struct {
+	// http client for sending unsigned proxied requests to the collector
+	httpClient *http.Client
+
+	proxyEndpoint string
```
anuraaga, line 29: It's not really a proxy in the HTTP proxy sense of the word, just `endpoint` is clearer I think

samplers/aws/xray/client.go
```diff
@@ -0,0 +1,76 @@
...
+}
+
+// newClient returns a http client with proxy endpoint
+func newClient(d string) *xrayClient {
```
anuraaga, line 33: Surely just `d` is going too far even in Go :P

samplers/aws/xray/client.go
```diff
@@ -0,0 +1,76 @@
...
+	proxyEndpoint := "http://" + d
+
+	p := &xrayClient{
```
anuraaga, line 36: Maybe just `return` this to avoid a comment about the letter `p`

samplers/aws/xray/remote_sampler_config.go
```diff
@@ -0,0 +1,71 @@
...
+	defaultPollingInterval = 300
+)
+
+// SamplerOption is a function that sets config on the sampler
+type Option func(options *config)
+
+type config struct {
+	proxyEndpoint                string
+	samplingRulesPollingInterval time.Duration
+	logger                       Logger
```
anuraaga, line 32: I see most usages using `log.Logger` though some `logr.Logger` too @Aneurysm9 is there a pattern for logs?

https://github.com/open-telemetry/opentelemetry-go/search?q=logger

samplers/aws/xray/client.go
```diff
@@ -0,0 +1,76 @@
...
+	}
+
+	return p
+}
+
+// getSamplingRules calls the collector(aws proxy enabled) for sampling rules
+func (p *xrayClient) getSamplingRules(ctx context.Context) (*getSamplingRulesOutput, error) {
+	req, err := http.NewRequestWithContext(ctx, http.MethodPost, p.proxyEndpoint+"/GetSamplingRules", nil)
+	if err != nil {
+		globalLogger.Printf("failed to create http request, %v\n", err)
```
anuraaga, line 48: Since we're returning the error, probably don't need this log message?

samplers/aws/xray/remote_sampler.go
```diff
@@ -0,0 +1,223 @@
...
+}
+
+func (rs *RemoteSampler) start(ctx context.Context) {
+	if !rs.pollerStart {
+		rs.pollerStart = true
+		rs.startRulePoller(ctx)
+	}
+}
+
+func (rs *RemoteSampler) startRulePoller(ctx context.Context) {
```
anuraaga, line 108: We should only have one poller for rules and targets, notably there should only be one goroutine that is fetching rules or targets based on their appropriate tick. This removes needing to think about any threadsafety issues.

samplers/aws/xray/rand.go
```diff
@@ -0,0 +1,114 @@
...
+	"encoding/binary"
+	"math/rand"
+	"sync"
+	"time"
+)
+
+var _ rand.Source = (*lockedSource)(nil)
+var _ rand.Source64 = (*lockedSource64)(nil)
+
+type lockedSource struct {
```
anuraaga, line 28: I am surprised we need an entire locked random to compute jitter. I don't see anything like it in this other library, do we really need this?

https://github.com/lthibault/jitterbug

samplers/aws/xray/sampling_rule.go
```diff
@@ -0,0 +1,79 @@
...
+// See the License for the specific language governing permissions and
+// limitations under the License.
+
+package xray
+
+import "sync"
+
+// ToDo: other fields will be used in business logic for remote sampling
+// centralizedRule represents a centralized sampling rule
+type centralizedRule struct {
```
anuraaga, line 21: The OTel version won't have centralized and local in the same sampler, so we should remove the word centralized from this codebase

samplers/aws/xray/sampling_rule.go
```diff
@@ -0,0 +1,79 @@
...
+// getSamplingRulesOutput is used to store parsed json sampling rules
+type getSamplingRulesOutput struct {
+	SamplingRuleRecords []*samplingRuleRecords `json:"SamplingRuleRecords"`
+}
+
+// updateRule updates the properties of the user-defined and default centralizedRule using the given
+// *properties.
+func (r *centralizedRule) updateRule(rule *ruleProperties) {
+	r.mu.Lock()
+	defer r.mu.Unlock()
```
anuraaga, line 75: I didn't look in detail but noticed there are many locks / unlocks throughout the code. This makes me suspect a design issue - we poll for a complete sampling configuration and can create it each time, so in principle you should only need one synchronized update to all the sampling state rather than updating each individual portion behind locks. It should make the code easier to reason about.

samplers/aws/xray/client.go
```diff
@@ -0,0 +1,76 @@
...
+		httpClient:    &http.Client{},
+		proxyEndpoint: proxyEndpoint,
+	}
+
+	return p
+}
+
+// getSamplingRules calls the collector(aws proxy enabled) for sampling rules
+func (p *xrayClient) getSamplingRules(ctx context.Context) (*getSamplingRulesOutput, error) {
+	req, err := http.NewRequestWithContext(ctx, http.MethodPost, p.proxyEndpoint+"/GetSamplingRules", nil)
```
anuraaga, line 46: I'm amazed that the API actually works with a nil body even though it defines one...

https://docs.aws.amazon.com/xray/latest/api/API_GetSamplingRules.html

I'd still recommend defining the struct, even if we never actually populate NextToken, to be consistent with that API spec, prevent possible confusion in the future, and be able to extend if needing that functionality later.

anuraaga (commented):
If it's possible, I would recommend a further split, first PR would just have the xrayClient

samplers/aws/xray/remote_sampler_config.go
bhautikpip: There's not standardized logging in go as of yet and we have tried to provide custom logging implementation here so that users can add their own logging logic.

samplers/aws/xray/client.go
bhautikpip: I have tried to add logging as much as possible mostly because this would help us in debugging. This logging lines would not flood customers since this would be a no-op logging for most cases unless users enable this.

samplers/aws/xray/remote_sampler.go
bhautikpip: Not sure about that because we typically refresh sampling rules every 5 mins and sampling targets every 10 seconds.

samplers/aws/xray/rand.go
bhautikpip: I have tried to add all the methods for this. Not sure using this lib is a good idea since we have like custom implementations. @Aneurysm9 any thoughts?

samplers/aws/xray/sampling_rule.go
bhautikpip: I think `centralized` here is not used to differentiate the `local` and `centralized` sampler. Since the rules are retrieved from remote centralized location I thought it would make sense to use `centralizedRule`.

samplers/aws/xray/remote_sampler_config.go
Aneurysm9: There is a facility for providing a `logr.Logger` that the SDK can use for internal logging, but there is as yet no mechanism for instrumentations or SDK extensions to access that instance.  It would probably be best here to also accept a `logr.Logger`, which provides a consistent interface and choice regarding underlying implementation.

samplers/aws/xray/sampling_rule.go
bhautikpip: I think the major intent of locks throughout the code is to make sure atomic access. Yeah We are writing/updating these values when we do synchronize update at every interval but at the same time we are also reading these values. So locks at each point would make these accesses safer I think.

bhautikpip:
> If it's possible, I would recommend a further split, first PR would just have the xrayClient

Thanks for your review :) It's hard to break down these PRs further. I have tried to abstract core logic as much possible since this PR would add all the data model and the next PR would have all the core logic.

samplers/aws/xray/rand.go
Aneurysm9: Jitterbug appears to depend on a `rand.Source` implementation, which is how it ignores concurrency issues.  It pushes that off to the source.  The [default source](https://pkg.go.dev/math/rand#pkg-overview) used by the package-global functions in the `math/rand` package is safe for concurrent use, but sources created with `NewSource()` are not and need locking like this to be made safe for concurrent use.  

I suppose the question here is whether the default source can be used.

samplers/aws/xray/remote_sampler_config.go
bhautikpip: Okay here I will remove the custom logging and then only accepts `logr.Logger` because I think accepting both logger implementation does not makes sense since `log4.Logger` will include all utilities.

samplers/aws/xray/rand.go
bhautikpip: Okay then if `jitterbug` is safe to use considering concurrency then show we remove this logic?

samplers/aws/xray/client.go
anuraaga: I mean won't the caller log the error? We shouldn't double log

samplers/aws/xray/rand.go
anuraaga: Ah I was referencing jitterbug just to look at a different implementation which looked much simpler. I don't think we need to use it as a dependency but if the default source can work then we should just use that I guess

samplers/aws/xray/remote_sampler.go
anuraaga: Yeah the same goroutine should be polling rules every 5 min *or* polling targets every 10s, but you shouldn't have two goroutines that may execute at the same time.

samplers/aws/xray/sampling_rule.go
anuraaga: Yeah I'm saying you should be able to atomically refresh all the state in one operation rather than processing each rule individually. Don't think of it as fetching individual rules, fetch the entire manifest, build a new manifest, and then swap it in

samplers/aws/xray/sampling_rule.go
anuraaga: If you have an adjective like centralized, you would expect to have a differen tadjective too like non-centralized, otherwise it adds some noise and some confusion. These are all just rules, reservoirs used by the xray sampler so we should be better off removing those

samplers/aws/xray/remote_sampler.go
bhautikpip: It won't execute at a same time and even if it will then those would be 2 separate processes right? I don't want to overburden the `getSamplingRules` and `getSamplingTargets` logic in a single goroutine. I think this would look much cleaner. Also, I'm not sure if we can combine those in a single goroutine programmatically as we would require both the logic to run in an infinite for loop.

anuraaga: > It won't execute at a same time

It will because the timing will overlap sometimes.

The logic including any potential cleanliness doesn't change that much I think

```
for {
  select {
    case <-rulesTicker.C():
      fetchRules()
    case <-targetsTicker.C()
      fetchTargets()
  }
}
```

This ensures the pollers are serialized onto a single thread to avoid simultaneously updating.

That being said, another thing to keep in mind is that targets fetch isn't a fixed tick, it is a value returned in the actual targets response

https://docs.aws.amazon.com/xray/latest/api/API_SamplingTargetDocument.html#xray-Type-SamplingTargetDocument-Interval

So presumably you can't use a ticker for the targets fetch, I'm not sure if there is any better pattern in Go but I guess it may involve a goroutine that sleeps and signals a channel

```
for {
  case := <-rulesTicker.C()
    fetchRules()
  case <-targetsChan
    fetchTargets()
}

fetchTargets() {
  ...
  nextFetch = min(rules.intervals)
  s.manifest = newManifest
  go func() {
    time.Sleep(nextFetch)
    targetsChan <- struct{}
}
```

samplers/aws/xray/rand.go
Aneurysm9: Using the default source in a library is generally avoided because it needs to be seeded to be effectively used in most cases but, as a library author, you shouldn't be making the choice to seed it for an application that might not want it seeded and you also can't guarantee that another library doesn't come along later and re-seed it with a fixed value.  I do really wish that new sources were also safe for concurrent access.

Can the problem be avoided by simply having more sources?  Can we have each actor with its own source and thus avoid concurrency concerns?

samplers/aws/xray/client.go
```diff
@@ -0,0 +1,79 @@
...
+	"encoding/json"
+	"fmt"
+	"net/http"
+)
+
+type xrayClient struct {
+	// http client for sending unsigned proxied requests to the collector
+	httpClient *http.Client
+
+	endpoint string
```
MrAlias, line 29: There is a lot of manual URL manipulation in this code. Using a dedicated and tested type here will help prevent bugs and provide needed untilities.

```suggestion
	endpoint *url.URL
```

samplers/aws/xray/client.go
```diff
@@ -0,0 +1,79 @@
...
+	// Unmarshalling json data to populate getSamplingTargetsOutput struct
+	var samplingRulesOutput getSamplingRulesOutput
+	err = json.Unmarshal(buf.Bytes(), &samplingRulesOutput)
+	if err != nil {
+		return nil, fmt.Errorf("xray client: unable to unmarshal the response body: %w", err)
+	}
+
+	err = output.Body.Close()
+	if err != nil {
+		globalLogger.Error(err, "failed to close http response body")
```
MrAlias, line 75: Return this error instead of logging it.

samplers/aws/xray/client.go
```diff
@@ -0,0 +1,79 @@
...
+	if err != nil {
+		return nil, fmt.Errorf("xray client: unable to read response body: %w", err)
+	}
+
+	// Unmarshalling json data to populate getSamplingTargetsOutput struct
+	var samplingRulesOutput getSamplingRulesOutput
+	err = json.Unmarshal(buf.Bytes(), &samplingRulesOutput)
+	if err != nil {
+		return nil, fmt.Errorf("xray client: unable to unmarshal the response body: %w", err)
+	}
```
MrAlias, line 71:
```suggestion
	var samplingRulesOutput getSamplingRulesOutput
	if err := json.NewDecoder(output.Body).Decode(&samplingRulesOutput); err != nil {
		return nil, fmt.Errorf("xray client: unable to unmarshal the response body: %w", err)
	}
```

samplers/aws/xray/client.go
```diff
@@ -0,0 +1,79 @@
...
+
+	err = output.Body.Close()
```
MrAlias, line 73: This should be defered right after the response is received to avoid returning without calling this and creating a memorry leak.

samplers/aws/xray/client.go
```diff
@@ -0,0 +1,79 @@
...
+	if err != nil {
+		globalLogger.Error(err, "failed to close http response body")
+	}
+
+	return &samplingRulesOutput, nil
```
MrAlias, line 78: The `samplingRulesOutput` value is never used, only a pointer to it. Declare a pointer value instead to avoid all the explicit indirection.

samplers/aws/xray/clock.go
```diff
@@ -0,0 +1,32 @@
...
+	Now() time.Time
+}
+
+// DefaultClock is an implementation of Clock interface.
+type DefaultClock struct{}
+
+// Now returns current time.
+func (t *DefaultClock) Now() time.Time {
+	return time.Now()
+}
```
MrAlias, line 32: These look to only exist to allow testing overrides. They do not look needed by the user. Please unexport them or move to an internal package.

samplers/aws/xray/client_test.go
```diff
@@ -0,0 +1,59 @@
...
+	ctx := context.Background()
+
+	// generate a test server so we can capture and inspect the request
+	testServer := httptest.NewServer(http.HandlerFunc(func(res http.ResponseWriter, req *http.Request) {
+		_, _ = res.Write([]byte(body))
+	}))
+
+	u, _ := url.Parse(testServer.URL)
```
MrAlias, line 35:
```suggestion
	u, err := url.Parse(testServer.URL)
	require.NoError(t, err)
```

samplers/aws/xray/client_test.go
```diff
@@ -0,0 +1,59 @@
...
+	ctx := context.Background()
+
+	// generate a test server so we can capture and inspect the request
+	testServer := httptest.NewServer(http.HandlerFunc(func(res http.ResponseWriter, req *http.Request) {
+		_, _ = res.Write([]byte(body))
```
MrAlias, line 32:
```suggestion
		_, err := res.Write([]byte(body))
		require.NoError(t, err)
```

MrAlias, line 32: Can we make this easier to read and use a `[]byte{}` instead of string which is converted?

```suggestion
	body := []byte(`{
  "NextToken": null,
  "SamplingRuleRecords": [
    {
      "CreatedAt": 0,
      "ModifiedAt": 1639517389,
      "SamplingRule": {
        "Attributes": {},
        "FixedRate": 0.5,
        "HTTPMethod": "*",
        "Host": "*",
        "Priority": 10000,
        "ReservoirSize": 60,
        "ResourceARN": "*",
        "RuleARN": "arn:aws:xray:us-west-2:836082170990:sampling-rule/Default",
        "RuleName": "Default",
        "ServiceName": "*",
        "ServiceType": "*",
        "URLPath": "*",
        "Version": 1
      }
    },
    {
      "CreatedAt": 1637691613,
      "ModifiedAt": 1643748669,
      "SamplingRule": {
        "Attributes": {},
        "FixedRate": 0.09,
        "HTTPMethod": "GET",
        "Host": "*",
        "Priority": 1,
        "ReservoirSize": 3,
        "ResourceARN": "*",
        "RuleARN": "arn:aws:xray:us-west-2:836082170990:sampling-rule/test-rule",
        "RuleName": "test-rule",
        "ServiceName": "test-rule",
        "ServiceType": "local",
        "URLPath": "/aws-sdk-call",
        "Version": 1
      }
    },
    {
      "CreatedAt": 1639446197,
      "ModifiedAt": 1639446197,
      "SamplingRule": {
        "Attributes": {},
        "FixedRate": 0.09,
        "HTTPMethod": "*",
        "Host": "*",
        "Priority": 100,
        "ReservoirSize": 100,
        "ResourceARN": "*",
        "RuleARN": "arn:aws:xray:us-west-2:836082170990:sampling-rule/test-rule-1",
        "RuleName": "test-rule-1",
        "ServiceName": "*",
        "ServiceType": "*",
        "URLPath": "*",
        "Version": 1
      }
    }
  ]
}`)
	ctx := context.Background()

	// generate a test server so we can capture and inspect the request
	testServer := httptest.NewServer(http.HandlerFunc(func(res http.ResponseWriter, req *http.Request) {
		_, _ = res.Write(body)
```

samplers/aws/xray/client_test.go
```diff
@@ -0,0 +1,59 @@
...
+	}))
+
+	u, _ := url.Parse(testServer.URL)
+	client := newClient(u.Host)
+
+	samplingRules, _ := client.getSamplingRules(ctx)
```
MrAlias, line 38:
```suggestion
	samplingRules, err := client.getSamplingRules(ctx)
	require.NoError(t, err)
```

samplers/aws/xray/rand.go
```diff
@@ -0,0 +1,114 @@
...
+// Rand is an interface for a set of methods that return random value.
+type Rand interface {
+	Int63n(n int64) int64
+	Intn(n int) int
+	Float64() float64
+}
+
+// DefaultRand is an implementation of Rand interface.
+// It is safe for concurrent use by multiple goroutines.
+type DefaultRand struct{}
```
MrAlias, line 94: These do not appear to need to be exported. Can they be unexported.

samplers/aws/xray/remote_sampler.go
```diff
@@ -0,0 +1,224 @@
...
+	"github.com/go-logr/logr"
+	"time"
+
+	sdktrace "go.opentelemetry.io/otel/sdk/trace"
+)
+
+// global variable for logging
+var globalLogger logr.Logger
+
+// RemoteSampler is an implementation of SamplingStrategy.
```
MrAlias, line 31: What is a SamplingStrategy? Can you update this documentation to explain what the `RemoteSampler` represents or does instead of saying only what it implements?

samplers/aws/xray/remote_sampler.go
```diff
@@ -0,0 +1,224 @@
...
+type RemoteSampler struct {
+	// List of known centralized sampling rules
+	manifest *centralizedManifest
+
+	// proxy is used for getting quotas and sampling rules
```
MrAlias, line 36: This comment doesn't match the code. Stale?

samplers/aws/xray/remote_sampler.go
```diff
@@ -0,0 +1,224 @@
...
+
+	sdktrace "go.opentelemetry.io/otel/sdk/trace"
+)
+
+// global variable for logging
+var globalLogger logr.Logger
+
+// RemoteSampler is an implementation of SamplingStrategy.
+type RemoteSampler struct {
+	// List of known centralized sampling rules
```
MrAlias, line 33:
```suggestion
	// List of known centralized sampling rules.
```

samplers/aws/xray/remote_sampler.go
```diff
@@ -0,0 +1,224 @@
...
+	manifest *centralizedManifest
+
+	// proxy is used for getting quotas and sampling rules
+	xrayClient *xrayClient
+
+	// pollerStart, if true represents rule and target pollers are started
```
MrAlias, line 39:
```suggestion
	// pollerStart, if true represents rule and target pollers are started.
```

samplers/aws/xray/remote_sampler.go
```diff
@@ -0,0 +1,224 @@
...
+	pollerStart bool
+
+	// samplingRules polling interval, default is 300 seconds
```
MrAlias, line 42:
```suggestion
	// samplingRulesPollingInterval polling interval, default is 300 seconds
```

samplers/aws/xray/remote_sampler.go
```diff
@@ -0,0 +1,224 @@
...
+	clientID string
+
+	// Provides system time
+	clock Clock
+}
+
+// Compile time assertion that remoteSampler implements the Sampler interface.
+var _ sdktrace.Sampler = (*RemoteSampler)(nil)
+
+// NewRemoteSampler returns a centralizedSampler which decides to sample a given request or not.
```
MrAlias, line 55: Please do not include unexported types in exported type documentation.

Can you update this to describe to a user with no context what the returned type does (i.e. how it samples)?

samplers/aws/xray/remote_sampler.go
```diff
@@ -0,0 +1,224 @@
...
+	"time"
+
+	sdktrace "go.opentelemetry.io/otel/sdk/trace"
+)
+
+// global variable for logging
+var globalLogger logr.Logger
+
+// RemoteSampler is an implementation of SamplingStrategy.
+type RemoteSampler struct {
```
MrAlias, line 32: Is an empty `RemoteSampler` a valid sampler? Should this be unexported and the `New*` function updated to return an `sdktrace.Sampler`?

samplers/aws/xray/remote_sampler.go
```diff
@@ -0,0 +1,224 @@
...
+	clock := &DefaultClock{}
+
+	m := &centralizedManifest{
+		rules: []*centralizedRule{},
+		index: map[string]*centralizedRule{},
+		clock: clock,
+	}
+
+	remoteSampler := &RemoteSampler{
+		pollerStart:                  false,
```
MrAlias, line 78:
```suggestion
```

samplers/aws/xray/remote_sampler.go
```diff
@@ -0,0 +1,224 @@
...
+	// starts the rule poller
+	remoteSampler.start(ctx)
+
+	return remoteSampler, nil
+}
+
+func (rs *RemoteSampler) ShouldSample(parameters sdktrace.SamplingParameters) sdktrace.SamplingResult {
+	// ToDo: add business logic for remote sampling
+
+	return sdktrace.SamplingResult{}
```
MrAlias, line 95: If the eventual plan is to default to parent based sample, please update this to return that now.

samplers/aws/xray/remote_sampler.go
```diff
@@ -0,0 +1,224 @@
...
+}
+
+func (rs *RemoteSampler) refreshManifest(ctx context.Context) (err error) {
+	// Explicitly recover from panics since this is the entry point for a long-running goroutine
+	// and we can not allow a panic to propagate to the application code.
+	defer func() {
+		if r := recover(); r != nil {
+			// Resort to bring rules array into consistent state.
+			//cs.manifest.sort()
+
```
MrAlias, line 141:
```suggestion
```

MrAlias (commented):
This is only a partial review.

Please address build and lint errors in addition to the comments made.

samplers/aws/xray/remote_sampler.go
```diff
@@ -209,14 +201,15 @@ func (rs *RemoteSampler) refreshManifest(ctx context.Context) (err error) {
...
-	// Prune inactive rules
-	rs.manifest.prune(actives)
+	// Re-sort to fix matching priorities.
+	tempManifest.sort()
 
-	// Re-sort to fix matching priorities
-	rs.manifest.sort()
+	// assign temp manifest to original copy/one sync refresh.
+	rs.manifest.mu.Lock()
+	rs.manifest.rules = tempManifest.rules
```
anuraaga, line 209: This should look something like `rs.manifest = tempManifest` rather than assigning to multiple fields, probably by moving the lock up to the sampler instead of the manifest.

Makefile
```diff
@@ -139,14 +139,14 @@ test-short:
...
-	done
-	$(TOOLS_DIR)/misspell -w $(ALL_DOCS)
+# lint: $(TOOLS_DIR)/golangci-lint $(TOOLS_DIR)/misspell lint-modules
+# 	set -e; for dir in $(ALL_GO_MOD_DIRS); do \
+# 	  echo "golangci-lint in $${dir}"; \
+# 	  (cd "$${dir}" && \
+# 	    $(TOOLS_DIR)/golangci-lint run --fix && \
+# 	    $(TOOLS_DIR)/golangci-lint run); \
+# 	done
+# 	$(TOOLS_DIR)/misspell -w $(ALL_DOCS)
```
MrAlias, line 149: This is not a viable approach to fixing lint error. Please revert this and address any outstanding issues.

MrAlias (changes requested):

Makefile
bhautikpip: This was intended for local testing. I forgot to uncomment it. I will uncomment it now. Thanks!

samplers/aws/xray/client.go
```diff
@@ -0,0 +1,73 @@
...
+	"bytes"
+	"context"
+	"encoding/json"
+	"fmt"
+	"net/http"
+	"net/url"
+)
+
+type xrayClient struct {
+	// http client for sending unsigned proxied requests to the collector
```
anuraaga, line 27:
```suggestion
	// http client for sending sampling requests to the collector
```

samplers/aws/xray/client.go
```diff
@@ -0,0 +1,73 @@
...
+	httpClient *http.Client
+
+	endpoint *url.URL
+}
+
+// newClient returns an HTTP client with proxy endpoint
+func newClient(d string) *xrayClient {
```
anuraaga, line 34:
```suggestion
func newClient(addr string) *xrayClient {
```

samplers/aws/xray/client.go
```diff
@@ -0,0 +1,73 @@
...
+	endpoint := "http://" + d
```
anuraaga, line 35: This should probably check if the address already has a scheme. If that should be strictly forbidden, then there should be some validation closer to the user I guess.

Note that while not the first version, eventually this sampler needs to be able to support more configuration, for example TLS or auth headers, as users don't necessarily run collectors as plaintext sidecars.

samplers/aws/xray/client.go
```diff
@@ -0,0 +1,73 @@
...
+
+	endpointURL, err := url.Parse(endpoint)
+	if err != nil {
+		globalLogger.Error(err, "unable to parse endpoint from string")
```
anuraaga, line 39: `newClient` needs to return `error` this is a fatal issue

samplers/aws/xray/client.go
```diff
@@ -0,0 +1,73 @@
...
+
+	return &xrayClient{
+		httpClient: &http.Client{},
+		endpoint:   endpointURL,
+	}
+}
+
+// getSamplingRules calls the collector(aws proxy enabled) for sampling rules
+func (p *xrayClient) getSamplingRules(ctx context.Context) (*getSamplingRulesOutput, error) {
+	statisticsByte, err := json.Marshal(getSamplingRulesInput{})
```
anuraaga, line 50: statistics?

samplers/aws/xray/client.go
```diff
@@ -0,0 +1,73 @@
...
+	if err != nil {
+		return nil, err
+	}
+	body := bytes.NewReader(statisticsByte)
+
+	req, err := http.NewRequestWithContext(ctx, http.MethodPost, p.endpoint.String()+"/GetSamplingRules", body)
```
anuraaga, line 56: Store `p.endpoint.String()+/GetSamplingRules` (and targets later) in the struct rather than the endpoint to avoid recomputing this every time

samplers/aws/xray/client.go
```diff
@@ -0,0 +1,73 @@
...
+	}
+
+	return &xrayClient{
+		httpClient: &http.Client{},
+		endpoint:   endpointURL,
+	}
+}
+
+// getSamplingRules calls the collector(aws proxy enabled) for sampling rules
+func (p *xrayClient) getSamplingRules(ctx context.Context) (*getSamplingRulesOutput, error) {
```
anuraaga, line 49: Nit `c` or `cl` or `client` seems more appropriate than `p`

samplers/aws/xray/client.go
```diff
@@ -0,0 +1,73 @@
...
+	}
+
+	output, err := p.httpClient.Do(req)
+	if err != nil {
+		return nil, fmt.Errorf("xray client: unable to retrieve sampling settings: %w", err)
+	}
+	defer output.Body.Close()
+
+	var samplingRulesOutput *getSamplingRulesOutput
+	if err := json.NewDecoder(output.Body).Decode(&samplingRulesOutput); err != nil {
```
anuraaga, line 68: This seems to be passing a pointer to a pointer - I'm surprised this seems to actually work , but think you don't need `&`, alternatively you might not need to return a pointer here since it should get moved to the caller

samplers/aws/xray/remote_sampler.go
```diff
@@ -0,0 +1,219 @@
...
+				return
+			}
+		}
+	}()
+}
+
+func (rs *remoteSampler) refreshManifest(ctx context.Context) (err error) {
+	// Compute 'now' before calling GetSamplingRules to avoid marking manifest as
+	// fresher than it actually is.
+	now := rs.clock.now().Unix()
```
anuraaga, line 144: Store `time.Time`, don't use `Unix()` anywhere here since we want to use monotonic time

samplers/aws/xray/rand.go
```diff
@@ -0,0 +1,114 @@
...
+	"math/rand"
+	"sync"
+	"time"
+)
+
+var _ rand.Source = (*lockedSource)(nil)
+var _ rand.Source64 = (*lockedSource64)(nil)
+
+type lockedSource struct {
+	mu  sync.Mutex
```
anuraaga, line 29: This conversation was marked resolved but I don't think it was - do we really need to implement this source? I guess @Aneurysm9's suggestion is having a new source per jittered ticker

samplers/aws/xray/remote_sampler.go
```diff
@@ -0,0 +1,219 @@
...
+	// pollerStart, if true represents rule and target pollers are started.
+	pollerStart bool
+
+	// samplingRulesPollingInterval, default is 300 seconds.
+	samplingRulesPollingInterval time.Duration
+
+	// Unique ID used by XRay service to identify this client.
+	clientID string
+
+	// Provides system time.
```
anuraaga, line 51:
```suggestion
	// Provides time.
```

Not sure but I think system time usually implies a non-monotonic time

samplers/aws/xray/remote_sampler.go
```diff
@@ -0,0 +1,219 @@
...
+	sdktrace "go.opentelemetry.io/otel/sdk/trace"
+)
+
+// global variable for logging
+var globalLogger logr.Logger
+
+// remoteSampler is a sampler for AWS X-Ray which polls sampling rules and sampling targets
+// to make a sampling decision based on rules set by users on AWS X-Ray console
+type remoteSampler struct {
+	// List of known centralized sampling rules.
```
anuraaga, line 36:
```suggestion
	// manifest is the list of known centralized sampling rules.
```

samplers/aws/xray/remote_sampler.go
```diff
@@ -0,0 +1,219 @@
...
+	manifest *manifest
+
+	// xrayClient is used for getting quotas and sampling rules.
+	xrayClient *xrayClient
+
+	// pollerStart, if true represents rule and target pollers are started.
+	pollerStart bool
```
anuraaga, line 43: `pollerStarted`, `active`, `polling`

samplers/aws/xray/remote_sampler.go
```diff
@@ -0,0 +1,219 @@
...
+	// starts the rule poller
+	remoteSampler.start(ctx)
+
+	return remoteSampler, nil
+}
+
+func (rs *remoteSampler) ShouldSample(parameters sdktrace.SamplingParameters) sdktrace.SamplingResult {
+	// ToDo: add business logic for remote sampling
+
+	return sdktrace.TraceIDRatioBased(0.05).ShouldSample(parameters)
```
anuraaga, line 101: You'll need the configurable "initial sampler" even after implementing logic, I think it's appropriate for this PR

samplers/aws/xray/remote_sampler.go
```diff
@@ -0,0 +1,219 @@
...
+}
+
+func (rs *remoteSampler) Description() string {
+	return "remote sampling with AWS X-Ray"
```
anuraaga, line 105: This needs to be the same across languages and a programmatic string I think. I don't think we've spec'd this well yet but can probably be a bit closer to this for now

https://github.com/open-telemetry/opentelemetry-java-contrib/blob/main/aws-xray/src/main/java/io/opentelemetry/contrib/awsxray/AwsXrayRemoteSampler.java#L116

samplers/aws/xray/remote_sampler.go
```diff
@@ -0,0 +1,219 @@
...
+	go func() {
+		// Period = 300s, Jitter = 5s
+		t := newTicker(rs.samplingRulesPollingInterval, 5*time.Second)
+
+		// Periodic manifest refresh
+		for {
+			if err := rs.refreshManifest(ctx); err != nil {
+				globalLogger.Error(err, "Error occurred while refreshing sampling rules")
+			} else {
+				globalLogger.Info("Successfully fetched sampling rules")
```
anuraaga, line 126: I suspect this should be a lower level than Info (there's no V), @Aneurysm9?

samplers/aws/xray/remote_sampler.go
```diff
@@ -0,0 +1,219 @@
...
+			continue
+		}
+
+		if *records.SamplingRule.Version != int64(1) {
+			globalLogger.V(1).Info("Sampling rule without version 1 is not supported", "RuleName", *records.SamplingRule.RuleName)
+			failed = true
+			continue
+		}
+
+		if len(records.SamplingRule.Attributes) != 0 {
```
anuraaga, line 181: We support attributes in the OTel implementations

samplers/aws/xray/remote_sampler.go
```diff
@@ -0,0 +1,219 @@
...
+			globalLogger.V(1).Info("Sampling rule with non nil Attributes is not applicable", "RuleName", *records.SamplingRule.RuleName)
+			continue
+		}
+
+		if records.SamplingRule.ResourceARN == nil {
+			globalLogger.V(1).Info("Sampling rule without ResourceARN is not applicable", "RuleName", *records.SamplingRule.RuleName)
```
anuraaga, line 187: These resourceARN checks seem strange, if it's present we should use it in the same way as any other (not restricted to `*`) if not present we should ignore

samplers/aws/xray/remote_sampler.go
```diff
@@ -0,0 +1,219 @@
...
+		createErr := tempManifest.createRule(records.SamplingRule)
+		if createErr != nil {
+			failed = true
+			globalLogger.Error(createErr, "Error occurred creating/updating rule")
+		}
+	}
+
+	// Set err if updates failed
+	if failed {
+		err = errors.New("error occurred creating/updating rules")
```
anuraaga, line 206: Unused

samplers/aws/xray/remote_sampler.go
```diff
@@ -0,0 +1,219 @@
...
+	}
+
+	// Re-sort to fix matching priorities.
+	tempManifest.sort()
+	// Update refreshedAt timestamp
+	tempManifest.refreshedAt = now
+
+	// assign temp manifest to original copy/one sync refresh.
+	rs.mu.Lock()
```
anuraaga, line 215: This might be for another PR, but wondering if volatile write / read is an appropriate use for `atomic` in Go too, but not sure

samplers/aws/xray/remote_sampler_config.go
```diff
@@ -0,0 +1,78 @@
...
+
+// sets custom proxy endpoint
+func WithEndpoint(endpoint string) Option {
+	return func(o *config) {
+		o.endpoint = endpoint
+	}
+}
+
+// sets polling interval for sampling rules
+func WithSamplingRulesPollingInterval(polingInterval time.Duration) Option {
```
anuraaga, line 48:
```suggestion
func WithSamplingRulesPollingInterval(pollingInterval time.Duration) Option {
```

anuraaga: Validate it's positive

samplers/aws/xray/remote_sampler_config.go
```diff
@@ -0,0 +1,78 @@
...
+type Option func(options *config)
+
+type config struct {
+	endpoint                     string
+	samplingRulesPollingInterval time.Duration
+	logger                       logr.Logger
+}
+
+// sets custom proxy endpoint
+func WithEndpoint(endpoint string) Option {
```
anuraaga, line 41: Validate endpoint expectations

samplers/aws/xray/remote_sampler_config.go
```diff
@@ -0,0 +1,78 @@
...
+	}
+
+	for _, option := range opts {
+		option(cfg)
+	}
+
+	// setting global logger
+	globalLogger = cfg.logger
+	// set global verbosity to log info logs
+	stdr.SetVerbosity(1)
```
anuraaga, line 75: Don't modify global state in this library

samplers/aws/xray/sampling_rule.go
```diff
@@ -0,0 +1,72 @@
...
+// See the License for the specific language governing permissions and
+// limitations under the License.
+
+package xray
+
+// ToDo: other fields will be used in business logic for remote sampling
+// rule represents a centralized sampling rule
+type rule struct {
+	// Centralized reservoir for keeping track of reservoir usage
+	reservoir *reservoir
```
anuraaga, line 21: Probably should leave out reservoir from this PR

samplers/aws/xray/client.go
bhautikpip: I tried removing `&` and it throwed the unmarshal error.

samplers/aws/xray/rand.go
bhautikpip: Sure. will chat with @Aneurysm9 on best course of action for this.

samplers/aws/xray/remote_sampler.go
anuraaga: You just need a `WithInitialSampler` option and a field to store it in right? Not sure of the rearranging

samplers/aws/xray/remote_sampler_config.go
```diff
@@ -0,0 +1,112 @@
...
+		samplingRulesPollingInterval: defaultPollingInterval * time.Second,
+		logger:                       stdr.NewWithOptions(log.New(os.Stderr, "", log.LstdFlags|log.Lshortfile), stdr.Options{LogCaller: stdr.Error}),
+	}
+
+	for _, option := range opts {
+		option(cfg)
+	}
+
+	// setting global logger
+	globalLogger = cfg.logger
```
anuraaga, line 78: Just noticed this - we shouldn't have a globallogger, logger needs to be on the sampler object itself. This isn't a globally initialized library

samplers/aws/xray/client.go
anuraaga: This is a struct that won't be initialized many times so having 2 fields is basically no overhead, while computing a string on every poll is overhead.

samplers/aws/xray/client.go
jmacd: I believe json.Unmarshal() handles this correctly, but it would be nicer to see 
```
var samplingRulesOutput getSamplingRulesOutput
```
i.e., drop the `*`, not the `&`

samplers/aws/xray/reservoir.go
```diff
@@ -0,0 +1,43 @@
...
+// WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+// See the License for the specific language governing permissions and
+// limitations under the License.
+
+package xray
+
+// ToDo: other fields will be used in business logic for remote sampling
+// reservoir is a reservoir distributed among all running instances of the SDK
+type reservoir struct {
+	// Quota assigned to client
```
jmacd, line 20: (What's happening here, with all the commented-out fields?)

samplers/aws/xray/remote_sampler.go
jmacd: I agree, it makes sense to plug in an initial value even though this PR is already big, it's the natural default.

samplers/aws/xray/sampling_rule.go
```diff
@@ -0,0 +1,72 @@
...
+// distributed under the License is distributed on an "AS IS" BASIS,
+// WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+// See the License for the specific language governing permissions and
+// limitations under the License.
+
+package xray
+
+// ToDo: other fields will be used in business logic for remote sampling
+// rule represents a centralized sampling rule
+type rule struct {
```
jmacd, line 19: Will you place a reference to some documentation on the AWS side where `manifest` and `rule` are documented, please?

bhautikpip:
Closing this in favor of this complete PR for [remote sampling support](https://github.com/open-telemetry/opentelemetry-go-contrib/pull/1859)

Issue #1335, opened 2021-10-08
OTel HTTP server duration microseconds instead of milliseconds


victoraugustolls:
Hi! According to the [specs](https://github.com/open-telemetry/opentelemetry-specification/blob/main/specification/metrics/semantic_conventions/http-metrics.md), shouldn't `http.server.duration` be in milliseconds instead of microseconds?

If I understood correctly, the current `otelhttp` implementation is using `microseconds` instead (see [here](https://github.com/open-telemetry/opentelemetry-go-contrib/blob/1412a89570649ed3cea04d6b0598e913ac9018b8/instrumentation/net/http/otelhttp/common.go#L36)).

If that's the case, would it be ok for me to open a PR changing it?

Thanks in advance!

MrAlias:
Closed by https://github.com/open-telemetry/opentelemetry-go-contrib/pull/1537

PR #1537, opened 2021-12-21
Modified otel-http-server-duration from microseconds to milliseconds

rltoSD:
Description:
This PR is to modify otel http server duration from microseconds to milliseconds as per the specs mentioned in OpenTelemetry Go.

Link to tracking Issue:
issue #1335


Aneurysm9 (approved):

instrumentation/net/http/otelhttp/handler.go
```diff
@@ -195,7 +195,7 @@ func (h *Handler) ServeHTTP(w http.ResponseWriter, r *http.Request) {
 	h.counters[RequestContentLength].Add(ctx, bw.read, attributes...)
 	h.counters[ResponseContentLength].Add(ctx, rww.written, attributes...)
 
-	elapsedTime := time.Since(requestStartTime).Microseconds()
+	elapsedTime := float64(time.Since(requestStartTime)) / float64(time.Millisecond)
```
MrAlias, line 198:
```suggestion
	// Use floating point division here for higher precision (instead of Millisecond method).
	elapsedTime := float64(time.Since(requestStartTime)) / float64(time.Millisecond)
```

MrAlias (commented):
Please fix the build failure. I looks like an EOL char is needed in `handler.go`.

instrumentation/net/http/otelhttp/handler.go
```diff
@@ -232,4 +233,4 @@ func WithRouteTag(route string, h http.Handler) http.Handler {
 		span.SetAttributes(semconv.HTTPRouteKey.String(route))
 		h.ServeHTTP(w, r)
 	})
-}
+}
```
MrAlias, line 236: A new line character is needed at the end of the file.

rltoSD: Added it to the file thanks

dashpole (approved):

MrAlias (approved):

PR #1546, opened 2021-12-26
Extract region name from Cloud Run metadata server response

nikklassen:
Previously this was returning a full string like /projects/123/regions/r, which does not match up with what the semvconv.CloudRegionKey value is supposed to be.

Originally filed as GoogleCloudPlatform/opentelemetry-operations-go#227


Aneurysm9 (approved):

XSAM (approved):

dashpole (approved):
Thanks!

MrAlias (approved):

PR #1554, opened 2022-01-06
Remove the MinMaxSumCount from cortex and datadog exporter

hanyuancheung:
This is part of the #1478 todo list:
- [x] Remove the MinMaxSumCount from cortex.
- [ ] Remove Point aggregations from datadog exporter.
- [ ] Handle Histograms in the datadog exporter
- [ ] Remove Point aggregations from dogstatsd exporter.
- [ ] Handle Histograms in the dogstatsd exporter

Issue #1478, opened 2021-12-11
Update metrics exporters


MadVikingGod:
The metrics version V0.26.0 removes the Points aggregation and MinMaxSumCount aggregation from the API, and includes a new Histogram aggregation which is not interchangeable.  This prevented us from updating modules in the `experimental-metrics` module set.

Things that need to be addressed:

- [x] Remove the MinMaxSumCount from cortex.
- [ ] Remove Point aggregations from datadog exporter.
- [ ] Handle Histograms in the datadog exporter
- [ ] Remove Point aggregations from dogstatsd exporter.
- [ ] Handle Histograms in the dogstatsd exporter

hanyuancheung:
I think this issue has been finished, cause with the deprecation of the datadog exporter and removal of histogram support, we don't need to spend efforts trying to handle others. Could you please help to close it?

PR #1555, opened 2022-01-06
Remove point aggregations from Datadog Exporter

hanyuancheung:
Follow: #1554 
This is part of the #1478 todo list:
- [x] Remove the MinMaxSumCount from cortex.
- [x] Remove Point aggregations from datadog exporter.
- [ ] Handle Histograms in the datadog exporter
- [ ] Remove Point aggregations from dogstatsd exporter.
- [ ] Handle Histograms in the dogstatsd exporter


Aneurysm9:
I think this leaves this exporter without support for histograms.  I'd like to see that handled before moving forward.  It may be that the only reasonable way to handle it is to remove this exporter entirely, in which case this is wasted effort.

zhyChesterCheung:
> I think this leaves this exporter without support for histograms. I'd like to see that handled before moving forward. It may be that the only reasonable way to handle it is to remove this exporter entirely, in which case this is wasted effort.

Ok, I'll remove and handle the histogram type in a single PR at the same time. Thanks for the advice @Aneurysm9 .

MrAlias (approved):

Aneurysm9:
This PR appears to have no change to the target branch.  Please re-open, or open a new PR, with new commits if this is in error.

PR #1581, opened 2022-01-12
otelaws : Adding Dynamodb attributes

jennynilsen:
When appending the middlewares to a dynamodb client, no dynamodb attributes are added to the spans.

This PR :
- adds dynamodb attributes that correspond to the dynamodb operation to the spans
- allows someone to provide their own function to set custom attributes
- adds a way to determine the attributes to set based on service automatically


jennynilsen:
closed because of wrong email.

PR #1583, opened 2022-01-13
Handle Histograms in Datadog Exporter

hanyuancheung:
Follow: #1554 
This is part of the #1478 todo list:
- [x] Remove the MinMaxSumCount from cortex.
- [x] Remove Point aggregations from datadog exporter.
- [x] Handle Histograms in the datadog exporter
- [ ] Remove Point aggregations from dogstatsd exporter.
- [ ] Handle Histograms in the dogstatsd exporter


hanyuancheung:
PTAL, when Codecov scans, it warns there are several lines not covered. It bothers me how I could fix it, thanks for the help.

exporters/metric/datadog/datadog.go
```diff
@@ -109,19 +109,13 @@ func (e *Exporter) Export(ctx context.Context, res *resource.Resource, ilr expor
 				tags = append(tags, tag)
 			}
 			switch agg := agg.(type) {
-			case aggregation.Points:
-				numbers, err := agg.Points()
+			case aggregation.Histogram:
```
jmacd, line 112: If you want to correctly handle Histograms for DD exporters, we need a way to change the `Aggregator` used to restore the former `exact` support. This is possible, however we do see a lot of change about to happen in the Metrics SDK and I'm concerned it will be a waste of your time.

In the longer term, an OTel user might want to install a more direct shim to the DD library by implementing the metrics API directly -- that is also about to change. However, if you did that, you'd lose the Views facility that is under development for the default SDK, and I would recommend DataDog to accept OTLP at its agent instead. 😁 

If you did bring back the Exact aggregator, I might suggest calculating Summaries instead. You could look at how it's handled in the OTC Statsd receiver: https://github.com/open-telemetry/opentelemetry-collector-contrib/blob/main/receiver/statsdreceiver/protocol/metric_translator.go#L70 
Then, you'd have all the DD client's functionality except for its handling of histograms, which would appear as quantiles over short intervals, which roughly approximates DD's old handling of histograms (pre-Dogsketch).

MrAlias:
With the deprecation of the datadog exporter and removal of histogram support in https://github.com/open-telemetry/opentelemetry-go-contrib/pull/1846 I think we can close this. The project direction is to remove this exporter in a future release so I don't think we should spend effort here trying to add back this functionality.

hanyuancheung:
> With the deprecation of the datadog exporter and removal of histogram support in #1846 I think we can close this. The project direction is to remove this exporter in a future release so I don't think we should spend effort here trying to add back this functionality.

Get it, I’ll close it.

PR #1843, opened 2022-02-15
bugfix/otelhttp: set TextMapPropagator to server for examples

1046102779:
```shell
curl -i "http://127.0.0.1:8080/users/2" --header "traceparent: 00-5448d4e7e9479e15be2a48e224694ee8-de850fb435003d32-01" -XGET
```

before PR:
image

after PR:
image


pellared (approved):
side note:
I do not know if this should not be the default global propagator so that the user would not need to call add this line each time

reference: https://github.com/open-telemetry/opentelemetry-specification/blob/main/specification/sdk-environment-variables.md#general-sdk-configuration

1046102779:
> gator so tha

it means use `OTEL_PROPAGATORS ` environment variable?

Here the user can also use the delegate within the local scope of otelhttp, but it needs to be changed to the `otelrestful.OTelFilter` method and pass in the config, and input `otelrestful.WithPropagators` param.

> side note: I do not know if this should not be the default global propagator so that the user would not need to call add this line each time
> 
> reference: https://github.com/open-telemetry/opentelemetry-specification/blob/main/specification/sdk-environment-variables.md#general-sdk-configuration

1046102779:
> side note: I do not know if this should not be the default global propagator so that the user would not need to call add this line each time
> 
> reference: https://github.com/open-telemetry/opentelemetry-specification/blob/main/specification/sdk-environment-variables.md#general-sdk-configuration

But I didn't find the use of the `OTEL_PROPAGATORS` environment variable in Opentelemetry (golang)

image

pellared:
@1046102779 My comment is not directly related to this PR. IMO the OTel spec says that the SDK should default to these propagators and the user should not need to specify them by hand. Your PR is OK 👍 I will create an issue in OTel SDK.

Aneurysm9:
> IMO the OTel spec says that the SDK should default to these propagators and the user should not need to specify them by hand.

This is in direct conflict with the [API specification](https://github.com/open-telemetry/opentelemetry-specification/blob/main/specification/context/api-propagators.md#global-propagators):

> The OpenTelemetry API MUST use no-op propagators unless explicitly configured otherwise.

MrAlias (approved):

PR #1857, opened 2022-02-24
fix: add missing platform and cloud provider values

anupamdalmia10:
Add missing cloud.platform value to AWS Lambda, GCP Compute Engine, GCP Cloud Run, GCP GKE resource detectors and add cloud.provider to GCP GKE resource detector.


MrAlias (approved):

dashpole (approved):
lgtm for GCP detectors

anupamdalmia10:
Fixed the failing test cases

anupamdalmia10:
Looks like test cases from different module are failing...

Aneurysm9 (approved):

Aneurysm9:
> Looks like test cases from different module are failing...

Tests in `instrumentation/github.com/aws/aws-lambda-go/otellambda` are failing due to the change in detected resource attributes.  Can you please update the test expectations.

anupamdalmia10:
> > Looks like test cases from different module are failing...
> 
> Tests in `instrumentation/github.com/aws/aws-lambda-go/otellambda` are failing due to the change in detected resource attributes. Can you please update the test expectations.

working on it

anupamdalmia10:
> > > Looks like test cases from different module are failing...
> > 
> > 
> > Tests in `instrumentation/github.com/aws/aws-lambda-go/otellambda` are failing due to the change in detected resource attributes. Can you please update the test expectations.
> 
> working on it

@Aneurysm9 , updated the test cases in the concerned package

anupamdalmia10:
Please approve workflow

anupamdalmia10:
This particular check "codecov/patch" is failing because there is a change in both gce and gke detectors and the test cases/files for those do not exist as of now. Shall those be added as a part of this PR or are the changes good to be merged?

MrAlias:
> This particular check "**codecov/patch**" is failing because there is a change in both **_gce_** and **_gke_** detectors and the test cases/files for those do not exist as of now. Shall those be added as a part of this PR or are the changes good to be merged?

It would be appreciated if you wanted to add coverage for these here.

anupamdalmia10:
> > This particular check "**codecov/patch**" is failing because there is a change in both **_gce_** and **_gke_** detectors and the test cases/files for those do not exist as of now. Shall those be added as a part of this PR or are the changes good to be merged?
> 
> It would be appreciated if you wanted to add coverage for these here.

Sure @MrAlias , I'll work on the test cases and update the PR

anupamdalmia10:
@MrAlias , there is another PR in progress: https://github.com/open-telemetry/opentelemetry-go-contrib/pull/1584

It will depend on it how the test cases are going to be organised further in fir gcp detectors.

detectors/gcp/gce.go
```diff
@@ -28,30 +28,41 @@ import (
...
+	}
+}
 
 // compile time assertion that GCE implements the resource.Detector interface.
 var _ resource.Detector = (*GCE)(nil)
 
 // Detect detects associated resources when running on GCE hosts.
 func (gce *GCE) Detect(ctx context.Context) (*resource.Resource, error) {
-	if !metadata.OnGCE() {
+	if !gce.onGCE() {
```
MrAlias, line 48: This is an exported type in a stable package. It needs to be considered what will happen for an existing user that is declaring a new variable of this type and they upgrade: the `onGCE` field will be nil and this will panic.

This also exports a new variable (I'm not sure why is this isn't `func` declaration?). I would suggest we try and preserve the existing initialization behavior of the `GCE` type as best we can to support existing uses and not include this new `var`. Also, if we ultimately want to add this, it should not be done in a PR dedicated to the addition (not in a PR adding semconv values).

I expect this was done to accommodate mocking while testing. Instead of this approach, I would suggest abstract the `metadata` package functions used to be unexported package variables that a test could overwrite.

detectors/gcp/gke.go
```diff
@@ -27,18 +27,29 @@ import (
...
 
 // GKE collects resource information of GKE computing instances
-type GKE struct{}
+type GKE struct {
+	mc     metadataClient
+	getenv func(string) string
+}
+
+// NewCloudRun creates a CloudRun detector.
+func NewGKE() *GKE {
```
MrAlias, line 36: Similar here.

MrAlias (changes requested):

PR #1880, opened 2022-03-03
Implemented logic for Opentracing baggage propagation

jaronoff97:
What
-----------------------
This PR implements the logic for extracting baggage in the opentracing propagator. 

Why
-----------------------
Users of the opentracing baggage propagator want to extract baggage so that they can effectively migrate from opentracing to opentelemetry.

Other Implementations:
-----------------------

[Java](https://github.com/open-telemetry/opentelemetry-java/blob/dfb4a74/extensions/trace-propagators/src/main/java/io/opentelemetry/extension/trace/propagation/OtTracePropagator.java#L108-L124)
[Python](https://github.com/open-telemetry/opentelemetry-python-contrib/blob/3137c1f/propagator/opentelemetry-propagator-ot-trace/src/opentelemetry/propagators/ot_trace/__init__.py#L95-L107)
[Javascript](https://github.com/open-telemetry/opentelemetry-js-contrib/blob/6c2378c/propagators/opentelemetry-propagator-ot-trace/src/OTTracePropagator.ts#L105-L114)


jaronoff97:
Pending CLA authorization from Lightstep right now.

jmacd (approved):

carlosalberto:
Looks great, thanks!

jaronoff97:
@carlosalberto  one result of this behavior is that if baggage from the opentracing side is invalid, we don't create a context from remote span. Should we instead change [this block](https://github.com/open-telemetry/opentelemetry-go-contrib/blob/9f13bf5ff514eaa4b7e76798bf5cc62f93820f5b/propagators/ot/ot_propagator.go#L98-L103) to be something like:

``` go
         bags, err := extractBags(carrier)
	if err != nil {
		return trace.ContextWithRemoteSpanContext(ctx, sc)
	}
	ctx = baggage.ContextWithBaggage(ctx, bags)
	return trace.ContextWithRemoteSpanContext(ctx, sc)
```

carlosalberto:
@jaronoff97 Oh, good call - indeed, invalid `Baggage` shouldn't prevent a valid `SpanContext` from being returned.

propagators/ot/ot_propagator.go
```diff
@@ -93,18 +94,42 @@ func (o OT) Extract(ctx context.Context, carrier propagation.TextMapCarrier) con
...
+	}
+	ctx = baggage.ContextWithBaggage(ctx, bags)
 	return trace.ContextWithRemoteSpanContext(ctx, sc)
 }
 
 func (o OT) Fields() []string {
 	return []string{traceIDHeader, spanIDHeader, sampledHeader}
 }
 
+// extractBags reconstructs the baggage information from opentracing
```
MrAlias, line 110:
```suggestion
// extractBags extracts OpenTracing baggage information from carrier.
```

propagators/ot/ot_propagator.go
```diff
@@ -93,18 +94,42 @@ func (o OT) Extract(ctx context.Context, carrier propagation.TextMapCarrier) con
...
+		if err != nil {
+			return emptyBags, err
+		}
+		members = append(members, member)
+	}
+	bags, err := baggage.New(members...)
+	if err != nil {
+		return emptyBags, err
+	}
+	return bags, nil
```
MrAlias, line 130:
```suggestion
	return baggage.New(members...)
```

MrAlias (approved):

propagators/ot/ot_propagator.go
```diff
@@ -93,18 +94,42 @@ func (o OT) Extract(ctx context.Context, carrier propagation.TextMapCarrier) con
...
+	var members []baggage.Member
+	for _, key := range carrier.Keys() {
+		lowerKey := strings.ToLower(key)
+		if !strings.HasPrefix(lowerKey, baggageHeaderPrefix) {
+			continue
+		}
+		strippedKey := strings.TrimPrefix(lowerKey, baggageHeaderPrefix)
+		member, err := baggage.NewMember(strippedKey, carrier.Get(key))
+		if err != nil {
+			return emptyBags, err
```
Aneurysm9, line 122: Should this accumulate into a `multierr.Error` and just drop the bad entry rather than potentially losing all baggage members?

jaronoff97: I was thinking the same thing. Will update with that change.

Aneurysm9 (approved):

propagators/ot/ot_propagator.go
```diff
@@ -20,6 +20,8 @@ import (
 	"fmt"
 	"strings"
 
+	"go.uber.org/multierr"
+
 	"go.opentelemetry.io/otel/baggage"
 
 	"go.opentelemetry.io/otel/propagation"
```
hanyuancheung, line 27:
```suggestion

	"go.opentelemetry.io/otel/baggage"
	"go.opentelemetry.io/otel/propagation"
```

hanyuancheung (approved):
lgtm

PR #1890, opened 2022-03-07
AWS X-Ray could contain only Root value

regeda:
The XRay propagator always expects the Parent value:
```
X-Amzn-Trace-Id: Root=1-5759e988-bd862e3fe1be46a994272793;Parent=53995c3f42cd8ad8;Sampled=1
```

The AWS spec says that the Parent value might be absent and the
following headers are also correct:
```
X-Amzn-Trace-Id: Root=1-5759e988-bd862e3fe1be46a994272793
X-Amzn-Trace-Id: Root=1-5759e988-bd862e3fe1be46a994272793;Sampled=1
```

This change makes the span id from the first 8 bytes of the trace id if the
Parent value is not exist.

### DIFF
--- a/propagators/aws/xray/propagator.go
+++ b/propagators/aws/xray/propagator.go
@@ -54,10 +54,13 @@ var (
 )
 
 // Propagator serializes Span Context to/from AWS X-Ray headers.
+// @see
+//   https://docs.aws.amazon.com/xray/latest/devguide/xray-concepts.html#xray-concepts-tracingheader
 //
 // Example AWS X-Ray format:
-//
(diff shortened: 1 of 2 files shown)


regeda:
The current implementation considers the following trace as invalid:
```
X-Amzn-Trace-Id: Root=1-5759e988-bd862e3fe1be46a994272793
```
bc no `Parent`

Then the library generates a new one with a random `Parent` and the origin `Root` is lost.

My approach keeps the origin `Root` and adds the `Parent` derived from the root to respect validation rules: both `Root` and `Parent` should exist

anuraaga:
@willarmiros

PR #1920, opened 2022-03-16
Remove deprecated exporters

MrAlias:

PR #1848, opened 2022-02-19
[otelecho] Use SpanStatusFromHTTPStatusCodeAndSpanKind on the middleware

sgirones:
Fix the `otelecho` middleware and follow the OpenTelemetry specification.

For servers, 4XX status codes must not be treated as errors. OpenTelemetry's `semconv` package provides the `SpanStatusFromHTTPStatusCodeAndSpanKind` so the status code is treated different depending on whether the `SpanKind` is `server` or not

PR #1582, opened 2022-01-12
otelaws: adding dynamodb attributes

jennynilsen:
When appending the middlewares to a dynamodb client, no dynamodb attributes are added to the spans.

This PR :

adds dynamodb attributes that correspond to the dynamodb operation to the spans
allows someone to provide their own function to set custom attributes
adds a way to determine the attributes to set based on service automatically

PR #1921, opened 2022-03-16
Release v1.5.0

MrAlias:
Blocked on https://github.com/open-telemetry/opentelemetry-go-contrib/pull/1920

---

### Added

- Added the `go.opentelemetry.io/contrib/samplers/jaegerremote` package.
  This package implements the Jaeger remote sampler for OpenTelemetry Go. (#936)
- DynamoDB spans created with the `go.opentelemetry.io/contrib/instrumentation/github.com/aws/aws-sdk-go-v2/otelaws` package now have the appropriate database attributes added for the operation being performed.
  These attributes are detected automatically, but it is also now possible to provide a custom function to set attributes using `WithAttributeSetter`. (#1582)
- Add resource detector for GCP cloud function. (#1584)
- Add OpenTracing baggage extraction to the OpenTracing propagator in `go.opentelemetry.io/contrib/propagators/ot`. (#1880)

### Fixed

- Fix the `echo` middleware by using `SpanKind.SERVER` when deciding the `SpanStatus`.
  This makes `4xx` response codes to not be an error anymore. (#1848)


Aneurysm9 (approved):

MrAlias:
> LGTM
> 
> Non-blocking, but probably issue worthy: I noticed that the samplers didn't have a `version.go`. I don't know if we want to keep up with including one for every package, as it just makes these reviews harder, or if we want to change it (for the non-stable packages).

I'll add one here for consistency.

XSAM (approved):
lgtm

PR #1973, opened 2022-03-21
Use server span when determining status from code

kitschysynq:
Switches to `SpanStatusFromHTTPStatusCodeAndSpanKind` with a span kind
of `SpanKindServer` so that 4xx status codes don't get labelled as
errors. Since these codes represent client errors, they make it
difficult to differentiate bad server behavior from bad client behavior.


kitschysynq:
Sure thing, thanks for looking!

kitschysynq:
I pushed the CHANGELOG entry but it's not clear to me how to test the result. Do you have any advice for checking the state of a span after the handler and middleware have returned?

MrAlias:
> I pushed the CHANGELOG entry but it's not clear to me how to test the result. Do you have any advice for checking the state of a span after the handler and middleware have returned?

You'll probably need to create a test case similar to this: https://github.com/open-telemetry/opentelemetry-go-contrib/blob/826e886bfc8e16c6aa88abf17f17bcfc394dbf70/instrumentation/github.com/gorilla/mux/otelmux/test/mux_test.go#L35

The `test` module is used to check integration and will be able to determine the final span state the default SDK will produce.

kitschysynq:
Thanks for the info. Added that test and caught the branch up with main.

MrAlias (approved):

hickeyma (approved):

instrumentation/github.com/gorilla/mux/otelmux/test/mux_test.go
```diff
@@ -69,6 +73,33 @@ func TestSDKIntegration(t *testing.T) {
...
+	assertSpan(t, sr.Ended()[0],
+		"/does/not/exist",
+		trace.SpanKindServer,
+		attribute.String("http.server_name", "foobar"),
+		attribute.Int("http.status_code", http.StatusNotFound),
+		attribute.String("http.method", "GET"),
+		attribute.String("http.target", "/does/not/exist"),
+		attribute.String("http.route", "/does/not/exist"),
+	)
+	assert.NotEqual(t, sr.Ended()[0].Status().Code, codes.Error)
```
Aneurysm9, line 99: Should this assert that the code is equal to `codes.Unset`?  Pretty sure that instrumentation like this is never supposed to set `codes.Ok` and that's the only other option.

Aneurysm9 (approved):

instrumentation/github.com/gorilla/mux/otelmux/test/mux_test.go
kitschysynq: I agree that it doesn't make any sense to return `codes.Ok` but it's not explicitly stated in the `semconv` docs that it won't. I went for this because it DOES explicitly state:
```
Exclude 4xx for SERVER to set the appropriate status.
```
I'm happy to update it if you think it's preferable to `assert.Equal(..., codes.Unset)`.

kitschysynq: Updated the test, and caught up with main

PR #1974, opened 2022-03-21
Build tooling refactor

MrAlias:
- Update `Makefile` to match `opentelemetry-go`
- Add vanity import checks
- Add missing generation step for zpages resource
  - Lock down missing dependency on `esc` used in the generation target


XSAM (approved):
lgtm

PR #2137, opened 2022-03-30
feature: use peroperationsamplerupdater first

### LINKED ISSUE #2135: jaegerremote should use perOperationSamplerUpdater first when update strategy from remote
## description
Hi, thanks to those who contributed to jaegerremote.
But I found that its behavior is inconsistent with what jaeger describes in the sampling section of the official documentation in the scenario where I want to disable certain operation's trace through per operation strategy

## problem 
version: go.opentelemetry.io/contrib/samplers/jaegerremote v0.1.0
collector strategy.json file content
```
{
    "service_strategies": [
        {
            "service": "serviceA",
            "type": "probabilistic",
            "param": 0.9
        },
        {
            "service": "serviceB",
            "type": "ratelimiting",
            "param": 5
        }
    ],
    "default_strategy": {
        "type": "probabilistic",
        "param": 1,
        "operation_strategies": [
            {
                "operation": "metrics",
                "type": "probabilistic",
                "param": 0.0
            }
        ]
    }
}
```
otel will query strategy at jaeger agent  server `localhost:5778/sampling?service=serviceA`
the response
```
{
  "strategyType": 0,
  "probabilisticSampling": {
    "samplingRate": 0.9
  },
  "operationSampling": {
    "defaultSamplingProbability": 0.9,
    "defaultLowerBoundTracesPerSecond": 0,
    "perOperationStrategies": [
      {
        "operation": "metrics",
        "probabilisticSampling": {
          "samplingRate": 0
        }
      }
    ],
    "defaultUpperBoundTracesPerSecond": 0
  }
}
```
Everything is correct up to here. As described in the jaeger official documentation (shown below). When starting a span on serviceA operation=/metrics, this span should not be sampled. But actually it is sampled.
image

Then I read the code related to the jaegerremote sampler and found that its workflow is as follows.
1. init sampler with multiple updaters
```
func New(
	serviceName string,
	opts ...Option,
) *Sampler {
	options := newConfig(opts...)
	sampler := &Sampler{
		config:      options,
		serviceName: serviceName,
		doneChan:    make(chan *sync.WaitGroup),
	}
	go sampler.pollController()
	return sampler
}

// newConfig returns an appropriately configured config.
func newConfig(options ...Option) config {
	c := config{
		sampler:                 newProbabilisticSampler(0.001),
		// omit for simple reading ...	
		posParams: perOperationSamplerParams{
			MaxOperations:            defaultSamplingMaxOperations,
			OperationNameLateBinding: defaultSamplingOperationNameLateBinding,
		},
	}
	for _, option := range options {
		option.apply(&c)
	}
	c.updaters = append(c.updaters,
		&perOperationSamplerUpdater{
			MaxOperations:            c.posParams.MaxOperations,
			OperationNameLateBinding: c.posParams.OperationNameLateBinding,
		})

	return c
}
```
2. Regularly pull strategy from the remote
**Here is the point**
3. Use multiple updaters to update the strategy. When one sampler is successfully updated, it returns.
```
// NB: this function should only be called while holding a Write lock
func (s *Sampler) updateSamplerViaUpdaters(strategy interface{}) error {
	for _, updater := range s.updaters {
		sampler, err := updater.Update(s.sampler, strategy)
		if err != nil {
			return err
		}
		if sampler != nil { // here it return !!!
			s.sampler = sampler
			return nil
		}
	}
	return fmt.Errorf("unsupported sampling strategy %+v", strategy)
}
```
There is actually a problem here. In fact, the agent integrates the configuration of strategy in the dimension of service. If there is a configuration of perOperation, it should be used first. When there is no perOperation configuration, choose one of the other two.

## my solution
So I think we should put peroperationupdater in front of updater when initializing Sampler.

By the way, another small problem, in the scenario where I want to disable certain operation's trace through per operation strategy, ratelimiter will still let go of a request.

vuuihc:
resolve #2135


dmathieu:
Do you think you could add tests for this behavior?

vuuihc:
> Do you think you could add tests for this behavior?

sure, I will add tests

vuuihc:
Just added a test for this behavior. 
btw, Ratelimiter always allows the first request to go through, and I found corresponding examples in tests, but I don't quite understand why such a feature is needed, or I think it may be a misleading bug, can someone tell me why this is Is it designed this way, or is this a bug?
If it's a designed feature, I'll remove the corresponding changes

vuuihc:
@dmathieu @Aneurysm9 @jmacd This time I run `make precommit` and `make ci` before push, please review the code  and rerun the workflow, thanks!

samplers/jaegerremote/sampler_remote_options.go
```diff
@@ -49,15 +49,13 @@ func newConfig(options ...Option) config {
 			OperationNameLateBinding: defaultSamplingOperationNameLateBinding,
 		},
 	}
+	c.updaters = append([]samplerUpdater{&perOperationSamplerUpdater{
+		MaxOperations:            c.posParams.MaxOperations,
+		OperationNameLateBinding: c.posParams.OperationNameLateBinding,
+	}}, c.updaters...)
```
dmathieu, line 55: This shouldn't be moved before the options loop. Doing so means any updater specified in `WithUpdaters` will never be used.

vuuihc: > This shouldn't be moved before the options loop. Doing so means any updater specified in `WithUpdaters` will never be used.

`withUpdaters` will overwrite `updaters`, so it has the highest priority. On the contrary, putting it after options will change the effect of `withUpdaters`, which will fail some tests.

vuuihc:
@dmathieu @jmacd @Aneurysm9 I'm First-time contributors, please rerun the tests workflow thanks. I replied to the reviewed comment, please tell me if I'm wrong, thanks

jmacd (approved):

samplers/jaegerremote/sampler_remote_options.go
Aneurysm9: `withUpdaters` is not exported and only used in tests.  Fix the tests that use it.  This absolutely needs to happen after the options are applied as `WithMaxOperations` and `WithOperationNameLateBinding` _are_ exported and would not be effective if this is done before option application.

Aneurysm9 (changes requested):

samplers/jaegerremote/sampler_remote_options.go
vuuihc: got it, done, thanks for your explanation.

vuuihc:
@dmathieu @Aneurysm9 The modifications you requested have been completed, please take the time to review the code. Thank you

Aneurysm9 (approved):

dmathieu (approved):

PR #2138, opened 2022-03-31
Add request count metric with status code label

nabokihms:
Signed-off-by: m.nabokikh <maksim.nabokikh@flant.com>

Closes https://github.com/open-telemetry/opentelemetry-go-contrib/issues/711

This PR adds a counter for the total amount of requests with the http_status_code label. It will help us to be aware of the ratio of errors.

I decided to avoid adding status code to other metrics because it looks excessive. Instead, counting the number of requests seems more sensible.

P.S.:
I accidentally closed the previous PR https://github.com/open-telemetry/opentelemetry-go-contrib/pull/771


CHANGELOG.md
```diff
@@ -13,6 +13,7 @@ This project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0.htm
 ### Added
 
 - The project is now tested against Go 1.18 (in addition to the existing 1.16 and 1.17) (#1976)
+- Add total requests counter by response status code for the `net/http` instrumentation. (#771)
```
MrAlias, line 16:
```suggestion
- Add total a counter metric instrument to the `go.opentelemetry.io/contrib/instrumentation/net/http/otelhttp` instrumentation that counts the number or requests and annotates them by their response code. (#771, #2138)
```

MrAlias (commented):
This metric overlaps with the information that will be communicated by the `http.server.active_requests`: https://github.com/open-telemetry/opentelemetry-specification/blob/v1.9.0/specification/metrics/semantic_conventions/http-metrics.md#http-server

I'm hesitant to accept this addition if we are ultimately going to comply with the OTel semantic conventions and there will be duplication. Should this be implemented as an `UpDownCounter` and track the active requests? Do the semantic conventions need to be updated before we add this?

Aneurysm9:
The spec recommends that the status code be added to the duration metrics, which can provide a count when aggregated as a histogram.  Should we do that and rely on the user to configure a view to do dimension reduction if histograms with status codes is a cardinality concern?

nabokihms:
From my point of view, it would be better to avoid using updown counter for errors, because you cannot apply rate functions to it.

Adding status code label for histograms sounds ok to me. It is a common practice. Should we go this way then?

nabokihms:
@MrAlias @Aneurysm9, just a friendly reminder. Is it ok to add the status code label to the server latency metric?

Aneurysm9:
> @MrAlias @Aneurysm9, just a friendly reminder. Is it ok to add the status code label to the server latency metric?

I'm good with that approach.

nabokihms:
I've just added the status code label, so everything should be good now.

nabokihms:
Folks, any updates on this one? It would be best if we resolve all concerns about this PR before the next release 🙏

geek (approved):

nabokihms:
I rebased this branch on the main. Just a friendly reminder, still waiting for the feedback 😸

dmathieu:
> Already implemented:

Your link is setting attributes for the spans, not metrics.

nabokihms:
@dmathieu hello there! A test is not included for this PR. Do you want me to add one, or should I wait for the initial review?

dmathieu:
I believe adding a test would make the review easier.

nabokihms:
@dmathieu I checked this out, and it seems like the issue is fixed by the following code
https://github.com/open-telemetry/opentelemetry-go-contrib/blob/c52452d90ab05a5a92f54398c0093376883cca2a/instrumentation/net/http/otelhttp/handler.go#L219-L221

[All thanks ](https://github.com/open-telemetry/opentelemetry-go-contrib/pull/3182) to @MrAlias. Now everything works as expected. Thanks to everyone. It was a two years journey that came to an end (I'm going to close the PR).

PR #2164, opened 2022-04-06
otelhttp: nil check wrappedBody.Close

adamdecaf:
We're experimenting with open-telemetry and ran into a panic where `*wrappedBody.Close()` can panic. This seems to be code from 2+ years ago, so I'm surprised we ran into it. 

I'm happy to sign the CLA if this change will be accepted.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -22,6 +22,10 @@ This project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0.htm
 - Upgraded all dependencies on stable modules from `go.opentelemetry.io/otel` from v1.5.0 to v1.6.1. (#2134)
 - Upgraded all dependencies on metric modules from `go.opentelemetry.io/otel` from v0.27.0 to v0.28.0. (#1977)
 
+### Fixed
+
+- otelhttp: Avoid panic by adding nil check to `wrappedBody.Close` (#2164)
+
 ## [1.5.0/0.30.0/0.1.0] - 2022-03-16
 
 ### Added
--- a/instrumentation/net/http/otelhttp/transport_test.go
+++ b/instrumentation/net/http/otelhttp/transport_test.go
@@ -285,7 +285,7 @@ func TestWrappedBodyClosePanic(t *testing.T) {
 	s := new(span)
 	var body io.ReadCloser
 	wb := newWrappedBody(s, body)
-	assert.NoError(t, wb.Close())
+	assert.NotPanics(t, func() { wb.Close() }, "nil body should not panic on close")
 }
 
 func TestWrappedBodyCloseError(t *testing.T) {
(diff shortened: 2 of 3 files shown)


Aneurysm9 (approved):
Can you please add a CHANGELOG entry noting this fix?  Thanks!

adamdecaf:
@Aneurysm9 added, thanks!

adamdecaf:
I got a notification about an error with the GRPC instrument, but that's not part of my changes.

https://github.com/open-telemetry/opentelemetry-go-contrib/blob/f06b204c74976660004392462bed1a8c6b8171ca/instrumentation/google.golang.org/grpc/otelgrpc/test/interceptor_test.go#L396

CHANGELOG.md
```diff
@@ -19,6 +19,10 @@ This project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0.htm
 - Upgraded all dependencies on stable modules from `go.opentelemetry.io/otel` from v1.5.0 to v1.6.1. (#2134)
 - Upgraded all dependencies on metric modules from `go.opentelemetry.io/otel` from v0.27.0 to v0.28.0. (#1977)
 
+### Fixed
+
+- otelhttp: nil check wrappedBody.Close
```
Aneurysm9, line 24:
```suggestion
- otelhttp: Avoid panic by adding nil check to `wrappedBody.Close` (#2164)
```

instrumentation/net/http/otelhttp/transport_test.go
```diff
@@ -279,6 +279,10 @@ func TestWrappedBodyClose(t *testing.T) {
 	wb := &wrappedBody{span: trace.Span(s), body: readCloser{}}
 	assert.NoError(t, wb.Close())
 	s.assert(t, true, nil, codes.Unset, "")
+
+	var body io.ReadCloser
```
dmathieu, line 283: How about making this its own test (like the following one, `TestWrapperBodyCloseError`)?
It'd make it easier to see why a failure happens.

XSAM: Yeah, I think it is better.

adamdecaf: Updated.

CHANGELOG.md
adamdecaf: Updated.

instrumentation/net/http/otelhttp/transport_test.go
```diff
@@ -281,6 +281,13 @@ func TestWrappedBodyClose(t *testing.T) {
 	s.assert(t, true, nil, codes.Unset, "")
 }
 
+func TestWrappedBodyClosePanic(t *testing.T) {
+	s := new(span)
+	var body io.ReadCloser
+	wb := newWrappedBody(s, body)
+	assert.NoError(t, wb.Close())
```
MrAlias, line 288:
```suggestion
	assert.NotPanics(t, func() { wb.Close() }, "nil body should not panic on close")
```

MrAlias (approved):

Issue #2166, opened 2022-04-07
Go-kit does not propagate incoming `traceparent`


drewtoto:
**What version of Go and opentelemetry-go-contrib are you using?**
go 1.17
opentelemetry-go-contrib  v0.31.0

**What operating system and processor architecture are you using?**
linux/amd64

**What did you do?**
Added tracing middleware to go-kit endpoints (gRPC) provided by `go.opentelemetry.io/contrib/instrumentation/github.com/go-kit/kit/otelkit` package using `otelkit.EndpointMiddleware(otelkit.WithOperation("operation"))(my-gokit-endpoint)`

**What did you expect to see?**
Expected to see the `otelkit` package extract the `traceparent` from the incoming `context` and make the _go-kit_ created span a child of the parent (the gRPC caller in my case).  Expected that `otelkit.EndpointMiddleware` would use `Extract` from the `go.opentelemetry.io/otel/propagation` package to parse and use the `traceparent`.

**What did you see instead?**
The tracing span from the caller (gRPC client from a separate process) was _NOT_ propagated and used as the trace parent for the _go-kit_ created span.  **NOTE**: `traceparent` header is present in the incoming `context`.

**Additional comments**
This feels like I'm doing something wrong as distributed tracing is a core benefit for implementing tracing to begin with.  I understand that this community is relatively small at the moment which may be the reason for the limited documentation, but, trying to figure out if this functionality is _baked in_ or if I'm just doing something wrong was quite an effort.

Ultimately, my solution was to implement a **propagation middleware** to wrap the `otelkit.EndpointMiddleware`.  Though this solution is relatively simple, it still feels like this functionality _should_ be baked in.

What am I doing wrong?

My solution below:

```
package endpoint

import (
	"context"

	"github.com/go-kit/kit/endpoint"
	"go.opentelemetry.io/contrib/instrumentation/github.com/go-kit/kit/otelkit"
	"go.opentelemetry.io/otel"
	"go.opentelemetry.io/otel/propagation"
	"google.golang.org/grpc/metadata"
)

func MakeTracingEndpoint(endpoint endpoint.Endpoint, operation string) endpoint.Endpoint {
	return PropagationMiddleware()(
		otelkit.EndpointMiddleware(otelkit.WithOperation(operation))(endpoint),
        )
}

type MetadataCarrier metadata.MD

// Compile time check that MapCarrier implements the TextMapCarrier.
[...]

PR #2167, opened 2022-04-08
Add trace propagation for otelkit endpoints (#2166)

drewtoto:
Trace propagation for otelkit currently supports:
* gRPC via GrpcPropagationMiddleware
* HTTP via HttpPropagationMiddleware


hanyuancheung (approved):

drewtoto:
Is there any action I have to do here?  Github is telling me this branch has conflicts with `go.mod` and `test/go.mod` but I'm not seeing any conflicts - am I just missing something here?

MrAlias:
> Is there any action I have to do here? Github is telling me this branch has conflicts with `go.mod` and `test/go.mod` but I'm not seeing any conflicts - am I just missing something here?

You need to merge the `main` upstream branch into this branch. Successfully doing so will require you to resolve the change conflicts between the two branches.

MrAlias (commented):
File permissions for the CHANGELOG, go.mod, and go.sum have changed. Please revert to the original permissions.

instrumentation/github.com/go-kit/kit/otelkit/propagation.go
```diff
@@ -0,0 +1,82 @@
...
+	"net/http"
+
+	"github.com/go-kit/kit/endpoint"
+	"google.golang.org/grpc/metadata"
+
+	"go.opentelemetry.io/otel"
+	"go.opentelemetry.io/otel/propagation"
+)
+
+type MultimapCarrier map[string][]string
```
MrAlias, line 28: This has the same underlying type as the [`HeaderCarrier`](https://github.com/open-telemetry/opentelemetry-go/blob/0f3ab76ff31af174bd0f8af54f194b91a80479f5/propagation/propagation.go#L69-L89). Why do we need to include this and not just wrap a `map[string][]string` as a `HeaderCarrier`?

instrumentation/github.com/go-kit/kit/otelkit/propagation.go
```diff
@@ -0,0 +1,82 @@
...
+	for k := range *c {
+		keys = append(keys, k)
+	}
+	return keys
+}
+
+// GrpcPropagationMiddleware uses gRPC metadata from the incoming context,
+// if it exists, and extracts the traceparent to propagate context information
+// that enables distributed tracing.
+func GrpcPropagationMiddleware() endpoint.Middleware {
```
MrAlias, line 59:
```suggestion
func GRPCPropagationMiddleware() endpoint.Middleware {
```

instrumentation/github.com/go-kit/kit/otelkit/propagation.go
drewtoto: Absolutely... Thanks for pointing this out (sorry, I didn't realize `http.Header` was the same underlying).

drewtoto: Ahh, it looks like a new type is needed after all - `http.Header` canonicalizes the keys passed into `Get` [^1][^2] and `metadata.MD`, which is used to store the **traceparent** header in gRPC, converts all keys to lowercase [^3][^4][^5].

The idea behind using `type MultimapCarrier map[string][]string` versus `type MetadataCarrier metadata.MD` was to relieve the dependency on `google.golang.org/grpc` package in case this type would become part of **opentelemetry-go** for other gRPC-enabled transport libraries to use, just as `HeaderCarrier` can be used for HTTP transport.

[^1]: https://pkg.go.dev/net/http#Header.Get
[^2]: https://cs.opensource.google/go/go/+/master:src/net/http/header.go;l=48 
[^3]: https://pkg.go.dev/google.golang.org/grpc/metadata#MD.Get
[^4]: https://pkg.go.dev/google.golang.org/grpc/metadata#MD.Set
[^5]: https://pkg.go.dev/google.golang.org/grpc/metadata#New

drewtoto:
> File permission and lint issues need to be addressed before this is ready to be comprehensively reviewed.

Thank you for this comment - I am in the process of _fixing_ file permissions (my IDE is changing these and it's really frustrating) and adding the extra tests for code coverage.

PR #2182, opened 2022-04-13
otelmongo: set attribute based on hostname or IP address

### LINKED ISSUE #2165: [otelmongo] should use `AttributeNetPeerIP` when hostname is a IP
https://github.com/open-telemetry/opentelemetry-go-contrib/blob/1e34aa359ff6cfe4ba2a44775d261311ea024197/instrumentation/go.mongodb.org/mongo-driver/mongo/otelmongo/mongo.go#L54

should use `AttributeNetPeerIP` when hostname is a IP

DiptoChakrabarty:
Fixes #2165 

Set the appropriate attribute based on whether hostname is an Ip address or not


instrumentation/go.mongodb.org/mongo-driver/mongo/otelmongo/mongo.go
```diff
@@ -51,10 +52,17 @@ func (m *monitor) Started(ctx context.Context, evt *event.CommandStartedEvent) {
 		semconv.DBSystemMongoDB,
 		semconv.DBOperationKey.String(evt.CommandName),
 		semconv.DBNameKey.String(evt.DatabaseName),
-		semconv.NetPeerNameKey.String(hostname),
 		semconv.NetPeerPortKey.Int(port),
 		semconv.NetTransportTCP,
 	}
+
+	addr := net.ParseIP(hostname)
```
dmathieu, line 59: I wonder if this wouldn't be more efficient with a regex?

instrumentation/go.mongodb.org/mongo-driver/mongo/otelmongo/mongo.go
```diff
@@ -51,10 +52,17 @@ func (m *monitor) Started(ctx context.Context, evt *event.CommandStartedEvent) {
...
+	if addr != nil {
+		attrs = append(attrs, semconv.NetPeerIPKey.String(hostname))
```
pellared, line 61: Can you please cover this scenario in the unit tests? It would be good to cover both IPv4 and IPv6.

MrAlias:
@DiptoChakrabarty are you still able to work on this? It looks abandoned. If so, please address the feedback that has been given. If not, please close the PR.

PR #2210, opened 2022-04-21
fix: do not swallow error in otel echo middleware and return it to other middlewares as expected. #2209

### LINKED ISSUE #2209: OtelEcho swallows the error and does not pass it to other middlewares
## Background

Hi, we were trying to use OpenTelemetry [echo middleware](https://github.com/open-telemetry/opentelemetry-go-contrib/tree/main/instrumentation/github.com/labstack/echo/otelecho) provided in this repository in our project. However, using it causes misfunction in other middlewares.  For instance, while trying we were using another middleware to send errors to our current apm tool and it stopped recording errors.

It seems that the current implementation of echo middleware swallows the error and does not pass it to other middleware (https://github.com/open-telemetry/opentelemetry-go-contrib/blob/main/instrumentation/github.com/labstack/echo/otelecho/echo.go#L101). 

## Proposed Solution

- Return the error received from the next middleware call.
- Also don't call the registered HTTP error handler since it is automatically called if the error received from the next middleware call is returned.

hcelaloner:
fix: do not swallow error in otel echo middleware and return it to other middlewares as expected.

Middleware was swallowing the error and it was not passing it to previous middlewares as expected.
Also, when we return the error from the middleware, HTTP error handler is called automatically.

Co-authored-by: Erkan Zileli<erkan.zileli@trendyol.com>
Signed-off-by: Celal Öner <celal.oner@trendyol.com>

Fixes #2209

### DIFF
--- a/instrumentation/github.com/labstack/echo/otelecho/echo.go
+++ b/instrumentation/github.com/labstack/echo/otelecho/echo.go
@@ -89,8 +89,6 @@ func Middleware(service string, opts ...Option) echo.MiddlewareFunc {
 			err := next(c)
 			if err != nil {
 				span.SetAttributes(attribute.String("echo.error", err.Error()))
-				// invokes the registered HTTP error handler
-				c.Error(err)
 			}
 
 			attrs := semconv.HTTPAttributesFromHTTPStatusCode(c.Response().Status)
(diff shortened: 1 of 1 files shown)


MadVikingGod (approved):

hcelaloner:
Closing pull request since the issue is now resolved by https://github.com/open-telemetry/opentelemetry-go-contrib/pull/3656

PR #2222, opened 2022-04-25
Upgrade semconv to v1.10.0

MrAlias:
Part of https://github.com/open-telemetry/opentelemetry-go/issues/2861

- [x] Upgrade semconv to v1.10.0
- [x] Fix for https://github.com/open-telemetry/opentelemetry-specification/pull/1973

PR #2180, opened 2022-04-11
[otelhttp] Remove propagation header injection

TylerHelmuth:
Current otelhttp instrumentation leaks propagation headers in the response headers.  This behavior is not defined so it should be removed.  Fixes #2163

PR #2255, opened 2022-04-28
Release v1.7.0/v0.32.0

MrAlias:
### Added

- Consistent probability sampler implementation. (#1379)

### Changed

- Upgraded all `semconv` package use to `v1.10.0`. This includes a backwards incompatible change for the `otelgocql` package to conform with the specification [change](https://github.com/open-telemetry/opentelemetry-specification/pull/1973). The `db.cassandra.keyspace` attribute is now transmitted as the `db.name` attribute. (#2222)

### Fixed

- Fix the `otelmux` middleware by using `SpanKindServer` when deciding the `SpanStatus`. This makes `4xx` response codes to not be an error anymore. (#1973)
- Fixed jaegerremote sampler not behaving properly with per operation strategy set. (#2137)
- Stopped injecting propagation context into response headers in otelhttp. (#2180)


samplers/jaegerremote/version.go
```diff
@@ -16,7 +16,7 @@ package jaegerremote // import "go.opentelemetry.io/contrib/samplers/jaegerremot
 
 // Version is the current release version of the Jaeger remote sampler.
 func Version() string {
-	return "0.30.0"
+	return "0.2.0"
```
Aneurysm9, line 19: The other two samplers seem to be missing `version.go`, should we add them here?

Aneurysm9 (approved):

samplers/jaegerremote/version.go
MrAlias: Will do :+1:

willarmiros:
Does this mean the AWS remote sampler has been released @MrAlias @Aneurysm9? If so I can't find it in the release notes

PR #2258, opened 2022-05-03
Add autoprop package to provide configuration of propagators with useful defaults and envar support

MrAlias:
### Use case

```go
otel.SetTextMapPropagator(autoprop.NewTextMapPropagator())
```

to replace

```go
otel.SetTextMapPropagator(propagation.NewCompositeTextMapPropagator(propagation.TraceContext{}, propagation.Baggage{}))
```

- Defaults to TraceContext with Baggage
- Arguments to `NewTextMapPropagator` act the same as `NewCompositeTextMapPropagator` and override defaults
- `OTEL_PROPAGATORS` overrides defaults and arguments
- Custom 3rd-party propagators can be registered with `RegisterTextMapPropagator` and then supported in the parsed propagator from `OTEL_PROPAGATORS`

Resolve https://github.com/open-telemetry/opentelemetry-go/issues/1265
Resolve https://github.com/open-telemetry/opentelemetry-go/issues/2304
Related to https://github.com/open-telemetry/opentelemetry-go/issues/1404


pellared:
I think it also resolves https://github.com/open-telemetry/opentelemetry-go/issues/2304

pellared:
What do you think of adding `OTEL_PROPAGATORS` support to the "non-contrib" version of OTel Go? It could be done in API-wise backward compatible way. I imagine something like this could be more user-friendly (probably this could be done better)

1. `propagation.NewCompositeTextMapPropagator` could already support `OTEL_PROPAGATORS`. `propagation.Register` is used to register more mappings
2. `tracecontext` `baggage` `none` are supported OOTB. `propagation.Unregister` is used if one would like to change any of the default mappings.
3. The exporter packages like `jaeger` could call `propagation.Register` in its `init()`.

This way existing users would get the `OTEL_PROPAGATORS` without changing their code.

MrAlias:
> What do you think of adding `OTEL_PROPAGATORS` support to the "non-contrib" version of OTel Go SDK? I imagine something like this could be done (probably this can be improved)
> 
> 1. `propagation.NewTextMapPropagator` supports `OTEL_PROPAGATORS`. `propagation` is a new package `go.opentelemetry.io/otel/sdk/propagation` as such code cannot live in OTel API according to https://github.com/open-telemetry/opentelemetry-specification/blob/main/specification/context/api-propagators.md#global-propagators
> 2. `tracecontext` `baggage` `none` are supported OOTB. The default propagators are `tracecontext,baggage`.
> 3. `propagation.Register` is used to register more mappings.
> 4. `propagation.Unregister` is used if one would like to change any of the default mappings.
> 5. The exporter packages like `jaeger` call `propagation.Register` in its `init()`. This couples the exporter packages to the SDK but I see some of them are already coupled to the SDK e.g. all detectors are coupled to `go.opentelemetry.io/otel/sdk/resource`.

I would not support this alternate approach adding the propagator setup package to the SDK. The contrib repository is intended to contain such things and adding something such a divergence to the specification would add nonconformity and confusion to the project.

As for the implementation, by only half supporting the values `OTEL_PROPAGATORS` is specified to support we would only support half of the specification. Requiring users to register packages to have full support is not how the specification is defined. Furthermore, it is an attempt to add convince by adding side-effects. A user imports the package and instead of being explicit about registering or using the propagator it becomes a side-effect of the import. If a user would like to setup side-effects with their own propagators they can do so with `autoprop`. They are not limited because of this approach.

I do not see that alternate approach as an improvement to this.

pellared:
> such a divergence to the specification would add nonconformity and confusion to the project

Just a note. This is not against the specification. The specification only defines the API requirements.
E.g. is what has been done in OTel .NET SDK: https://github.com/open-telemetry/opentelemetry-dotnet/blob/2e63fd99a8c8263bf3ae3a8bd70c042580bada22/src/OpenTelemetry/Sdk.cs#L32-L36

> by only half supporting the values `OTEL_PROPAGATORS` is specified to support we would only support half of the specification

The specification only says what are the "known values". The default values and `none` SHOULD be supported.

>  Requiring users to register packages to have full support is not how the specification is defined

I agree here. I was thinking about the trace exporters (instead of propagators) 🤦My bad - sorry. The contrib could simply have a helper to register all propagators supported by contrib repository.

Still, I think that it would be better if the SDK would support `OTEL_PROPAGATORS` with support for the OTel default propagators.

MrAlias:
> > such a divergence to the specification would add nonconformity and confusion to the project
> 
> This is not against the specification. The specification only defines the API requirements.

Right, this is not defined by the specification. What happens when it is? What happens when other SIGs do it differently? Why would a user expect to find this in the SDK when other SIGs do not offer this?

> > by only half supporting the values `OTEL_PROPAGATORS` is specified to support we would only support half of the specification
> 
> The specification only says what are the "known values". The default values and `none` SHOULD be supported. Reference: https://github.com/open-telemetry/opentelemetry-specification/blob/main/specification/sdk-environment-variables.md

I don't follow. If we support half the known values we sill only support half. What about the other half of known values? The user expects these values to supported if we support others, they are both known values.

> Still, I think that it would be better if the SDK would support `OTEL_PROPAGATORS` ...

I agree on this.

> with support for the OTel default propagators.

This is where you lose me. Again, supporting half the known values is supporting half of user expectation. This adds frustration and confusion. This is precisely why we did not add support for this environment variable in the past.

> Changing the proposal:
> 
> 1. `propagation.NewTextMapPropagator` supports `OTEL_PROPAGATORS`. `propagation` is a new package `go.opentelemetry.io/otel/sdk/propagation` as such code cannot live in OTel API according to https://github.com/open-telemetry/opentelemetry-specification/blob/main/specification/context/api-propagators.md#global-propagators

Again, this is a non-starter for me. I would not accept a change to add a `propagation` package to `go.opentelemetry.io/otel/sdk`.

pellared:
> Right, this is not defined by the specification. What happens when it is? What happens when other SIGs do it differently? Why would a user expect to find this in the SDK when other SIGs do not offer this?

Most of the other SDKs are already doing it. It was discussed here: https://github.com/open-telemetry/opentelemetry-go/issues/2611

> I don't follow. If we support half the known values we sill only support half. What about the other half of known values? The user expects these values to supported if we support others, they are both known values.

We should always document what we support and what we do not support. E.g. OTel JS has similar approach
- https://github.com/open-telemetry/opentelemetry-js/blob/cfda625e83a164c9e19745392879c32aedcfe76f/packages/opentelemetry-sdk-trace-base/src/BasicTracerProvider.ts#L55-L62
- https://github.com/open-telemetry/opentelemetry-js/blob/df0cee9b7fb59fbe5182b69667c00aeb984236ed/packages/opentelemetry-core/src/utils/environment.ts#L129

The downside of current that using `autoprop` would be that it is also dependent on propagators not required by the specification. So far I have seen a lot of caution when bringing potentially necessary dependencies. 

PS. I am not requesting any changes nor blocking this PR. I think it would be good to see what others think.

pellared:
I had a call with @MrAlias and we discussed 👆 

The current approach with `autoprop` is safer as

1. We create less packages - less confusion for the users
2. Having it in contrib allows to make breaking changes (as it is still v0)

propagators/autoprop/registry.go
```diff
@@ -0,0 +1,90 @@
...
+		none: propagation.NewCompositeTextMapPropagator(),
+	},
+}
+
+// registry maintains an index of propagator names to TextMapPropagator
+// implementations that is safe for concurrent use by multiple goroutines
+// without additional locking or coordination.
+type registry struct {
+	mu    sync.Mutex
+	index map[string]propagation.TextMapPropagator
```
pellared, line 60: nit:
the name `index` may a little confusing for map/hashmap/dictionary structure as some languages enable to iterate by index (not only by key). How about renaming it e.g. to `dict`?

propagators/autoprop/propagator_test.go
```diff
@@ -0,0 +1,88 @@
...
+	} else {
+		t.Cleanup(func() { os.Unsetenv(key) })
+	}
+}
+
+type handler struct {
+	err error
+}
+
+var _ otel.ErrorHandler = (*handler)(nil)
```
pellared, line 48: nit: this line is not needed.  `otel.SetErrorHandler(h)` already checks that

pellared (approved):

propagators/autoprop/registry.go
MrAlias: I'll switch to `names`. `dict` also has conflict with other language meaning.

propagators/autoprop/registry.go
```diff
@@ -0,0 +1,90 @@
...
+	}
+	r.names[key] = value
+}
+
+// RegisterTextMapPropagator sets the TextMapPropagator p to be used when the
+// OTEL_PROPAGATORS environment variable contains the propagator name. This
+// allows the default supported environment TextMapPropagators to be extended
+// with 3rd-part implementations.
+func RegisterTextMapPropagator(name string, p propagation.TextMapPropagator) {
+	envRegistry.store(name, p)
```
dashpole, line 89: Do we care that this could override existing propagators, or that there can be naming conflicts?

propagators/autoprop/propagator_test.go
MrAlias: Hmm, not sure why this is still here(?). I thought I pushed a change to remove this ...

propagators/autoprop/propagator.go
```diff
@@ -0,0 +1,129 @@
...
+// propagators to use.
+const otelPropagatorsEnvKey = "OTEL_PROPAGATORS"
+
+// NewTextMapPropagator returns a new TextMapPropagator composited by props or
+// one defined by the OTEL_PROPAGATORS environment variable. The
+// TextMapPropagator defined by OTEL_PROPAGATORS, if set, will take precedence
+// to the once composited by props.
+//
+// The propagators supported with the OTEL_PROPAGATORS environment variable by
+// default are: tracecontext, baggage, b3, b3multi, jaeger, xray, ottrace, and
```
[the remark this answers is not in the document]
dashpole: Cool, thanks.  I'll have to follow-up there to add google's propagator :).

propagators/autoprop/registry.go
MrAlias: It was intentional to allow overrides of defaults or existing registrations. This would allow 3rd party implementations to be used instead of the default propagators. I tried to capture this intention in the function docs, does it need to be made clearer?

As for the name collision, I was assuming if a user registered a propagator and used that name they would get what they expected: the name collision is resolved by the override being returned. This seemed like the intuitive behavior, but I'm open to suggestions.

propagators/autoprop/example_test.go
```diff
@@ -0,0 +1,84 @@
...
+	sort.Strings(fields)
+	fmt.Println(fields)
+	// Output: [baggage traceparent tracestate x-b3-flags x-b3-sampled x-b3-spanid x-b3-traceid]
+}
+
+func ExampleNewTextMapPropagator_environment() {
+	// Propagators set for the OTEL_PROPAGATORS environment variable take
+	// precedence and will override any arguments passed to
+	// NewTextMapPropagator.
+	_ = os.Setenv("OTEL_PROPAGATORS", "b3,baggage")
```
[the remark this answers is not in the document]
MrAlias: I'm not certain. I'll need to look into this. I'll update or add another comment afterwards.

propagators/autoprop/registry.go
dashpole: Ah, I read "extended with 3rd-part implementations" as meaning you can add additional third party propagators.  Maybe "This allows adding additional environment TextMapPropagators, or replacing the default TextMapPropagators with 3rd-party implementations."

(also s/part/party)

propagators/autoprop/example_test.go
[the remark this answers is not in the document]
MrAlias: Looking into panicking on our own for parallel tests will not work. The parallel flag is not exported by the `testing` package. I'll look into merging https://github.com/open-telemetry/opentelemetry-go/pull/2917, include a similar fix here, and update to use the Go 1.17 specific function.

propagators/autoprop/registry.go
MrAlias: Docs updated.

propagators/autoprop/example_test.go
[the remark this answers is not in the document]
dashpole: Thanks!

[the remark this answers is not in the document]
MrAlias: Blocked by #2314

propagators/autoprop/registry.go
Aneurysm9: I'm wary of allowing overriding propagator implementations that already have well-known names.  That seems like a recipe for user confusion.  I think that if someone wants to register a third-party implementation they should register it with a new name.

pellared: I imagine people (or even distributions) may want to provide their own implementation of given propagator e.g. for bugfixing, better logging, improving performance etc

MrAlias: > I'm wary of allowing overriding propagator implementations that already have well-known names. That seems like a recipe for user confusion.

My thought here was that it would be the user doing the registration and therefore the user would know there was an override.

Though, this comment made me wonder if there could be a situation where a library that a user is using in addition to `autoprop` that sets and override without the users knowledge. I imagine that library would document their registration if they were a good actor, but if they were a bad actor it might be done maliciously.

The alternative, where a user will use a different name for a similar propagator might be a better solution with this in mind. It makes the alternate implementation use explicit and the configuration of propagators one-sided (only with environment variables).

What do you think the behavior should be on a duplicate registration? Just drop?

propagators/autoprop/example_test.go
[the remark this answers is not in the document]
MrAlias: Updated to use `T.Setenv` for the unit tests.

propagators/autoprop/registry.go
MrAlias: From SIG meeting:

- Dropping the attempt to double register still leaves open the possibility of a malicious user registering their `init` prior to a users' code and having the users' code silently not register. This is not the behavior wanted.
- Panicking seems the safest and will clearly tell the user when there are conflicts.
  - The downside is that non-malicious just accidental collisions cannot be ignored. For instance, two packages register the same propagator with the same name and the user imports them both. The user does not care about the conflict, they just care that one was successful.
- Return an error and allow the user to handle the error.
   - Allowing the user to decide how to handle conflicts allows the most flexibility and still provides _some_ security.
   - Errors are often ignored or unhanded (just logged). This might not be the right user interface given the sensitive data a propagator could have access to.
   - Even if an error is returned, it is going to be the dependent library that needs to appropriately handle this, not the user importing that library. The only safe way for that library to handle the situation would seem for them to panic themselves. Which would indicate that returning an error here only pushes the security concern to dependencies (who may not recognize it as such and not handle it appropriately).

I'm planning to change to a panic based on this. Open to alternative opinions though.

MrAlias: Updated, please take another look.

dashpole (approved):

Issue #746, opened 2021-04-14
Extending functional options


pellared:
## Why

It would be good to provide guidance on how to extend on top of functional options API.

## What

Make vendor-specific (which may be not OTel compliant) extensibility "easy" and user-friendly.

The whole idea behind https://github.com/signalfx/splunk-otel-go is to create convenient and easy to use helpers that are working on top of OTel libraries. For example:

1. `splunkhttp.NewHandler` may set default otelhttp options
2. the user does not need to call `otelhttp.NewHandler`

I have experimented with 4 approaches. I will post them as separate comments so you can comment/upvote/downvote each one individually.

All of these approaches allow achieving the following usage code:

```go
handler = splunkhttp.NewHandler(handler, "server", otelhttp.WithTracerProvider(provider), splunkhttp.WithTraceResponseHeader(false))
```

The `splunkhttp.WithOTelOpts(....)` call is no longer needed.

All of the approaches are not making any breaking change.

I posted this issue because I think we should strive to use similar extensibility approaches across different vendors. 

### Blocking extensibility by design?

Is there any reason to do [such things](https://github.com/open-telemetry/opentelemetry-go/blob/b09df84a24b35cb5f287a501e31d9dec324be5a9/trace/config.go#L43-L46)? I would say it is blocking extensibility on purpose, but why? BTW even if I would like to do so I would rather unexport this method: https://github.com/open-telemetry/opentelemetry-go/blob/b09df84a24b35cb5f287a501e31d9dec324be5a9/trace/config.go#L41.

**Answered [here](https://github.com/open-telemetry/opentelemetry-go-contrib/issues/746#issuecomment-820434852).**

### Why functional options cannot return an error?

One of the main advantages of [functional options pattern](https://dave.cheney.net/2014/10/17/functional-options-for-friendly-apis) is that it can be also used to encapsulate validation. How we would notify the client that his configuration passed to https://github.com/open-telemetry/opentelemetry-go-contrib/blob/e1a7c4781baf5716364bc5661052c5746e87e344/instrumentation/net/http/otelhttp/handler.go#L60 is invalid?

**Answered [here](https://github.com/open-telemetry/opentelemetry-go-contrib/issues/746#issuecomment-820434852).**

pellared:
## Approach 1 - `WithNop` dummy option

Add `WithNop` functional option to make it easier to do a composition. 

Example: https://github.com/signalfx/splunk-otel-go/pull/37

**Pros:**
[...]

PR #2269, opened 2022-05-09
Change otelecho to wrap otelhttp

terev:
### Summary
This pr changes the otelecho instrumentation to wrap the otelhttp instrumentation rather than implement it's own. This has many benefits in my opinion as the otelecho instrumentation wasn't doing anything novel. Wrapping the otelhttp instrumentation provides metrics in addition to a more standard trace middleware. I don't really expect this pr to be accepted as is because it kind of goes against the definition of an instrumentation package. Is there a different/better approach that should be taken here?


instrumentation/github.com/labstack/echo/otelecho/echo_test.go
```diff
@@ -23,6 +23,8 @@ import (
 	"net/http/httptest"
 	"testing"
 
+	"go.opentelemetry.io/contrib/instrumentation/net/http/otelhttp"
+
 	"github.com/labstack/echo/v4"
 
 	"github.com/stretchr/testify/assert"
```
hanyuancheung, line 30: This part should be formatted properly.

instrumentation/github.com/labstack/echo/otelecho/example/server.go
```diff
@@ -19,6 +19,8 @@ import (
 	"log"
 	"net/http"
 
+	"go.opentelemetry.io/contrib/instrumentation/net/http/otelhttp"
+
 	"github.com/labstack/echo/v4"
 
 	"go.opentelemetry.io/contrib/instrumentation/github.com/labstack/echo/otelecho"
```
hanyuancheung, line 26: ditto

instrumentation/github.com/labstack/echo/otelecho/test/echo_test.go
```diff
@@ -22,6 +22,8 @@ import (
 	"net/http/httptest"
 	"testing"
 
+	"go.opentelemetry.io/contrib/instrumentation/net/http/otelhttp"
+
 	"github.com/labstack/echo/v4"
 
 	"github.com/stretchr/testify/assert"
```
hanyuancheung, line 29: ditto

hanyuancheung (approved):
overall lgtm

instrumentation/github.com/labstack/echo/otelecho/echo.go
```diff
@@ -15,89 +15,60 @@
...
-		cfg.TracerProvider = otel.GetTracerProvider()
-	}
-	tracer := cfg.TracerProvider.Tracer(
-		tracerName,
-		oteltrace.WithInstrumentationVersion(SemVersion()),
-	)
-	if cfg.Propagators == nil {
-		cfg.Propagators = otel.GetTextMapPropagator()
-	}
+func Middleware(service string, opts ...otelhttp.Option) echo.MiddlewareFunc {
```
pellared, line 35: What if we would like to have an "echo"-specific option?

Maybe better to use the approach that was proposed by @MadVikingGod some time ago: https://github.com/open-telemetry/opentelemetry-go-contrib/issues/746#issuecomment-821660165?

terev: @pellared I've added an option within the otelecho pkg to set otelhttp options. This would allow for any "echo" specific options to be added in the future. I went this way rather than mirroring otelhttp options as that doesn't seem to scale well and makes upgrades a pain. However the user experience of adding an otelhttp option isn't great i feel the benefit outweighs that con. Do you have any suggestions here?

Since otelhttp is locked to a version it should be easy to upgrade the package as is necessary.

instrumentation/github.com/labstack/echo/otelecho/config.go
```diff
@@ -39,30 +37,27 @@ func (o optionFunc) apply(c *config) {
...
 }
 
-// WithTracerProvider specifies a tracer provider to use for creating a tracer.
-// If none is specified, the global provider is used.
-func WithTracerProvider(provider oteltrace.TracerProvider) Option {
-	return optionFunc(func(cfg *config) {
-		if provider != nil {
-			cfg.TracerProvider = provider
-		}
+func WithOTelHTTPOptions(options ...otelhttp.Option) Option {
```
pellared, line 50: this approach was criticized in the issue that I linked before. for example see: https://github.com/open-telemetry/opentelemetry-go-contrib/issues/746#issuecomment-821235264

instrumentation/github.com/labstack/echo/otelecho/echo.go
pellared: >  I feel the benefit outweighs that con

It was criticized e.g. here https://github.com/open-telemetry/opentelemetry-go-contrib/issues/746#issuecomment-820402688

terev: @pellared can you please indicate the changes you think should be made here?

pellared: Instead of having `func WithOTelHTTPOptions(options ...otelhttp.Option) Option {` we should have functions like `WithPropagators`, `WithMeterProvider`, `WithMessageEvents`

Reference: https://github.com/open-telemetry/opentelemetry-go-contrib/issues/746#issuecomment-821235264

pellared:
I advise you to consider joining the SIG meeting to present the PR and ask for feedback. You can also ask for reviews on Slack. See: https://github.com/open-telemetry/opentelemetry-go/blob/main/CONTRIBUTING.md

The comments I left are based on an issue that I was working on a long time ago. Maybe it is worth the revisit the topic. Nobody even "upvoted" my comments so maybe they are no longer invalid.

instrumentation/github.com/labstack/echo/otelecho/echo.go
terev: @pellared sorry for the delay. i've just pushed a change to mirror the otelhttp options in otelecho now

CHANGELOG.md
```diff
@@ -8,6 +8,12 @@ This project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0.htm
 
 ## [Unreleased]
 
+### Changed
+
+- `otelecho` middleware to wrap `otelhttp` tracer implementation rather than implement its own. This is a backwards incompatible change
+  in most cases as `Middleware` now only accepts `otelhttp` options rather than options specific to it's package. This means that the `otelecho`
+  package now supports http server metrics provided by `otelhttp`.
```
pellared, line 15: can you add `(#2296)` add the end to make finding this PR easier?

pellared (commented):
The new options in `config_otelhttp.go` are not covered by unit tests.
It is not a blocker for me. However, you can consider adding them as I do not know if other reviewers will share my opinion. Still, I will approve this PR even without these tests 😉 

I also left comments to `CHANGELOG.md` which are easy to address. I will take another look once these comments are addressed.

pellared (approved):

hanyuancheung (approved):

terev:
I ran precommit to fix the mod tidy issue.

pellared (approved):

terev:
it looks like a linter failed because of a timeout on an unrelated module. anything else this pr needs to get in?

dmathieu:
Unfortunately, the lint job is a bit flaky at the moment 😢 
I reran it. But it seems the failure now is indeed related to your changes.

pellared (approved):

dmathieu (approved):

pellared:
@open-telemetry/go-maintainers PTAL

instrumentation/github.com/labstack/echo/otelecho/config.go
```diff
@@ -39,30 +38,29 @@ func (o optionFunc) apply(c *config) {
...
-// If none is specified, the global provider is used.
-func WithTracerProvider(provider oteltrace.TracerProvider) Option {
-	return optionFunc(func(cfg *config) {
-		if provider != nil {
-			cfg.TracerProvider = provider
-		}
+// WithSkipper specifies a skipper for allowing requests to skip generating spans.
+func WithSkipper(skipper func(c echo.Context) bool) Option {
+	return optionFunc(func(conf *config) {
+		conf.otelhttpOptions = append(conf.otelhttpOptions, otelhttp.WithFilter(func(request *http.Request) bool {
```
Aneurysm9, line 54: How does this interact if the user also provides an `otelhttp.WithFilter(...)` option?

instrumentation/github.com/labstack/echo/otelecho/echo.go
```diff
@@ -15,90 +15,63 @@
...
-	semconv "go.opentelemetry.io/otel/semconv/v1.10.0"
-	oteltrace "go.opentelemetry.io/otel/trace"
+	"go.opentelemetry.io/otel/trace"
 )
 
+type echoCtxKey int
+
 const (
-	tracerKey  = "otel-go-contrib-tracer-labstack-echo"
-	tracerName = "go.opentelemetry.io/contrib/instrumentation/github.com/labstack/echo/otelecho"
```
Aneurysm9, line 33: Will this no longer identify as the `otelecho` instrumentation?

instrumentation/github.com/labstack/echo/otelecho/echo.go
```diff
@@ -15,90 +15,63 @@
...
-				c.Error(err)
-			}
-
-			attrs := semconv.HTTPAttributesFromHTTPStatusCode(c.Response().Status)
-			spanStatus, spanMessage := semconv.SpanStatusFromHTTPStatusCodeAndSpanKind(c.Response().Status, oteltrace.SpanKindServer)
-			span.SetAttributes(attrs...)
-			span.SetStatus(spanStatus, spanMessage)
-
+			ctx := context.WithValue(c.Request().Context(), echoContextCtxKey, c)
+			otelhttp.NewHandler(handlerFunc, service, conf.otelhttpOptions...).
```
Aneurysm9, line 58: This is creating a new `otelhttp.Handler` for every request.  That seems like a lot of overhead.  Has this been benchmarked?

terev: I've addressed this in my latest changes. It now passes required echo context and next handler through context.

instrumentation/github.com/labstack/echo/otelecho/echo.go
terev: It no longer creates its own spans with a tracer so these aren't used anymore.

CHANGELOG.md
```diff
@@ -8,9 +8,20 @@ This project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0.htm
...
+  chain to tag all routes with `echo.Context.Path()`. (#2269)
+
+### Changed
+
+- `otelecho` middleware to wrap `otelhttp` tracer implementation rather than implement its own. This means that the `otelecho`
+  package now supports http server metrics provided by `otelhttp`. The following changes are considered breaking: (#2269)
+  - Span names differ from the previous version by default. However, passing `otelecho.WithSpanNameFormatter(otelecho.PathSpanNameFormatter)` as an option
+    will produce the previous behaviour.
+  - Route tags differ from the previous version by default. However, passing `otelecho.WithRouteTagFromPath()` will produce the
+    previous behaviour. Or add the `WithRouteTag()` middleware to your handler to tag a specific route differently.
```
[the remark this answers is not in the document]
terev: Was able to accomplish this in my latest changes.

matoous:
@terev just out of curiosity, was this closed for lack of review or another reason?

terev:
@matoous Yes mainly because it was not getting reviewed and possibly stale. Very happy to resubmit if others see value in it too.

matoous:
@terev I definitely do and would love to see this merged, also happy to contribute. Maybe @pellared could help get this through (please)?

pellared:
I am reopening it and assigning to myself. However, I guess I will have to time to look at it in about a month 😬 
Anyway I will mention the PR during the SIG meeting.
@terev Can you make update the PR before tomorrow SIG meeting?

PS. Sorry for inconvenience.

terev:
Thanks for trying to get this some love for us but I just noticed this change in main https://github.com/open-telemetry/opentelemetry-go-contrib/blob/main/instrumentation/net/http/otelhttp/handler.go#L65 . This makes me think my change is no longer needed. Having the ability to create a net/http middleware directly allows us to wrap it and use it without need for a specific package. Once this is released we should be able to do:

```golang
mw := echo.WrapMiddleware(otelhttp.NewMiddleware("operation", opts...))
server.Use(mw)
```

pellared:
@terev I think the "missing feature" of using `otelhttp.NewMiddleware` is that the route tags (like `assert.Contains(t, attrs, attribute.String("http.route", "/user/:id"))`) would be lost.

terev:
@pellared Right that's a good point. maybe we need a variant of WithRouteTag that can be used as a middleware as well? I'd be happy to make that change.

matoous:
@terev exactly as @pellared wrote. If you don't have time or capacity for it, please let me know, I would be happy to contribute.

terev:
@matoous @pellared Submitted a new pull request here https://github.com/open-telemetry/opentelemetry-go-contrib/pull/4017

PR #2270, opened 2022-05-11
Actually test the passed options in otelsarama

dmathieu:
This test is passing the default otel-provided tracer provider and propagators, which is the same as the default value.
So, if the method didn't do anything, it wouldn't be failing.

This change uses custom tracer providers and propagators to actually test the method.

### DIFF
--- a/instrumentation/github.com/Shopify/sarama/otelsarama/go.mod
+++ b/instrumentation/github.com/Shopify/sarama/otelsarama/go.mod
@@ -8,5 +8,6 @@ require (
 	github.com/Shopify/sarama v1.32.0
 	github.com/stretchr/testify v1.7.1
 	go.opentelemetry.io/otel v1.7.0
+	go.opentelemetry.io/otel/sdk v1.7.0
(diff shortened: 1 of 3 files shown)


instrumentation/github.com/Shopify/sarama/otelsarama/go.mod
```diff
@@ -8,5 +8,6 @@ require (
 	github.com/Shopify/sarama v1.32.0
 	github.com/stretchr/testify v1.7.1
 	go.opentelemetry.io/otel v1.7.0
+	go.opentelemetry.io/otel/sdk v1.7.0
```
pellared, line 11: the tests using the SDK are in `instrumentation/github.com/Shopify/sarama/otelsarama/test`

dmathieu: I can't move it into the test subpackage, as the `config` struct is private.
I've taken another approach, setting up a fake tracerprovider/tracer to avoid having the SDK.

pellared (approved):

MrAlias (approved):

PR #2310, opened 2022-05-23
Unify and improve GCP resource detection, second attempt

dashpole:
This PR supersedes https://github.com/GoogleCloudPlatform/opentelemetry-operations-go/pull/383 and https://github.com/open-telemetry/opentelemetry-go-contrib/pull/2223.  It was also discussed at the May 5th sig meeting.

It implements resource detection for GCE, GKE, GAE, Cloud Run, and Cloud Functions using a single detector.  The underlying logic is [implemented](https://github.com/GoogleCloudPlatform/opentelemetry-operations-go/tree/main/detectors/gcp) (and [integration tested](https://github.com/GoogleCloudPlatform/opentelemetry-operations-go/blob/e2cce7b04a10906fca4fd38b95ba824a085d61e2/e2e-test-server/endtoendserver/server.go#L153) in GCP environments) in [GoogleCloudPlatform/opentelemetry-operations-go](https://github.com/GoogleCloudPlatform/opentelemetry-operations-go).  The underlying library provides a function, CloudPlatform, which detects which platform it is on, and then one function for each attribute.  The reason for the split between "library" and detector is so that the library functions can be integration tested in real GCP environments, which is difficult to do in upstream repositories.

Unfortunately, the current resource detector was already marked stable, so we won't be able to remove existing types, etc.   I've added deprecation warnings to them to encourage users to move to the new implementation.

When users switch to the new detector, there are a few breaking changes:

* No longer supports detecting k8s.pod.name, container.name, k8s.namespace.name when running on k8s. These are expected to be set outside of this resource detector as described in the README.
* No longer sets service.namespace to "cloud-run-managed" on Cloud Run.
* Sets faas.name instead of service.name on Cloud Run.
* Sets faas.id instead of service.instance.id on Cloud Run.

Additions:

* Adds detector for Google App Engine
* Correctly sets cloud.platform on GKE and Cloud Run
* Detects cloud.availability_zone/region when running on GKE with workload identity enabled (the GCE metadata server is not accessible in that scenario)
* Detects faas.version on cloud run and cloud functions.

### DIFF
--- a/detectors/gcp/README.md
+++ b/detectors/gcp/README.md
@@ -1,0 +1,57 @@
+# GCP Resource detector
+
+The GCP resource detector supports detecting resources on:
+
+ * Google Compute Engine (GCE)
+ * Google Kubernetes Engine (GKE)
+ * Google App Engine (GAE)
+ * Cloud Run
+ * Cloud Functions
+
+## Usage
+
+```golang
+ctx := context.Background()
+// Detect your resources
+res, err := resource.New(ctx,
+    // Use the GCP resource detector!
+    resource.WithDetectors(gcp.NewDetector()),
+    // Keep the default detectors
+    resource.WithTelemetrySDK(),
+    // Add your own custom attributes to identify your application
+    resource.WithAttributes(
+        semconv.ServiceNameKey.String("my-application"),
+        semconv.ServiceNamespaceKey.String("my-company-frontend-team"),
+    ),
+)
+if err != nil {
+    // Handle err
+}
+// Use the resource in your tracerprovider (or meterprovider)
+tp := trace.NewTracerProvider(
+    // ... other options
+    trace.WithResource(res),
+)
+```
+
+## Setting Kubernetes attributes
+
+Previous iterations of GCP resource detection attempted to detect
+container.name, k8s.pod.name and k8s.namespace.name.  When using this detector,
+you should use this in your Pod Spec to set these using
+OTEL_RESOURCE_ATTRIBUTES:
+
+```yaml
+env:
+- name: POD_NAME
+  valueFrom:
+    fieldRef:
+      fieldPath: metadata.name
+- name: NAMESPACE_NAME
+  valueFrom:
+    fieldRef:
+      fieldPath: metadata.namespace
+- name: CONTAINER_NAME
+  value: my-container-name
+- name: OTEL_RESOURCE_ATTRIBUTES
+  value: k8s.pod.name=$(POD_NAME),k8s.namespace.name=$(NAMESPACE_NAME),k8s.container.name=$(CONTAINER_NAME)
--- a/detectors/gcp/README.md
+++ b/detectors/gcp/README.md
@@ -37,9 +37,9 @@ tp := trace.NewTracerProvider(
 ## Setting Kubernetes attributes
 
 Previous iterations of GCP resource detection attempted to detect
-container.name, k8s.pod.name and k8s.namespace.name.  When using this detector,
+`container.name`, `k8s.pod.name` and `k8s.namespace.name`.  When using this detector,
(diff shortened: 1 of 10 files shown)


detectors/gcp/README.md
```diff
@@ -0,0 +1,57 @@
...
+tp := trace.NewTracerProvider(
+    // ... other options
+    trace.WithResource(res),
+)
+```
+
+## Setting Kubernetes attributes
+
+Previous iterations of GCP resource detection attempted to detect
+container.name, k8s.pod.name and k8s.namespace.name.  When using this detector,
```
MrAlias, line 40:
```suggestion
`container.name`, `k8s.pod.name` and `k8s.namespace.name`.  When using this detector,
```

detectors/gcp/README.md
```diff
@@ -0,0 +1,57 @@
...
+you should use this in your Pod Spec to set these using
+OTEL_RESOURCE_ATTRIBUTES:
```
MrAlias, line 42:
```suggestion
`OTEL_RESOURCE_ATTRIBUTES`:
```

detectors/gcp/types.go
```diff
@@ -0,0 +1,39 @@
...
+// distributed under the License is distributed on an "AS IS" BASIS,
+// WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+// See the License for the specific language governing permissions and
+// limitations under the License.
+
+package gcp // import "go.opentelemetry.io/contrib/detectors/gcp"
+
+import "github.com/GoogleCloudPlatform/opentelemetry-operations-go/detectors/gcp"
+
+// gcpDetector can detect attributes of GCP environments
```
MrAlias, line 19:
```suggestion
// gcpDetector can detect attributes of GCP environments.
```

detectors/gcp/detector.go
```diff
@@ -0,0 +1,135 @@
...
+		return nil, err
+	}
+	attributes := []attribute.KeyValue{semconv.CloudProviderGCP, semconv.CloudAccountIDKey.String(projectID)}
+
+	switch d.detector.CloudPlatform() {
+	case gcp.GKE:
+		attributes = append(attributes, semconv.CloudPlatformGCPKubernetesEngine)
+		v, locType, err := d.detector.GKEAvailabilityZoneOrRegion()
+		if err != nil {
+			return nil, err
```
MrAlias, line 60: Based on [this](https://github.com/open-telemetry/opentelemetry-go/blob/a50cf6aadd582f9760c578e2c4b5230b6c30913d/sdk/resource/auto.go#L35-L40):

> 	// Detect returns an initialized Resource based on gathered information.
>	// If the source information to construct a Resource contains invalid
>	// values, a Resource is returned with the valid parts of the source
>	// information used for initialization along with an appropriately
>	// wrapped ErrPartialResource error.

Shouldn't this return a partial resource and a wrapped `ErrPartialResource` in these error cases?

MrAlias (approved):

dmathieu (approved):

dashpole:
@jaredjenkins can you expand on what you are unable to do?  The previous detector was using environment variables to set those attributes.

You can achieve the previous behavior by adding:

```yaml
- name: OTEL_RESOURCE_ATTRIBUTES
  value: k8s.pod.name=$(HOSTNAME),k8s.namespace.name=$(NAMESPACE),k8s.container.name=$(CONTAINER_NAME)
```

In your pod spec.

k8s.* attributes are not GCP-specific, and thus don't belong in the GCP detector.

dashpole:
Got it, that makes sense.  I think the best path forward would be https://github.com/open-telemetry/opentelemetry-go-contrib/issues/4136

PR #2339, opened 2022-06-06
fix(otelhttp): link incoming span context for public endpoints

### LINKED ISSUE #2338: bug(otelhttp): WithPublicEndpoint does not link incoming span context
The [otelhttp.WithPublicEndpoint](https://pkg.go.dev/go.opentelemetry.io/contrib/instrumentation/net/http/otelhttp#WithPublicEndpoint) should link parent span as per comment:
```go
// WithPublicEndpoint configures the Handler to link the span with an incoming
// span context. If this option is not provided, then the association is a child
// association instead of a link.
func WithPublicEndpoint() Option {
		c.SpanStartOptions = append(c.SpanStartOptions, trace.WithNewRoot())
	})
}
```

But there is no logic for linking, `trace.WithNewRoot()` completely discards parent span.

Probably there should be new boolean `config.SpanLinkParent` that will be set by this option and will explicitly link incoming span context if true.

ernado:
Fix #2338


instrumentation/net/http/otelhttp/test/handler_test.go
```diff
@@ -141,3 +142,47 @@ func TestHandlerRequestWithTraceContext(t *testing.T) {
...
 }
+
+func TestWithPublicEndpoint(t *testing.T) {
+	spanRecorder := tracetest.NewSpanRecorder()
+	provider := sdktrace.NewTracerProvider(
+		sdktrace.WithSpanProcessor(spanRecorder),
+	)
+	remoteSpan := trace.SpanContextConfig{
+		TraceID: trace.TraceID{0x01},
+		SpanID:  trace.SpanID{0x01},
```
MrAlias, line 153: Why not set the `Remote` field here?

```suggestion
		SpanID:  trace.SpanID{0x01},
		Remote: true,
```

MrAlias (approved):
Overall looks good to me, one small nit.

instrumentation/net/http/otelhttp/test/handler_test.go
ernado: Missed that, thank you!
Fixed

PR #2341, opened 2022-06-06
Update gcp detection library to v0.32.1, and fix GAE

dashpole:
Update github.com/GoogleCloudPlatform/opentelemetry-operations-go/detectors/gcp to v0.32.1, which fixes resource detection on google app engine standard.  Since detection of region differs between app engine standard and flex, split it into two sections.


pellared (approved):

MrAlias (approved):

MrAlias:
Merging with only two approvals from the same company as this is an updated of a dependency.

PR #2342, opened 2022-06-09
Setup PublicEndpointFn option

dmathieu:
Marking all endpoints as public is a pretty good way to avoid orphan spans.
However, this is an all-or-nothing situation, which is rarely accurate. For example, a public API could have both external and internal customers.
External customers' requests need to be considered as public, while internal ones need not to.

This introduces a `WithPublicEndpointFn` option, which allows setting a specific request as public or not based on the return value of a function.
A specific TraceContext baggage could then be set on all internal requests, to keep the context while discarding it for any request which doesn't have it.


instrumentation/net/http/otelhttp/config.go
```diff
@@ -108,7 +109,16 @@ func WithMeterProvider(provider metric.MeterProvider) Option {
...
 	return optionFunc(func(c *config) {
 		c.PublicEndpoint = true
-		c.SpanStartOptions = append(c.SpanStartOptions, trace.WithNewRoot())
+	})
+}
+
+// WithPublicEndpointFn runs with every request, and allows conditionnally
+// configuring the Handler to link the span with an incoming span context. If
+// this option is not provided or returns false, then the association is a
+// child association instead of a link.
```
dashpole, line 118: We should note that WithPublicEndpoint takes precedence over WithPublicEndpointFn

dashpole (commented):
I can see this being useful for https://github.com/kubernetes/kubernetes/issues/103186 :).

dashpole (approved):

hanyuancheung (approved):

instrumentation/net/http/otelhttp/test/handler_test.go
```diff
@@ -190,6 +190,87 @@ func TestWithPublicEndpoint(t *testing.T) {
...
+
+			rr := httptest.NewRecorder()
+			h.ServeHTTP(rr, r)
+			assert.Equal(t, 200, rr.Result().StatusCode)
+
+			// Recorded span should be linked with an incoming span context.
+			assert.NoError(t, spanRecorder.ForceFlush(ctx))
+			spans := spanRecorder.Ended()
+			tt.spansAssert(t, sc, spans)
+
```
hanyuancheung, line 273:
```suggestion
}

```

Aneurysm9 (approved):

Issue #2481, opened 2022-07-02
Request to Add Instrumentation for otelhttp


jeremyje:
## Background

**Package Link**: go.opentelemetry.io/contrib/instrumentation/net/http/otelhttp

Add instrumentation and tracing to net/http

### Why can this instrumentation not be included in the package itself?
No it's part of the go standard library
## Proposed Solution

Use the RequestCount metric already defined in otelhttp.

### Tracing

- attributes:
  - <add proposed attributes or remove>
- events:
  - <add proposed events or remove>
- links:
  - <add proposed links or remove>

### Metrics

Instruments

- <instrument name>: <describe what the instrument will measure>
  - type: <propose instrument type information>
  - unit: <propose instrument unit>
  - description: <propose instrument description>
  - attributes:
    - <add proposed attributes or remove>

### Prior Art

- <list other established instrumentation for this package that can be referenced>

## Tasks

- Code complete:
  - [ ] Comprehensive unit tests.
  - [ ] End-to-end integration tests.
  - [ ] Tests all passing.
  - [ ] Instrumentation functionality verified.
- Documented
  - [ ] Added to the [OpenTelemetry Registry](https://opentelemetry.io/registry/)
  - [ ] README included for the module describing high-level purpose.
  - [ ] Complete documentation of all public API including package documentation.
  - [ ] [Instrumentation documentation](https://github.com/open-telemetry/opentelemetry-go-contrib/blob/main/instrumentation/README.md#instrumentation-packages) updated.
- Examples
  - [ ] `Dockerfile` file to build example application.
  - [ ] `docker-compose.yml` to run example in a docker environment to demonstrate instrumentation.

PR #2482, opened 2022-07-02
Support http.server.request_count metric in otelhttp.

jeremyje:
Fixes https://github.com/open-telemetry/opentelemetry-go-contrib/issues/2481

### DIFF
--- a/instrumentation/net/http/otelhttp/handler.go
+++ b/instrumentation/net/http/otelhttp/handler.go
@@ -100,6 +100,9 @@ func (h *Handler) createMeasures() {
 	h.counters = make(map[string]syncint64.Counter)
 	h.valueRecorders = make(map[string]syncfloat64.Histogram)
 
+	requestCounter, err := h.meter.SyncInt64().Counter(RequestCount)
+	handleErr(err)
+
 	requestBytesCounter, err := h.meter.SyncInt64().Counter(RequestContentLength)
 	handleErr(err)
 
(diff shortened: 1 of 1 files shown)


hanyuancheung (approved):

instrumentation/net/http/otelhttp/handler.go
```diff
@@ -202,6 +206,7 @@ func (h *Handler) ServeHTTP(w http.ResponseWriter, r *http.Request) {
 
 	// Add metrics
 	attributes := append(labeler.Get(), semconv.HTTPServerMetricAttributesFromHTTPRequest(h.operation, r)...)
+	h.counters[RequestCount].Add(ctx, 1, attributes...)
```
Aneurysm9, line 209: Is this necessary?  Can't the same information be extracted from the count of the server latency histogram?

Aneurysm9 (changes requested):

instrumentation/net/http/otelhttp/handler.go
RangelReale: I'm using Dynatrace, at least in their UI I can't seem to find a way to get the count out of the `http.server.duration` metric, is it required to be available by opentelemetry?

RangelReale:
Can this keep to be considered? I have no good way of getting the request count on my provider without this metric, which seems to be required by the spec.

jeremyje:
I ended up deleting this PR unintentionally because I purged all my repositories. My list got way too long. 
If there's still interest in getting this merged I can recreate this PR but there was some pushback so it was never merged.

https://github.com/open-telemetry/opentelemetry-go-contrib/pull/2482/files still fortunately shows the diff.

cc: @Aneurysm9 @RangelReale @hanyuancheung

dmathieu:
I'm not seeing this metric being required in the spec.
https://github.com/open-telemetry/opentelemetry-specification/blob/main/specification/metrics/semantic_conventions/http-metrics.md

RangelReale:
> I'm not seeing this metric being required in the spec. https://github.com/open-telemetry/opentelemetry-specification/blob/main/specification/metrics/semantic_conventions/http-metrics.md

Hmm you're right, I think I was confusing it with another thing. It isn't even mentioned there.

The only thing I'm afraid is adding it to my code, and in the future this being added to the library and it counting the request 2 times.

Issue #2259, opened 2022-05-04
jaeger.sampler.UpdateSampler() log output


tttoad:
## Background

The log output statement in the UpdateSampler function is commented out.

## Proposed Solution
Maybe we can use the otel.Handle function to print the log

tttoad:
![image](https://user-images.githubusercontent.com/14021602/167790083-16ee3803-e8db-404d-9760-107141f354d7.png)
In code  https://github.com/open-telemetry/opentelemetry-go-contrib/blob/main/samplers/jaegerremote/sampler_remote.go#L149 , the output of the log is commented out...
The log should be output.

PR #2566, opened 2022-07-09
Add logging options to jaeger.sampler

tttoad:
Add logging options to jaeger.sampler.
#2259


Aneurysm9:
I second the recommendation to use `Logr`.  I do not want to see each and every instrumentation creating and maintaining a logging interface.  This is something that we should do more to support at the API level, but in the absence of specification I think that adding a `WithLogger(logr.Logger)` Option to this instrumentation will allow the end user to decide which logger to use and how to configure it without requiring that we take on a significant maintenance overhead.

tttoad:
> I second the recommendation to use `Logr`. I do not want to see each and every instrumentation creating and maintaining a logging interface. This is something that we should do more to support at the API level, but in the absence of specification I think that adding a `WithLogger(logr.Logger)` Option to this instrumentation will allow the end user to decide which logger to use and how to configure it without requiring that we take on a significant maintenance overhead.

I agree to use a unified logging interface...
If we choose to use `logr`, we need a uniform specification on how to set the level.
"log.info" is not used in `jaeger.remote, so I'll try using `logr`.

Aneurysm9:
I'm not sure I understand the concern.  The `logr.Logger` interface provides `Info()` and `Error()` methods, which should be sufficient for this use.  If more granularity is required then the `V()` interface can be used and the OTel log data model provides a [mapping of severity levels to severity names](https://github.com/open-telemetry/opentelemetry-specification/blob/main/specification/logs/data-model.md#displaying-severity).

tttoad:
I'm not sure I understand it correctly.
logr.Logger is a struct.When using the `WithLogger(logr.Logger)` option, the end user needs to implement the `logr.LogSink` interface.
When using the `WithLogger(logr.Logger)` option, the end user needs to implement the `logr.LogSink` interface.
The `LogSink` interface is as follows.
```
type LogSink interface {
	Init(info RuntimeInfo)
	Enabled(level int) bool
	Info(level int, msg string, keysAndValues ...interface{})
	Error(err error, msg string, keysAndValues ...interface{})
	WithValues(keysAndValues ...interface{}) LogSink
	WithName(name string) LogSink
}
```
The end-user needs to clearly understand the `v-level`. (It is possible that the user also uses logr, but the definition of `v-level` is different).
https://github.com/go-logr/logr#what-if-my-v-levels-differ-between-libraries
I prefer to provide the following interface:
```
type Logger interface {
	Info(msg string,values ...interface{})
        Error(err error,msg string,values ...interface{}) 
	Debug(msg string,values ...interface{})
        .....
}
```

tttoad:
@dmathieu @Aneurysm9 PTAL. Should I add changelog or use skiplog label？

samplers/jaegerremote/sampler_remote_options.go
```diff
@@ -112,6 +118,13 @@ func WithSamplingRefreshInterval(samplingRefreshInterval time.Duration) Option {
 	})
 }
 
+// WithSamplingLogger creates a Option that initializes sampling logger.
+func WithSamplingLogger(logger Logger) Option {
```
MrAlias, line 122: This should accept a `"github.com/go-logr/logr".Logger` (similar to [this](https://github.com/open-telemetry/opentelemetry-go/blob/096a1624184285edbf9c82c53ea0b068895915ef/internal_logging.go#L23-L26)).

There should not be a need to export our own `Logger` interface here.

tttoad: done.

samplers/jaegerremote/sampler_remote_options.go
```diff
@@ -48,6 +54,7 @@ func newConfig(options ...Option) config {
 			MaxOperations:            defaultSamplingMaxOperations,
 			OperationNameLateBinding: defaultSamplingOperationNameLateBinding,
 		},
+		logger: stdr.NewWithOptions(log.New(os.Stderr, "", log.LstdFlags|log.Lshortfile), stdr.Options{LogCaller: stdr.Error}),
```
Aneurysm9, line 57: Use a no-op logger here.  It is bad form for libraries to be emitting output to STDOUT.  The user can provide a logger using the new Option.

samplers/jaegerremote/sampler_remote_options.go
```diff
@@ -112,6 +119,13 @@ func WithSamplingRefreshInterval(samplingRefreshInterval time.Duration) Option {
 	})
 }
 
+// WithSamplingLogger creates a Option that initializes sampling logger.
+func WithSamplingLogger(logger logr.Logger) Option {
```
Aneurysm9, line 123:
```suggestion
// WithgLogger creates a Option that initializes the logger.
func WithLogger(logger logr.Logger) Option {
```

The logger is not sampling unless the user has configured it to be, but that's beyond the scope of this interface.

samplers/jaegerremote/sampler_remote.go
```diff
@@ -105,7 +104,7 @@ func (s *Sampler) ShouldSample(p trace.SamplingParameters) trace.SamplingResult
 // go-routines it may have started.
 func (s *Sampler) Close() {
 	if swapped := atomic.CompareAndSwapInt64(&s.closed, 0, 1); !swapped {
-		otel.Handle(fmt.Errorf("repeated attempt to close the sampler is ignored"))
+		s.logger.Error(nil, "repeated attempt to close the sampler is ignored")
```
Aneurysm9, line 107:
```suggestion
		s.logger.Info("repeated attempt to close the sampler is ignored")
```

If there is no `error` then don't use `Error()`.  It can be emitted at the info level.  That said, why change this away from the error handling system?

tttoad: I don't know. It's just that the original code generated an error, so I used Error(). I also think using Info is a better choice.

samplers/jaegerremote/sampler_remote_options.go
tttoad: I think it should output to STDOUT. (aws/xray is the same)
If the user does not configure this option this error will not be visible.
I think most users will use the default option.

samplers/jaegerremote/sampler_remote_options.go
tttoad: You're right.

CHANGELOG.md
```diff
@@ -8,9 +8,9 @@ This project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0.htm
 
 ## [Unreleased]
 
-### Changed
-
```
MrAlias, line 12: Please return the deleted section of the changelog.

CHANGELOG.md
```diff
@@ -8,9 +8,9 @@ This project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0.htm
...
 - Upgraded all `semconv` package use to `v1.12.0`. (#2589)
+- The `samplers/jaegerremote` added Logger option.
+  Outputs the log of the sampler.UpdateSampler() function. (#2566)
```
MrAlias, line 13: This needs to be included in a new `Added` section.

MrAlias, line 13:
```suggestion
- The `WithLogger` option to `go.opentelemetry.io/contrib/samplers/jaegerremote` to allow users to pass a `logr.Logger` and have operations logged. (#2566)
```

samplers/jaegerremote/sampler_remote.go
```diff
@@ -105,7 +104,7 @@ func (s *Sampler) ShouldSample(p trace.SamplingParameters) trace.SamplingResult
 // go-routines it may have started.
 func (s *Sampler) Close() {
 	if swapped := atomic.CompareAndSwapInt64(&s.closed, 0, 1); !swapped {
-		otel.Handle(fmt.Errorf("repeated attempt to close the sampler is ignored"))
+		s.logger.Info("repeated attempt to close the sampler is ignored")
```
MrAlias, line 107: This changed from an error to an info log message? That does not seem appropriate.

samplers/jaegerremote/sampler_remote.go
```diff
@@ -151,20 +150,20 @@ func (s *Sampler) setSampler(sampler trace.Sampler) {
...
+		s.logger.Error(err, "failed to parse sampling strategy response")
 		return
 	}
 
 	s.Lock()
 	defer s.Unlock()
 
 	if err := s.updateSamplerViaUpdaters(strategy); err != nil {
-		//c.logger.Infof("failed to handle sampling strategy response %+v. Got error: %v", res, err)
+		s.logger.Error(err, fmt.Sprintf("failed to handle sampling strategy response %+v.", res))
```
MrAlias, line 166: Use the structure nature of the logger. This allows logging implementation to determine how they want to display this information.

```suggestion
		s.logger.Error(err, "failed to handle sampling strategy response", "response", res)
```

samplers/jaegerremote/sampler_remote_options.go
MrAlias: Agreed, please use a no-op logger here. Debugging will require a positive action to enable logging, the instrumentation should not spew output to a console by default. Especially since that is not the current behavior.

samplers/jaegerremote/sampler_remote_options.go
```diff
@@ -112,6 +119,13 @@ func WithSamplingRefreshInterval(samplingRefreshInterval time.Duration) Option {
 	})
 }
 
+// WithLogger creates a Option that initializes the logger.
```
MrAlias, line 122:
```suggestion
// WithLogger configures the sampler to log operation and debug information with logger.
```

samplers/jaegerremote/sampler_remote_test.go
```diff
@@ -112,6 +115,7 @@ func TestRemoteSamplerOptions(t *testing.T) {
 	initSampler := newProbabilisticSampler(0.123)
 	fetcher := new(fakeSamplingFetcher)
 	parser := new(samplingStrategyParserImpl)
+	logger := stdr.NewWithOptions(log.New(os.Stderr, "", log.LstdFlags|log.Lshortfile), stdr.Options{LogCaller: stdr.Error})
```
MrAlias, line 118: Please use `testr` instead: https://pkg.go.dev/github.com/go-logr/logr@v1.2.3/testr#New

samplers/jaegerremote/sampler_remote.go
tttoad: I think it's good to use info level here. The close function should only be called once, and multiple calls won't have any effect.

samplers/jaegerremote/logger.go
```diff
@@ -0,0 +1,40 @@
...
+// WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+// See the License for the specific language governing permissions and
+// limitations under the License.
+
+package jaegerremote // import "go.opentelemetry.io/contrib/samplers/jaegerremote"
+
+import "github.com/go-logr/logr"
+
+// NullLogger is implementation of the Logger interface that is no-op
+var NullLoggger = &nullLogger{}
```
MrAlias, line 20: Please use the [`Discard`](https://pkg.go.dev/github.com/go-logr/logr#Discard) logger instead of adding this here.

samplers/jaegerremote/sampler_remote.go
MrAlias: If you would like to make a behavior change please do so in a dedicated PR that has supporting and objective motivation.

Otherwise, please limit this PR to scope of adding a logger option to this sampler.

samplers/jaegerremote/logger.go
tttoad: done.

.github/dependabot.yml
```diff
@@ -1,651 +1,3 @@
...
-    labels:
-      - dependencies
-      - go
-      - Skip Changelog
-    schedule:
-      interval: weekly
-      day: sunday
+dbotconf generate: /home/lx/work/opentelemetry-go-contrib/samplers/jaegerremote/go.mod:9: require <<<<<<<: version "HEAD" invalid: must be of the form v1.2.3
+/home/lx/work/opentelemetry-go-contrib/samplers/jaegerremote/go.mod:12: usage: require module/path v1.2.3
+/home/lx/work/opentelemetry-go-contrib/samplers/jaegerremote/go.mod:16: require >>>>>>>: version "3b73c686d463b57fe457153bd702edb7242f1048" invalid: must be of the form v1.2.3
```
MrAlias, line 3: Please fix this generation error.

samplers/jaegerremote/sampler_remote.go
MrAlias: The original code generated an error because `otel.Handle` accepts an error, not a log message string. Please update to use @Aneurysm9 suggestion.

MrAlias (changes requested):

tttoad:
@MrAlias  I think I addressed all of your concerns. PTAL

MrAlias:
Please restore the dependabot configuration: https://github.com/open-telemetry/opentelemetry-go-contrib/pull/2566#discussion_r1013026356

tttoad:
> Please restore the dependabot configuration: [#2566 (comment)](https://github.com/open-telemetry/opentelemetry-go-contrib/pull/2566#discussion_r1013026356)

Sorry, I forgot to commit it. It has been restored.  PTAL

MrAlias (approved):

Aneurysm9 (approved):

PR #2571, opened 2022-07-12
Add IgnoreSpanWithNamesSampler

### LINKED ISSUE #355: Investigate GRPC Tracer to support filtering or exclude the Health Checking Protocol.

ymotongpoo:
Ignoring spans based on is the common case when the system contains a lot of endpoints.
One of those case is dropping gRPC health check requests. As described in [this issue](https://github.com/open-telemetry/opentelemetry-python-contrib/issues/373) (another repository), health check requests often pollutes the trace dashboard, which leads:

* Reducing discoverability of trace on the dashboard
* More costs in SaaS

Some instrumentation has hook points for filter spans, but having framework/middlware independent filter should help more general cases.

This fixes #355.


ymotongpoo:
@dmathieu thanks for the comments.

> I don't think any custom sampler needs to be in this repository. Especially when [the specification](https://github.com/open-telemetry/opentelemetry-specification/blob/fee583ec99d35693693c5a598b873d563256d70b/specification/trace/sdk.md#built-in-samplers) mentions a specific list of built-in samplers.

Could elaborate a bit more why this can be the rationale not to have custom samplers in this repository? Should all the instrumentation/propagation/detectors have to do with some 3rd party tools? Also, the spec is just listing up "built-in" samplers and I can't come up with the connection between it and the reason not to have custom samplers here. 

If yes, this feature could go under `otelgrpc`, because the initial motivation here is to exclude unnecessary health check traces generated by `otelgrpc`'s interceptors. (#355 and https://github.com/open-telemetry/opentelemetry-python-contrib/issues/373), but I tried to make it more general because filtering spans is very common use case and it doesn't make sense to me for all project to have the similar custom samplers at hand just to achieve filtering.

dmathieu:
I'm not saying your sampler doesn't have a use. But any code within this repository has to be maintained after it's added. So we can't add any and every sampler (or instrumentation/propagation/detector).

ymotongpoo:
Thank you Damien for the reply. I understand your point. I would like to hear your thoughts about the issue #355.

Do you think if a similar feature is pull-requested under `otelgrpc` to solve #355, does it make sense to you?

dmathieu:
It's only my personal opinion. But I do think an option for `otelgrpc` to have a way to filter out some operations from generating spans could be useful.
That was started in #512, so starting from the review comments in there would be ideal.

ymotongpoo:
Thank you Damien for the comment. Yes, I looked up #512 first and switched to more generic way as I stated above.

I'll mention both #355 and #512 for the next pr.

dashpole:
An approach similar to #512 is probably best.

If you want something generic today, you could write a filtering SpanProcessor similar to the example: https://github.com/open-telemetry/opentelemetry-go/blob/575e1bb27025c73fd76f1e6b9dc2727b85867fdc/sdk/trace/span_processor_filter_example_test.go#L82, but i'm not sure if that is something we would want to keep in contrib.  Other span processors have been published as their own repo: https://github.com/MrAlias/flow

ymotongpoo:
closing this because I created another PR #2572 that adopted the whole discussion here.

PR #2572, opened 2022-07-13
Add gRPC Filter

ymotongpoo:
After the discussion in #2571 and looking up the whole discussion in #512, I changed the direction just to focus on solving the issue #355 and other possible demand to filter the request to trace.

Most of the implementation are from #512. I tried to follow the package structure to match [`otelhttp.Filter`](https://github.com/open-telemetry/opentelemetry-go-contrib/blob/main/instrumentation/net/http/otelhttp/common.go) but because of the difference between `http.Request` and the objects passed to four types of interceptors, some clumsy structs/functions remain to the surface.


instrumentation/google.golang.org/grpc/otelgrpc/filters/interceptorinfo.go
```diff
@@ -0,0 +1,102 @@
...
+// InterceptorInfo is the union of all arguments to four types of
+// gRPC interceptors, except for function types and function arguments
+// of invoker and streamer:
+// * invoker  grpc.UnaryInvoker (UnaryClient)
+// * streamer grpc.Streamer (StreamClient)
+// * stream   grpc.ServerStream (StreamServer)
+// * handler  grpc.UnaryHandler | grpc.StreamHandler (UnaryServer, StreamServer)
+// * req, reply, srv interface{} (UnaryClient, UnaryServer, StreamClient)
+// * cc       *grpc.ClientConn.
+type InterceptorInfo struct {
```
dashpole, line 43: It feels a bit strange to make this and NewXInterceptorInfo functions public, as we don't expect users to ever need to use or call them.  Would it be better to leave filters in the otelgrpc package?  It also eliminates the `filters.Filter` stutter for naming.

dashpole (commented):
lgtm otherwise

instrumentation/google.golang.org/grpc/otelgrpc/filters/interceptorinfo.go
ymotongpoo: Yes, I totally agree! This is because I created `filters` package to follow the package structure in `otelhttp`. (It has `otelhttp.filters`) I moved around the NewXXXInterceptors and InterceptorInfo back and forth among `otelgrpc`, `otelgrpc/internal` and `otelgrpc/filters` to eliminate unnecessary public members, but I gave up and settled down to the current state. If it's totally ok to move everything under `otelgrpc` and `otelgrpc/internal` directories, then I'm happy to do it.

ymotongpoo:
@dashpole thanks for the comment.

@Aneurysm9 does it make sense to you to have everything for filter under `otelgrpc` and `otelgrpc/internal` to eliminate unnecessary public functions and struct? I'm asking you because you submit `otelhttp/filters` and also suggested the common object for filters in #512.

ymotongpoo:
@dashpole I moved everything under `otelgrpc` and now unnecessary public members are eliminated. Could you approve this?

ymotongpoo:
@dashpole Thank you David for the review and approval!

To maintainers: should I change anything else?

CHANGELOG.md
```diff
@@ -8,6 +8,10 @@ This project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0.htm
 
 ## [Unreleased]
 
+### Added
+
+- Add gRPC trace `Filter` to the `go.opentelemetry.io/contrib/instrumentation/google.golang.org/grpc/otelgrpc` package to provide the way to filter the traces automatically generated in interceptors. (#2572)
```
pellared, line 13: nit: This is a very long line 😄 Can you split it?

instrumentation/google.golang.org/grpc/otelgrpc/filters.go
```diff
@@ -0,0 +1,204 @@
...
+		case unaryServer:
+			p = splitFullMethod(i.usinfo.FullMethod)
+		case streamServer:
+			p = splitFullMethod(i.ssinfo.FullMethod)
+		case unaryClient, streamClient:
+			p = splitFullMethod(i.method)
+		default:
+			p = gRPCPath{
+				method: i.method,
+			}
```
pellared, line 109: I see this pattern repeated in 5 places. Consider removing the duplication?

pellared (approved):

CHANGELOG.md
MrAlias: The standard in this file is to put one sentence per line. This conforms with that style and should not be changed.

instrumentation/google.golang.org/grpc/otelgrpc/filters.go
```diff
@@ -0,0 +1,204 @@
...
+// true only if none of the Filters in the list return true.
+func None(fs ...Filter) Filter {
+	return func(i *interceptorInfo) bool {
+		for _, f := range fs {
+			if f(i) {
+				return false
+			}
+		}
+		return true
+	}
```
Aneurysm9, line 84:
```suggestion
	return Not(Any(fs...))
```

Isn't this the same?  If so, do we need the additional interface surface?

instrumentation/google.golang.org/grpc/otelgrpc/filters.go
```diff
@@ -0,0 +1,204 @@
...
+			}
+		}
+		return strings.HasPrefix(p.service, pre)
+	}
+}
+
+// NotHealthCheck returns a Filter that returns true if the request's
+// is not health check defined by gRPC Health Checking Protocol.
+// https://github.com/grpc/grpc/blob/master/doc/health-checking.md
+// This is the short of Not(MethodName("grpc.health.v1.Health/Check")).
```
Aneurysm9, line 201:
```suggestion
```

Seems irrelevant and already inaccurate.  No need to carry documentation details that are evident from the code.

instrumentation/google.golang.org/grpc/otelgrpc/filters/interceptorinfo.go
Aneurysm9: I put the filter implementations in a separate package because most often they'll be referred to by users and `filters.MethodPrefix()` makes more sense to me than `otelhttp.MethodPrefix()`, doubly so for the meta-filters.  `otelhttp.Any()` is less indicative of behavior than `filters.Any()`.  

As for exporting the interceptor info struct, I think it might be worthwhile to do so as without it users cannot implement their own filters.  That would also necessitate exporting its fields.  I think I'm fine with not exporting it for now, but if there's a way to ensure this isn't a one-way door and we can change our minds later that would be great.

ymotongpoo:
@pellared @Aneurysm9 thank you both for the comments and suggestions. I updated the branch and it should look better now.

instrumentation/google.golang.org/grpc/otelgrpc/filters.go
ymotongpoo: Thank you for pointing this out. I added a method to interceptorInfo to resolve this.

ymotongpoo (commented):
reply to the review comments

instrumentation/google.golang.org/grpc/otelgrpc/filters/interceptorinfo.go
dashpole: I think:

* `filters.MethodPrefix` feels better than `otelgrpc.MethodPrefix`
* `filters.Any` feels better than `otelgrpc.Any` (or `otelgrpc.FilterAny`)
* `otelgrpc.Filter` feels better than `filters.Filter`
* `otelgrpc.InterceptorInfo` feels better than `filters.InterceptorInfo`

To get that, we would need to make InterceptorInfo (and its fields) public, but could keep NewXInterceptorInfo private, which I'm OK with.  Would that work @Aneurysm9?

ymotongpoo:
@Aneurysm9 @dashpole I'm trying to move filters to the subsidiary `filters` package and keep `newXXXInterceptorInfo` private in `otelgrpc`, but that requires either of the followings to achieve the restructuring:

1. to expose all fields of `filters.InterceptorInfo` to create the object in `newXXXInterceptorInfo`
2. to prepare a public constructor method to create the `filters.InterceptorInfo` object

Either way make things uglier, and either of them has pros/cons.

* Option 1:
  * Implementation is straight forward, just replacing private field names with public ones
  * It unnecessary exposes fields
* Option 2:
  * We can keep the fields of `filter.InterceptorInfo` private
  * The constructor requires the values of all fields as arguments regardless of actual fields used (eg. for StreamServer, it only uses `ssinfo` and `typ`, so they need to explicitly fill the other arguments with zero values when calling the constructor.) 
  * The original motivation we keep `newXXXInterceptorInfo` functions private is not to expose the unnecessary functions for users, but the intention will be a bit ruined.

All filters and `InterceptorInfo` struct need to be in the same package, i.e. `filters` because it causes cyclic dependency if we keep `InterceptorInfo` in `otelgrpc`, so we need to figure out some ways to construct `filters.InterceptorInfo` objects out of the package.

jmacd (approved):

ymotongpoo:
@Aneurysm9 Kindly ping for your comments on the visibility of the fields `IntercepterInfo`, as I can't move forward without your comments. It's been 3 weeks since the beginning of the thread and it would be great if I can make this PR get ready for the merge.

ymotongpoo:
@MrAlias thank you for arranging this PR as the milestone. Could you help accelerate the review? I'm ready to change the code but I can't without the review from @Aneurysm9 but I have no idea how I can get help but for waiting.

instrumentation/google.golang.org/grpc/otelgrpc/interceptorinfo.go
```diff
@@ -0,0 +1,137 @@
...
+// * invoker  grpc.UnaryInvoker (UnaryClient)
+// * streamer grpc.Streamer (StreamClient)
+// * stream   grpc.ServerStream (StreamServer)
+// * handler  grpc.UnaryHandler | grpc.StreamHandler (UnaryServer, StreamServer)
+// * req, reply, srv interface{} (UnaryClient, UnaryServer, StreamClient)
+// * cc       *grpc.ClientConn.
+type interceptorInfo struct {
+	ctx    context.Context
+	method string
+	desc   *grpc.StreamDesc
```
MadVikingGod, line 48: Is this used anywhere?  I see a few places it's set, but I don't think it's accessed by any filters.

ymotongpoo: Thanks for the comment! As you point out, it's not used in this PR! And first I left it so that others can utilize it later for other filters, `grpc.StreamDesc` is only used in the initialization to register stream, so we can remove this.

ymotongpoo:
@Aneurysm9 could you reply to this comment so that I can move forward to the acceptable change?
https://github.com/open-telemetry/opentelemetry-go-contrib/pull/2572#issuecomment-1197595873

Aneurysm9:
Apologies for the delay, I was on vacation.

> All filters and `InterceptorInfo` struct need to be in the same package, i.e. `filters` because it causes cyclic dependency if we keep `InterceptorInfo` in `otelgrpc`, so we need to figure out some ways to construct `filters.InterceptorInfo` objects out of the package.

Not necessarily.  There could be a third package that contains `InterceptorInfo`.  This would be similar to the `otelhttp` instrumentation that uses the `http.Request` as an argument to its filters.  

It might also be possible to, as @dashpole suggested, use `otelgrpc.Filter` as the type and `filter.*` implement that type.  I think that would then only have `otelgrpc` depending on itself, `filter` depending on `otelgrpc`, and then user code depending on both, as required.

ymotongpoo:
@Aneurysm9 I appreciate your comment. I hope you had a great vacation.

> Not necessarily. There could be a third package that contains `InterceptorInfo`. This would be similar to the `otelhttp` instrumentation that uses the `http.Request` as an argument to its filters.

The reason `http.Request` works in `otelhttp` is because it exposes its fields. If it's okay for `IntercepterInfo` to expose its fields as well, I'm +1 to the decision.

Also per declaring `otelgrpc.Filter` type, `filter.*` is referring to `interceptorInfo`'s fields as of current implementation, so from this point of view, `InterceptorInfo`'s fields needs to be public or accessible from public methods.

Given, does it sound good to expose the fields (I think preparing public methods is better)?

ymotongpoo:
@Aneurysm9 I moved all filters under `filters` package. Those filters are in `otelgrpc.Filter` type and uses `otelgrpc.InterceptorInfo`. Also, because `InterceptorInfo` got smaller than what it was, I removed `newXXXX` constructors and directly instantiated `InterceptorInfo`, which are enabled by exposing its fields. This expose was required for `fitler_test.go` as well.

instrumentation/google.golang.org/grpc/otelgrpc/interceptorinfo.go
```diff
@@ -0,0 +1,47 @@
...
+
+import (
+	"google.golang.org/grpc"
+)
+
+// interceptorType is the flag to define which gRPC interceptor
+// the interceptorInfo object is.
+type InterceptorType uint8
+
+const (
```
dashpole, line 25: Should we introduce a UndefinedClient type as the default?  Seems like generally a good practice since this is exposed to users so it doesn't default to UnaryClient.

dashpole (approved):
Great work, and thanks for your patience, @ymotongpoo!  Just a few nits.

instrumentation/google.golang.org/grpc/otelgrpc/interceptorinfo.go
ymotongpoo: Agreed. Because `InterceptorType` has server types as well, so how about `UndefinedInterceptor` as default?

dashpole: Yes, that makes more sense

ymotongpoo:
@Aneurysm9 Kindly ping because it's been a week since your last comment and I updated the code regarding it.

petenorth:
Open Telemetry in Go for grpc is unusable without this PR for our organization.

ymotongpoo:
@Aneurysm9 @MadVikingGod @MrAlias hi, all maintainers, let me know how I can proceed from here. It's been two weeks since @Aneurysm9's last comment and I'm totally stalling.

For interface simplicity, if we don't think about abstraction for future, we can simply use method name as common input for `otelgrpc.Filter` and eliminate `InterceptorInfo` itself, as commented in https://github.com/open-telemetry/opentelemetry-go-contrib/pull/512#discussion_r602181665.

Of course, if we are to use `context.Context` for more filters, we can add it in the current implementation of `InterceptorInfo`.

```go
type InterceptorInfo struct {
	Ctx              context.Context
	Method           string
	UnaryServerInfo  *grpc.UnaryServerInfo
	StreamServerInfo *grpc.StreamServerInfo
	Typ              InterceptorType
}
```

Another option is just to keep method name and context when we instantiate `InterceptorInfo`, to eliminate `InterceptorType` switch. This enables `InterceptorInfo` to be simpler:

```go
type InterceptorInfo struct {
    Ctx    context.Context
    // Method should be either of the followings:
    // - method passed to unary/stream client interceptor function
    // - info.FullMethodName, where info is Unary/StreamServerInfo passed to server interceptor function
    Method string
}
```

Aneurysm9 (approved):
Proposed a function name change.  Otherwise LGTM.  Implement the change now or as a follow-on PR (or tell me I'm crazy and bring receipts :) )

instrumentation/google.golang.org/grpc/otelgrpc/filters/filters.go
```diff
@@ -0,0 +1,158 @@
...
+	return func(i *otelgrpc.InterceptorInfo) bool {
+		p := splitFullMethod(i)
+		return strings.HasPrefix(p.service, pre)
+	}
+}
+
+// NotHealthCheck returns a Filter that returns true if the request's
+// is not health check defined by gRPC Health Checking Protocol.
+// https://github.com/grpc/grpc/blob/master/doc/health-checking.md
+func NotHealthCheck() otelgrpc.Filter {
```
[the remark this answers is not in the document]
ymotongpoo: Thank you for the comment! For readability perspective, I agree with your point and the shortcut I wanted is too small (i.e. `NotHealthCheck()` instead of `Not(HealthCheck())`), so I change this part as you suggest.

ymotongpoo:
@Aneurysm9 Thank you for the approval! Yes, I'm happy to create a follow-up PR for better implementation :)

instrumentation/google.golang.org/grpc/otelgrpc/interceptor.go
```diff
@@ -73,10 +73,19 @@ func UnaryClientInterceptor(opts ...Option) grpc.UnaryClientInterceptor {
 		invoker grpc.UnaryInvoker,
 		callOpts ...grpc.CallOption,
 	) error {
+		cfg := newConfig(opts)
+		i := &InterceptorInfo{
+			Method: method,
+			Typ:    UnaryClient,
+		}
+		if cfg.Filter != nil && !cfg.Filter(i) {
```
Jthomas54, line 81: This applies to all similar logic, but doesn't this negate the activation of the filter?  `HealthCheck()` would return `true` so this would evaluate as false, stopping the logic from triggering.  Perhaps it worked differently before code suggestions?

I just happened to be looking for similar logic and pulled this code in to play with.

ymotongpoo: As commented in https://github.com/open-telemetry/opentelemetry-go-contrib/pull/2672#issuecomment-1217311105, the issue is that the comment for `HealthCheck` is not right, and other filters are working as intended.

Aneurysm9: @ymotongpoo can you put up a quick PR to fix the comment?  Thanks.

Issue #2591, opened 2022-07-20
OpenTelemetry Go Launcher


JamieDanielson:
As discussed briefly in CNCF Slack and in the Go SIG meeting, we would like to contribute a vendor-neutral launcher to streamline configuration of the Go SDK.

Please review the [design document for the OpenTelemetry Go Launcher](https://docs.google.com/document/d/10X8pjFL3aXpF6s6BSTPK2LZpb__oAgi3Xj193OG6kS4/edit?usp=sharing) for further discussion. A branch (WIP) for further review of the code can be found in the [Honeycomb fork of this repo](https://github.com/honeycombio/opentelemetry-go-contrib/tree/launcher/launcher).

PR #2592, opened 2022-07-21
[feature] Design doc for Go Launcher

JamieDanielson:
During today's SIG meeting, we discussed moving the design doc linked on #2591 into a PR for a central location and ease of leaving feedback on specific sections of the doc.

A branch (WIP) for further review of the code can be found in the [Honeycomb fork of this repo](https://github.com/honeycombio/opentelemetry-go-contrib/tree/launcher/launcher).

---

**Note**: The "launcher/init" code has been moved into its own repo in the Honeycombio org, [honeycombio/otel-launcher-go](https://github.com/honeycombio/otel-launcher-go). Please direct any feedback there as we continue to iterate on it. As discussed with the SIG, we will revisit the contribution in a few months.

### DIFF
--- a/design/0000-launcher.md
+++ b/design/0000-launcher.md
@@ -19,7 +19,7 @@ The vendor-neutral launcher will live in the opentelemetry-go-contrib repo, and 
 ### Getting started
 
 ```shell
-go get github.com/opentelemetry-go-contrib/otel-launcher-go/launcher
+go get go.opentelemetry.io/contrib/launcher
 ```
 
 ### Configure
@@ -27,7 +27,7 @@ go get github.com/opentelemetry-go-contrib/otel-launcher-go/launcher
 Minimal setup - by default will send all telemetry to localhost:4317, with the ability to set environment variables for endpoint, headers, resource attributes, and more as listed in the configuration options noted below:
 
 ```go
-import "github.com/opentelemetry-go-contrib/otel-launcher-go/launcher"
+import "go.opentelemetry.io/contrib/launcher"
 
 func main() {
     lnchr, err := launcher.ConfigureOpentelemetry()
@@ -38,7 +38,7 @@ func main() {
 Or set headers directly in code instead:
 
 ```go
-import "github.com/opentelemetry-go-contrib/otel-launcher-go/launcher"
+import "go.opentelemetry.io/contrib/launcher"
 
 func main() {
     lnchr, err := launcher.ConfigureOpentelemetry(
@@ -86,7 +86,7 @@ Minimal setup, which sends to the Honeycomb endpoint and requires `HONEYCOMB_API
 
 ```go
 import (
-    "github.com/opentelemetry-go-contrib/otel-launcher-go/launcher"
+    "go.opentelemetry.io/contrib/launcher"
     _ "github.com/honeycombio/otel-launcher-go/launcher/honeycomb"
 )
 
@@ -100,7 +100,7 @@ Alternatively, to set the Honeycomb API key directly in code:
 
 ```go
 import (
-    "github.com/opentelemetry-go-contrib/otel-launcher-go/launcher"
+    "go.opentelemetry.io/contrib/launcher"
     "github.com/honeycombio/otel-launcher-go/launcher/honeycomb"
 )
 
--- a/design/0000-launcher.md
+++ b/design/0000-launcher.md
@@ -1,0 +1,122 @@
+# Launcher
+
+## What is it?
+
+The Go Launcher (name TBD) is a configuration layer that chooses default values for configuration options that many OpenTelemetry users would ultimately configure manually, allowing for minimal code to quickly instrument with OpenTelemetry.
+
+## Motivation
+
+The goal of this Launcher is to help users that aren't familiar with OpenTelemetry quickly ramp up on what they need to get going and instrument their applications. [There is currently a lot of boilerplate code required to use OpenTelemetry](https://opentelemetry.io/docs/instrumentation/go/manual/#initializing-a-new-tracer), and users must make decisions early on before instrumentation works - for example, they must determine and find the appropriate exporters, set resource attributes, etc. This launcher is intended to simplify the setup and minimize decision-making overhead to be able to quickly get started.
+
+Separately, vendors may choose to provide a vendor-specific package compatible with this launcher for easier exporting to their own backend. Neither the launcher nor the vendor-specific package are required to use OpenTelemetry, and using the launcher does not require the end user to use a vendor-specific package.
+
+## Explanation
+
+The Go Launcher provides a single function to reduce the amount of code that needs to be written to get started, and provides additional, straightforward options to add in that still make additional configuration easy.
+
+The vendor-neutral launcher will live in the opentelemetry-go-contrib repo, and vendors may create separately hosted vendor-specific configuration packages that can be used with the launcher.
+
+### Getting started
+
+```shell
+go get github.com/opentelemetry-go-contrib/otel-launcher-go/launcher
+```
+
+### Configure
+
+Minimal setup - by default will send all telemetry to localhost:4317, with the ability to set environment variables for endpoint, headers, resource attributes, and more as listed in the configuration options noted below:
+
+```go
+import "github.com/opentelemetry-go-contrib/otel-launcher-go/launcher"
+
+func main() {
+    lnchr, err := launcher.ConfigureOpentelemetry()
+    defer lnchr.Shutdown()
+}
+```
+
+Or set headers directly in code instead:
+
+```go
+import "github.com/opentelemetry-go-contrib/otel-launcher-go/launcher"
+
+func main() {
+    lnchr, err := launcher.ConfigureOpentelemetry(
+        launcher.WithServiceName("service-name"),
+        launcher.WithHeaders(map[string]string{
+            "service-auth-key": "value",
+            "service-useful-field": "testing",
+        }),
+    )
+    defer lnchr.Shutdown()
+}
+```
+
+### Configuration Options
+
+| Config Option               | Env Variable                        | Required | Default              |
+| --------------------------  | ----------------------------------- | -------- | -------------------- |
+| WithServiceName             | OTEL_SERVICE_NAME                   | y        | -                    |
+| WithServiceVersion          | OTEL_SERVICE_VERSION                | n        | -                    |
+| WithHeaders                 | OTEL_EXPORTER_OTLP_HEADERS          | n        | {}                   |
+| WithTracesExporterEndpoint  | OTEL_EXPORTER_OTLP_TRACES_ENDPOINT  | n        | localhost:4317       |
+| WithTracesExporterInsecure  | OTEL_EXPORTER_OTLP_TRACES_INSECURE  | n        | false                |
+| WithMetricsExporterEndpoint | OTEL_EXPORTER_OTLP_METRICS_ENDPOINT | n        | localhost:4317       |
+| WithMetricsExporterInsecure | OTEL_EXPORTER_OTLP_METRICS_INSECURE | n        | false                |
+| WithLogLevel                | OTEL_LOG_LEVEL                      | n        | info                 |
+| WithPropagators             | OTEL_PROPAGATORS                    | n        | tracecontext,baggage |
+| WithResourceAttributes      | OTEL_RESOURCE_ATTRIBUTES            | n        | -                    |
+| WithMetricsReportingPeriod  | OTEL_EXPORTER_OTLP_METRICS_PERIOD   | n        | 30s                  |
+| WithMetricsEnabled          | OTEL_METRICS_ENABLED                | n        | true                 |
+| WithTracesEnabled           | OTEL_TRACES_ENABLED                 | n        | true                 |
+| WithProtocol                | OTEL_EXPORTER_OTLP_PROTOCOL         | n        | grpc                 |
+
+### Additional Configuration Options
+
+Not yet implemented but still desired configuration options would include:
+
+- `OTEL_EXPORTER_OTLP_PROTOCOL`: `http/protobuf`
+    - ideally, selecting this would change the default endpoint to localhost:4318
+
+### Using a Vendor-Specific Package
+
+Vendors may create and maintain convenience configuration packages to more easily setup for export to a specific backend. For example, using a Honeycomb package would look like this:
+
+Minimal setup, which sends to the Honeycomb endpoint and requires `HONEYCOMB_API_KEY` environment variable:
+
+```go
+import (
+    "github.com/opentelemetry-go-contrib/otel-launcher-go/launcher"
+    _ "github.com/honeycombio/otel-launcher-go/launcher/honeycomb"
+)
+
+func main() {
+    lnchr, err := launcher.ConfigureOpentelemetry()
+    defer lnchr.Shutdown()
+}
+```
+
+Or, with headers set directly in code:
+
+```go
+import (
+    "github.com/opentelemetry-go-contrib/otel-launcher-go/launcher"
+    "github.com/honeycombio/otel-launcher-go/launcher/honeycomb"
+)
+
+func main() {
+    lnchr, err := launcher.ConfigureOpentelemetry(
+        launcher.WithServiceName("service-name"),
+        honeycomb.WithApiKey("api-key"),
+    )
+    defer lnchr.Shutdown()
+}
+```
+
+## Trade-offs and mitigations
+
+This launcher is intentionally providing default configuration options, which may not precisely map out to the final configuration an end user desires. As such, there may be dependencies in the package that are not being used. For example, the default configuration will bring in both gRPC and HTTP exporters; using gRPC will result in HTTP being included but not used; similarly, using HTTP/protobuf will result in the gRPC dependency being pulled in but unused.
+
+There still exists the ability to change the default configuration with environment variables or in-code configuration options, but there is not an easy way to remove unused dependencies without also keeping the minimal configuration option.
+
+If removing unused dependencies is required, the user can avoid using the launcher and initialize OTel manually, which is the current experience today.
--- a/design/0000-launcher.md
+++ b/design/0000-launcher.md
@@ -56,7 +56,7 @@ func main() {
 
 | Config Option               | Env Variable                        | Required | Default              |
 | --------------------------  | ----------------------------------- | -------- | -------------------- |
-| WithServiceName             | OTEL_SERVICE_NAME                   | y        | -                    |
+| WithServiceName             | OTEL_SERVICE_NAME                   | n*       | unknown_service:go   |
 | WithServiceVersion          | OTEL_SERVICE_VERSION                | n        | -                    |
 | WithHeaders                 | OTEL_EXPORTER_OTLP_HEADERS          | n        | {}                   |
 | WithTracesExporterEndpoint  | OTEL_EXPORTER_OTLP_TRACES_ENDPOINT  | n        | localhost:4317       |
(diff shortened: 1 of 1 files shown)


design/0000-launcher.md
```diff
@@ -0,0 +1,121 @@
...
+    )
+    defer lnchr.Shutdown()
+}
+```
+
+### Configuration Options
+
+| Config Option               | Env Variable                        | Required | Default              |
+| --------------------------  | ----------------------------------- | -------- | -------------------- |
+| WithServiceName             | OTEL_SERVICE_NAME                   | y        | -                    |
```
MrAlias, line 59: When issues are created from this, we need to ensure this can also be set via `OTEL_RESOURCE_ATTRIBUTES`

design/0000-launcher.md
```diff
@@ -0,0 +1,121 @@
...
+| WithServiceVersion          | OTEL_SERVICE_VERSION                | n        | -                    |
+| WithHeaders                 | OTEL_EXPORTER_OTLP_HEADERS          | n        | {}                   |
+| WithTracesExporterEndpoint  | OTEL_EXPORTER_OTLP_TRACES_ENDPOINT  | n        | localhost:4317       |
+| WithTracesExporterInsecure  | OTEL_EXPORTER_OTLP_TRACES_INSECURE  | n        | false                |
+| WithMetricsExporterEndpoint | OTEL_EXPORTER_OTLP_METRICS_ENDPOINT | n        | localhost:4317       |
+| WithMetricsExporterInsecure | OTEL_EXPORTER_OTLP_METRICS_INSECURE | n        | false                |
+| WithLogLevel                | OTEL_LOG_LEVEL                      | n        | info                 |
+| WithPropagators             | OTEL_PROPAGATORS                    | n        | tracecontext,baggage |
```
MrAlias, line 67: Ideally this will be extensible. Possibly by using the [`autoprop`](https://pkg.go.dev/go.opentelemetry.io/contrib/propagators/autoprop) package to allow users to register their own propagators.

design/0000-launcher.md
```diff
@@ -0,0 +1,121 @@
...
+```
+
+### Configuration Options
+
+| Config Option               | Env Variable                        | Required | Default              |
+| --------------------------  | ----------------------------------- | -------- | -------------------- |
+| WithServiceName             | OTEL_SERVICE_NAME                   | y        | -                    |
+| WithServiceVersion          | OTEL_SERVICE_VERSION                | n        | -                    |
+| WithHeaders                 | OTEL_EXPORTER_OTLP_HEADERS          | n        | {}                   |
+| WithTracesExporterEndpoint  | OTEL_EXPORTER_OTLP_TRACES_ENDPOINT  | n        | localhost:4317       |
```
MrAlias, line 62: The `OTEL_TRACES_EXPORTER` and `OTEL_METRICS_EXPORTER` (and eventually `OTEL_LOGS_EXPORTER`) [environment variables](https://github.com/open-telemetry/opentelemetry-specification/blob/main/specification/sdk-environment-variables.md#exporter-selection) should be supported. Ideally, in the same way the propagators are: by allowing users to register their own and then call them here. This likely means an "autoexport" similar to `autoprop` needs to be created as an item for this project.

design/0000-launcher.md
```diff
@@ -0,0 +1,121 @@
...
+| WithTracesExporterInsecure  | OTEL_EXPORTER_OTLP_TRACES_INSECURE  | n        | false                |
+| WithMetricsExporterEndpoint | OTEL_EXPORTER_OTLP_METRICS_ENDPOINT | n        | localhost:4317       |
+| WithMetricsExporterInsecure | OTEL_EXPORTER_OTLP_METRICS_INSECURE | n        | false                |
```
MrAlias, line 65: These options seem generic across all exports yet they pair with OTLP exporter environment variables. How do we plan to support other exporter configuration?

MrAlias:
From the SIG meeting: The next steps for this are:

- Ratify this design by getting enough approvals to merge it
- Building a [project](https://docs.github.com/en/issues/trying-out-the-new-projects-experience/about-projects) to track all work needed.
- Build the project on a separate feature branch (not included in the SIG meeting, but should be done)
- Merge the completed project into the `main` branch and release.

design/0000-launcher.md
Aneurysm9: This will be important to get right, otherwise the promise of vendor neutrality is hollow and only applies to vendors that accept OTLP.

design/0000-launcher.md
```diff
@@ -0,0 +1,121 @@
...
+### Using a Vendor-Specific Package
+
+Vendors may create and maintain convenience configuration packages to more easily setup for export to a specific backend. For example, using a Honeycomb package would look like this:
+
+Minimal setup, which sends to the Honeycomb endpoint and requires `HONEYCOMB_API_KEY` environment variable:
+
+```go
+import (
+    "github.com/opentelemetry-go-contrib/otel-launcher-go/launcher"
+    _ "github.com/honeycombio/otel-launcher-go/launcher/honeycomb"
```
Aneurysm9, line 90: How does this work?  Does the `launcher` package export some API that can be used by the `honeycomb` package in `init()`, thus allowing the import for side effects pattern?  What is that API?

design/0000-launcher.md
```diff
@@ -0,0 +1,121 @@
...
+
+```go
+import (
+    "github.com/opentelemetry-go-contrib/otel-launcher-go/launcher"
+    "github.com/honeycombio/otel-launcher-go/launcher/honeycomb"
+)
+
+func main() {
+    lnchr, err := launcher.ConfigureOpentelemetry(
+        honeycomb.WithApiKey("api-key"),
```
Aneurysm9, line 109: I assume this would return a `launcher.Option`, probably by wrapping `launcher.WithHeaders()` like this:

```go
func WithAPIKey(key string) launcher.Option {
  return launcher.WithHeaders(map[string]string{"HONEYCOMB_API_KEY": key})
}
```

Would the `launcher.Option` type operate on an exported `launcher.Config` type allowing the implementation of wholly-new option implementations, or would everything need to be done in terms of existing `launcher.Option` implementations?  We've traditionally not exported the config type to avoid unintended/unsupportable behavior, but this might be a place where doing it make sense.

design/0000-launcher.md
```diff
@@ -0,0 +1,121 @@
...
+    )
+    defer lnchr.Shutdown()
+}
+```
+
+## Trade-offs and mitigations
+
+This launcher is intentionally providing default configuration options, which may not precisely map out to the final configuration an end user desires. As such, there may be dependencies in the package that are not being used. For example, the default configuration will bring in both gRPC and HTTP exporters; using gRPC will result in HTTP being included but not used; similarly, using HTTP/protobuf will result in the gRPC dependency being pulled in but unused.
```
Aneurysm9, line 117: Could we decouple the launcher framework from the included components and then provide stock distributions that the user could select from in the same way they'd select a vendor distribution?

```go
import (
  go.opentelemetry.io/contrib/launcher
  _ go.opentelemetry.io/contrib/launcher/otlp/grpc // or just /otlp for both HTTP and gRPC
)
```

This adds slightly more boilerplate, but also makes things somewhat more consistent.

design/0000-launcher.md
```diff
@@ -0,0 +1,121 @@
...
+            "service-useful-field": "testing",
+        }),
+    )
+    defer lnchr.Shutdown()
+}
+```
+
+### Configuration Options
+
+| Config Option               | Env Variable                        | Required | Default              |
```
Aneurysm9, line 57: I don't see any options related to the `BatchSpanProcessor`.  I'd assume that would be used and the user might need a way to modify it.  

This makes me wonder more generally what the transition path would be for a user who outgrows the launcher and wants to convert to more explicit initialization.  It would be awesome (though maybe overkill) if there was a `lnchr.Generate()` method that could emit the code that would be necessary to explicitly instantiate the launcher.

design/0000-launcher.md
```diff
@@ -0,0 +1,121 @@
...
+## Explanation
+
+The Go Launcher provides a single function to reduce the amount of code that needs to be written to get started, and provides additional, straightforward options to add in that still make additional configuration easy.
+
+The vendor-neutral launcher will live in the opentelemetry-go-contrib repo, and vendors may create separately hosted vendor-specific configuration packages that can be used with the launcher.
+
+### Getting started
+
+```shell
+go get github.com/opentelemetry-go-contrib/otel-launcher-go/launcher
```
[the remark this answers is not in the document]
JamieDanielson: Updated in https://github.com/open-telemetry/opentelemetry-go-contrib/pull/2592/commits/e96e2cc1089c0ba291097271b287ad90fd2d4fda

design/0000-launcher.md
JamieDanielson: The launcher uses the `BatchSpanProcessor` by default. There currently is not a plan to swap out the batch processor with, for example, the `SimpleSpanProcessor`. In our other distros (Java, .NET) we have been using the batch processor with no option to swap it out, and have not yet received any feedback requesting to make it configurable.

The `Generate()` function you mention does sound like a great stretch goal, but it may be too much for the initial implemenetation here.

design/0000-launcher.md
JamieDanielson: The goal would be to support all `OTEL_*` environment variables. The interface for setting configuration options in code, like `WithTracesExporterEndpoint`, would probably remain more limited.

design/0000-launcher.md
JamieDanielson: The initial intent was to have this specific to OTEL and OTLP, not to have it extend to other exporters, as OTLP is vendor-neutral but yes, would require accepting the OTLP format. Other exporter APIs could be added later on though. As a stop-gap, the collector can be used to translate to other formats.

design/0000-launcher.md
JamieDanielson: Yes, the goal is to support all `OTEL_*` environment variables and as such, honor a service name set in `OTEL_RESOURCE_ATTRIBUTES`. Agreed that it shouldn't really be required per se, as it should have the default of `unknown_service:<process>`. I will update this.

design/0000-launcher.md
JamieDanielson: This assumption is correct, that it would return a `launcher.Option`.

Regarding unintended/unsupportable behavior, vendors can add config validation for their own use cases. For example, our [vendor package (WIP)](https://github.com/honeycombio/otel-launcher-go/blob/mike/reorganise/honeycomb/honeycomb.go#L45) has a `launcher.ValidateConfig` that returns errors for invalid headers.

design/0000-launcher.md
JamieDanielson: We would still want to use the package that exposes both http and grpc launchers. As far as requiring the user to add their own imports here, we feel like this takes away from the point of the launcher. If someone feels really strongly about this they can use the SDK without the launcher.

design/0000-launcher.md
JamieDanielson: I've clarified the service name requirement in https://github.com/open-telemetry/opentelemetry-go-contrib/pull/2592/commits/e03964910c117672ef15f1d6a9c289c2068fc205

design/0000-launcher.md
JamieDanielson: If we're understanding this right, the `autoprop` package seems like the right move for specifying propagators, and later, the `autoexporter`(?) package will be useful for specifying exporters. Since `autoprop` is already merged in we'll be able to use that sooner rather than later, and for now keep the exporters limited to http and grpc as planned.

JamieDanielson:
Thanks for all the feedback here. Please let me know if there are any other items to address for the initial design document, or anything else I can do to help clarify any questions.

design/0000-launcher.md
Aneurysm9: I certainly wouldn't suggest using the `SimpleSpanProcessor`, but the `BatchSpanProcessor` does have several configuration options.  We should consider where the breakpoint should be that forces users to drop the simple setup path and use fully manual SDK configuration.  I think that adjusting batch size or timing should probably be on the simple side of that line.

dashpole:
I'm still not thrilled with the idea of a new code config surface when we already have one.  I would prefer a solution which:

* Doesn't introduce a new set of options for either the core library or for vendors (as GCP, I don't want to maintain 2 sets of options either).
* Can set up OpenTelemetry with a few lines `go` code.
* Can support "auto" vendor exporters and detector usage

I'll propose an alternative for consideration here:

### The launcher
Introduce a new package, `launcher`, which each has two functions:
```golang
func NewTracerProvider(opts ...sdktrace.TracerProviderOption) *TracerProvider { ...
```

By default, it uses an OTLP exporter, and keeps all other defaults.  This should make it possible to do:
```golang
otel.SetTracerProvider(launcher.NewTracerProvider())
defer otel.GetTracerProvider().Shutdown()
```

It is trivial for users to switch between this and the sdktrace TracerProvider.  However, it isn't easy for vendors to use the launcher (yet).

### Auto exporter config

Introduce a package which (TBD) allows configuration of ~any exporter without code changes.  It defaults to using the OTLP exporter.  Update the simple launcher above to use this exporter by default, which does not change default behavior of the simple launcher.

### Auto resource config

Introduce package which (TBD) allows configuration of ~any resource detector without code changes.  It defaults to using the default set of OTel detectors.  Update the simple launcher above to use this resource by default, which does not change default behavior of the simple launcher.

cartermp:
Bikeshedding alert:

What if instead of `launcher` it's called `otelinit`?

To me, `launcher` doesn't mean a whole lot. But `otelinit` does: initialize opentelemetry.

design/0000-launcher.md
```diff
@@ -0,0 +1,136 @@
...
+Minimal setup, which sends to the Honeycomb endpoint and requires `HONEYCOMB_API_KEY` environment variable:
+
+```go
+import (
+    "go.opentelemetry.io/contrib/launcher"
+    _ "github.com/honeycombio/otel-launcher-go/launcher/honeycomb"
+)
+
+func main() {
+    lnchr, err := launcher.ConfigureOpentelemetry()
```
dmathieu, line 109: I'm not super fond of silently imported packages. How about something like this:


```golang
lnchr, err := launcher.ConfigureOpentelemetry(
    honey.Configure(),
)
```

MikeGoldsmith: Yeah, we've had some feedback to suggest the same. We liked the idea behind it when reviewing other sources, eg sql drivers, but it's easy to miss and not see why it's not working as expected.

dmathieu: Both could be supported?

It seems to me that an anonymous import could easily be missed as well.
To help with that, the launcher could provide errors if something appears to be wrong, such as no vendor/provider being setup.

cartermp: Both are supported today - you can use our layer without the anonymous import and add a line of code similar to what you have. We (honeycomb) may consider not enabling the anonymous import, or at least making it not the documented/preferred way, since it can be easily missed. Good feedback that you feel the same way.

Stepping back a bit, how prescriptive do you think we (opentelemetry contributors) should be here? Should we define a set of must-have/must-not-haves for vendor layers to follow?

dmathieu: As this feature is meant to help folks get started, I think things should be as simple and clear as possible, and avoid providing several ways to do the same thing (which goes in opposition with my previous comment, I know).
It also feels to me that we should try to let users know about anything that we can detect which seems wrong, such as telemetry configured with no provider, or a missing service name for example.
Anything that may be confusing to users when the get started, and that we can automatically detect could be provided as an information to them.

JamieDanielson:
> Bikeshedding alert:
> 
> What if instead of `launcher` it's called `otelinit`?
> 
> To me, `launcher` doesn't mean a whole lot. But `otelinit` does: initialize opentelemetry.

[I've renamed Launcher to Initialization Layer](https://github.com/open-telemetry/opentelemetry-go-contrib/pull/2592/commits/735553361d395adb3a88e6d06d2551ba94b88f3c), `otelinit` in code, so we can see how that looks. Any objections?

design/0000-launcher.md
```diff
@@ -0,0 +1,136 @@
...
+
+Or set headers directly in code instead:
+
+```go
+import "go.opentelemetry.io/contrib/otelinit"
+
+func main() {
+    init, err := otelinit.ConfigureOpentelemetry(
+        otelinit.WithServiceName("service-name"),
+        otelinit.WithHeaders(map[string]string{
```
dashpole, line 54: Is this implying that there is a common set of configuration for all exporters?  What happens if I set WithHeaders and also specify the `stdout` exporter?

design/0000-launcher.md
```diff
@@ -0,0 +1,136 @@
...
+            "service-auth-key": "value",
+            "service-useful-field": "testing",
+        }),
+    )
+    defer init.Shutdown()
+}
+```
+
+### Configuration Options
```
dashpole, line 63: For expoter-options below, do these assume OTLP is the exporter, or would they be expected to work with "registered" exporters as well?

JamieDanielson:
This design has now been updated to include the `autoexport` package ([currently in another PR](https://github.com/open-telemetry/opentelemetry-go-contrib/pull/2753/files)) and the `autoprop` package. The intent is to use the configuration options available with those instead of re-creating them here, so the OTLP-specific exporter configuration options have been dropped from this as well.

This design should now allow for setting defaults (e.g. otlp/grpc/tracecontext/baggage) while still allowing for the ability to override - some override code configuration examples are provided for exporter and propagator. This should help keep the promise of minimal code necessary to initialize, without adding too much extra code surface to maintain, and avoiding the huge leap to manually configure various options for a more customized setup.

Please review the updated document and let me know if there are open questions that have not been addressed here that require changes to this doc. Thanks!

design/0000-launcher.md
JamieDanielson: Design has been updated to remove OTLP-specific configuration options as they would be handled with the exporter itself; the design now specifies using the autoexport package.

design/0000-launcher.md
JamieDanielson: Design has been updated to remove OTLP-specific configuration options as they would be handled with the exporter itself; the design now specifies using the autoexport package.

design/0000-launcher.md
JamieDanielson: Design has been updated to remove OTLP-specific configuration options as they would be handled with the exporter itself; the design now specifies using the autoexport package.

design/0000-launcher.md
JamieDanielson: Design has been updated to remove OTLP-specific configuration options as they would be handled with the exporter itself; the design now specifies using the autoexport package.

design/0000-launcher.md
JamieDanielson: Design has been updated to remove OTLP-specific configuration options as they would be handled with the exporter itself; the design now specifies using the autoexport package.

MikeGoldsmith:
@open-telemetry/go-approvers looking for feedback on this design doc now we've updated to use autoexport 👍🏻

design/0000-launcher.md
```diff
@@ -0,0 +1,120 @@
...
+
+Alternatively, set environment variables or set variables directly in code to override defaults:
+
+```go
+import "go.opentelemetry.io/contrib/otelinit"
+
+func main() {
+    init, err := otelinit.ConfigureOpentelemetry(
+        otelinit.WithTracesExporter(new trace.SpanExporter("jaeger")) // if using non-default exporter
+        otelinit.WithPropagators(b3.New()) // if using non-default propagator
```
seankhliao, line 54: do the imports for `trace` and `b3` come from api / sdk or something under `otelinit`?

JamieDanielson:
**Note**: The "launcher/init" code has been moved into its own repo in the Honeycombio org, [honeycombio/otel-launcher-go](https://github.com/honeycombio/otel-launcher-go). Please direct any feedback there as we continue to iterate on it. As discussed with the SIG, we will revisit the contribution in a few months.

PR #2593, opened 2022-07-22
Add TextMapPropagator function to autoprop

### LINKED ISSUE #2590: Request to expose string based configuration in autoprop module.
## Background

Currently autoprop module only supports setting the propagator either by passing actual propagators objects or via environment variable OTEL_PROPAGATOR.

This limits consumers of the module to expose these settings as external configuration strings.

 For example OTel Collector using this autoprop can only configure the autoprop module either by introducing 

1) it's own configuration semantics and later convert the configuration strings into propagator objects to pass them to the autoprop or 

2) Set the environment variables from withing the process by taking the values as configuration strings. 

But having an api that takes in a string with values equivalent to what gets passed in OTEL_PROPAGATOR let consumers expose the sdks settings as configuration. 

see https://github.com/open-telemetry/opentelemetry-collector/issues/5572

MrAlias:
Resolve #2590

The collector plans to add configuration that will use a propagator describe in a text configuration file. That text needs to be converted into a propagator. This adds the `GetTextMapPropagator` function to do that conversion.


Aneurysm9 (approved):

dashpole (approved):

propagators/autoprop/registry.go
```diff
@@ -115,3 +116,54 @@ func RegisterTextMapPropagator(name string, p propagation.TextMapPropagator) {
...
+// function for more information) or a default (tracecontext, baggage, b3,
+// b3multi, jaeger, xray, or ottrace).
+//
+// If "none" is included in the arguments, or no names are provided, the
+// returned TextMapPropagator will be a no-operation implementation.
+//
+// An error is returned for any un-registered names. The remaining, known,
+// names will be used to compose a TextMapPropagator that is returned with the
+// error.
+func GetTextMapPropagator(names ...string) (propagation.TextMapPropagator, error) {
```
pellared, line 132: Should it not be named `TextMapPropagator` (usage `autoprop.TextMapPropagator(...)`)

Reference: https://go.dev/doc/effective_go#Getters

MrAlias: sure

pellared (approved):

PR #2594, opened 2022-07-23
sarama: fix span name and add payload size sem conv

srikanthccv:
Updates the span name to follow messaging semantic conventions and adds message payload size attribute


hanyuancheung (approved):

XSAM (approved):
lgtm

instrumentation/github.com/Shopify/sarama/otelsarama/producer.go
```diff
@@ -245,6 +247,31 @@ func WrapAsyncProducer(saramaConfig *sarama.Config, p sarama.AsyncProducer, opts
...
+	}
+	var size int
+	if version >= 2 {
+		size = maximumRecordOverhead
+		for _, h := range msg.Headers {
+			size += len(h.Key) + len(h.Value) + 2*binary.MaxVarintLen32
+		}
+	} else {
+		size = producerMessageOverhead
+	}
```
Aneurysm9, line 265:
```suggestion
	size := producerMessageOverhead
	if version >= 2 {
		size = maximumRecordOverhead
		for _, h := range msg.Headers {
			size += len(h.Key) + len(h.Value) + 2*binary.MaxVarintLen32
		}
	}
```

Since `producerMessageOverhead` is a magic number only used here not sure whether it hurts to inline it into the `size` declaration. v0v

instrumentation/github.com/Shopify/sarama/otelsarama/producer.go
```diff
@@ -245,6 +247,31 @@ func WrapAsyncProducer(saramaConfig *sarama.Config, p sarama.AsyncProducer, opts
 	return wrapped
 }
 
+func msgPayloadSize(msg *sarama.ProducerMessage, kafkaVersion sarama.KafkaVersion) int {
```
Aneurysm9, line 250: There are a lot of magic numbers in here and it's not clear why/when `binary.MaxVarintLen32` or `binary.MaxVarintLen64` are used.  Some comments would be helpful.

dmathieu:
You mention you're updating the span name to follow semantic conventions.
Is there a spec specifying those conventions for kafka, or examples of other libraries using the new name?

srikanthccv:
> Is there a spec specifying those conventions for kafka, or examples of other libraries using the new name?

I didn't lookup other libraries. There is no spec specifying those conventions for kafka or (any specific message system).

dmathieu:
I'm not sure I understand how this is an update to "follow messaging semantic conventions" if no other libraries is doing the same thing, and no conventions are defined?

srikanthccv:
>if no other libraries is doing the same thing

There is more likely chance other instrumentations are also not following the semconv so I didn't bother looking at them.

>and no conventions are defined?

I am not sure if they are going to add naming conventions for every messaging system. There is a general guideline which says span name should `<destination (topic)> <operation name>`. More [here](https://github.com/open-telemetry/opentelemetry-specification/blob/main/specification/trace/semantic_conventions/messaging.md).

dmathieu (approved):

srikanthccv:
I think this has enough approvals and comments are addressed @open-telemetry/go-instrumentaiton-maintainers Can someone merge this?

srikanthccv:
Can this be merged so that It will be part of the next release?  I would like to use this feature.

PR #2617, opened 2022-08-01
[net/http] add http.server.active_requests metric

fatsheep9146:
Signed-off-by: Ziqi Zhao <zhaoziqi9146@gmail.com>

Enhance instrumentation library [otelhttp](https://github.com/open-telemetry/opentelemetry-go-contrib/tree/main/instrumentation/net/http/otelhttp) for `net/http` to satisfy opentelemetry [http server metric semantic conventions](https://github.com/open-telemetry/opentelemetry-specification/blob/main/specification/metrics/semantic_conventions/http-metrics.md#http-server).

This PR added missed metric `http.server.active_requests` for server

### DIFF
--- a/instrumentation/net/http/otelhttp/handler.go
+++ b/instrumentation/net/http/otelhttp/handler.go
@@ -111,8 +111,12 @@ func (h *Handler) createMeasures() {
 	serverLatencyMeasure, err := h.meter.SyncFloat64().Histogram(ServerLatency)
 	handleErr(err)
 
+	activeRequest, err := h.meter.SyncInt64().UpDownCounter(ActiveRequest)
+	handleErr(err)
+
 	h.counters[RequestContentLength] = requestBytesCounter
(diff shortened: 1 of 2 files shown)


instrumentation/net/http/otelhttp/handler.go
```diff
@@ -201,12 +205,15 @@ func (h *Handler) ServeHTTP(w http.ResponseWriter, r *http.Request) {
 	labeler := &Labeler{}
 	ctx = injectLabeler(ctx, labeler)
 
+	// Add metrics
+	attributes := append(labeler.Get(), semconv.HTTPServerMetricAttributesFromHTTPRequest(h.operation, r)...)
```
Aneurysm9, line 209: This breaks the use of `Labeler` to allow handlers to add labels to the instrumentation.

Aneurysm9 (changes requested):

PR #2623, opened 2022-08-05
fix jaegerremote parser to use jsonpb unmarshaler

tonybase:
Failed to parse sampling strategy response: 
json: cannot unmarshal string into Go struct field SamplingStrategyResponse.strategyType of type api_v2.SamplingStrategyType:
```json
{"strategyType":"PROBABILISTIC","probabilisticSampling":{"samplingRate":1},"operationSampling":{"defaultSamplingProbability":1,"defaultLowerBoundTracesPerSecond":0,"perOperationStrategies":[{"operation":"/health","probabilisticSampling":{"samplingRate":0}},{"operation":"/metrics","probabilisticSampling":{"samplingRate":0}}}],"defaultUpperBoundTracesPerSecond":0}}

```


samplers/jaegerremote/sampler_remote.go
```diff
@@ -26,6 +26,7 @@ import (
 	"sync/atomic"
 	"time"
 
+	"github.com/gogo/protobuf/jsonpb"
```
hanyuancheung, line 29:
```suggestion
	"github.com/gogo/protobuf/jsonpb"
	
```

hanyuancheung (approved):

samplers/jaegerremote/sampler_remote.go
tonybase: Because the generate uses gogoproto, Protobuf V2 cannot be used

hanyuancheung:
This PR should implement in your local-copy version, instead of being implemented in contrib. If you use jsonpb to Marshal, you should also copy and reimplement the jsonpb Unmarshal. But I don't think this is the standard way according to the otel specification. PTAL.

hanyuancheung (changes requested):
@MrAlias how do you think about it?

dmathieu:
This should include an automated test

PR #2643, opened 2022-08-08
WIP Add runtime/metrics source option to instrumentation/runtime metrics

### LINKED ISSUE #2624: Improvements to the instrumentation/runtime package
## Background

Go-1.16 introduced a new `runtime/metrics` package that provides official metric names for all of the Go runtime properties worthy of monitoring. Ideally, I think we should migrate from the current `instrumentation/runtime` metrics to use exactly the names given by the Go team. This will require deprecation of the old names, which is not directly supported by OTel's schema migration support.

The runtime/metrics API is well suited to OTel's asynchronous instrument model, except for the appearance of 4 histograms (1 of them a Gauge histogram). These issues are being tracked at https://github.com/open-telemetry/opentelemetry-specification/issues/2713 and https://github.com/open-telemetry/opentelemetry-specification/issues/2714.

For the non-histogram metric values, these could easily be turned into OTel asynchronous instruments and callbacks. However, this looks like an area where users may want some configurability. Which would users prefer to see:

1. One package that has all runtime metrics, and another package that has just the essentials?
2. One package that has all runtime metrics, where many of them are disabled by default?

I think I would prefer the second option. Ideally, I would like to see OTel specify "API Hints" for instrumentation to supply their own default Views. This would allow an instrumentation package to instrument more than is enabled by default. See https://github.com/open-telemetry/opentelemetry-specification/issues/2229

jmacd:
Fixes #2624 

The code in `instrumentation/metrics` was contributed a long time ago. The Go team has since introduced runtime metrics in an official way. This PR adds optional support for the new metrics. There is a problem in the old implementation because it was written before the current SDK specification; it is no longer correct to use synchronous instruments from async callbacks--where there are more than one reader concerned. This leaves the old code as-is and focuses on the desired future outcome without changing current default behavior.

I propose that we deprecate the old and adopt the new eventually, however there are a number of gotchas.

There are four histogram instruments, but we lack a way to record them asynchronously. Moreover, one is a Gauge Histogram which OTel has not specified.

https://github.com/open-telemetry/opentelemetry-specification/issues/2713
https://github.com/open-telemetry/opentelemetry-specification/issues/2714

There are questions about what is a UpDownCounter and what is a Gauge. They are all currently UpDownCounters, but that is no guarantee. Presently, if a non-cumulative metric is Uint64, it becomes UpDownCounter, and if the value is Float64 it becomes a Gauge. We have no way to test this Gauge path presently because the go-1.19 runtime currently produces all Uint64 metrics and Float64Hist metrics, no Float64 metrics.

Lastly, there are two cases in 1.19 where there are duplicate metric names, after removing units. This creates a technical problem for OTLP -> Prometheus -> OTLP; currently, the only way to do this that satisfies both groups' specifications is to use an empty unit string when the unit is in the metric name itself. Thus, when there is a conflict of this nature, the units are kept in the name and the OTel unit is empty. This is an area for development between OpenTelemetry and OpenMetrics/Prometheus.


instrumentation/runtime/doc.go
```diff
@@ -12,21 +12,80 @@
...
+// The following metrics are generated in go-1.19.
+//
+// Name                                                    Unit         Instrument
+// -----------------------------------------------------------------------------
+// process.runtime.go.cgo.go-to-c-calls                    {calls}      Counter[int64]
+// process.runtime.go.gc.cycles.automatic                  {gc-cycles}  Counter[int64]
+// process.runtime.go.gc.cycles.forced                     {gc-cycles}  Counter[int64]
+// process.runtime.go.gc.cycles.total                      {gc-cycles}  Counter[int64]
+// process.runtime.go.gc.heap.allocs.bytes                 (*)          Counter[int64]
+// process.runtime.go.gc.heap.allocs.objects               (*)          Counter[int64]
```
dashpole, line 54: Here is a note about prometheus naming I forgot in our earlier conversation, which may help us: Counts of a thing (e.g. objects, or requests) are [considered unit-less in prometheus](https://prometheus.io/docs/practices/naming/#metric-names), and thus aren't impacted by unit naming conventions.  For example, in this case, we could have the two metrics:

* `process.runtime.go.gc.heap.allocs` (unit = bytes), and 
* `process.runtime.go.gc.heap.allocs.objects` (unit = {objects})

which would produce:

* `process_runtime_go_gc_heap_allocs_bytes_total` (unit = bytes)
* `process_runtime_go_gc_heap_allocs_objects_total` (no unit)

The `{objects}` unit isn't round-tripped, but that isn't possible given prom's current unit standards, and also is better (I think) than no units at all.

dashpole: See https://github.com/open-telemetry/opentelemetry-collector-contrib/tree/main/pkg/translator/prometheus#metric-name for the current (alpha) strategy for unit handling in the prometheus receiver/exporter.

jmacd: Nice, will do.

instrumentation/runtime/doc.go
```diff
@@ -12,21 +12,80 @@
...
-//   runtime.go.mem.heap_idle     (bytes)    Bytes in idle (unused) spans
-//   runtime.go.mem.heap_inuse    (bytes)    Bytes in in-use spans
-//   runtime.go.mem.heap_objects  -          Number of allocated heap objects
-//   runtime.go.mem.heap_released (bytes)    Bytes of idle spans whose physical memory has been returned to the OS
-//   runtime.go.mem.heap_sys      (bytes)    Bytes of heap memory obtained from the OS
-//   runtime.go.mem.live_objects  -          Number of live objects is the number of cumulative Mallocs - Frees
-//   runtime.uptime               (ms)       Milliseconds since application was initialized
+//   process.runtime.go.cgo.calls         -          Number of cgo calls made by the current process
+//   process.runtime.go.gc.count          -          Number of completed garbage collection cycles
+//   process.runtime.go.gc.pause_ns       (ns)       Amount of nanoseconds in GC stop-the-world pauses
```
dashpole, line 23: Could we remove the _ns suffix (since it is implied by the unit already)?  In Prom, it would currently produce `process_runtime_go_gc_pause_ns_nanoseconds`.  Same comment for the total below.

instrumentation/runtime/doc.go
```diff
@@ -12,21 +12,80 @@
...
+// older implementation will be removed no sooner than January 2024.
+//
+// The following metrics are generated in go-1.19.
+//
+// Name                                                    Unit         Instrument
+// -----------------------------------------------------------------------------
+// process.runtime.go.cgo.go-to-c-calls                    {calls}      Counter[int64]
+// process.runtime.go.gc.cycles.automatic                  {gc-cycles}  Counter[int64]
+// process.runtime.go.gc.cycles.forced                     {gc-cycles}  Counter[int64]
+// process.runtime.go.gc.cycles.total                      {gc-cycles}  Counter[int64]
```
dashpole, line 52: Should we omit total from the name (since it is implied?)

instrumentation/runtime/doc.go
jmacd: Done.

instrumentation/runtime/doc.go
jmacd: Well, this is the _OLD_ library's behavior and I don't want to touch it.

instrumentation/runtime/doc.go
jmacd: This raises more issues, unfortunately, and we can't separate this from the ongoing debate about "direction" attributes. I studied these two metrics with a `.total`, and it turns out they are duplicative. For every `.total` metric there are a number of other metrics that sum to the total.

This leaves options: 
1. Drop the total metric? Declare that quantities like gc.cycles.forced and gc.cycles.automatic are separate?
2. Drop the total metric? Merge all the things it totals into the same metric with an attribute for each?

jmacd: For the record, I prefer to drop the total and combine all sub-components into a single metric. This will take some work.

jmacd: This is now done. There are 13 timeseries with name `process.runtime.go.memory.classes` and `class=...`, there are 2 timeseries with name `process.runtime.go.gc.cycles` and `cycle=automatic` or `cycle=forced`.

jmacd:
I feel this deserves more attention, and I don't feel the group has sufficient bandwidth while the metrics Alpha is still under development. I will bring this up for discussion, meanwhile will be testing a _new package_ copied from this PR in the lightstep/otel-launcher-go repo.

jmacd:
Lightstep will work to submit their downstream fork of this code to the upstream repository. For now, see https://github.com/lightstep/otel-launcher-go/tree/main/lightstep/instrumentation

PR #2646, opened 2022-08-12
gRPC: evaluate config only once

ash2k:
Currently options are evaluated twice on each request. Change code to evaluate options only a single time when an interceptor is created.


dashpole (approved):
Thanks!

ash2k:
A changelog entry seem to be unnecessary for this one? It's an internal optimization.

ash2k:
Conflict resolved, ready to be merged.

MrAlias (approved):

Issue #1028, opened 2021-08-23
[Performance] gRPC instrumentation: interceptor causes full payload to marshal into memory


a-feld:
[This call causes the proto to get marshalled into memory](https://github.com/open-telemetry/opentelemetry-go-contrib/blob/be89a1bdcc5b8696c914688b3c5c0e2d9e45ee84/instrumentation/google.golang.org/grpc/otelgrpc/interceptor.go#L52)

Stack Trace:
```
go.opentelemetry.io/collector/model/internal/data/protogen/collector/metrics/v1.(*ExportMetricsServiceRequest).Marshal (opentelemetry-collector/model/internal/data/protogen/collector/metrics/v1/metrics_service.pb.go:238)
google.golang.org/protobuf/internal/impl.legacyMarshal (google.golang.org/protobuf@v1.27.1/internal/impl/legacy_message.go:404)
google.golang.org/protobuf/proto.MarshalOptions.size (google.golang.org/protobuf@v1.27.1/proto/size.go:43)
google.golang.org/protobuf/proto.MarshalOptions.Size (google.golang.org/protobuf@v1.27.1/proto/size.go:26)
google.golang.org/protobuf/proto.Size (google.golang.org/protobuf@v1.27.1/proto/size.go:16)
github.com/golang/protobuf/proto.Size (github.com/golang/protobuf@v1.5.2/proto/wire.go:18)
go.opentelemetry.io/contrib/instrumentation/google.golang.org/grpc/otelgrpc.messageType.Event (go.opentelemetry.io/contrib/instrumentation/google.golang.org/grpc/otelgrpc@v0.21.0/interceptor.go:52)
go.opentelemetry.io/contrib/instrumentation/google.golang.org/grpc/otelgrpc.UnaryServerInterceptor.func1 (go.opentelemetry.io/contrib/instrumentation/google.golang.org/grpc/otelgrpc@v0.21.0/interceptor.go:336)
go.opentelemetry.io/collector/model/internal/data/protogen/collector/metrics/v1._MetricsService_Export_Handler (opentelemetry-collector/model/internal/data/protogen/collector/metrics/v1/metrics_service.pb.go:218)
google.golang.org/grpc.(*Server).processUnaryRPC (google.golang.org/grpc@v1.39.0/server.go:1292)
google.golang.org/grpc.(*Server).handleStream (google.golang.org/grpc@v1.39.0/server.go:1617)
[...]

PR #2647, opened 2022-08-12
Support proto v2 message format, drop v1 format

ash2k:
Fixes https://github.com/open-telemetry/opentelemetry-go-contrib/issues/1028.

Currently, the size is only calculated for v1 messages and not for v2. This is because type assertion checks for the v1 interface.

We have two options:

- Add type check for v2 interface and support both v1 and v2 messages.
- Drop support for v1 and support v2 only.

This PR proposes to drop v1 support as the API has been deprecated for quite a while and most users should have migrated to v2 already. This should fix the performance problem by never calculating the size for v1 messages. Instead, the size will be efficiently calculated for v2 messages.


jmacd (approved):

dmathieu (approved):

Aneurysm9 (approved):

ash2k:
Rebased to resolve conflicts. Ready for merge.

PR #2672, opened 2022-08-16
Remove filter negate

Jthomas54:
While attempting to use the new filter from #2572, I noticed that the `HealthCheck` filter still recorded spans even though it was meant to filter them out.  I believe it stemmed from a code suggestion that changed the methods name, but not the related logic.

I update the tests as best I could.

### DIFF
--- a/instrumentation/google.golang.org/grpc/otelgrpc/interceptor.go
+++ b/instrumentation/google.golang.org/grpc/otelgrpc/interceptor.go
@@ -78,7 +78,7 @@ func UnaryClientInterceptor(opts ...Option) grpc.UnaryClientInterceptor {
 			Method: method,
 			Typ:    UnaryClient,
 		}
-		if cfg.Filter != nil && !cfg.Filter(i) {
+		if cfg.Filter != nil && cfg.Filter(i) {
 			return invoker(ctx, method, req, reply, cc, callOpts...)
 		}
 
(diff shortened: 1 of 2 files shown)


dashpole:
cc @ymotongpoo

Aneurysm9 (changes requested):
I think the issue is actually that the comment didn't get updated. The implementation is correct and the `HealthCheck()` filter should return `true` for health checks and cause them to be instrumented.  Use `Not(HealthCheck())` to exclude them.

ymotongpoo:
@Aneurysm9 yes, I forgot to update the comment!

@Jthomas54 the behavior is correct and the comment for HealthCheck is written in opposite way. It should be:

```
// HealthCheck returns a Filter that returns true if the request's
// is health check defined by gRPC Health Checking Protocol.
// https://github.com/grpc/grpc/blob/master/doc/health-checking.md
```

Removing "not" from the comment should work.

Jthomas54:
Ah, didn't realize there was a `Not` option.  Perhaps this isn't the best place to ask, but if you'll indulge me a second.  The purpose of the filter is to stop the otelgrpc interceptor for creating a span correct?  If that is the case, it seems odd that any filter wouldn't appear to work without first passing it through the `Not` option.  I think is is probably more a question of API design rather than function perhaps.

ymotongpoo:
> The purpose of the filter is to stop the otelgrpc interceptor from creating a span correct?

You can use filter for that purpose (and I implemented the gRPC one for that purpose) but the intention of `otelgrpc.Filter` is opposite. It is the option to intercept spans that apply to the filter, which is consistent with [`otelhttp.Filter`](https://pkg.go.dev/go.opentelemetry.io/contrib/instrumentation/net/http/otelhttp#Filter)

> If that is the case, it seems odd that any filter wouldn't appear to work without first passing it through the Not option.

That is the reason why `cfg.Filter != nil` is the first condition. The default is to intercept all spans.

Jthomas54:
Background helps!  Thanks, closing.

PR #2730, opened 2022-09-05
[otelgin] support additional span options to otelgin middleware

hjweddie:
We have used otelgin in our production environment, and find it may be more easy to reach opentelemetry best practises if we can set additional attributes in span. Like hostname, os, platform and so on.


pellared:
@hjweddie Can you please sign the CLA?

hjweddie:
> @hjweddie Can you please sign the CLA?

okay, signed

hanyuancheung (approved):

dmathieu (commented):
Could you also add a test for this, checking that the attributes are indeed there?

Also, this is nice. But it means those attributes are always static. They can't be tied to the request.
How about doing the same thing as for the net/http instrumentation, with a [WithSpanOptions](https://github.com/open-telemetry/opentelemetry-go-contrib/blob/main/instrumentation/net/http/otelhttp/config.go#L138) config?

hjweddie:
> WithSpanOptions

I understand that it would be much nice to add a test for checking attributes. But I do not sure how could I test it in unittest, because I do not know how to get span attributes. Do you have any ideas ?

dmathieu:
Using the [tracetest package](https://github.com/open-telemetry/opentelemetry-go/tree/main/sdk/trace/tracetest), you can configure a test tracer which will record all emitted spans. You can then inspect and assert on them.

See this test as an example: https://github.com/open-telemetry/opentelemetry-go-contrib/blob/fe241f3098fdcd4769aea55c6caf5a2999d853e0/instrumentation/net/http/otelhttp/test/transport_test.go#L40

hjweddie:
> tracetest package

[tracetest package](https://github.com/open-telemetry/opentelemetry-go/tree/main/sdk/trace/tracetest) is awesome.

I have changed WithAttributes to WithSpanOptions and add unittest for it.

I thinks it may be nice to use tracetest to enrich test in otelgin, not just assert on trace id and span id.   It may be done in another pr

CHANGELOG.md
```diff
@@ -20,6 +20,7 @@ This project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0.htm
 - Add gRPC trace `Filter` to the `go.opentelemetry.io/contrib/instrumentation/google.golang.org/grpc/otelgrpc` package to provide the way to filter the traces automatically generated in interceptors. (#2572)
 - The `TextMapPropagator` function to `go.opentelemetry.io/contrib/propagators/autoprop`.
   This function is used to return a composite `TextMapPropagator` from registered names (instead of having to specify with an environment variable). (#2593)
+- Add `WithAttributes` to `go.opentelemetry.io/contrib/github.com/gin-gonic/gin/otelgin` package for setting additional attributes to otelgin middleware
```
dmathieu, line 23:
```suggestion
- Add `WithSpanOptions` to `go.opentelemetry.io/contrib/github.com/gin-gonic/gin/otelgin` package to configure span options. (#2730)
```

instrumentation/github.com/gin-gonic/gin/otelgin/gintrace.go
```diff
@@ -60,12 +60,15 @@ func Middleware(service string, opts ...Option) gin.HandlerFunc {
...
-			oteltrace.WithAttributes(semconv.HTTPServerAttributesFromHTTPRequest(service, c.FullPath(), c.Request)...),
-			oteltrace.WithSpanKind(oteltrace.SpanKindServer),
-		}
+		opts := cfg.SpanStartOptions
+		opts = append(opts,
+			[]oteltrace.SpanStartOption{
+				oteltrace.WithAttributes(semconv.NetAttributesFromHTTPRequest("tcp", c.Request)...),
+				oteltrace.WithAttributes(semconv.EndUserAttributesFromHTTPRequest(c.Request)...),
+				oteltrace.WithAttributes(semconv.HTTPServerAttributesFromHTTPRequest(service, c.FullPath(), c.Request)...),
+				oteltrace.WithSpanKind(oteltrace.SpanKindServer),
```
dmathieu, line 69: This order of append means the default options always get precedence. So it's impossible to override the span kind. That may be fine though.

instrumentation/github.com/gin-gonic/gin/otelgin/gintrace_test.go
```diff
@@ -105,3 +108,39 @@ func TestPropagationWithCustomPropagators(t *testing.T) {
...
+	w := httptest.NewRecorder()
+
+	attrs := trace.WithAttributes(
+		[]attribute.KeyValue{
+			attribute.String("foo", "bar"),
+		}...,
+	)
+
+	router := gin.New()
+	router.Use(Middleware("foobar", WithTracerProvider(provider), WithPropagators(b3), WithSpanOptions(attrs)))
```
dmathieu, line 127: It doesn't feel like propagators are need for the test to pass?

instrumentation/github.com/gin-gonic/gin/otelgin/gintrace_test.go
```diff
@@ -105,3 +108,39 @@ func TestPropagationWithCustomPropagators(t *testing.T) {
...
+
+func TestWithCustomSpanOptions(t *testing.T) {
+	spanRecorder := tracetest.NewSpanRecorder()
+	provider := sdktrace.NewTracerProvider(sdktrace.WithSpanProcessor(spanRecorder))
+	b3 := b3prop.New()
+
+	r := httptest.NewRequest("GET", "/user/123", nil)
+	w := httptest.NewRecorder()
+
+	attrs := trace.WithAttributes(
```
dmathieu, line 120: This isn't attributes anymore. How about naming the variable `opts`, and setting something different than attributes as well?

instrumentation/github.com/gin-gonic/gin/otelgin/gintrace.go
```diff
@@ -60,12 +60,15 @@ func Middleware(service string, opts ...Option) gin.HandlerFunc {
...
 		}()
 		ctx := cfg.Propagators.Extract(savedCtx, propagation.HeaderCarrier(c.Request.Header))
-		opts := []oteltrace.SpanStartOption{
-			oteltrace.WithAttributes(semconv.NetAttributesFromHTTPRequest("tcp", c.Request)...),
-			oteltrace.WithAttributes(semconv.EndUserAttributesFromHTTPRequest(c.Request)...),
-			oteltrace.WithAttributes(semconv.HTTPServerAttributesFromHTTPRequest(service, c.FullPath(), c.Request)...),
-			oteltrace.WithSpanKind(oteltrace.SpanKindServer),
-		}
+		opts := cfg.SpanStartOptions
+		opts = append(opts,
```
dmathieu, line 64: No need to set a variable with `cfg.SpanStartOptions`. You can do directly `opts = append(cfg.SpanStartOptions, ...`

hjweddie: fixed it

hjweddie:
I try to change WithSpanOptions to two funs,  WithSpanStartOptions and WithSpanEndOptions.
it may be usefull if we can overwrite setting with span end options

instrumentation/github.com/gin-gonic/gin/otelgin/gintrace_test.go
hjweddie: fixed it

instrumentation/github.com/gin-gonic/gin/otelgin/gintrace.go
hjweddie: is it necessary to support overwrite otelgin settings, like span kind and some http related attributes ?

instrumentation/github.com/gin-gonic/gin/otelgin/gintrace_test.go
```diff
@@ -105,3 +109,54 @@ func TestPropagationWithCustomPropagators(t *testing.T) {
...
+		),
+	}
+
+	endOpts := []trace.SpanEndOption{
+		trace.WithTimestamp(spanEndTime),
+		trace.WithStackTrace(true), // capture the error with stack trace
+	}
+
+	router := gin.New()
+	router.Use(Middleware("foobar", WithTracerProvider(provider), WithSpanStartOptions(startOpts...), WithSpanEndOptions(endOpts...)))
```
dmathieu, line 139: I think it would be easier to read if there were two tests, one for each option.

hjweddie: ok, I change it into two tests, one for span start options, one for span end options

dmathieu (approved):

hjweddie:
Is there anything for mering this pr that I missed ?

hjweddie:
@jmacd @MrAlias @Aneurysm9 @evantorrie @XSAM Anyone helps reviewing this pr ?

Aneurysm9:
> We have used otelgin in our production environment, and find it may be more easy to reach opentelemetry best practises if we can set additional attributes in span. Like hostname, os, platform and so on.

I'm confused by this.  The listed attributes would be resource attributes and not set at span creation time.  I'm not aware of any other instrumentation that includes such options.  What would they be used for that does not involve setting resource attributes on individual spans rather than on the resource given to the `TracerProvider` when the SDK is initialized?

instrumentation/github.com/gin-gonic/gin/otelgin/option.go
```diff
@@ -22,8 +22,10 @@ import (
 )
 
 type config struct {
-	TracerProvider oteltrace.TracerProvider
-	Propagators    propagation.TextMapPropagator
+	TracerProvider   oteltrace.TracerProvider
+	Propagators      propagation.TextMapPropagator
+	SpanStartOptions []oteltrace.SpanStartOption
```
flc1125, line 27: I suggest adjusting this to private. `spanStartOptions`.

`SpanEndOptions`  is the same.

instrumentation/github.com/gin-gonic/gin/otelgin/option.go
```diff
@@ -57,3 +59,19 @@ func WithTracerProvider(provider oteltrace.TracerProvider) Option {
 		}
 	})
 }
+
+// WithSpanStartOptions configures an additional set of
+// trace.SpanOptions, which are applied to each new span.
+func WithSpanStartOptions(opts ...oteltrace.SpanStartOption) Option {
+	return optionFunc(func(cfg *config) {
+		cfg.SpanStartOptions = append(cfg.SpanStartOptions, opts...)
```
flc1125, line 67: My understanding of "with" is to cover, rather than to append.

If it were to append, I might consider naming it: `AppendSpanStartOptions`.

> `WithSpanEndOptions`  is the same.

flc1125 (changes requested):

instrumentation/github.com/gin-gonic/gin/otelgin/gintrace.go
```diff
@@ -60,18 +60,20 @@ func Middleware(service string, opts ...Option) gin.HandlerFunc {
 			c.Request = c.Request.WithContext(savedCtx)
 		}()
 		ctx := cfg.Propagators.Extract(savedCtx, propagation.HeaderCarrier(c.Request.Header))
-		opts := []oteltrace.SpanStartOption{
-			oteltrace.WithAttributes(semconv.NetAttributesFromHTTPRequest("tcp", c.Request)...),
-			oteltrace.WithAttributes(semconv.EndUserAttributesFromHTTPRequest(c.Request)...),
-			oteltrace.WithAttributes(semconv.HTTPServerAttributesFromHTTPRequest(service, c.FullPath(), c.Request)...),
-			oteltrace.WithSpanKind(oteltrace.SpanKindServer),
-		}
+		opts := append(cfg.SpanStartOptions,
```
flc1125, line 63: Should we consider adjusting it to: `append(default..., cfg.SpanStartOptions)`

Because we might want to override the default configuration.

flc1125 (changes requested):

pellared:
>  find it may be more easy to reach opentelemetry best practises if we can set additional attributes in span. Like hostname, os, platform and so on.

These are not valid use cases. Hostname, os, platform should be set as resource attributes. I think that the instrumentation libraries should follow the semantic conventions. You can always create additional spans with your custom attributes.

I am closing this PR as it is very old.

PR #2731, opened 2022-09-05
Jaeger Propagator: Fix accepting short trace and span IDs

### LINKED ISSUE #2729: Jaeger propagator doesn't accept trace ids with lengths other than 64 and 128 bit
### Summary
The Jaeger propagator currently doesn't accept trace ids with lengths other than exactly 64 or 128 bit, although the [Jaeger Propagation Format](https://www.jaegertracing.io/docs/1.37/client-libraries/#value) specifies that servers must 0-left-pad received hex-strings shorter than 32 hex characters. This results in an interoperability issue with the Python OpenTracing Jaeger Client, leading to received trace IDs wrongly being discarded.

### Details

We've run into an interoperability issue with a Python gRPC client instrumented with the OpenTracing Jaeger client and a Go gRPC server instrumented with OpenTelemetry: Some of the server traces were not associated with the client traces. Looking into this, I realized that all propagation values with 64-bit trace IDs with a leading `0` were being discarded. Assuming random distribution of the trace ID, every 16th 64-bit trace ID is being dropped (all IDs with a leading hex `0`).

The Python Jaeger Client generally [encodes trace IDs as variable-length integers](https://github.com/jaegertracing/jaeger-client-python/blob/4.8.0/jaeger_client/codecs.py#L206-L220). This may not only lead to the observed single zeroes in the beginning being dropped, but may result in multiple hex zeroes at the beginning of the trace ID being omitted.

Given this observed issue and the Jaeger trace ID specification requiring 0-padding up to 128 bits, I suggest changing the [current implementation](https://github.com/open-telemetry/opentelemetry-go-contrib/blob/v1.9.0/propagators/jaeger/jaeger_propagator.go#L110-L117) to left-pad with a variable numbers of zeros to always reach 128 bits (if the trace ID is shorter than 128 bits). The current implementation only supports adding 0-padding for exactly 64 bits. I may later open a PR changing this.

mfrister:
Fixes #2729.

While implementing the fix for trace IDs, i noticed that span IDs have the same requirement, so I fixed them as well.

As an optimization, we could keep the special case for 64-bit tracing IDs with a constant padding. I'm not sure it's worth it, so I haven't implemented this for now.


propagators/jaeger/jaeger_data_test.go
```diff
@@ -162,12 +162,6 @@ var invalidExtractHeaders = []extractTest{
 			jaegerHeader: fmt.Sprintf("%s:%s:0:1", traceID32Str+"0000", spanIDStr),
 		},
 	},
-	{
-		name: "trace ID length is not 32 or 16",
```
Aneurysm9, line 166: Please replace this with a valid test.

hanyuancheung:
@mfrister Could you please sign the CLA?

mfrister:
> @mfrister Could you please sign the CLA?

Sure, I'm already working on it. Will need a corporate CLA, so this might take a few days.

propagators/jaeger/jaeger_data_test.go
mfrister: There is another test `trace ID length > 32` and the propagator now accepts all shorter trace IDs that were previously invalid (ignoring the special case for an empty one that's handled differently).

I've added a test for a 15 char hex string trace ID.

mfrister:
@Aneurysm9 I've got the CLA signed and I've added a changelog entry.

Aneurysm9 (approved):

CHANGELOG.md
```diff
@@ -12,6 +12,7 @@ This project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0.htm
 
 - Rename the `Typ` field of `"go.opentelemetry.io/contrib/instrumentation/google.golang.org/grpc/otelgrpc".InterceptorInfo` to `Type`. (#2688)
 - Use Go 1.19 as the default version for CI testing/linting. (#2675)
+- Fix the Jaeger propagator rejecting trace IDs that are both shorter than 128 bits and not exactly 64 bits long (while not being 0). Also fix the propagator rejecting span IDs shorter than 64 bits. This fixes compatibility with Jaeger clients encoding trace and span IDs as variable-length hex strings, [as required by the Jaeger propagation format](https://www.jaegertracing.io/docs/1.37/client-libraries/#value).
```
dmathieu, line 15:
```suggestion
- Fix the Jaeger propagator rejecting trace IDs that are both shorter than 128 bits and not exactly 64 bits long (while not being 0). Also fix the propagator rejecting span IDs shorter than 64 bits. This fixes compatibility with Jaeger clients encoding trace and span IDs as variable-length hex strings, [as required by the Jaeger propagation format](https://www.jaegertracing.io/docs/1.37/client-libraries/#value). (#2731)
```

dmathieu (approved):

CHANGELOG.md
mfrister: Thanks, added.

CHANGELOG.md
```diff
@@ -12,6 +12,7 @@ This project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0.htm
 
 - Rename the `Typ` field of `"go.opentelemetry.io/contrib/instrumentation/google.golang.org/grpc/otelgrpc".InterceptorInfo` to `Type`. (#2688)
 - Use Go 1.19 as the default version for CI testing/linting. (#2675)
+- Fix the Jaeger propagator rejecting trace IDs that are both shorter than 128 bits and not exactly 64 bits long (while not being 0). Also fix the propagator rejecting span IDs shorter than 64 bits. This fixes compatibility with Jaeger clients encoding trace and span IDs as variable-length hex strings, [as required by the Jaeger propagation format](https://www.jaegertracing.io/docs/1.37/client-libraries/#value). (#2731)
```
MrAlias, line 15:
```suggestion

### Fixed

- Fix the Jaeger propagator rejecting trace IDs that are both shorter than 128 bits and not exactly 64 bits long (while not being 0).
  Also fix the propagator rejecting span IDs shorter than 64 bits.
  This fixes compatibility with Jaeger clients encoding trace and span IDs as variable-length hex strings, [as required by the Jaeger propagation format](https://www.jaegertracing.io/docs/1.37/client-libraries/#value). (#2731)
```

MrAlias (approved):

CHANGELOG.md
mfrister: Updated, thanks! Sorry I didn't check the changelog format better before, I hope it's ok now.

hanyuancheung (approved):

mfrister:
Is there anything missing for this to get merged?

PR #2748, opened 2022-09-12
[net/http] Allows adding custom request atributes to otelhttp metrics

RangelReale:
Option sample:

```go
otelhttp.WithMetricAttributes(func(params *otelhttp.MetricAttributesParams) []attribute.KeyValue {
	return semconv.HTTPServerAttributesFromHTTPRequest(params.Operation, "", params.Request)
}),
```


instrumentation/net/http/otelhttp/config.go
```diff
@@ -37,6 +38,7 @@ type config struct {
 	Meter             metric.Meter
 	Propagators       propagation.TextMapPropagator
 	SpanStartOptions  []trace.SpanStartOption
+	MetricAttributes  []func(string, *http.Request, int) []attribute.KeyValue
```
Aneurysm9, line 41:
```suggestion
	MetricAttributeFns  []func(string, *http.Request, int) []attribute.KeyValue
```

The name should make clear that it is storing functions that will produce metric attributes and not the attributes themselves.

instrumentation/net/http/otelhttp/config.go
```diff
@@ -141,6 +143,14 @@ func WithSpanOptions(opts ...trace.SpanStartOption) Option {
 	})
 }
 
+// WithMetricAttributes adds a function that will be called to add attributes
+// to the request metrics.
+func WithMetricAttributes(metricAttributes func(operation string, r *http.Request, statusCode int) []attribute.KeyValue) Option {
```
Aneurysm9, line 148: Should this define a structure to be passed to the function rather than a set of parameters?  This approach doesn't leave any room for providing additional information to the attribute functions later.

instrumentation/net/http/otelhttp/config.go
```diff
@@ -141,6 +143,14 @@ func WithSpanOptions(opts ...trace.SpanStartOption) Option {
 	})
 }
 
+// WithMetricAttributes adds a function that will be called to add attributes
+// to the request metrics.
```
Aneurysm9, line 147: This should include a warning to avoid high cardinality metrics unless the user is sure that their destination system can handle them.

instrumentation/net/http/otelhttp/handler.go
```diff
@@ -45,6 +45,7 @@ type Handler struct {
 	meter             metric.Meter
 	propagators       propagation.TextMapPropagator
 	spanStartOptions  []trace.SpanStartOption
+	metricAttributes  []func(string, *http.Request, int) []attribute.KeyValue
```
Aneurysm9, line 48:
```suggestion
	metricAttributeFns  []func(string, *http.Request, int) []attribute.KeyValue
```

Same here with the name.

instrumentation/net/http/otelhttp/test/config_test.go
```diff
@@ -15,16 +15,25 @@
...
+	"github.com/stretchr/testify/require"
+	"go.opentelemetry.io/otel/attribute"
+	"go.opentelemetry.io/otel/propagation"
+	"go.opentelemetry.io/otel/sdk/metric/metrictest"
+	sdktrace "go.opentelemetry.io/otel/sdk/trace"
 
 	"go.opentelemetry.io/contrib/instrumentation/net/http/otelhttp"
 	"go.opentelemetry.io/otel/sdk/trace"
 	"go.opentelemetry.io/otel/sdk/trace/tracetest"
+	semconv "go.opentelemetry.io/otel/semconv/v1.12.0"
```
hanyuancheung, line 36: `go fmt` the imports block.

hanyuancheung:
Pls add a Changelog for this PR.

MrAlias:
> but we didn't have any function to set metric options, and the only option available for metrics are the attributes, so this was the reasoning.

Understood. The reason there are no options to set metric attributes is because the `Labler` exists. Adding an option to set metric attributes would duplicate functionality, that is why it does not exist.

pellared:
@RangelReale FYI I added this PR to the OpenTelemetry Go SIG meeting's agenda.

pellared:
Closing per agreement during SIG meeting. @RangelReale if you want you can create a follow-up issue or even join the SIG meeting to discuss it.

Issue #2625, opened 2022-08-05
Rename or remove runtime.uptime metric from instrumentation/runtime package


jmacd:
In #1549 added a "process." prefix to the metrics that had been prefixed "runtime.go" already., 
The "runtime.uptime" metric was not renamed, but it also did not have the "runtime.go" prefix.

In my opinion, `rutime.uptime` should be removed and OTel would ideally specify a resource attribute that would allow deriving the uptime at the moment of any data point, see https://github.com/open-telemetry/opentelemetry-specification/issues/1273. My preference is to remove "runtime.uptime".

An alternative is to rename this "process.runtime.go.uptime", since OTel hasn't specified a cross-lang process uptime metric presently.

AkhigbeEromo:
Would like to work on this issue @jmacd @MrAlias

MrAlias:
The uptime metric was deprecated in https://github.com/open-telemetry/opentelemetry-go-contrib/releases/tag/v1.29.0. Closing.

PR #2756, opened 2022-09-16
Release Preparation: Update runtime instrumentation to use new SDK

MrAlias:
Part of https://github.com/open-telemetry/opentelemetry-go/issues/3176


jmacd (approved):
Note that I don't think we should restore these tests, instead write new ones as described in #2624 and #2625. I've drafted such tests in https://github.com/lightstep/otel-launcher-go/tree/main/lightstep/instrumentation/runtime.

PR #2835, opened 2022-10-05
Avoid getting a new Tracer for every RPC

bogdandrutu:
@open-telemetry/go-approvers I believe you should do this optimization for all your instrumentations, since it does not make any sense to retrieve a new Tracer (with same arguments) every request.


MrAlias (approved):

Aneurysm9 (approved):
Please add a changelog entry.  While this should be transparent to users, it could have impact (hopefully just positive performance impact) and they should be made aware.

XSAM (approved):
👍🏽

PR #2838, opened 2022-10-06
Optimize otelgrpc, avoid unnecessary allocations and copies

bogdandrutu:
Depends on https://github.com/open-telemetry/opentelemetry-go-contrib/pull/2835

### DIFF
--- a/instrumentation/google.golang.org/grpc/otelgrpc/grpctrace.go
+++ b/instrumentation/google.golang.org/grpc/otelgrpc/grpctrace.go
@@ -135,29 +135,45 @@ func (s *metadataSupplier) Keys() []string {
 // Inject injects correlation context and span context into the gRPC
 // metadata object. This function is meant to be used on outgoing
 // requests.
+// Deprecated: Unnecessary public func.
 func Inject(ctx context.Context, md *metadata.MD, opts ...Option) {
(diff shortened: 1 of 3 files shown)


instrumentation/google.golang.org/grpc/otelgrpc/grpctrace.go
```diff
@@ -135,29 +135,45 @@ func (s *metadataSupplier) Keys() []string {
...
+		metadata: &md,
 	})
+	return metadata.NewOutgoingContext(ctx, md)
 }
 
 // Extract returns the correlation context and span context that
 // another service encoded in the gRPC metadata object with Inject.
 // This function is meant to be used on incoming requests.
+// Deprecated: Unnecessary public func.
 func Extract(ctx context.Context, md *metadata.MD, opts ...Option) (baggage.Baggage, trace.SpanContext) {
```
bogdandrutu, line 161: Decided to mark this funcs as deprecated, I can also remove them if you are ok with that, they should not be used outside but I am not sure about your policy.

For Extract was very interesting that you extract Baggage to add it back.

Aneurysm9 (approved):

dashpole (approved):

PR #2879, opened 2022-10-22
[otelaws] Add DynamoDB and SQS X-Ray specific attributes

lukestoward:
The AWS SDK v2 instrumentation library `otelaws` does not set SQS specific operational attributes. The goal here is to capture the SQS queue URL making it available to the AWS X-Ray exporter (in the collector) in order to provide more detailed telemetry data to X-Ray. Based on the PR comments, I've updated this PR to only include messaging attributes from the semantic conventions and not the AWS X-Ray bespoke attributes. This change will ensure the attributes are captured for SQS operations and made available to the exporters. Separately, we will need to update the collector exporter to extract the specific semantic convention attributes during the mapping between OTel and X-Ray.

Below is an example trace from the AWS X-Ray console that demonstrates what is possible once DynamoDB table and SQS Queue URL information is captured and exported to X-Ray.

**Before:**
image

**After:**
image

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -10,8 +10,7 @@ This project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0.htm
 
 ### Added
 
-- `instrumentation/github.com/aws/aws-sdk-go-v2/otelaws`: Add `aws.table_name` attribute to all supporting DynamoDB operations (#2879).
-- `instrumentation/github.com/aws/aws-sdk-go-v2/otelaws`: Add `aws.queue_url` & `messaging.url` attributes to all supporting SQS operations (#2879).
+- `instrumentation/github.com/aws/aws-sdk-go-v2/otelaws`: Add semconv `messaging.url` & `messaging.system`  attributes to all supporting SQS operations (#2879).
 
 ## [1.11.1/0.36.4/0.5.2]
 
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -10,8 +10,8 @@ This project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0.htm
 
 ### Added
 
-- `instrumentation/github.com/aws/aws-sdk-go-v2/otelaws`: Add `aws.table_name` attribute to all supporting DynamoDB operations.
-- `instrumentation/github.com/aws/aws-sdk-go-v2/otelaws`: Add `aws.queue_url` & `messaging.url` attributes to all supporting SQS operations.
+- `instrumentation/github.com/aws/aws-sdk-go-v2/otelaws`: Add `aws.table_name` attribute to all supporting DynamoDB operations (#2879).
+- `instrumentation/github.com/aws/aws-sdk-go-v2/otelaws`: Add `aws.queue_url` & `messaging.url` attributes to all supporting SQS operations (#2879).
 
 ## [1.11.1/0.36.4/0.5.2]
 
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -8,6 +8,11 @@ This project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0.htm
 
 ## [Unreleased]
 
+### Added
+
+- `instrumentation/github.com/aws/aws-sdk-go-v2/otelaws`: Add `aws.table_name` attribute to all supporting DynamoDB operations.
+- `instrumentation/github.com/aws/aws-sdk-go-v2/otelaws`: Add `aws.queue_url` & `messaging.url` attributes to all supporting SQS operations.
+
 ## [1.11.1/0.36.4/0.5.2]
 
 ### Added
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -10,7 +10,7 @@ This project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0.htm
 
 ### Added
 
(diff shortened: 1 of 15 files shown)


instrumentation/github.com/aws/aws-sdk-go-v2/otelaws/attributes.go
```diff
@@ -31,32 +31,45 @@ const (
 	RegionKey    attribute.Key = "aws.region"
 	ServiceKey   attribute.Key = "aws.service"
 	RequestIDKey attribute.Key = "aws.request_id"
+	TableNameKey attribute.Key = "aws.table_name"
```
dmathieu, line 34: This attribute key doesn't match the specification's requirement: https://github.com/anuraaga/opentelemetry-specification/blob/c38800cdc4b1551e8d64f262c80c1d625c4ebc29/specification/trace/semantic_conventions/instrumentation/aws-sdk.md#dynamodb

I'm not seeing SQS mentioned in there, but I think since other AWS services have semantic conventions described, so should SQS.

lukestoward: Agreed. These are pre-existing X-Ray specific attributes that predate the semantic conventions. Ideally, the X-Ray exporter would look for and favour the semantic convention key for Dynamo, but it's currently not doing that. Instead it's looking for `aws.table_name`. This is being populated by the Java auto-instrumentor and likely others, and IMO should be populated in this instrumentation library.

Aneurysm9: The X-Ray exporter should be changed to handle the translation from the OTel semantic conventions properly.  I'd rather not have this instrumentation emitting non-standard attributes, and twice at that since it is already emitting the attribute with the conventional name.

lukestoward: @Aneurysm9 ok, fair enough. How about I update this PR to only include the new `SQSAttributeSetter` and set the `semconv.MessagingURLKey` attribute? This way we're include the specific semantic convention key that is currently not being set and separately, I can open a PR for the X-Ray exporter to extract the semantic convention keys instead.

lukestoward: @Aneurysm9 / @dmathieu I have updated the PR to remove custom attributes and instead only included applicable attributes from the semantic conventions. Please could you re-review.

Aneurysm9 (approved):

CHANGELOG.md
```diff
@@ -8,6 +8,10 @@ This project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0.htm
 
 ## [Unreleased]
 
+### Added
+
+- `instrumentation/github.com/aws/aws-sdk-go-v2/otelaws`: Add semconv `messaging.url` & `messaging.system`  attributes to all supporting SQS operations (#2879).
```
MrAlias, line 13:
```suggestion
- Add the `messaging.url` & `messaging.system` attributes to all appropriate SQS operations in the `go.opentelemetry.io/contrib/instrumentation/github.com/aws/aws-sdk-go-v2/otelaws` package. (#2879)
```

instrumentation/github.com/aws/aws-sdk-go-v2/otelaws/attributes_test.go
```diff
@@ -0,0 +1,47 @@
...
+	service := "test-service"
+	attr := ServiceAttr(service)
+	assert.Equal(t, attribute.String("aws.service", service), attr)
+}
+
+func TestRequestIDAttr(t *testing.T) {
+	requestID := "test-request-id"
+	attr := RequestIDAttr(requestID)
+	assert.Equal(t, attribute.String("aws.request_id", requestID), attr)
+}
```
MrAlias, line 47: These are change detector tests. Consider replacing the attribute key with the variable used by the actual code.

instrumentation/github.com/aws/aws-sdk-go-v2/otelaws/sqsattributes_test.go
```diff
@@ -0,0 +1,195 @@
...
+func TestSQSDeleteMessageBatchInput(t *testing.T) {
+	input := middleware.InitializeInput{
+		Parameters: &sqs.DeleteMessageBatchInput{
+			QueueUrl: aws.String("test-queue-url"),
+		},
+	}
+
+	attributes := SQSAttributeSetter(context.TODO(), input)
+
+	assert.Contains(t, attributes, attribute.String("messaging.url", "test-queue-url"))
```
MrAlias, line 38: Same here. Consider using the semconv defined key so when name changes occur this doesn't break.

MrAlias (approved):

instrumentation/github.com/aws/aws-sdk-go-v2/otelaws/attributes_test.go
lukestoward: Personally, I prefer not referencing the `const` value for the local AWS attributes as this can allow for bugs to creep in. For example, if someone accidentally changes the value assigned to a particular `const` key, let's say a typo, then the tests will continue to pass. Conversely, when this is an intended change, the tests need to be updated as well. I feel the latter is worth protecting against the former when dealing with constants under our control.

instrumentation/github.com/aws/aws-sdk-go-v2/otelaws/sqsattributes_test.go
lukestoward: Updated.

CHANGELOG.md
lukestoward: Updated.

PR #2961, opened 2022-11-01
inject context before handling downstream for grpc/http server

freyayunfu:
With current implementation, http/grpc server does not propagate context to downstream anymore. It does not function well in distributed tracing scenario. For example, in our case, http endpoint forward request to grpc service, and grpc service query an external grpc service etc. With current implementation, these traces will be separate traces instead of one. I fixed the issue by injecting context before handling downstream in grpc/http server in this PR.


freyayunfu:
In my use case (http server --> grpc client --> grpc server), otelgrpc client got silently overwrite by other interceptor. So that I did this http server to grpc server hack to do context propagation. However, the right approach is to fix the grpc client otel client interceptor, so that we can do in process context propagation between http server to grpc client. Closing this PR to avoid confusion.

Issue #2958, opened 2022-11-01
Add custom Labeler to otelhttp Handler


zshehov:
Currently the `otelhttp.Handler`'s `ServeHTTP` depends directly on a concrete `Labeler` coming from `labeler.go`. This `Labeler` should be good enough for the general case but forces otelhttp users to always pay the price of going through mutex.Lock/Unlock + a slice allocation in the `Get()` method. It seems like a good idea to decouple `otelhttp.Handler` from the concrete `Labeler` and move to a `CustomLabeler` interface with a default value of the `labeler.go`'s `Labeler`

I'm not sure how we should go about this if this is to be implemented - is it worth to have a separate `labelers` package? Where is the `CustomLabeler` interface defined?

PR #2984, opened 2022-11-11
Add the ability to provide custom Lablers in the Http Handler

zshehov:
Addressing issue #2958

### DIFF
--- a/instrumentation/net/http/otelhttp/labeler.go
+++ b/instrumentation/net/http/otelhttp/labeler.go
@@ -23,10 +23,11 @@ import (
 
 type labelerContextKeyType int
 
+// LabelerContextKey is the key in the context.Context on which the Labeler instance would be placed.
 const LabelerContextKey labelerContextKeyType = 0
 
 // Labeler provides a way to hook custom attribute.KeyValue entries to a request context during the execution of the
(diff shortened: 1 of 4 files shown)


instrumentation/net/http/otelhttp/labeler.go
```diff
@@ -21,45 +21,57 @@ import (
...
 
-// Labeler is used to allow instrumented HTTP handlers to add custom attributes to
+type labelerContextKeyType int
+
+// LabelerContextKey is the key in the context.Context on which the Labeler instance would be placed.
+const LabelerContextKey labelerContextKeyType = 0
+
+// Labeler provides a way to hook custom attribute.KeyValue entries to a request context during the execution of the
+// request. The instance of the custom Labeler can be accessed from the request's context with the LabelerContextKey.
+type Labeler interface {
```
MadVikingGod, line 31: This is a breaking change to this package's API.  If this can be accomplished without changing the `Labeler` type it can be considered.

MadVikingGod (changes requested):

instrumentation/net/http/otelhttp/labeler.go
zshehov: Argh, true. Let's keep the `Labeler` a struct. Can you suggest a fitting name for the interface?

PR #3001, opened 2022-11-15
Feature: nats instrumentation

piotrpersona:
I would like to propose PR which adds methods to propagate TraceID and SpanID in nats message headers.

SpanContext data will be carried in nats message headers.

Publisher may inject current SpanContext to new or existing `*nats.Msg`.
Consumer may reconstruct `context.Context` or `trace.Span` from consumed `*nats.Msg`.


dmathieu:
We are trying to start enforcing a clear ownership for new contrib packages, as setting up a new package is great. But it needs someone to maintain it over time afterwards.
Why can't this be in its own repository? (or a contrib one owned by the nats contributors)

piotrpersona:
@dmathieu Thank you for the response.
What do you suggest? I looked up on how Shopify/sarama/otelsarama was created and tried to do the same for nats. I'm open to suggestions of how such contribution should be done.

pellared:
@piotrpersona PTAL https://github.com/open-telemetry/opentelemetry-go-contrib/blob/main/instrumentation/README.md#new-instrumentation

piotrpersona:
Hi @dmathieu @pellared I've read new instrumentation guidelines and everything is clear now.
It seems that the best way to achieve such instrumentation would be when `nats.go` package itself will provide some message middleware which may be called on each sent/received message. `nats.go` contributors have planned to implement this. Therefore, I may contribute to `nats.go` repo when it's done or just wait for it to be implemented.

This issue can be closed.

Issue #2840, opened 2022-10-07
Request to Add Instrumentation for google.golang.org/grpc


fatsheep9146:
## Background

`google.golang.org/grpc` is already instrumented with library
`go.opentelemetry.io/contrib/instrumentation/google.golang.org/grpc/otelgrpc`, but this library only covers trace. 

So this issue will try to complete all the metrics for `google.golang.org/grpc` which are already defined in https://github.com/open-telemetry/opentelemetry-specification/blob/main/specification/metrics/semantic_conventions/rpc-metrics.md

### Metrics

Instruments

### Server
- [ ] rpc.server.duration #2700 
- [ ] rpc.server.request.size
- [ ] rpc.server.response.size	
- [ ] rpc.server.requests_per_rpc
- [ ] rpc.server.responses_per_rpc	

### Client
- [ ] rpc.client.duration	
- [ ] rpc.client.request.size
- [ ] rpc.client.response.size
- [ ] rpc.client.requests_per_rpc
- [ ] rpc.client.responses_per_rpc	

## Tasks

- Code complete:
  - [ ] Comprehensive unit tests.
  - [ ] End-to-end integration tests.
  - [ ] Tests all passing.
  - [ ] Instrumentation functionality verified.
- Documented
  - [ ] Added to the [OpenTelemetry Registry](https://opentelemetry.io/registry/)
  - [ ] README included for the module describing high-level purpose.
  - [ ] Complete documentation of all public API including package documentation.
  - [ ] [Instrumentation documentation](https://github.com/open-telemetry/opentelemetry-go-contrib/blob/main/instrumentation/README.md#instrumentation-packages) updated.
- Examples
  - [ ] `Dockerfile` file to build example application.
  - [ ] `docker-compose.yml` to run example in a docker environment to demonstrate instrumentation.

dmathieu:
Related: #2782.

fatsheep9146:
I think this issue is now blocked by #197

Sovietaced:
Anyone mind if I add a metrics implementation to the client interceptors?

ash2k:
[...]

PR #2700, opened 2022-08-31
[otelgrpc] add metric rpc.server.duration support 

fatsheep9146:
Signed-off-by: Ziqi Zhao [zhaoziqi9146@gmail.com](mailto:zhaoziqi9146@gmail.com)

Enhance instrumentation library [otelgrpc](https://github.com/open-telemetry/opentelemetry-go-contrib/tree/main/instrumentation/google.golang.org/grpc/otelgrpc) for grpc to satisfy opentelemetry [grpc server metric semantic conventions](https://github.com/open-telemetry/opentelemetry-specification/blob/main/specification/metrics/semantic_conventions/rpc.md#rpc-server).

This pr is meant to add metric `rpc.server.duration` for the first step.

I use example code to test, and produce the data for unary and stream request
```
{
    "Name": "rpc.server.duration{service.name=unknown_service:main,telemetry.sdk.language=go,telemetry.sdk.name=opentelemetry,telemetry.sdk.version=1.9.0,instrumentation.name=go.opentelemetry.io/contrib/instrumentation/google.golang.org/grpc/otelgrpc,instrumentation.version=semver:0.34.0,net.peer.ip=127.0.0.1,net.peer.port=55809,rpc.grpc.status_code=0,rpc.method=SayHelloServerStream,rpc.service=api.HelloService,rpc.system=grpc}",
    "Sum": 502.366387
},
{
    "Name": "rpc.server.duration{service.name=unknown_service:main,telemetry.sdk.language=go,telemetry.sdk.name=opentelemetry,telemetry.sdk.version=1.9.0,instrumentation.name=go.opentelemetry.io/contrib/instrumentation/google.golang.org/grpc/otelgrpc,instrumentation.version=semver:0.34.0,net.peer.ip=127.0.0.1,net.peer.port=55809,rpc.grpc.status_code=0,rpc.method=SayHelloBidiStream,rpc.service=api.HelloService,rpc.system=grpc}",
    "Sum": 253.394227
}
```

PR #3002, opened 2022-11-16
otelgrpc: Implement grpc.StatsHandler for trace instrumentation

### LINKED ISSUE #197: Implement `grpc.StatsHandler` for grpc instrumentation
See OpenCensus https://godoc.org/go.opencensus.io/plugin/ocgrpc

Much easier to use and implement, now:
```golang
s := grpc.NewServer(
		grpc.UnaryInterceptor(grpctrace.UnaryServerInterceptor(global.Tracer(""))),
		grpc.StreamInterceptor(grpctrace.StreamServerInterceptor(global.Tracer(""))),
	)
```
After:
```
s := grpc.NewServer(grpc.StatsHandler(&grpctrace.NewServerHandler(global.Tracer(""))))
```

fatsheep9146:
fixed #197
related #2840
Signed-off-by: Ziqi Zhao <zhaoziqi9146@gmail.com>


instrumentation/google.golang.org/grpc/otelgrpc/stats_handler.go
```diff
@@ -0,0 +1,184 @@
...
+	"sync/atomic"
+
+	"go.opentelemetry.io/contrib/instrumentation/google.golang.org/grpc/otelgrpc/internal"
+	"go.opentelemetry.io/otel/attribute"
+	"go.opentelemetry.io/otel/codes"
+	semconv "go.opentelemetry.io/otel/semconv/v1.12.0"
+	"go.opentelemetry.io/otel/trace"
+	grpc_codes "google.golang.org/grpc/codes"
+	"google.golang.org/grpc/stats"
+	"google.golang.org/grpc/status"
```
hanyuancheung, line 28: The import part should be divided into different partitions.

instrumentation/google.golang.org/grpc/otelgrpc/stats_handler.go
```diff
@@ -0,0 +1,184 @@
...
+)
+
+type grpcContext struct {
```
hanyuancheung, line 31:
```suggestion
type gRPCContext struct {
```
grpc is a Proper Noun. Rename it into gRPCContext.

instrumentation/google.golang.org/grpc/otelgrpc/stats_handler.go
```diff
@@ -0,0 +1,184 @@
...
+	messagesReceived int64
+	messagesSent     int64
+}
+
+// NewServerHandler create a stats.Handler for grpc server.
```
hanyuancheung, line 36:
```suggestion
// NewServerHandler creates a stats.Handler for gRPC server.
```

instrumentation/google.golang.org/grpc/otelgrpc/stats_handler.go
```diff
@@ -0,0 +1,184 @@
...
+func NewServerHandler(opts ...Option) stats.Handler {
+	h := &serverHandler{
+		config: newConfig(opts),
+	}
+
+	h.tracer = h.config.TracerProvider.Tracer(
+		instrumentationName,
+		trace.WithInstrumentationVersion(SemVersion()),
+	)
+
```
hanyuancheung, line 46:
```suggestion
```
Meaningless blank line, should be removed.

fatsheep9146:
@MrAlias @dashpole @dmathieu 
Please help review this pr, thanks :)

dashpole:
Thanks for tackling this!

I have a few questions. Can you update the description with the answers?

* Is this meant to have feature parity with the existing otelgrpc instrumentation, and eventually replace it?  Or is it meant to be something you use instead of otelgrpc instrumentation?  If it is replacing it, what is the plan for deprecating the old one?
* Is there anything you were unable to implement with the stats handler interface?

One other question:

Is it possible to run the same unit tests against a stats-handler-based implementation and the current implementation?  Or did you copy existing test cases when writing these tests?

fatsheep9146:
Sorry for my delay

> Is this meant to have feature parity with the existing otelgrpc instrumentation, and eventually replace it?  Or is it meant to be something you use instead of otelgrpc instrumentation?  If it is replacing it, what is the plan for deprecating the old one?

I think this feature is not to replace the existing one, but to provide a new way to instrument the grpc library. But in this new way with stats handler, we are able to implement more cases that the interceptors do not, for example, a failure to decode the message which is mentioned in https://github.com/open-telemetry/opentelemetry-go-contrib/issues/197#issuecomment-668377700.

> Is there anything you were unable to implement with the stats handler interface?

For now, I think stats handler has much more info about the grpc request, response and connection, so it definitely is a good place to add instrument logic. I can not answer this question based on what I know about the grpc stats handler.

> Is it possible to run the same unit tests against a stats-handler-based implementation and the current implementation? Or did you copy existing test cases when writing these tests?

Yes, I already add the same test like `interceptors` in test directory.

@dashpole

fatsheep9146:
@open-telemetry/go-approvers hi, is anyone able to review this pr? Thanks a lot

instrumentation/google.golang.org/grpc/otelgrpc/stats_handler.go
```diff
@@ -0,0 +1,184 @@
...
+	case *stats.Begin:
+	case *stats.InPayload:
+		if gctx != nil {
+			messageId = atomic.AddInt64(&gctx.messagesReceived, 1)
+		}
+		span.AddEvent("message",
+			trace.WithAttributes(
+				semconv.MessageTypeReceived,
+				semconv.MessageIDKey.Int64(messageId),
+			),
```
s-matyukevich, line 159: I think we now can also set the following 2 [attributes](https://github.com/open-telemetry/opentelemetry-go-contrib/blob/main/instrumentation/google.golang.org/grpc/otelgrpc/semconv.go#L34-L38), which, if I am reading the code right, are completely ignored. rs.Length and rs.WireLength properties should contain this info. 

Looks like those properties were deleted [here](https://github.com/open-telemetry/opentelemetry-go-contrib/pull/3168/files#diff-3218c106290c074c8331304ce04eb5006aa78fb8b913a2252e51930cfc7d3a1bL52) because caclulating them in interceptors was too expensive. With StatsHandler version we no longer have this problem.

instrumentation/google.golang.org/grpc/otelgrpc/stats_handler.go
```diff
@@ -0,0 +1,184 @@
...
+	attrs = append(attrs, mAttrs...)
+	ctx, _ = h.tracer.Start(
+		trace.ContextWithRemoteSpanContext(ctx, trace.SpanContextFromContext(ctx)),
+		name,
+		trace.WithSpanKind(trace.SpanKindServer),
+		trace.WithAttributes(attrs...),
+	)
+
+	gctx := gRPCContext{}
+	return context.WithValue(ctx, gRPCContext{}, &gctx)
```
ash2k, line 70: Here and blow: I think it'd be better to use a separate dedicated type for the key. Typically, an alias for `struct{}`. Benefits:
- It's confusing that this type is used for two purposes.
- As the key, it doesn't carry any data (it doesn't need to), but consumes 16 bytes of memory. This is pure waste that can be avoided by just using a separate type.

WDYT?

instrumentation/google.golang.org/grpc/otelgrpc/stats_handler.go
fatsheep9146: Done

instrumentation/google.golang.org/grpc/otelgrpc/stats_handler.go
```diff
@@ -0,0 +1,188 @@
...
+	tracer trace.Tracer
+}
+
+// TagRPC can attach some information to the given context.
+func (h *serverHandler) TagRPC(ctx context.Context, info *stats.RPCTagInfo) context.Context {
+	ctx = extract(ctx, h.config.Propagators)
+
+	attrs := []attribute.KeyValue{RPCSystemGRPC}
+	name, mAttrs := internal.ParseFullMethod(info.FullMethodName)
+	attrs = append(attrs, mAttrs...)
```
ash2k, line 61: I think it'd be slightly more performant to invert the append operation:
- Current code allocates `attrs`, `mAttrs` and appends them into a new slice. I.e. 3 allocated slices in the worst case.
- `mAttrs = append(mAttrs, RPCSystemGRPC)` allocates `mAttrs` and appends them into a new slice. I.e. 2 allocated slices in the worst case.

fatsheep9146: Thanks! You are right, I modify this place and another place which has similar problem as you said.
@ash2k

instrumentation/google.golang.org/grpc/otelgrpc/stats_handler.go
```diff
@@ -0,0 +1,185 @@
...
+// See the License for the specific language governing permissions and
+// limitations under the License.
+
+package otelgrpc // import "go.opentelemetry.io/contrib/instrumentation/google.golang.org/grpc/otelgrpc"
+
+import (
+	"context"
+	"sync/atomic"
+
+	grpc_codes "google.golang.org/grpc/codes"
```
pellared, line 21: why not simply?

```suggestion
	"google.golang.org/grpc/codes"
```

CHANGELOG.md
```diff
@@ -11,6 +11,7 @@ This project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0.htm
 ### Added
 
 - Add the new `go.opentelemetry.io/contrib/instrgen` package to provide auto-generated source code instrumentation. (#3068, #3108)
+- [otelgrpc] refactor otelgrpc to use grpc.StatsHandler. (#3002)
```
pellared, line 14: This is not a refactoring[^1] as it changes the output.
Can you describe the changes more precise?

[^1]: Refactoring is a disciplined technique for restructuring an existing body of code, altering its internal structure without changing its external behavior.

instrumentation/google.golang.org/grpc/otelgrpc/test/stats_handler_test.go
```diff
@@ -0,0 +1,131 @@
...
+
+	"go.opentelemetry.io/contrib/instrumentation/google.golang.org/grpc/otelgrpc"
+	"go.opentelemetry.io/otel/attribute"
+	"go.opentelemetry.io/otel/codes"
+	"go.opentelemetry.io/otel/sdk/trace"
+	"go.opentelemetry.io/otel/sdk/trace/tracetest"
+	semconv "go.opentelemetry.io/otel/semconv/v1.17.0"
+)
+
+func TestClientStatsHandler(t *testing.T) {
```
pellared, line 33: Why do we need this test? Is it not redunant with `TestStatsHandler`? `TestStatsHandler` also looks more "end to end" than this one.

pellared (commented):
Leaving a quick review feedback

CHANGELOG.md
fatsheep9146: Done

instrumentation/google.golang.org/grpc/otelgrpc/stats_handler.go
fatsheep9146: Because I also import `go.opentelemetry.io/otel/codes` which has same package name.

instrumentation/google.golang.org/grpc/otelgrpc/stats_handler.go
fatsheep9146: I use alias type `gRPCContextKey` as the key for gPRCContext, which is alias for `struct{}`
@ash2k

instrumentation/google.golang.org/grpc/otelgrpc/stats_handler.go
```diff
@@ -0,0 +1,187 @@
...
+}
+
+// TagRPC can attach some information to the given context.
+func (h *serverHandler) TagRPC(ctx context.Context, info *stats.RPCTagInfo) context.Context {
+	ctx = extract(ctx, h.config.Propagators)
+
+	name, attrs := internal.ParseFullMethod(info.FullMethodName)
+	attrs = append(attrs, RPCSystemGRPC)
+	ctx, _ = h.tracer.Start(
+		trace.ContextWithRemoteSpanContext(ctx, trace.SpanContextFromContext(ctx)),
```
ash2k, line 63: May I suggest to implement options to configure this behavior. Similar to how `net/http` instrumentation works, see:

https://github.com/open-telemetry/opentelemetry-go-contrib/blob/75d0c13f6129330f3b5b060791aa8cefdc553849/instrumentation/net/http/otelhttp/config.go#L106-L124

I.e. default to private endpoints (i.e. `parent` relationship between spans/traces) and switch to public (i.e. `link` relationship) when explicitly asked to.

p.s. I think maintainers should probably decide and document how this should work across all integrations to provide consistent experience to users.

dashpole (commented):
The code changes LGTM, but I think this needs to be surfaced in the README.  There, we should cover:

* Do we consider the new Stats Handlers experimental?
* Do we recommend users use interceptors, stats handlers, or both?
* What happens if users use both?  Will users get duplicate telemetry?  Will parent/child relationships between spans be broken because propagators are used in both interceptor and stats handlers instrumentation?  Or is the telemetry they get from stats handlers and interceptors complementary and non-overlapping?

Thanks for taking this on!

instrumentation/google.golang.org/grpc/otelgrpc/stats_handler.go
ash2k: It does maintain the existing behavior but at the same time now is the chance to introduce new default behavior, consistent with `net/http` instrumentation. If not now, then it'll be a breaking change for anyone who switches to the new implementation. Breaking change may be fine if this implementation is marked as experimental, I don't mind that or addressing it separately, I just need such an option. Thanks!

fatsheep9146: > It does maintain the existing behavior but at the same time now is the chance to introduce new default behavior, consistent with `net/http` instrumentation. If not now, then it'll be a breaking change for anyone who switches to the new implementation. Breaking change may be fine if this implementation is marked as experimental, I don't mind that or addressing it separately, I just need such an option. Thanks!

I will open another issue to tackle this problem after this pr is merged.

fatsheep9146:
@dashpole I start to try to answer the questions 

> Do we consider the new Stats Handlers experimental?

Yes, since the way of using this is a breaking change to user, so I think it is definitely experimental and need to be watched for a period of time to be stable.

> Do we recommend users use interceptors, stats handlers, or both?

I think we should recommend users to use stats handlers for two main reasons.

- **Functional advantages**: `stats` has more information for user to build more flexible and granular metric, for example
  - multiple different types of represent ”data length”: In [InPayLoad](https://pkg.go.dev/google.golang.org/grpc/stats#InPayload), there exists `Length`, `CompressedLength`, `WireLength` to denote the size of uncompressed, compressed payload data, with or without framing data. But in interceptors, we can only got uncompressed data, and this feature is also removed due to performance problem. https://github.com/open-telemetry/opentelemetry-go-contrib/pull/3168
  - more accurate timestamp: `InPayload.RecvTime` and `OutPayload.SentTime` records more accurate timestamp that server got and sent the message, the timestamp recorded by interceptors depends on the location of this interceptors in the total interceptor chain.
  - some other use cases: https://github.com/open-telemetry/opentelemetry-go-contrib/issues/197#issuecomment-668377700
 - **Performance advantages**: If too many interceptors are registered in a service, the interceptor chain can become too long, which increases the latency and processing time of the entire RPC call.

> What happens if users use both? Will users get duplicate telemetry? Will parent/child relationships between spans be broken because propagators are used in both interceptor and stats handlers instrumentation? Or is the telemetry they get from stats handlers and interceptors complementary and non-overlapping?
 
TBD

fatsheep9146:
Could you help check this pr again? @dashpole

dashpole:
Can you add your findings in https://github.com/open-telemetry/opentelemetry-go-contrib/pull/3002#issuecomment-1643278070 to a README?  I think it is important that users can understand the differences between the two implementations.

fatsheep9146:
> Can you add your findings in [#3002 (comment)](https://github.com/open-telemetry/opentelemetry-go-contrib/pull/3002#issuecomment-1643278070) to a README? I think it is important that users can understand the differences between the two implementations.

Absolutely!

fatsheep9146:
The README is already added, and I will add the example of using `stats.Handler` in [example](https://github.com/open-telemetry/opentelemetry-go-contrib/tree/main/instrumentation/google.golang.org/grpc/otelgrpc/example) after this pr is merged.

@dashpole

pellared:
nit: Can you please avoid doing force push as it makes reviewing harder? I cannot see the diff between reviews and the PR is quite large so it is easy to miss important issues in subsequent review integrations 😉

fatsheep9146:
> nit: Can you please avoid force pushing as it makes reviewing harder? I cannot see the diff between reviews and the PR is quite large so it is easy to miss important issues in subsequent review integrations. Just make regular merge commits without rebasing 😉

Sorry about this. I usually prefer to clean up commit history in the end. I will notice this in the future. @pellared

pellared:
> > nit: Can you please avoid force pushing as it makes reviewing harder? I cannot see the diff between reviews and the PR is quite large so it is easy to miss important issues in subsequent review integrations. Just make regular merge commits without rebasing 😉
> 
> Sorry about this. I usually prefer to clean up commit history in the end. I will notice this in the future. @pellared

> I usually prefer to clean up commit history in the end. 

We do "rebase and merge" so the git history is not a big issue for us.

PS. I am fine with cleaning up the git history - just avoid rebasing the commits that are already pushed 😄

fatsheep9146:
@dashpole I think the failed test is flaky, I have tried the test in may local laptop for many times. I still can not reproduce this fail. Could you help rerun the failed check?

fatsheep9146:
@open-telemetry/go-approvers could any approvers help review this pr again?

pellared:
@hanyuancheung Can you please review (as codeowner)?

fatsheep9146:
@hanyuancheung nicely ping for review

fatsheep9146:
@hanyuancheung ping for review, thanks

instrumentation/google.golang.org/grpc/otelgrpc/README.md
```diff
@@ -0,0 +1,20 @@
...
+
+- by gRPC Interceptors
+- [experimental] by gPRC `stats.Handler`
+
+You can see the example of both ways in directory `./example`
+
+**notice**: **Do not use both interceptors and stats handlers at the same time!** If so, you will get duplicated spans and the parent/child relationships between spans will also be broken.
+
+Although the implementation `stats.Handler` in experimental stage, we strongly still recommand you to use `stats.Handler`, mainly for two reasons:
+- **Functional advantages**: `stats.Handler`` has more information for user to build more flexible and granular metric, for example
```
ash2k, line 15:
```suggestion
- **Functional advantages**: `stats.Handler` has more information for user to build more flexible and granular metric, for example
```

ash2k (approved):
Really looking forward to using the new instrumentation.

fatsheep9146:
@open-telemetry/go-approvers 
Is there any possibilities we could push this pr forward?

hanyuancheung (approved):
LGTM

fatsheep9146:
The failed test is still flaky, I will solve this in another pr.

fatsheep9146:
@hanyuancheung could you help rerun the failed test?

fatsheep9146:
@pellared could you help merge this pr?

pellared:
@fatsheep9146 I just returned from a long PTO. I will do my best to look at this before end of this week.

fatsheep9146:
> @fatsheep9146 I just returned from a long PTO. I will do my best to look at this before end of this week.

Thanks!

instrumentation/google.golang.org/grpc/otelgrpc/README.md
```diff
@@ -0,0 +1,20 @@
+# gRPC instrumentation
+
+This library is the instrumentation library for `google.golang.org/grpc`
+
+For now you can instrument your program which use `google.golang.org/grpc` in two ways:
+
+- by gRPC Interceptors
+- [experimental] by gPRC `stats.Handler`
```
pellared, line 8: 1. typo
2. what is it noted as "experimental"?

```suggestion
- by gRPC `stats.Handler`
```

pellared (approved):
I propose to:

1. Merge this PR
2. Create an issue for https://github.com/open-telemetry/opentelemetry-go-contrib/pull/3002/files#r1209293763
3. Deprecate the interceptors so that they are available in next release.
4. After the next release remove the interceptors. 

@open-telemetry/go-approvers please upvote/downvote if you agree or not. Feel free to leave any additional feedback.

instrumentation/google.golang.org/grpc/otelgrpc/README.md
fatsheep9146: > 2\. what is it noted as "experimental"?

I wrote "experimental" to express that the feature is now not stable enough, and should be used for a period of time to reach the stable stage.

Do we have some common conventions to express this?
@pellared

fatsheep9146: > It would be better if we put the docs in `doc.go` instead of `README.md`. More:

I will move to doc.go

pellared: > I wrote "experimental" to express that the feature is now not stable enough, and should be used for a period of time to reach the stable stage.

The version of the module is `v0.*` 😉

fatsheep9146: > conventions

Got it

fatsheep9146: I already moved readme to doc.go

pellared: Lint is failing (probably you need to format `doc.go`)

fatsheep9146: > Lint is failing (probably you need to format `doc.go`)

Done

instrumentation/google.golang.org/grpc/otelgrpc/doc.go
pellared, line 1: These doc comments do not render correctly.

image


See here for more guidance how to use the syntax and test it: https://github.com/open-telemetry/opentelemetry-go/blob/main/CONTRIBUTING.md#documentation

instrumentation/google.golang.org/grpc/otelgrpc/doc.go
```diff
@@ -13,23 +13,33 @@
...
+
+  - more accurate timestamp: [stats.InPayLoad]'s "RecvTime" and [stats.OutPayLoad]'s "SentTime" records more accurate timestamp that server got and sent the message, the timestamp recorded by interceptors depends on the location of this interceptors in the total interceptor chain.
+
+  - some other use cases: for example, catch failure of decoding message.
+
+Performance advantages: If too many interceptors are registered in a service, the interceptor chain can become too long, which increases the latency and processing time of the entire RPC call.
+
+[stats.Handler]: https://pkg.go.dev/google.golang.org/grpc/stats#Handler
+[stats.InPayLoad]: https://pkg.go.dev/google.golang.org/grpc/stats#OutPayLoad
+[stats.OutPayLoad]: https://pkg.go.dev/google.golang.org/grpc/stats#InPayload
```
pellared, line 39:
```suggestion
[stats.InPayload]: https://pkg.go.dev/google.golang.org/grpc/stats#OutPayload
[stats.OutPayload]: https://pkg.go.dev/google.golang.org/grpc/stats#InPayload
```

and also update the casing in the references

pellared (approved):

instrumentation/google.golang.org/grpc/otelgrpc/stats_handler.go
```diff
@@ -0,0 +1,187 @@
...
+	"context"
+	"sync/atomic"
+
+	grpc_codes "google.golang.org/grpc/codes"
+	"google.golang.org/grpc/stats"
+	"google.golang.org/grpc/status"
+
+	"go.opentelemetry.io/contrib/instrumentation/google.golang.org/grpc/otelgrpc/internal"
+	"go.opentelemetry.io/otel/codes"
+	semconv "go.opentelemetry.io/otel/semconv/v1.12.0"
```
pellared, line 27: It should be updated to 
```suggestion
	semconv "go.opentelemetry.io/otel/semconv/v1.17.0"
```

pellared (changes requested):
I noticed that the stats handler misses the functionality you added here: https://github.com/open-telemetry/opentelemetry-go-contrib/pull/2700. Any reasons for that?

fatsheep9146:
> I noticed that the stats handler misses the functionality you added here: #2700. Any reasons for that?

I was plan to implement the stats.handler step by step, cover the trace first, and then cover the metrics. Which is also convenient for review. 

Do you think I should implement it in this pr or create another pr for metrics?

I'm open for both way. @pellared

instrumentation/google.golang.org/grpc/otelgrpc/stats_handler.go
fatsheep9146: OK

pellared:
> > I noticed that the stats handler misses the functionality you added here: #2700. Any reasons for that?
> 
> I was plan to implement the stats.handler step by step, cover the trace first, and then cover the metrics. Which is also convenient for review.
> 
> Do you think I should implement it in this pr or create another pr for metrics?
> 
> I'm open for both way. @pellared

During the SIG meetings we discussed that we would like to have a feature-parity before we deprecate it as part of https://github.com/open-telemetry/opentelemetry-go-contrib/pull/3002#pullrequestreview-1613182957.

Still I think we can tackle it in a separate PR.

Are there any other features which are missing when comparing with the interceptors?

fatsheep9146:
> > > I noticed that the stats handler misses the functionality you added here: #2700. Any reasons for that?
> > 
> > 
> > I was plan to implement the stats.handler step by step, cover the trace first, and then cover the metrics. Which is also convenient for review.
> > Do you think I should implement it in this pr or create another pr for metrics?
> > I'm open for both way. @pellared
> 
> During the SIG meetings we discussed that we would like to have a feature-parity before we deprecate interceptors as part of [#3002 (review)](https://github.com/open-telemetry/opentelemetry-go-contrib/pull/3002#pullrequestreview-1613182957).
> 
> Still I think we can tackle it in a separate PR.
> 
> Are there any other features which are missing when comparing with the interceptors?

Only metrics 

> > > I noticed that the stats handler misses the functionality you added here: #2700. Any reasons for that?
> > 
> > 
> > I was plan to implement the stats.handler step by step, cover the trace first, and then cover the metrics. Which is also convenient for review.
> > Do you think I should implement it in this pr or create another pr for metrics?
> > I'm open for both way. @pellared
> 
> During the SIG meetings we discussed that we would like to have a feature-parity before we deprecate interceptors as part of [#3002 (review)](https://github.com/open-telemetry/opentelemetry-go-contrib/pull/3002#pullrequestreview-1613182957).
> 
> Still I think we can tackle it in a separate PR.
> 
> Are there any other features which are missing when comparing with the interceptors?

I think metric is the only missing feature, since I wrote a simple e2e test `TestStatsHandler` just like interceptors to make sure the trace related function is complete.
@pellared

instrumentation/google.golang.org/grpc/otelgrpc/stats_handler.go
fatsheep9146: Done @pellared all checks are passed

pellared (approved):

dmathieu (approved):

fatsheep9146:
@pellared Thanks for creating the following issues!

hanyuancheung:
It seems that this PR could be merged first, and then we could continue the following work.

pellared:
@fatsheep9146 Merged 🎉 Thank you for your patience ❤️ You can follow-up with https://github.com/open-telemetry/opentelemetry-go-contrib/issues/4316 and https://github.com/open-telemetry/opentelemetry-go-contrib/issues/4317

fatsheep9146:
> @fatsheep9146 Merged 🎉 Thank you for your patience ❤️ You can follow-up with #4316 and #4317

Thanks! I will solve the follow-up ASAP. @pellared

mehdihadeli:
Hi,
When do you release new style of grpc otel? V0.44 doesn't have grpc metrics

dmathieu:
0.45.0 was released a few hour ago.
https://github.com/open-telemetry/opentelemetry-go-contrib/releases/tag/v1.20.0

pellared:
> 0.45.0 was released a few hour ago. https://github.com/open-telemetry/opentelemetry-go-contrib/releases/tag/v1.20.0

Yet metrics are not supported yet. See https://github.com/open-telemetry/opentelemetry-go-contrib/issues/4316

pellared:
> 0.45.0 was released a few hour ago. https://github.com/open-telemetry/opentelemetry-go-contrib/releases/tag/v1.20.0

@mehdihadeli  Yet metrics are not supported yet. See https://github.com/open-telemetry/opentelemetry-go-contrib/issues/4316. We hope to implement it for the next release.

PR #3003, opened 2022-11-18
Refactor handler_test tests into subtests to reduce repetition

dmathieu:
These tests were all based on similar setup and expectations. So I am refactoring them into subtests to reduce repetition.


hanyuancheung (approved):

instrumentation/net/http/otelhttp/handler_test.go
```diff
@@ -26,56 +27,64 @@ import (
...
+		name               string
+		handler            http.Handler
+		requestBody        io.Reader
+		expectedStatusCode int
+	}{
+		{
+			name: "implements flusher",
+			handler: otelhttp.NewHandler(
+				http.HandlerFunc(func(w http.ResponseWriter, r *http.Request) {
+					assert.Implements(t, (*http.Flusher)(nil), w)
```
MadVikingGod, line 41: The `t` here is the `*testing.T` of the parent test, not the subtest.  This means if there is a failure in one of these handlers it won't be attributed to the correct Subtest.

dmathieu: Good catch! I've added a `testing.T` argument.

dashpole (approved):

PR #3047, opened 2022-12-04
Add span name formatters to gocql instrumentation

brennantaylor:
Currently the observers hardcode the span names without allowing customization.

In particular the queries are just the statements executed and I have found these to contain whitespace like tabs and multi line statements with indentation.

By allowing span name formatters similar to otelhttp the user can customize the span names to their heart's desire.


instrumentation/github.com/gocql/gocql/otelgocql/observer.go
```diff
@@ -71,7 +74,7 @@ func (o *OTelQueryObserver) ObserveQuery(ctx context.Context, observedQuery gocq
 
 		ctx, span := o.tracer.Start(
 			ctx,
-			observedQuery.Statement,
+			o.spanNameFormatter(observedQuery),
```
MadVikingGod, line 77: I would be a bit more defensive about these not being nil.  If a user does `With*SpanNameFormatter(nil)` this will panic.

I would suggest adding a nil check either after apply's or before used in creating the Observers.

hanyuancheung (approved):

pellared:
Closing as `otelgoclq` is deprected per https://github.com/open-telemetry/opentelemetry-go-contrib/pull/4164.

PR #3090, opened 2022-12-15
Implementing xrayTraceIDRatioBasedsampler

alanprot:
Signed-off-by: Alan Protasio <approtas@amazon.com>

Hi..

The default Otel TraceIDRatioBased does not work with XRay as it make the decision based on the first 8 bytes of the traceId  -> which are not random on traces generated with the `xray.NewIDGenerator()`.

This PR creates a similar sampler that works on XRAY as it makes the decision based on the last 8 bytes of the trace id.

### DIFF
--- a/samplers/aws/xray/go.mod
+++ b/samplers/aws/xray/go.mod
@@ -6,6 +6,7 @@ require (
 	github.com/go-logr/logr v1.2.3
 	github.com/go-logr/stdr v1.2.2
 	github.com/stretchr/testify v1.8.1
+	go.opentelemetry.io/contrib/propagators/aws v1.12.0
 	go.opentelemetry.io/otel v1.11.2
 	go.opentelemetry.io/otel/sdk v1.11.2
 	go.opentelemetry.io/otel/trace v1.11.2
--- a/samplers/aws/xray/go.mod
+++ b/samplers/aws/xray/go.mod
@@ -18,3 +18,5 @@ require (
 	golang.org/x/sys v0.0.0-20220919091848-fb04ddd9f9c8 // indirect
 	gopkg.in/yaml.v3 v3.0.1 // indirect
 )
+
+replace go.opentelemetry.io/contrib/propagators/aws => ../../../propagators/aws
--- a/samplers/aws/xray/traceid_ratio_sampler.go
+++ b/samplers/aws/xray/traceid_ratio_sampler.go
@@ -1,3 +1,17 @@
+// Copyright The OpenTelemetry Authors
+//
+// Licensed under the Apache License, Version 2.0 (the "License");
+// you may not use this file except in compliance with the License.
+// You may obtain a copy of the License at
+//
+//     http://www.apache.org/licenses/LICENSE-2.0
+//
+// Unless required by applicable law or agreed to in writing, software
+// distributed under the License is distributed on an "AS IS" BASIS,
+// WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+// See the License for the specific language governing permissions and
+// limitations under the License.
+
 package xray
 
 import (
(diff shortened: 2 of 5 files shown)


samplers/aws/xray/go.mod
```diff
@@ -6,6 +6,7 @@ require (
 	github.com/go-logr/logr v1.2.3
 	github.com/go-logr/stdr v1.2.2
 	github.com/stretchr/testify v1.8.1
+	go.opentelemetry.io/contrib/propagators/aws v1.12.0
```
Aneurysm9, line 9: Please ensure `make crosslink` is run to add replace statements for modules in this repo.

samplers/aws/xray/traceid_ratio_sampler.go
```diff
@@ -0,0 +1,73 @@
...
+	"go.opentelemetry.io/otel/trace"
+)
+
+type traceIDRatioBasedSampler struct {
+	sdktrace.Sampler
+	max  uint64
+	desc string
+}
+
+// NewTraceIDRatioBased creates a sampler based on random number.
```
Aneurysm9, line 32:
```suggestion
// NewTraceIDRatioBased creates a sampler based on (presumably) random parts of the TraceID.
```

Aneurysm9 (commented):
Just a couple nits.

samplers/aws/xray/go.mod
alanprot: done

Aneurysm9 (approved):

willarmiros:
This is interesting to see, because I know in Java the TraceIdRatioBased sampler reads from the trailing bytes rather than the leading bytes by default. We even [have a test for it](https://github.com/aws-observability/aws-otel-java-instrumentation/blob/a011b8cc29ee32b7f668c04ccfdf64cd30de467c/awsagentprovider/src/test/java/software/amazon/opentelemetry/javaagent/providers/AwsTracerConfigurerTest.java#L60-L84). So it makes me wonder, is the correct fix to make a new sampler for X-Ray IDs only (which may be very confusing for customers to use), or could we change the default traceIdRatioBased sampler to just read from the trailing bytes in the main Go SDK?

alanprot:
> Thanks for creating this PR! This is interesting to see, because I know in Java the TraceIdRatioBased sampler reads from the trailing bytes rather than the leading bytes by default. We even [have a test for it](https://github.com/aws-observability/aws-otel-java-instrumentation/blob/a011b8cc29ee32b7f668c04ccfdf64cd30de467c/awsagentprovider/src/test/java/software/amazon/opentelemetry/javaagent/providers/AwsTracerConfigurerTest.java#L60-L84). So it makes me wonder, is the correct fix to make a new sampler for X-Ray IDs only (which may be very confusing for customers to use), or could we change the default traceIdRatioBased sampler to just read from the trailing bytes in the main Go SDK?

I had the same question but I guess I took the safer route.. idk if changing the default one will break some other use cases ..

What the others think ?

Aneurysm9:
I think it would be ideal to fix in the base sampler.  I'm not aware of anything that should break if the position of the sampled bits changes.  https://github.com/open-telemetry/opentelemetry-go/pull/3557 may replace this PR.

samplers/aws/xray/traceid_ratio_sampler.go
```diff
@@ -0,0 +1,73 @@
...
+		return sdktrace.NeverSample()
+	}
+
+	return &traceIDRatioBasedSampler{
+		desc: fmt.Sprintf("xrayTraceIDRatioBasedSampler{%v}", fraction),
+		max:  uint64(fraction * math.MaxUint64),
+	}
+}
+
+func (s *traceIDRatioBasedSampler) ShouldSample(p sdktrace.SamplingParameters) sdktrace.SamplingResult {
```
hanyuancheung, line 49: Please add a note for exported func.

hanyuancheung (approved):

tomwans:
Take a look at a related PR: https://github.com/open-telemetry/opentelemetry-go/pull/3557 which brings the Go SDK in line with the Java and Python SDKs, which all by default sample from the trailing bytes.

PR #3092, opened 2022-12-17
Fix panic in AWS ECS detector if container ARN is not valid

DarthSim:
If the container ARN is invalid or empty, the ECS detector panics in `containerMetadata.ContainerARN[:strings.LastIndex(containerMetadata.ContainerARN, ":")]`: `strings.LastIndex` returns `-1`, and the detector panics with `index out of scope`. I added a check for this case.


hanyuancheung (approved):

Aneurysm9:
I believe this was addressed via #3167.  Please open a new issue/PR if there are additional considerations.

PR #3133, opened 2023-01-04
Bump grpc to 1.53

skitt:
This drops the indirect dependency on cloud.google.com/go, which helps reduce the dependency tree (see
https://github.com/kubernetes/kubernetes/issues/113366 and https://github.com/kubernetes/kubernetes/pull/114772 for the wider context).

Signed-off-by: Stephen Kitt <skitt@redhat.com>


instrumentation/google.golang.org/grpc/otelgrpc/go.mod
```diff
@@ -7,22 +7,23 @@ require (
 	go.opentelemetry.io/otel v1.11.2
 	go.opentelemetry.io/otel/metric v0.34.0
 	go.opentelemetry.io/otel/trace v1.11.2
-	google.golang.org/grpc v1.51.0
+	google.golang.org/grpc v1.53.0-dev.0.20230103192020-f2fbb0e07ebf
```
MrAlias, line 10: This is not a stable and tested release made by the project.

MrAlias (changes requested):
Thanks for your interest in this project. This bumps a dependency to the latest commit on the main branch of the project. Unless there is a security concern, we do no normally jump to a dependency upgrade like this. We plan to upgrade this dependency when that project makes a stable release after testing their implementation.

instrumentation/google.golang.org/grpc/otelgrpc/go.mod
skitt: Yes, I should have noted that in the PR description — this is a draft PR precisely because I’m waiting for the grpc release, I just wanted to see what the impact was (including CI).

hanyuancheung (approved):

skitt:
Does this need a changelog entry? AFAICT dependency updates are no longer mentioned in `CHANGELOG.md`.

PR #3170, opened 2023-01-19
Update metric use the new API

MadVikingGod:
With the landing of the new [metric API](https://github.com/open-telemetry/opentelemetry-go/pull/3584) uses of the old api need to be adjusted.

This PR can't be considered until there is a release of the metric API, but will be needed when these versions are updated to that release.

If you would like to test it you can use this go.work: https://gist.githubusercontent.com/MadVikingGod/0045dd54a9fba8f8104459f79454e4ad/raw/50da79ebea42f3d176e7ecb00cc3b34aa4741187/go.work


MrAlias:
For reference: https://github.com/open-telemetry/opentelemetry-go-contrib/pull/2589

Something like `go get pkgname@commit` should allow you to pull in the latest in `main` from `otel`.

instrumentation/runtime/runtime.go
```diff
@@ -120,7 +118,7 @@ func Start(opts ...Option) error {
 
 func (r *runtime) register() error {
 	startTime := time.Now()
-	uptime, err := r.meter.AsyncInt64().UpDownCounter(
+	uptime, err := r.meter.Int64ObservableUpDownCounter(
```
matej-g, line 121: I don't think it's worth opening a separate PR, but I noticed this should be just an async counter since it's monotonic. Potentially worth changing as part of switching to new API:
```suggestion
	uptime, err := r.meter.Int64ObservableCounter(
```

MrAlias: I would suggest fixing this in a follow-up PR.

MrAlias (approved):

Aneurysm9 (approved):

PR #3180, opened 2023-01-24
Add HTTP verb to Beego span names

luismiramirez:
Beego span names now contain the HTTP verb the route was called with. This improves categorisation of traces, as in a REST environment the same route pattern can be called with different verbs triggering different parts of apps' business logic.

### DIFF
--- a/instrumentation/github.com/astaxie/beego/otelbeego/beego.go
+++ b/instrumentation/github.com/astaxie/beego/otelbeego/beego.go
@@ -16,6 +16,7 @@ package otelbeego // import "go.opentelemetry.io/contrib/instrumentation/github.
 
 import (
 	"context"
+	"fmt"
 	"net/http"
 
 	"go.opentelemetry.io/otel/codes"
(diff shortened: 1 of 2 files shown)


dmathieu:
I don't know if we should make this change when even the net/http instrumentation doesn't do it that way.
https://github.com/open-telemetry/opentelemetry-go-contrib/blob/main/instrumentation/net/http/otelhttp/handler.go#L58-L60

Also, you can already override this value in your own code: https://github.com/open-telemetry/opentelemetry-go-contrib/blob/905ac8fb4e2191ff5f6c42cb0862b34ac4997ce7/instrumentation/github.com/astaxie/beego/otelbeego/config.go#L91

luismiramirez:
@dmathie I understand. What about adding the route pattern to the `http.route` as other web framework instrumentations do?

dmathieu:
The `http.route` parameter should definitely be set if it's available.
That is actually in the specification: https://github.com/open-telemetry/opentelemetry-specification/blob/main/specification/trace/semantic_conventions/http.md#http-server-semantic-conventions

luismiramirez:
Brilliant, I'll open a PR to set the `http.route` when available. Thanks for the feedback!

PR #3183, opened 2023-01-26
[samplers/jaegerremote] Change parser to support enums as both strings and numbers

### LINKED ISSUE #3184: [samplers/jaegerremote] Invalid expectation of JSON format
## Background

Originated from https://github.com/open-telemetry/opentelemetry-specification/issues/3126

The official [Jaeger remote sampling protocol][1] returns `strategyType` enums as strings. Both `jaeger_agent:5778/sampling` and `jaeger_collector:14268/api/sampling` endpoints return that format. OTel Collector also returns the same format ([code][2], [unit test][3]). There is another, long-deprecated, format with enums as numbers, which is historically returned from `jaeger_agent:5778/` endpoint (only very early versions of Jaeger SDKs have used this in 2015, before Thrift 0.9.3 was released):

```shell
# official
$ curl "http://localhost:5778/sampling?service=foo"
{"strategyType":"PROBABILISTIC","probabilisticSampling":{"samplingRate":1}}

# deprecated, legacy
$ curl "http://localhost:5778/?service=foo"
{"strategyType":0,"probabilisticSampling":{"samplingRate":1}}
```

[1]: https://www.jaegertracing.io/docs/1.41/apis/#remote-sampling-configuration-stable
[2]: https://github.com/open-telemetry/opentelemetry-collector-contrib/blob/395c731d066c830772afe099fcf7e0d4e4d4bee4/extension/jaegerremotesampling/internal/http.go#L117
[3]: https://github.com/open-telemetry/opentelemetry-collector-contrib/blob/8049c8c9ca22a722568cb16f3494094d8fb4da9a/extension/jaegerremotesampling/internal/http_test.go#L90

## Problem
This extension expects the legacy format with enums as numbers, as demonstrated by this PR: #3183. Thus the SDK extension is completely incompatible with the OTel Collector's extension `jaegerremotesampling`, and is only compatible with the legacy endpoint in Jaeger Agent.

## Proposed Solution

* Option 1: fix this extension to expect the official format (potentially breaking change for existing deployments)
* Option 2: add a configuration option
* Option 3: sniff the payload and recognize both formats (e.g., the way Jaeger generates the legacy format is by first rendering the official format and then doing a dump string substitution to replace string enums with numbers). I would go the other way in this change, numbers to strings, so that official gRPC JSON parsing could be used.

yurishkuro:
Resolves #3184

* Switch from `encoding/json` to `gogoproto/jsonpb` parser that works for both string and number enums in JSON


samplers/jaegerremote/sampler_remote_test.go
```diff
@@ -535,3 +535,55 @@ func getSamplingStrategyResponse(strategyType jaeger_api_v2.SamplingStrategyType
...
+			s := res.(*jaeger_api_v2.SamplingStrategyResponse)
+			test.assert(t, s)
+		})
+	}
+}
+
+func TestSamplingStrategyParserImpl_Error(t *testing.T) {
+	json := `{"strategyType":"foo_bar","probabilisticSampling":{"samplingRate":0.42}}`
+	any, err := new(samplingStrategyParserImpl).Parse([]byte(json))
+	require.Error(t, err, "output: %+v", any)
```
Aneurysm9, line 587:
```suggestion
	val, err := new(samplingStrategyParserImpl).Parse([]byte(json))
	require.Error(t, err, "output: %+v", val)
```

`any` is a builtin type and can't be used as a variable name.

Aneurysm9 (approved):
Other than the linter not liking your variable names, this LGTM.

samplers/jaegerremote/sampler_remote_test.go
yurishkuro: GH seems confused, the two most recent commits already changed this but it keeps showing stale file version (and running the workflow against it)

MrAlias (approved):

yurishkuro:
can this be merged?

hanyuancheung (approved):

Narenderbhcu:
Hi, Can you please provide the release plan for this changes?

Issue #3513, opened 2023-02-28
Point release of go.opentelemetry.io/contrib/instrumentation/net/http/otelhttp v0.35.0 needed


MrAlias:
In order to unblock the Kubernetes project, the fix from https://github.com/open-telemetry/opentelemetry-go-contrib/pull/2983 needs to be back-ported to v0.35.0 for the go.opentelemetry.io/contrib/instrumentation/net/http/otelhttp package.

### Proposal

1. Create a branch from the v0.35.1 release
2. Cherry-pick the fix from #2983 to that branch
3. Tag the branch with `instrumentation/net/http/otelhttp/v0.35.1`

dashpole:
The initial bug from upstream: https://github.com/kubernetes/kubernetes/issues/113791.  Because of that, the tracing feature in k8s is not useable until we get this fix.

I tried to update the otelhttp version here: https://github.com/kubernetes/kubernetes/pull/115208 but discovered that it was blocked by https://github.com/kubernetes/kubernetes/issues/113366, which looks like it will take a while to resolve.

MrAlias:
I created the [`release-otelhttp-v0.35.1`](https://github.com/open-telemetry/opentelemetry-go-contrib/tree/release-otelhttp-v0.35.1) for this.

MrAlias:
Point release complete: https://github.com/open-telemetry/opentelemetry-go-contrib/tree/instrumentation/net/http/otelhttp/v0.35.1

PR #2983, opened 2022-11-10
otelhttp: Treat http.NoBody the same as a nil body

dashpole:
Fixes https://github.com/open-telemetry/opentelemetry-go-contrib/issues/2982

PR #3514, opened 2023-02-28
Cherrypick #2983: treat http.NoBody the same as a nil body

dashpole:
Part of https://github.com/open-telemetry/opentelemetry-go-contrib/issues/3513

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -6,7 +6,11 @@ The format is based on [Keep a Changelog](https://keepachangelog.com/en/1.0.0/).
 
 This project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0.html).
 
-## [Unreleased]
+## [0.35.1]
+
+### Fixed
+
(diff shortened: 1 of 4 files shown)


MrAlias (approved):

Aneurysm9 (approved):
Will also need  a PR to `main` adding the changelog entry.

dashpole:
do I also need to update otelhttp version.go?  do I need to update version.yaml?  Or should I leave the changelog as Unreleased?

MrAlias:
> do I also need to update otelhttp version.go? do I need to update version.yaml? Or should I leave the changelog as Unreleased?

Good question. I think `version.go` needs to be updated in the `otelhttp` directory.

I'm not sure `version.yaml` needs to be updated as I plan on tagging this outside of our tooling.

The changelog looks good to me given this is the end of the road for this branch's development.

dashpole:
Updated version.go

Issue #2755, opened 2022-09-16
Add integration testing to the host instrumenation


MrAlias:
Integration testing for the `host` instrumentation was removed in #2754 

When https://github.com/open-telemetry/opentelemetry-go/issues/3031 is resolved, a new `host/test` package need to be created and integration testing for the instrumentation added to it.

MrAlias:
Blocked by https://github.com/open-telemetry/opentelemetry-go/issues/3407

MrAlias:
Unblocked. Use a manual-reader to collect expected metrics and the `metricdatatest` package to compare expected values (or just compare directly).

Petrie:
hi @MrAlias , I would like to work on this issue, please assign it to me.

JAFFASH99:
hi mr alias . id like to work on this doc plz assign it to me also.

PR #3517, opened 2023-03-01
Add unit tests for host instrumentation

Petrie:
Resove https://github.com/open-telemetry/opentelemetry-go-contrib/issues/2755


dmathieu:
Adding tests shouldn't need a changelog entry.
Also, what is preventing those tests from residing in the same package as the code?

Petrie:
@dmathieu 
1. change log will be removed
2. By adding these tests to a separate `test` sub-directory, we can avoid unnecessary dependencies on the metric SDK in the `host` directory.

.github/dependabot.yml
```diff
@@ -10,6 +10,132 @@ updates:
     schedule:
       interval: weekly
       day: sunday
+  - package-ecosystem: docker
```
MrAlias, line 13: Can you split these changes into their own PR? I'm guessing they are coming from updating the build tools and are good changes, just they are unrelated to the changes included in this PR.

Petrie: Done. New PR https://github.com/open-telemetry/opentelemetry-go-contrib/pull/3522 for Dockerfile dependabot config.

instrumentation/host/test/host_test.go
```diff
@@ -0,0 +1,265 @@
...
+	// Validate process times:
+	// User times are in range
+	require.LessOrEqual(t, processBefore.User, processUser)
+	require.GreaterOrEqual(t, processAfter.User, processUser)
+	// System times are in range
+	require.LessOrEqual(t, processBefore.System, processSystem)
+	require.GreaterOrEqual(t, processAfter.System, processSystem)
+	// Ranges are not empty
+	require.NotEqual(t, processAfter.System, processBefore.System)
+	require.NotEqual(t, processAfter.User, processBefore.User)
```
pellared, line 130: In this block we could use `assert.` here instead of `require.`.

instrumentation/host/test/host_test.go
```diff
@@ -0,0 +1,265 @@
...
+	require.Equal(t, 1, len(hostAfter))
+	// User times are in range
+	require.LessOrEqual(t, hostBefore[0].User, systemUser)
+	require.GreaterOrEqual(t, hostAfter[0].User, systemUser)
+	// System times are in range
+	require.LessOrEqual(t, hostBefore[0].System, systemSystem)
+	require.GreaterOrEqual(t, hostAfter[0].System, systemSystem)
+	// Ranges are not empty
+	require.NotEqual(t, hostAfter[0].System, hostBefore[0].System)
+	require.NotEqual(t, hostAfter[0].User, hostBefore[0].User)
```
pellared, line 147: In this block we could use `assert.` here instead of `require.`.

pellared (commented):
The tests are failing (or are flaky)

pellared:
@Petrie, are you still willing to work on this?

Issue #3388, opened 2023-02-16
Mongo instrumentation's 'db.statement' tag could be leaking sensitive data


dubonzi:
The Mongo instrumentation adds a ['db.statement' tag](https://github.com/open-telemetry/opentelemetry-go-contrib/blob/main/instrumentation/go.mongodb.org/mongo-driver/mongo/otelmongo/mongo.go#L59) to spans containing the raw query which might leak sensitive information. I see a `TODO` in the sanitization function that would be doing the obfuscation of these values, and since it's not yet done, maybe this feature should be disabled by default to avoid leaks by users that are unaware of this.

dmathieu:
You are absolutely right. Do you want to open a PR?

dubonzi:
I could, you mean to disable the feature by default?

dmathieu:
I think we should first disable the feature. Implementing obfuscation is going to take longer.

dubonzi:
Done.

PR #3519, opened 2023-03-02
Disable adding the mongo 'db.statement' tag by default

dubonzi:
As of now, the 'db.statement' tag is not obfuscated, which can lead to sensitive information being leaked through the tag.

See #3388.

### DIFF
--- a/instrumentation/go.mongodb.org/mongo-driver/mongo/otelmongo/config.go
+++ b/instrumentation/go.mongodb.org/mongo-driver/mongo/otelmongo/config.go
@@ -33,7 +33,8 @@ type config struct {
 // newConfig returns a config with all Options set.
 func newConfig(opts ...Option) config {
 	cfg := config{
-		TracerProvider: otel.GetTracerProvider(),
+		TracerProvider:           otel.GetTracerProvider(),
+		CommandAttributeDisabled: true,
 	}
 	for _, opt := range opts {
 		opt.apply(&cfg)
@@ -68,7 +69,8 @@ func WithTracerProvider(provider trace.TracerProvider) Option {
 }
 
 // WithCommandAttributeDisabled specifies if the MongoDB command is added as an attribute to Spans or not.
-// The MongoDB command will be added as an attribute to Spans by default if this option is not provided.
+// This is disabled by default and the MongoDB command will not be added as an attribute
+// to Spans if this option is not provided.
 func WithCommandAttributeDisabled(disabled bool) Option {
 	return optionFunc(func(cfg *config) {
 		cfg.CommandAttributeDisabled = disabled


dmathieu:
Thank you! 🌷 
This will require a changelog entry.

dubonzi:
Done.

CHANGELOG.md
```diff
@@ -16,6 +16,7 @@ This project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0.htm
 
 - Dropped compatibility testing for [Go 1.18].
   The project no longer guarantees support for this version of Go. (#3813)
+- Adding the `db.statement` tag to spans in `go.opentelemetry.io/contrib/instrumentation/go.mongodb.org/mongo-driver/mongo/otelmongo` is now disabled by default.
```
dmathieu, line 19:
```suggestion
- Adding the `db.statement` tag to spans in `go.opentelemetry.io/contrib/instrumentation/go.mongodb.org/mongo-driver/mongo/otelmongo` is now disabled by default. (#3519)
```

dmathieu (commented):
(sorry for the back and forth)

I think we should update the comment in https://github.com/open-telemetry/opentelemetry-go-contrib/blob/3ffe346226682df627916139633ab973f214bb1c/instrumentation/go.mongodb.org/mongo-driver/mongo/otelmongo/mongo.go#L110 to say something like:

```
// TODO sanitize values where possible, then reenable `db.statement` span attributes default.

CHANGELOG.md
dubonzi: Done!

dmathieu (approved):

pellared (changes requested):
Personally, I agree that we should be "secure by default".
Unfortunately, it is currently against the OTel specification.

https://github.com/open-telemetry/opentelemetry-specification/blob/main/specification/trace/semantic_conventions/database.md#call-level-attributes

Can you first try addressing it in the OTel specification? I would support it 😉 

Related PR in spec: https://github.com/open-telemetry/opentelemetry-specification/pull/1659

pellared:
Related issue in OTel spec: https://github.com/open-telemetry/opentelemetry-specification/issues/3210

dubonzi:
I added a comment in support of being able to have it disabled by default if the instrumentation doesn't implement obfuscation.

pellared:
@dubonzi There is also an open PR here: https://github.com/open-telemetry/opentelemetry-specification/pull/3127

dubonzi:
> Related issue in OTel spec: [open-telemetry/opentelemetry-specification#3104](https://github.com/open-telemetry/opentelemetry-specification/issues/3104)

@dmathieu I see you tried to link to this PR in a comment but you linked to the wrong one.

dubonzi:
While the specification issue seems to be moving forward, I was thinking, would it be acceptable to use datadog's [obfuscation library](https://github.com/DataDog/datadog-agent/tree/main/pkg/obfuscate) that is used in their [agent](https://github.dev/DataDog/datadog-agent/blob/main/pkg/trace/agent/obfuscate.go#L93) to do the sanitization?

pellared:
> While the specification issue seems to be moving forward, I was thinking, would it be acceptable to use datadog's [obfuscation library](https://github.com/DataDog/datadog-agent/tree/main/pkg/obfuscate) that is used in their [agent](https://github.dev/DataDog/datadog-agent/blob/main/pkg/trace/agent/obfuscate.go#L93) to do the sanitization?

Obfuscation on "raw data" is never perfect. We could use similar feature as opt-in, but we should not use it by default.

dubonzi:
Can we move forward with this now that the specification issue was accepted?

pellared (approved):
Side note: this package has no unit tests 😬

MrAlias (approved):

PR #3535, opened 2023-03-05
[WIP] Add go-resty instrumentation

dubonzi:
This adds instrumentation for the [go-resty client](https://github.com/go-resty/resty). 

I have a question, do I need to include examples? I added an examples folder locally but it's being ignored by git, but I see that other instrumentations have it.

Still a work in progress but any feedback is welcome!


dubonzi:
> See https://github.com/open-telemetry/opentelemetry-go-contrib/blob/main/instrumentation/README.md#new-instrumentation

That's my bad, I honestly didn't see those guidelines, I should've paid more attention before starting working on this. I can try contacting the maintainers for the possibility of including the instrumentation there. 

If I read correctly, in order for the instrumentation to be added to the registry, it has to be hosted by the maintainers, right? If I end up creating it on my own, would it be a valid candidate?

dmathieu:
The registry at https://opentelemetry.io/ecosystem/registry/ doesn't require libraries to be hosted within this repository. You can have it within any account.

dubonzi:
Thank you, I'm gonna close this and move it to its own repository then.

PR #3580, opened 2023-03-13
Allow respWriterWrapper to call ResponseWriter.WriteHeader multiple times

jarlefosen:
otelhttp respWriterWrapper blocks multiple calls to underlying WriteHeader which makes it harder to detect bugs in code where multiple calls to WriteHeader would be caught and logged by net/http.

This PR attempts to expose incorrect code by not blocking multiple calls to the underlying `ResponseWriter.WriteHeader`.


jarlefosen:
You're right. I've implemented tests to ensure we propagate WriteHeader calls to underlying ResponseWriters regardless of respWriterWrapper state.

instrumentation/net/http/otelhttp/wrap.go
```diff
@@ -86,10 +86,9 @@ func (w *respWriterWrapper) Write(p []byte) (int, error) {
 }
 
 func (w *respWriterWrapper) WriteHeader(statusCode int) {
-	if w.wroteHeader {
-		return
+	if !w.wroteHeader {
+		w.wroteHeader = true
+		w.statusCode = statusCode
```
pellared, line 91: If feel that this change is not correct.

You always call `w.ResponseWriter.WriteHeader(statusCode)` but on the other side you assume that it is not possible to change the status code after it was initially set (becasuse `w.statusCode = statusCode` is inside an `if` statement).

If there would be a `ResponseWriter` that allows setting the status code multiple times then this PR does not fix it.

Personally, I would rather consider changing this code to:

```go
func (w *respWriterWrapper) WriteHeader(statusCode int) {
	w.statusCode = statusCode
	w.ResponseWriter.WriteHeader(statusCode)
	w.wroteHeader = true
}
```

but then it would lead to strange behavior when the default `ResponseWriter` is used (but there would be logs from http package)...

jarlefosen:
Thanks for the comment @pellared. I'll answer in the main thread to avoid losing context if the [comment is resolved](https://github.com/open-telemetry/opentelemetry-go-contrib/pull/3580#discussion_r1135140231)

The primary goal for me is to have our system behave the same with and without otelhttp.

Today my code can work nicely with multiple calls to WriteHeader since `otelhttp` is a middleware that ensures the ResponseWriter never gets multiple WriteHeader calls.

This is not the responsibility of OpenTelemetry in my opinion, and can lead to unknown bad implementations of HTTP Handlers where WriteHeader is called multiple times, or it is called after `ResponseWriter.Write` without the developer noticing it. _Which incidentally is what happened in our codebase without our knowledge for some time._

I believe that otelhttp should report the initially written statusCode, as it does today, and not change reported status based on bad implementations. Changing the reported statusCode on consecutive calls to WriteHeader would lead to discrepancies between the attributes on the span vs what was actually sent to the client.

The goal of otelhttp would be to report the first written statusCode on the span without filtering out future WriteHeader calls.

If it is something we believe should be implemented I'll go ahead and change the code to document behavior in more detail.

pellared:
@jarlefosen I think it makes sense especially that now I see that I do not believe anyone expects the `ResponseWriter` to handle multiple calls to `WriteHeader` properly. For example from https://pkg.go.dev/net/http#NewResponseController, we have:

> The ResponseWriter should be the original value passed to the Handler.ServeHTTP method, or have an Unwrap method returning the original ResponseWriter.

Documenting behavior in more detail seems fine.

instrumentation/net/http/otelhttp/wrap.go
pellared: Per https://github.com/open-telemetry/opentelemetry-go-contrib/pull/3580#issuecomment-1468123160

I think that it would be good enough to just leave a comment describing the behavior in more detail e.g. to avoid a regression bug.

jarlefosen:
@pellared I have finally gotten back to this and updated comments and documentation.
PR is updated and branch is rebased on main.

CHANGELOG.md
```diff
@@ -12,6 +12,10 @@ This project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0.htm
 
 - Add the new `go.opentelemetry.io/contrib/instrgen` package to provide auto-generated source code instrumentation. (#3068)
 
+### Fixed
+
+- Enable `otelhttp.Handler` to propagate multiple WriteHeader calls while persisting initial statusCode in `go.opentelemetry.io/contrib/instrumentation/net/http/otelhttp`. (#3580)
```
pellared, line 17:
```suggestion
- Enable `otelhttp.Handler` to propagate multiple `WriteHeader` calls while persisting the initial HTTP status code in `go.opentelemetry.io/contrib/instrumentation/net/http/otelhttp`. (#3580)
```

instrumentation/net/http/otelhttp/wrap.go
```diff
@@ -85,11 +85,15 @@ func (w *respWriterWrapper) Write(p []byte) (int, error) {
 	return n, err
 }
 
+// WriteHeader persists initial statusCode for span attribution.
+// All calls to WriteHeader will be propagated to the underlying ResponseWriter
+// but will not overwrite persisted statusCode.
```
pellared, line 90:
```suggestion
// and will persist the statusCode from the first call.
```

pellared (approved):

CHANGELOG.md
jarlefosen: Thank you so much. I struggled so much finding the words to describe this tiny change.

pellared (approved):
Thanks for your patience and contribution 👍

hanyuancheung (approved):

MrAlias (approved):

Issue #3603, opened 2023-03-18
[jaegerremote] probabilistic strategy is not working as expected


vietdoan:
### Description
We configured our Jaeger remote sampler as below
```go
jaegerRemoteSampler := jaegerremote.New(
	cfg.ServiceName,
	jaegerremote.WithSamplingServerURL(cfg.Tracing.SamplingServerURL),
	jaegerremote.WithSamplingRefreshInterval(5*time.Minute), // decrease polling interval to get quicker feedback
	jaegerremote.WithInitialSampler(sdktrace.TraceIDRatioBased(cfg.Tracing.SamplingFraction)),
)
```
Configuration for `sampling.strategies-file`
```json
{
     "default_strategy": {
        "type": "probabilistic",
        "param": 0.0,
        "operation_strategies": [
          {
            "operation": "/health",
            "type": "probabilistic",
            "param": 0.0
          },
          {
            "operation": "/metrics",
            "type": "probabilistic",
            "param": 0.0
          }
        ]
      }
}
```
When I tried generating the traces for a `/liveness` endpoint, 1 sampling trace of each application container was still reported to jaeger backend every 5 minutes 
![telegram-cloud-photo-size-5-6098420048252220869-y](https://user-images.githubusercontent.com/13875142/226095247-f93aff22-c599-4d5a-838c-77933b9f3ee9.jpg)
### Environment
OS: Linux Based OS
Go Version: 1.19
opentelemetry-go version: v1.14.0
### Expected behavior
There is no sampling trace according to the probability configured in `sampling.strategies-file`

This issue might be part of #3029

PR #3604, opened 2023-03-18
Fix rate limiting sampler to drop trace when lower bound is zero

vietdoan:
Every time we fetch config, a new `rateLimitingSampler` is created and then it always accept a trace even though the `lowerbound` is zero. As `guaranteedThroughputProbabilisticSampler` leverages both `probabilisticSampler` and `rateLimitingSampler`, the sampler still sample trace with zero probability as described in #3603


MrAlias (approved):

MrAlias:
Please add a changelog entry for the fix.

hanyuancheung (approved):

dmathieu (approved):

MrAlias (approved):

Issue #3521, opened 2023-03-02
Sync DynamoDB instrumentation attributes with Python instrumentation


SenorMooMoo:
The current instrumentation of DynamoDB does not set the same attributes:

- Span name is different, does not include the operation like in Python (DynamoDB.GetItem vs DynamoDB)
- rpc.method is set to aws.operation in Go
- rpc.service is set to aws.service in Go

dmathieu:
I haven't investigated this yet. But we should not "sync with python".
We should sync with the specification, if we don't yet: https://github.com/open-telemetry/opentelemetry-specification/blob/66f8c4c884d75ed3882af4f31ab036271bdd2a94/specification/trace/semantic_conventions/instrumentation/aws-sdk.md

dmathieu:
It appears the span name will be fixed with #3582.

PR #3617, opened 2023-03-21
Sync aws sdk attributes with semantic conventions

nemoshlag:
Solves #3521 for DynamoDB and AWS SDK.

Making AWS SDK attributes consistent with RPC clients:

1 Rename "aws.operation" attribute to "rpc.method"
2 Rename "aws.service" attribute to "rpc.service"
3 Add "rpc.system" attribute and set it to "aws-api" value

According to [semconv](https://github.com/open-telemetry/opentelemetry-specification/blob/main/specification/trace/semantic_conventions/instrumentation/aws-sdk.md)


instrumentation/github.com/aws/aws-sdk-go-v2/otelaws/attributes.go
```diff
@@ -48,7 +54,7 @@ func RegionAttr(region string) attribute.KeyValue {
 	return RegionKey.String(region)
 }
 
-// ServiceAttr returns the AWS service attribute.
+// ServiceAttr returns the RPC service attribute.
```
dmathieu, line 57: I'm not sure this doc should change. Even though the attributes change, mentioning AWS here feels better.

dmathieu (approved):

CHANGELOG.md
```diff
@@ -10,6 +10,7 @@ This project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0.htm
 
 ### Added
 
+- AWS SDK add `rpc.system` attribute. (#3617)
```
MrAlias, line 13: Please use the full package import path instead of just saying "AWS SDK". That way users can search the changelog to find changes related to the package they are using.

instrumentation/github.com/aws/aws-sdk-go-v2/otelaws/attributes.go
```diff
@@ -38,9 +37,14 @@ var servicemap = map[string]AttributeSetter{
 	sqs.ServiceID:      SQSAttributeSetter,
 }
 
+// SystemAttr return the RPC system attribute.
+func SystemAttr(operation string) attribute.KeyValue {
+	return semconv.RPCSystemKey.String(operation)
```
MrAlias, line 42: Is there ever a case where this shouldn't just use `"aws-api"` as the system?

nemoshlag: It seems to be the only value for aws sdk instrumentations according to [semantic convention](https://github.com/open-telemetry/opentelemetry-specification/blob/main/specification/trace/semantic_conventions/instrumentation/aws-sdk.md#common-attributes)

MrAlias: Then let's have this just return a constant and not accept agreements, right?

PR #3658, opened 2023-03-28
Add WithResource option for the remote sampler constructor

jj22ee:
To be consistent across the OTel SDK XRay samplers, the service name and cloud platform fields of the Remote Sampler should be able to be retrieved from a Resource. Currently, These fields are only configurable through the Remote Sampler constructor.

Context of Issue:
The resource of the TracerProvider is not available to the sampler ([GH Issue](https://github.com/open-telemetry/opentelemetry-specification/issues/1588)), and it does not look like it will be anytime soon. As an alternative, the resource should be able to be passed into the sampler instantiation.


dmathieu:
Could you add tests?

jj22ee:
Closed as I will create a new PR that may be potentially different.

PR #3676, opened 2023-04-04
Func literal instrumentation fix

pdelewski:
Fixing following case 
```
func foo() {
  func() {}
}
```
func inside foo is anonymous one and in fact it's not valid Go program, however otel calls were incorrectly added inside body of it


MrAlias (approved):

hanyuancheung (approved):

pdelewski:
Moving all this work to Go compile instrumentation sig

PR #3680, opened 2023-04-04
Instantiate XRay Remote Sampler with OTel Resource

jj22ee:
To be consistent across the OTel SDK XRay samplers, the service name and cloud platform fields of the Remote Sampler should be able to be retrieved from a Resource. Currently, These fields are only configurable through the Remote Sampler constructor.

Context of Issue:
The resource of the TracerProvider is not available to the sampler ([GH Issue](https://github.com/open-telemetry/opentelemetry-specification/issues/1588)), and it does not look like it will be anytime soon. As an alternative, the resource should be able to be passed into the sampler instantiation.


samplers/aws/xray/remote_sampler.go
```diff
@@ -83,6 +93,54 @@ func NewRemoteSampler(ctx context.Context, serviceName string, cloudPlatform str
 	return remoteSampler, nil
 }
 
+// NewRemoteSamplerWithResource returns a sampler which decides to sample a given request or not
+// based on the sampling rules set by users on AWS X-Ray console. Sampler also periodically polls
+// sampling rules and sampling targets.
+// NOTE: ctx passed in NewRemoteSamplerWithResource API is being used in background go routine. Cancellation to this context can kill the background go routine.
+// This takes in an OTel resource in order to configure service name and cloud platform for service rule matching. Their default values are empty strings if they are not in the resource.
+// Guide on AWS X-Ray remote sampling implementation (https://aws-otel.github.io/docs/getting-started/remote-sampling#otel-remote-sampling-implementation-caveats).
+func NewRemoteSamplerWithResource(ctx context.Context, otelResource *resource.Resource, opts ...Option) (sdktrace.Sampler, error) {
```
srprash, line 102: Can we mark the `NewRemoteSampler` method as deprecated in the favor of this one?

srprash (approved):

samplers/aws/xray/remote_sampler.go
jj22ee: Marked as deprecated

hanyuancheung (approved):

PR #3682, opened 2023-04-05
instrgen fixing imports

pdelewski:
Remove import that only is needed in `main.go` from the other test data files so it will not fail the build.

### DIFF
--- a/instrgen/driver/testdata/basic/fib.go
+++ b/instrgen/driver/testdata/basic/fib.go
@@ -12,7 +12,7 @@
 // See the License for the specific language governing permissions and
 // limitations under the License.
 
-//nolint:all
+//nolint:all // Linter is executed at the same time as tests which leads to race conditions and failures.
 package main
 
 import (
--- a/instrgen/driver/testdata/basic/goroutines.go
+++ b/instrgen/driver/testdata/basic/goroutines.go
@@ -17,8 +17,6 @@ package main
 
 import (
 	"fmt"
-
-	"go.opentelemetry.io/contrib/instrgen/rtlib"
 )
 
 func goroutines() {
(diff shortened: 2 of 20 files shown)


pellared:
How about removing `//nolint:all` and addressing all the issues? Would it be too much?

pdelewski:
> How about removing `//nolint:all` and addressing all the issues? Would it be too much?

`//nolint:all` was added for different reason. Linter is executed at the same time as tests which leads to race conditions and failures

MrAlias:
Why are these imports removed?

MrAlias:
It looks like the tests are failing. Can you take a look at the data race?

pdelewski:
> Why are these imports removed?

This import is needed only in `main.go`, they are leftovers and if you will try to build this test program compiler will complain about unused imports

pdelewski:
> It looks like the tests are failing. Can you take a look at the data race?

They are unrelated to `instrgen`

MrAlias (approved):

pellared (approved):

pellared:
> > How about removing `//nolint:all` and addressing all the issues? Would it be too much?
> 
> `//nolint:all` was added for different reason. Linter is executed at the same time as tests which leads to race conditions and failures

Then it would be good to add the reason in a comment e.g. like `//nolint:all // Linter is executed at the same time as tests which leads to race conditions and failures.` Possibly also track it in some issue to resolve it at some point.

pdelewski:
> > > How about removing `//nolint:all` and addressing all the issues? Would it be too much?
> > 
> > 
> > `//nolint:all` was added for different reason. Linter is executed at the same time as tests which leads to race conditions and failures
> 
> Then it would be good to add the reason in a comment e.g. like `//nolint:all // Linter is executed at the same time as tests which leads to race conditions and failures.` Possibly also track it in some issue to resolve it at some point.

Good point.

PR #3684, opened 2023-04-07
Disable "taking" from reservoir if there is no capacity, regardless of borrow value

jj22ee:
When `borrow=true`, the Golang remote sampler allows a quota of 1req/s. However, this also applies when capacity is 0 (which implies that quota should always be 0, as quota is a portion of capacity). For context, `borrow=true` when quota is unknown but will be determined later.

This fix makes sure that the sampler cannot "take" when capacity is 0, which will also disable "borrowing". That is, when capacity is 0, do not take from reservoir at all, regardless of borrow = true/false.


srprash (approved):
LGTM

Aneurysm9 (approved):

MrAlias (approved):

PR #3685, opened 2023-04-07
Align gRPC server status code to span status code

FaranIdo:
Following an update to OTEL spec, align gRPC server instrumentation to the new span status code

https://github.com/open-telemetry/opentelemetry-specification/blob/f5a269fd3de948656c085c25d83342c78c8dc99c/specification/trace/semantic_conventions/rpc.md#grpc-status

Pull Request in the OTEL spec: https://github.com/open-telemetry/opentelemetry-specification/pull/3333


dmathieu:
Could you add a changelog entry, as well as tests?

FaranIdo:
@dmathieu Yes, on it, will add it later

FaranIdo:
Hi @dmathieu, I have added tests and a changelog entry for the changes I made. 
Additionally, I noticed that there were no tests for the StreamServerInterceptor flow, so I went ahead and added them.

hanyuancheung (approved):

instrumentation/google.golang.org/grpc/otelgrpc/test/interceptor_test.go
```diff
@@ -583,39 +583,150 @@ func TestStreamClientInterceptorWithError(t *testing.T) {
 	assert.Equal(t, codes.Error, span.Status().Code)
 }
 
-func TestServerInterceptorError(t *testing.T) {
+var serverChecks = []struct {
+	grpcCode grpc_codes.Code
+	spanCode codes.Code
```
pellared, line 588: Can we rename it to `wantSpanCode` to indicate that this is what we are expecting to get?

instrumentation/google.golang.org/grpc/otelgrpc/test/interceptor_test.go
```diff
@@ -583,39 +583,150 @@ func TestStreamClientInterceptorWithError(t *testing.T) {
...
+	name     string
```
pellared, line 589: Would it not be more readable to rename it to `method`?

instrumentation/google.golang.org/grpc/otelgrpc/test/interceptor_test.go
```diff
@@ -583,39 +583,150 @@ func TestStreamClientInterceptorWithError(t *testing.T) {
...
 	tp := trace.NewTracerProvider(trace.WithSpanProcessor(sr))
 	usi := otelgrpc.UnaryServerInterceptor(otelgrpc.WithTracerProvider(tp))
-	deniedErr := status.Error(grpc_codes.PermissionDenied, "PERMISSION_DENIED_TEXT")
-	handler := func(_ context.Context, _ interface{}) (interface{}, error) {
-		return nil, deniedErr
-	}
-	_, err := usi(context.Background(), &grpc_testing.SimpleRequest{}, &grpc.UnaryServerInfo{}, handler)
-	require.Error(t, err)
-	assert.Equal(t, err, deniedErr)
+	for _, check := range serverChecks {
```
pellared, line 632: How about adding a `t.Run(check.grpcCode.String(), func ...`? 
Then the `require` as far as I remember should not break the whole "test suite" and we could use `require` instead of `if !assert... { continue}` in some places.

PS. Same comment for the second test.

instrumentation/google.golang.org/grpc/otelgrpc/test/interceptor_test.go
```diff
@@ -583,39 +583,150 @@ func TestStreamClientInterceptorWithError(t *testing.T) {
...
+		if assert.True(t, codeAttr.Valid(), "attributes contain gRPC status code") {
+			assert.Equal(t, attribute.Int64Value(int64(check.grpcCode)), codeAttr.Value)
+		}
+
+		// validate events and their attributes
+		assert.Len(t, span.Events(), 2)
+		assert.ElementsMatch(t, []attribute.KeyValue{
+			attribute.Key("message.type").String("SENT"),
+			attribute.Key("message.id").Int(1),
+		}, span.Events()[1].Attributes)
```
pellared, line 676: Can we create some helper assertion methods to remove some duplication? I think it would also increase the readability.

instrumentation/google.golang.org/grpc/otelgrpc/test/interceptor_test.go
FaranIdo: `getSpanFromRecorder` is always called in this test file with the `name` variable. So I wanted to follow this convention
I agree it makes line 696 (`grpc.StreamServerInfo{FullMethod: check.name}`) a bit less readable, but I think consistency matters more here.
Maybe a better change will be to remove the duplicated name + method in the client's tests, but I preferred to avoid refactoring them.

instrumentation/google.golang.org/grpc/otelgrpc/test/interceptor_test.go
FaranIdo: done

dmathieu (approved):

instrumentation/google.golang.org/grpc/otelgrpc/test/interceptor_test.go
```diff
@@ -583,24 +583,67 @@ func TestStreamClientInterceptorWithError(t *testing.T) {
...
-	span, ok := getSpanFromRecorder(sr, "")
-	if !ok {
-		t.Fatalf("failed to export error span")
+func assertServerSpan(t *testing.T, span trace.ReadOnlySpan, wantSpanCode codes.Code, grpcCode grpc_codes.Code, grpcErr error) {
+	// validate span status
+	assert.Equal(t, wantSpanCode, span.Status().Code)
+	if span.Status().Code == codes.Error {
+		assert.Contains(t, grpcErr.Error(), span.Status().Description)
+	} else {
+		assert.Empty(t, span.Status().Description)
```
pellared, line 643: I would prefer to add `wantSpanStatusDescription` to `serverChecks` and change this to:

```suggestion
	assert.Equal(t, wantSpanStatusDescription, span.Status().Description)
```
Having too much conditional in tests logic makes them more obscure, harder to maintain, and less trustworthy.

More on this topic: http://xunitpatterns.com/Conditional%20Test%20Logic.html

instrumentation/google.golang.org/grpc/otelgrpc/test/interceptor_test.go
```diff
@@ -583,24 +583,67 @@ func TestStreamClientInterceptorWithError(t *testing.T) {
...
+		name:         "unauthenticated",
+	},
+}
+
+func assertServerErr(t *testing.T, err error, grpcCode grpc_codes.Code) {
+	// the error should be nil only if the gRPC code is OK, otherwise it should be the same as the gRPC error
+	if grpcCode == grpc_codes.OK {
+		assert.NoError(t, err)
+	} else {
+		assert.Equal(t, grpcCode, status.Code(err))
```
pellared, line 633: Can we add `grpcErr` to `serverChecks`?

So that it can be used here in the assertion by simply

```go
   assert.Equal(t, grpcCode, err)
```

and use the same error when setting up the handler:

```go
handler := func(_ context.Context, _ interface{}) (interface{}, error) {
	return nil, check.grpcErr
}
```

instrumentation/google.golang.org/grpc/otelgrpc/test/interceptor_test.go
```diff
@@ -583,24 +583,67 @@ func TestStreamClientInterceptorWithError(t *testing.T) {
...
 	}
-	_, err := usi(context.Background(), &grpc_testing.SimpleRequest{}, &grpc.UnaryServerInfo{}, handler)
-	require.Error(t, err)
-	assert.Equal(t, err, deniedErr)
+}
 
-	span, ok := getSpanFromRecorder(sr, "")
-	if !ok {
-		t.Fatalf("failed to export error span")
+func assertServerSpan(t *testing.T, span trace.ReadOnlySpan, wantSpanCode codes.Code, grpcCode grpc_codes.Code, grpcErr error) {
```
pellared, line 637: The common pattern is to have the "want" arguments first (just after `t`). Here we have the args mixed.

FaranIdo:
@pellared @dmathieu @dashpole - I fixed the CR comments.

instrumentation/google.golang.org/grpc/otelgrpc/test/interceptor_test.go
```diff
@@ -583,39 +583,137 @@ func TestStreamClientInterceptorWithError(t *testing.T) {
...
-	tp := trace.NewTracerProvider(trace.WithSpanProcessor(sr))
-	usi := otelgrpc.UnaryServerInterceptor(otelgrpc.WithTracerProvider(tp))
-	deniedErr := status.Error(grpc_codes.PermissionDenied, "PERMISSION_DENIED_TEXT")
-	handler := func(_ context.Context, _ interface{}) (interface{}, error) {
-		return nil, deniedErr
-	}
-	_, err := usi(context.Background(), &grpc_testing.SimpleRequest{}, &grpc.UnaryServerInfo{}, handler)
-	require.Error(t, err)
-	assert.Equal(t, err, deniedErr)
+var serverChecks = []struct {
```
MrAlias, line 586: This is mising the following test cases: `UNKNOWN`, `INVALID_ARGUMENT`, `DEADLINE_EXCEEDED`, `ALREADY_EXISTS`, `RESOURCE_EXHAUSTED`, `FAILED_PRECONDITION`, `ABORTED`, `OUT_OF_RANGE`, `UNAVAILABLE`, `DATA_LOSS`.

FaranIdo: I was thinking that adding all gRPC status codes is a bit overkill so I included only samples of the major ones. But I can include all of them if you think otherwise.

FaranIdo:
@MrAlias added all gRPC statuses.
I also did small refactor in the test by removing the name + grpcErr from the `serverChecks` variable. Now it is much more simple -  just grpcCode, span status + span description.

instrumentation/google.golang.org/grpc/otelgrpc/test/interceptor_test.go
FaranIdo: I managed to remove this "if" branching and avoid duplicating it in `serverChecks` - cause `status.Error` returns nil if input code is "OK", so  the handler + the assert will use `nil` anyway in case of OK status

pellared (approved):
Thank you for your contribution 👍

MrAlias (approved):

PR #3706, opened 2023-04-14
Instrgen gencode sema validation

pdelewski:
The goal of this PR is to validate AST after each transformation


pdelewski:
Closing it, as obsolete due to recent redesign

Issue #3536, opened 2023-03-06
otelhttp grpc ResourceExhausted - resources not being freed


KasonBraley:
When wrapping an http server with the [otelhttp](https://github.com/open-telemetry/opentelemetry-go-contrib/tree/main/instrumentation/net/http/otelhttp) handler, these errors occur and do not stop. The message size continues to grow and never frees:

```
context deadline exceeded: rpc error: code = ResourceExhausted desc = grpc: received message larger than max (5325126 vs. 4194304)
context deadline exceeded: rpc error: code = ResourceExhausted desc = grpc: received message larger than max (7255095 vs. 4194304)
context deadline exceeded: rpc error: code = ResourceExhausted desc = grpc: received message larger than max (9129195 vs. 4194304)
context deadline exceeded: rpc error: code = ResourceExhausted desc = grpc: received message larger than max (10831746 vs. 4194304)
context deadline exceeded: rpc error: code = ResourceExhausted desc = grpc: received message larger than max (12704602 vs. 4194304)
...
context deadline exceeded: rpc error: code = ResourceExhausted desc = grpc: received message larger than max (33605778 vs. 4194304)
```

Full stacktrace (seems to be related to metrics):
```
{
    "level": "error",
    "ts": 1678141390.0813837,
    "caller": "reporting/otel.go:37",
    "msg": "context deadline exceeded: rpc error: code = ResourceExhausted desc = grpc: received message larger than max (33605778 vs. 4194304)",
[...]

Issue #3071, opened 2022-12-08
otelgrpc high metric cardinality from rpc.server.duration


rcrowe:
Since upgrading to **v1.12.0/0.37.0/0.6.0** & specifically https://github.com/open-telemetry/opentelemetry-go-contrib/pull/2700 we've seen an explosion in the cardinality of our metrics.

1. Metric recorded: https://github.com/open-telemetry/opentelemetry-go-contrib/blob/ad04d9c71cb8974374ba8e3ee454fdaaddc3e345/instrumentation/google.golang.org/grpc/otelgrpc/interceptor.go#L346
2. Attributes are taken from `spanInfo`: https://github.com/open-telemetry/opentelemetry-go-contrib/blob/ad04d9c71cb8974374ba8e3ee454fdaaddc3e345/instrumentation/google.golang.org/grpc/otelgrpc/interceptor.go#L331
3. Peer attributes added: https://github.com/open-telemetry/opentelemetry-go-contrib/blob/ad04d9c71cb8974374ba8e3ee454fdaaddc3e345/instrumentation/google.golang.org/grpc/otelgrpc/interceptor.go#L476-L479

Is the intention that these types of high cardinality issues are handled in userland, by us using a view (or blocking at the ingestion layer) to strip labels from certain metrics after we find them problematic? On one hand I can understand that being acceptable, on the other, this feels like pushing a problem downstream to consumers that will suddenly get hit by a metric that is most likely going to grow very wide.

CC @fatsheep9146

rcrowe:
Missed https://github.com/open-telemetry/opentelemetry-specification/blob/main/specification/metrics/semantic_conventions/rpc-metrics.md#attributes which seems to suggest this is working as expected

lucasoares:
How to remove this metric? It is exposing 7000 metrics in the exporter.

regeda:
After `v0.38.0` release, I see the same issue with `otelhttp` package. We avoid this upgrade bc it blows up the memory consumption drastically.

MrAlias:
You can use a [`View`](https://pkg.go.dev/go.opentelemetry.io/otel/sdk/metric#NewView) to drop the aggregation entirely or filter the attributes causing your cardinality explosion.

regeda:
> You can use a [`View`](https://pkg.go.dev/go.opentelemetry.io/otel/sdk/metric#NewView) to drop the aggregation entirely or filter the attributes causing your cardinality explosion.

[...]

PR #3730, opened 2023-04-18
Fix cardinality issue for rpc.server.duration (#3536)

chchaffin:
Added an optional parameter for allowing peer information to be added as an attribute. By default the gRPC interceptor will ignore peer details for the sake of performance.

Related issue: https://github.com/open-telemetry/opentelemetry-go-contrib/issues/3536

### DIFF
--- a/instrumentation/google.golang.org/grpc/otelgrpc/config.go
+++ b/instrumentation/google.golang.org/grpc/otelgrpc/config.go
@@ -43,8 +43,9 @@ type config struct {
 	TracerProvider trace.TracerProvider
 	MeterProvider  metric.MeterProvider
 
-	meter             metric.Meter
-	rpcServerDuration instrument.Int64Histogram
+	meter                           metric.Meter
+	rpcServerDuration               instrument.Int64Histogram
+	includeRPCServerDurationPeerCtx bool
 }
 
 // Option applies an option value for a config.
@@ -125,6 +126,20 @@ func (o meterProviderOption) apply(c *config) {
 	if o.mp != nil {
 		c.MeterProvider = o.mp
 	}
+}
+
+type includePeerAddrAsAttribute struct {
+	use bool
+}
+
+// WithPeerAddrAsAttribute returns an Option to use peer details when
+// recording with the meter rpc.server.duration.
+func WithPeerAddrAsAttribute(peerAddrAsAttribute bool) Option {
+	return includePeerAddrAsAttribute{use: peerAddrAsAttribute}
+}
+
+func (o includePeerAddrAsAttribute) apply(c *config) {
+	c.includeRPCServerDurationPeerCtx = o.use
 }
 
 // WithMeterProvider returns an Option to use the MeterProvider when
--- a/.gitignore
+++ b/.gitignore
@@ -4,6 +4,7 @@ Thumbs.db
 .tools/
 .idea/
 .vscode/
+vendor/
 *.iml
 *.so
 coverage.*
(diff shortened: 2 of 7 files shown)


chchaffin:
I'm working on getting the CLA handled

chchaffin:
Ran `make precommit` and all linting and formatting tests passed.

hanyuancheung (approved):

instrumentation/google.golang.org/grpc/otelgrpc/config.go
```diff
@@ -127,6 +128,20 @@ func (o meterProviderOption) apply(c *config) {
...
 }
 
+type includePeerAddrAsAttribute struct {
+	use bool
+}
+
+// WithPeerAddrAsAttribute returns an Option to use peer details when
+// recording with the meter rpc.server.duration.
+func WithPeerAddrAsAttribute(peerAddrAsAttribute bool) Option {
+	return includePeerAddrAsAttribute{use: peerAddrAsAttribute}
```
dmathieu, line 138: I'm having doubts about this solution. The option seems extremely specific, and I wonder if we could find a more generic way of doing it.

pellared:
Reference https://github.com/open-telemetry/opentelemetry-specification/blob/main/specification/metrics/semantic_conventions/rpc-metrics.md#attributes

I have no time to review this PR, but at quick glance it looks that peer attributes should be added by default.

chchaffin:
> I have no time to review this PR, but at quick glance it looks that peer attributes should be added by default.
> 
> Reference https://github.com/open-telemetry/opentelemetry-specification/blob/main/specification/metrics/semantic_conventions/rpc-metrics.md#attributes

Then perhaps the spec should change? The port and ip change constantly in a cloud container environment, which I think is the primary use case. Correct me if I'm wrong though. If the desired default is to include the host and ip then clients are going to run into the same issue I have and will undoubtedly have very little insight as to why their metric payloads keep growing. I'm open to ideas though

instrumentation/google.golang.org/grpc/otelgrpc/config.go
chchaffin: The meter here is specific to one type of interceptor. I could apply this to the spans as well if that's more preferable.

pellared:
The spec currently says:

>  To avoid high cardinality, implementations should prefer the most stable of `net.peer.name` or
`net.sock.peer.addr`, depending on expected deployment profile.  For many cloud applications, this is likely
`net.peer.name` as names can be recycled even across re-instantiation of a server with a different `ip`.
>
> For client-side metrics `net.peer.port` is required if the connection is IP-based and the port is available (it describes the server port they are connecting to).
For server-side spans `net.peer.port` is optional (it describes the port the client is connecting from).
Furthermore, setting [net.transport][] is required for non-IP connection like named pipe bindings.

Take a look if this what is defined in the specification meets your expectations.

If yes, then it would be good to describe in the PR description which part of the specification this PR addressed.

If not, then maybe it would be worth to discuss it on the specification level.

chchaffin:
> 

@pellared so by default we need to have one of these two attributes set per the spec requirement:

```
Additional attribute requirements: At least one of the following sets of attributes is required:

[net.sock.peer.addr](https://github.com/open-telemetry/opentelemetry-specification/blob/main/specification/trace/semantic_conventions/span-general.md)
[net.peer.name](https://github.com/open-telemetry/opentelemetry-specification/blob/main/specification/trace/semantic_conventions/span-general.md)
```

It looks like the code, by default, tries to first set the higher cardinality attributes first https://github.com/open-telemetry/opentelemetry-go-contrib/blob/6682591fcd676a3950ff04358163f672af3284af/instrumentation/google.golang.org/grpc/otelgrpc/interceptor.go#L475-L480

Would you prefer to always set either `net.sock.peer.addr` and `net.peer.name`. Then, if the client specifies the higher cardinality attributes set those?

pellared:
Related issue https://github.com/open-telemetry/opentelemetry-go-contrib/issues/3071

Issue #2782, opened 2022-09-20
Should we have expert, or generalist handlers?


dmathieu:
## Background

Right now, our handlers are generalists. They perform both tracing and metrics (with some/most handlers only doing tracing).
This is introducing some logic duplication in those specific bits of code, as some metrics reproduce the same behavior a trace is going to provide.

For HTTP handlers for example, [the specification](https://github.com/open-telemetry/opentelemetry-specification/blob/main/specification/metrics/semantic_conventions/http-metrics.md) is asking us to provide metrics that can already be correlated from traces.
So it seems to me (that may just be me) that folks using tracing will not care about the duplicated data in metrics.

Doing both traces and metrics in the same code blocks also makes the libraries a bit harder to read. And a better separation of concerns would improve readability.

Hence this proposal to offer alternative ways.

## Proposal A

This first proposal consists in providing two handlers for each instrumentation. a `MetricHandler`, and a `TraceHandler`, each managing only the appropriate signal.
Someone loading only metrics will not get any traces from their component instrumentation.

For example, the [grpc instrumentation](https://github.com/open-telemetry/opentelemetry-go-contrib/tree/main/instrumentation/google.golang.org/grpc/otelgrpc) will introduce 4 new methods:

* `WithTraceStreamInterceptor`
* `WithTraceUnaryInterceptor`
* `WithMetricStreamInterceptor`
* `WithMetricUnaryInterceptor`

And deprecate the existing `WithStreamInterceptor` and `WithUnaryInterceptor` methods (which can alias to the tracer ones during the deprecation period since this specific instrumentation has no metrics. Instrumentations with metrics may need to duplicate their code during that period).

## Proposal B

This second proposal consists in still providing a single public handler, with private handlers that would be loaded only if metric/tracer providers are configured.
There would be no visible change to users of the library, they would keep using `WithStreamInterceptor` and `WithUnaryInterceptor`.

Under the hood however, we would setup 4 private methods:

* `traceStreamInterceptor`
* `taceUnaryInterceptor`
* `metricStreamInterceptor`
* `metricUnaryInterceptor`

The public handler would analyse whether metrics and traces are configured or not, and automatically load the appropriate handlers.

This proposal doesn't require any deprecations, since no public methods would need to be changed.

## Proposal C: Do nothing

This third proposal keeps the current status quo. Each instrumentation has a single handler, which can do both traces and metrics.

fatsheep9146:
[...]

PR #3741, opened 2023-04-24
[otelhttp] add options to disable traces/metrics in otelhttp.NewHandler

RangelReale:
`otelhttp.NewHandler` handles both traces and metrics.

When we use other libraries for traces, for instance, `otelmux`, which doesn't have metrics, we would like to use the default metrics to be consistent, but currently it is not possible to use just metrics.

I tried to add a `noop` tracer provider option, but [it seems that it is not intended to be used this way](https://github.com/open-telemetry/opentelemetry-go/issues/4027).

This PR adds 2 new options: `WithDisableTraces` and `WithDisableMetrics`. The "disable" name is used to keep compatilibity with the default being "enabled".

### DIFF
--- a/instrumentation/net/http/otelhttp/config.go
+++ b/instrumentation/net/http/otelhttp/config.go
@@ -44,6 +44,8 @@ type config struct {
 	Filters           []Filter
 	SpanNameFormatter func(string, *http.Request) string
 	ClientTrace       func(context.Context) *httptrace.ClientTrace
+	DisableTraces     bool
+	DisableMetrics    bool
 
 	TracerProvider trace.TracerProvider
 	MeterProvider  metric.MeterProvider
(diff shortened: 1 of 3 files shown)


pellared:
Do we want to the same for:

- https://github.com/open-telemetry/opentelemetry-go-contrib/tree/main/instrumentation/google.golang.org/grpc/otelgrpc
- https://github.com/open-telemetry/opentelemetry-go-contrib/tree/main/instrumentation/github.com/gocql/gocql/otelgocql
- https://github.com/open-telemetry/opentelemetry-go-contrib/tree/main/instrumentation/github.com/astaxie/beego/otelbeego

and potentially any future instrumentation libraries that would support both traces and metrics instrumentation?

RangelReale:
> Do we want to the same for:
> 
> * https://github.com/open-telemetry/opentelemetry-go-contrib/tree/main/instrumentation/google.golang.org/grpc/otelgrpc
> * https://github.com/open-telemetry/opentelemetry-go-contrib/tree/main/instrumentation/github.com/gocql/gocql/otelgocql
> * https://github.com/open-telemetry/opentelemetry-go-contrib/tree/main/instrumentation/github.com/astaxie/beego/otelbeego
> 
> and potentially any future instrumentation libraries that would support both traces and metrics instrumentation?

Or maybe having 2 separate handlers, one for traces and one for metrics, and a "combined" one which has both, and keeps compatibility with the current one?

dmathieu:
Related proposal: #2782.

pellared:
Closing per https://github.com/open-telemetry/opentelemetry-go/issues/4027#issuecomment-1527170258

PR #3768, opened 2023-05-02
otelgrpc: Add SpanStartOptions

### LINKED ISSUE #3683: otelgrpc: add WithSpanOptions
`otelgrpc` interceptors make spans, but those spans are not configurable. `otelhttp` has `WithSpanOptions` function which can be used to configure spans in otelhttp middleware both for transport and handler. `otelgrpc` interceptors ought to have the same functionality.

dubonzi:
Closes #3683.

### DIFF
--- a/instrumentation/google.golang.org/grpc/otelgrpc/config.go
+++ b/instrumentation/google.golang.org/grpc/otelgrpc/config.go
@@ -37,10 +37,11 @@ type Filter func(*InterceptorInfo) bool
 
 // config is a group of options for this instrumentation.
 type config struct {
-	Filter         Filter
-	Propagators    propagation.TextMapPropagator
-	TracerProvider trace.TracerProvider
-	MeterProvider  metric.MeterProvider
+	Filter           Filter
+	Propagators      propagation.TextMapPropagator
+	TracerProvider   trace.TracerProvider
+	MeterProvider    metric.MeterProvider
+	SpanStartOptions []trace.SpanStartOption
 
 	ReceivedEvent bool
 	SentEvent     bool
@@ -168,4 +169,18 @@ func (m messageEventsProviderOption) apply(c *config) {
 //   - SentEvents: Record the number of bytes written after every gRPC write operation.
 func WithMessageEvents(events ...Event) Option {
 	return messageEventsProviderOption{events: events}
+}
+
+type spanStartOption struct{ opts []trace.SpanStartOption }
+
+func (o spanStartOption) apply(c *config) {
+	if o.opts != nil {
+		c.SpanStartOptions = append(c.SpanStartOptions, o.opts...)
+	}
+}
+
+// WithSpanOptions configures an additional set of
+// trace.SpanOptions, which are applied to each new span.
+func WithSpanOptions(opts ...trace.SpanStartOption) Option {
+	return spanStartOption{opts}
 }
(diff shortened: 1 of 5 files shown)


instrumentation/google.golang.org/grpc/otelgrpc/interceptor.go
```diff
@@ -84,12 +85,11 @@ func UnaryClientInterceptor(opts ...Option) grpc.UnaryClientInterceptor {
 		}
 
 		name, attr := spanInfo(method, cc.Target())
-		var span trace.Span
-		ctx, span = tracer.Start(
+		startOpts := append(defaultStartOpts, trace.WithAttributes(attr...))
```
ash2k, line 88: This is not safe to do as this function can be called concurrently from multiple goroutines and the backing array of the `defaultStartOpts` might be modified concurrently by this `append`. This can only happen if the slice happens to have unused capacity i.e. the array is re-used rather than re-allocated, but there is a chance nonetheless. I hope this makes sense.

A solution is to explicitly allocate a new slice, copy elements from `defaultStartOpts` and append `trace.WithAttributes()`.

dubonzi: You are correct, I didn't notice that when moving the slice creation to outside of the inner function, and that can definitely happen. My idea was to avoid calling `trace.WithSpanKind` and `cfg.SpanStartOptions...` for every span since those options never change. Do you think thats a worthwhile optmization?
If not, I can always revert to creating the whole slice for each span which would avoid this problem.

ash2k: It probably is a minor but worthwhile optimization. The code will be allocating a slice one way or another so no harm in making it a bit more efficient even if a bit more clunky.

ash2k: On the other hand, maybe Go compiler can optimize/inline the allocation/call of the option function and there is no point. I'll leave this to maintainers to decide.

dubonzi: Sorry for leaving this unsolved, I chose to revert the changes to avoid the problem.

instrumentation/google.golang.org/grpc/otelgrpc/config.go
```diff
@@ -135,6 +136,20 @@ func WithMeterProvider(mp metric.MeterProvider) Option {
 	return meterProviderOption{mp: mp}
 }
 
+type spanStartOption struct{ opts []trace.SpanStartOption }
+
+func (o spanStartOption) apply(c *config) {
+	if o.opts != nil {
```
ash2k, line 142: This `if` is not actually needed, right?

Everything else looks good, thanks!

dubonzi: You are correct.

ash2k (approved):
LGTM, but I'm not a maintainer here.

pellared:
@dubonzi Can you please update the branch (resolve conflicts)?

@dashpole @hanyuancheung Can you please review (as codeowners)?

dubonzi:
> @dubonzi Can you please update the branch (resolve conflicts)?
> 
> @dashpole @hanyuancheung Can you please review (as codeowners)?

All done

pellared (approved):

hanyuancheung (approved):

PR #3771, opened 2023-05-04
Update HTTP span names to match semantic conventions

### LINKED ISSUE #726: Spans created by HTTP instrumentation package does not conform to guidelines for span names
According to [the OpenTelemetry specification](https://github.com/open-telemetry/opentelemetry-specification/blob/v1.1.0/specification/trace/semantic_conventions/http.md#name) HTTP spans MUST follow the overall [guidelines for span names](https://github.com/open-telemetry/opentelemetry-specification/blob/v1.1.0/specification/trace/api.md#span).

For example, the current implementation of HTTP instrumentation package creates Spans with the name _GET_ or _POST_, even though the specification says that _HTTP {METHOD_NAME}_ should be used as Span name.

> HTTP client spans SHOULD be using conservative, low cardinality names formed from the available parameters of an HTTP request, such as "HTTP {METHOD_NAME}"

JamieDanielson:
### What problem is this PR solving

- Closes #726 

### Short description of the changes

Current [semantic conventions](https://opentelemetry.io/docs/reference/specification/trace/semantic_conventions/http/#name) for http span names do not include "HTTP" and suggest including the route.

This PR:

- updates http span name to include URL path
- removes "HTTP" from the span name
- adds a function for config testing that matches the `defaultTransportFormatter`

Because the `defaultTransportFormatter` is not publicly accessible outside the package, I couldn't pull it into the `config_test` directly, which is why I created a matching function in the tests. Without this, the tests were hardcoded and were not failing as expected when I made changes. This is still somewhat brittle because it doesn't change when the `defaultTransportFormatter` function changes, but it's more reliable than it was.

### How to verify this has the expected result

HTTP Span names should change from, for example, `HTTP GET` to `GET /hello`

### DIFF
--- a/instrumentation/net/http/otelhttp/transport.go
+++ b/instrumentation/net/http/otelhttp/transport.go
@@ -16,7 +16,6 @@ package otelhttp // import "go.opentelemetry.io/contrib/instrumentation/net/http
 
 import (
 	"context"
-	"fmt"
 	"io"
 	"net/http"
 	"net/http/httptrace"
@@ -80,9 +79,9 @@ func (t *Transport) applyConfig(c *config) {
 func defaultTransportFormatter(_ string, r *http.Request) string {
 	path := r.URL.Path
 	if path == "" {
-		return fmt.Sprintf("HTTP %s", r.Method)
+		return r.Method
 	}
-	return fmt.Sprintf("HTTP %s %s", r.Method, path)
+	return r.Method + " " + path
 }
 
 // RoundTrip creates a Span and propagates its context via the provided request's headers
--- a/instrumentation/net/http/otelhttp/transport.go
+++ b/instrumentation/net/http/otelhttp/transport.go
@@ -16,6 +16,7 @@ package otelhttp // import "go.opentelemetry.io/contrib/instrumentation/net/http
 
 import (
 	"context"
+	"fmt"
 	"io"
 	"net/http"
 	"net/http/httptrace"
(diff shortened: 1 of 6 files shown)


instrumentation/net/http/otelhttp/transport.go
```diff
@@ -77,7 +77,11 @@ func (t *Transport) applyConfig(c *config) {
 }
 
 func defaultTransportFormatter(_ string, r *http.Request) string {
-	return "HTTP " + r.Method
+	path := r.URL.Path
```
Aneurysm9, line 80: I don't think we can use the path (e.g., `/users/1234`) here.  The default HTTP handler instrumentation doesn't have route information (e.g., `/users/{id}`) available to it and so I think we need to use just the method.

> Instrumentation MUST NOT default to using URI path as span name, but MAY provide hooks to allow custom logic to override the default span name.

JamieDanielson: Ah, 🤦 That makes sense. So I can drop the path part and just keep the change that removes HTTP.

JamieDanielson: Looking at this some more... I guess that's what the `WithRouteTag` is available for in the handler. Checking out [this example](https://github.com/open-telemetry/opentelemetry-go-contrib/blob/a08e95c39ab3858bfba02ace6ea0a21a989a5cf3/instrumentation/net/http/otelhttp/handler_example_test.go#L59), this satisfies the "MAY provide hooks to allow custom logic to override the default span name". 

```go
...
	otelhttp.WithRouteTag("/hello/:name", http.HandlerFunc(
		func(w http.ResponseWriter, r *http.Request) {
			ctx := r.Context()
...
```

PR #3805, opened 2023-05-11
Add go.work

pellared:


pellared:
Workaround for the issue: https://github.com/open-telemetry/opentelemetry-go-contrib/pull/3805/commits/08efee79b019a397ab678a3f5dce1ee4ae1ccee6

pellared:
SIG meeting:

We will not commit `go.work`, because of the lack of confidence on Go workspaces quality. However, we will add a Make target that will create a `go.work` to facilitate local development.

PR #3816, opened 2023-05-15
instrgen : initial otlp, zipkin exporter support

pdelewski:


instrgen/rtlib/rtlib.go
```diff
@@ -42,30 +62,109 @@ func NewTracingState() TracingState {
...
+
+		bsp := trace.NewBatchSpanProcessor(traceExporter)
+		tracingState.Tp = trace.NewTracerProvider(
+			trace.WithSampler(trace.AlwaysSample()),
+			trace.WithResource(res),
+			trace.WithSpanProcessor(bsp),
+		)
+	default:
+		// fallback to file exporting
+		tracingState.File, err = os.Create(traceFile)
```
MrAlias, line 144: Having traces end up in a local file that is not in any standardize format doesn't seem ideal to be the default. Why not have the OTLP exporter be the default?

pdelewski: Yes, you are right. I just left local file as it was available from the beginning.

MrAlias:
From SIG meeting: We can leverage https://github.com/open-telemetry/opentelemetry-go-contrib/tree/main/exporters/autoexport here.

pdelewski:
Moving all this work to Go compile instrumentation sig

PR #3817, opened 2023-05-16
Add semconvutil and use it instead of httpconv and netconv

pellared:
## Why

Towards https://github.com/open-telemetry/opentelemetry-go/issues/4081 (point 1)

Closes https://github.com/open-telemetry/opentelemetry-go/issues/3744

It will make updating `semconv` in https://github.com/open-telemetry/opentelemetry-go a lot easier. This repo will only contain the "currently used" semconv code. Updating to newer will also be easier as we will change existing code instead of generating new code.

## What

1. Add `semconvutil` as a internal shared package which is not to be published (Go templates).
2. Copy `semconvutil` package (via `go generate` + `gotmpl`) and use it instead of `httpconv` and `netconv` packages which are planned to be deprecated.

The "copy package" via go generate approach is used in order to not introduce dependencies between modules. More: https://github.com/open-telemetry/opentelemetry-go/issues/3846

PR #3820, opened 2023-05-17
[chore] Run porto after go generate

pellared:
The same is done in the "main" (non-contrib) repository: https://github.com/open-telemetry/opentelemetry-go/blob/main/Makefile#L86

This can help addressing https://github.com/open-telemetry/opentelemetry-go-contrib/pull/3817#discussion_r1195242653


instrgen/driver/testdata/interface/serializer/interface.go
```diff
@@ -13,7 +13,7 @@
 // limitations under the License.
 
 //nolint:all // Linter is executed at the same time as tests which leads to race conditions and failures.
-package serializer
+package serializer	// import "go.opentelemetry.io/contrib/instrgen/testdata/interface/serializer"
```
pellared, line 16: @pdelewski I do not understand why `instrgen_test.go` keeps replacing "space" to a "tab". Is it not a bug?

pellared: > Regarding tests, we can improve that, however it does not seem must have for now.

The problem is that the test execution modifies the `instrgen/driver/testdata/interface/app/impl.go` and `instrgen/driver/testdata/interface/serializer/interface.go` files that you can accidently commit.

Makefile
```diff
@@ -68,18 +68,21 @@ $(TOOLS)/dbotconf: PACKAGE=go.opentelemetry.io/build-tools/dbotconf
...
 .PHONY: generate build
 
 generate: $(OTEL_GO_MOD_DIRS:%=generate/%)
 generate/%: DIR=$*
-generate/%: | $(STRINGER)
+generate/%: | $(STRINGER) $(PORTO)
 	@echo "$(GO) generate $(DIR)/..." \
 		&& cd $(DIR) \
-		&& PATH="$(TOOLS):$${PATH}" $(GO) generate ./...
+		&& PATH="$(TOOLS):$${PATH}" $(GO) generate ./... && $(PORTO) -w .
```
pellared, line 85: This approach adds the vanity imports even to `testdata`. See https://github.com/open-telemetry/opentelemetry-go-contrib/pull/3820#discussion_r1196279532

Maybe we could simply do something like following instead?


```make
.PHONY: generate go-generate build

generate: go-generate vanity-import-fix

go-generate: $(OTEL_GO_MOD_DIRS:%=generate/%)
go-generate/%: DIR=$*
go-generate/%: | $(STRINGER) $(PORTO)
	@echo "$(GO) generate $(DIR)/..." \
		&& cd $(DIR) \
		&& PATH="$(TOOLS):$${PATH}" $(GO) generate ./...
```make

pdelewski (approved):

Makefile
MrAlias: I'll wait for this ^ change before reviewing.

pellared: @MrAlias Done: https://github.com/open-telemetry/opentelemetry-go-contrib/pull/3820/commits/80bd9cdf06ff98567dca692b06fafc29e37704db

instrgen/driver/testdata/interface/serializer/interface.go
pellared: I reverted he changes as they are not longer needed but I suggest at looking at this problem anyway 😉 https://github.com/open-telemetry/opentelemetry-go-contrib/pull/3820/commits/42181219d44cdb80e8fd1fcb2a2cbe99f5a2e829

MrAlias (approved):

PR #3873, opened 2023-05-22
fix otelhttp client span end

mjlzz:
Issue: otelhttp client does not end span when err is nil, users will not get the right spans
This PR fix the issue

### DIFF
--- a/instrumentation/net/http/otelhttp/transport.go
+++ b/instrumentation/net/http/otelhttp/transport.go
@@ -104,6 +104,7 @@ func (t *Transport) RoundTrip(r *http.Request) (*http.Response, error) {
 	opts := append([]trace.SpanStartOption{}, t.spanStartOptions...) // start with the configured options
 
 	ctx, span := tracer.Start(r.Context(), t.spanNameFormatter("", r), opts...)
+	defer span.End()
 
 	if t.clientTrace != nil {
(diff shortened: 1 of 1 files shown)


pellared (changes requested):
What are the repro steps?

The span ends when the caller closes the request body. https://github.com/open-telemetry/opentelemetry-go-contrib/blob/8850f949daf5fe8a1906fe395d23937c2b417c67/instrumentation/net/http/otelhttp/transport.go#L188

mjlmton:
please close the PR.
it seems like I did not handle the response in the right way,  wrong code like below:
`_, err := client.Do(req)`
`if err != nil {`
`        panic(err)`
`}`
this leave a question: if user don't need to handle response, they still should close response manually, right?

dmathieu:
Yes, you must always close the response after dealing with it. But that's not something required by otelhttp, it's something required by Go.

https://pkg.go.dev/net/http

> The client must close the response body when finished with it

schmurfy:
I just stumbled on this and while closing the body is indeed required by go to avoid leaks the behaviour should be documented somewhere.

dmathieu:
Here for example?
https://pkg.go.dev/go.opentelemetry.io/contrib/instrumentation/net/http/otelhttp#Transport.RoundTrip

PR #3876, opened 2023-05-23
Avoid NPE when ClientConn is nil 

pravindra:
- this can happen when using grpchan's inprocess channel for example

### DIFF
--- a/instrumentation/google.golang.org/grpc/otelgrpc/interceptor.go
+++ b/instrumentation/google.golang.org/grpc/otelgrpc/interceptor.go
@@ -233,6 +233,15 @@ func (w *clientStream) sendStreamEvent(eventType streamEventType, err error) {
 	}
 }
 
+func target(cc *grpc.ClientConn) string {
+	// Since cc is ClientConn instead of ClientConnInterface, it can be nil in some
+	// situations.
+	if cc == nil {
+		return "unknown"
+	}
+	return cc.Target()
+}
+
 // StreamClientInterceptor returns a grpc.StreamClientInterceptor suitable
 // for use in a grpc.Dial call.
 func StreamClientInterceptor(opts ...Option) grpc.StreamClientInterceptor {
@@ -258,7 +267,7 @@ func StreamClientInterceptor(opts ...Option) grpc.StreamClientInterceptor {
 			return streamer(ctx, desc, cc, method, callOpts...)
 		}
 
-		name, attr := spanInfo(method, cc.Target())
+		name, attr := spanInfo(method, target(cc))
 		var span trace.Span
 		ctx, span = tracer.Start(
 			ctx,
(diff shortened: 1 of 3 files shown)


instrumentation/google.golang.org/grpc/otelgrpc/interceptor.go
```diff
@@ -233,6 +233,15 @@ func (w *clientStream) sendStreamEvent(eventType streamEventType, err error) {
 	}
 }
 
+func target(cc *grpc.ClientConn) string {
+	// Since cc is ClientConn instead of ClientConnInterface, it can be nil in some
+	// situations.
+	if cc == nil {
+		return "unknown"
```
pellared, line 240: Why `unknown`? This will have an impact on the produced telemetry. I guess that it will set improper peer attributes. See: https://github.com/open-telemetry/opentelemetry-go-contrib/blob/ee8b97343ce4e86ee0025de77e897aa94a22f974/instrumentation/google.golang.org/grpc/otelgrpc/interceptor.go#L466

instrumentation/google.golang.org/grpc/otelgrpc/test/interceptor_test.go
```diff
@@ -526,6 +526,54 @@ func TestStreamClientInterceptorCancelContext(t *testing.T) {
...
+	reply := &grpc_testing.SimpleResponse{}
+
+	// send and receive fake data
+	for i := 0; i < 10; i++ {
+		_ = streamClient.SendMsg(req)
+		_ = streamClient.RecvMsg(reply)
+	}
+
+	// close client stream
+	_ = streamClient.CloseSend()
```
pellared, line 574: Where are the assertions for the spans that were actually created?

pellared (commented):
The PR gets rid of the panic, but I do not think it handles this case in a proper way.

instrumentation/google.golang.org/grpc/otelgrpc/interceptor.go
```diff
@@ -233,6 +233,15 @@ func (w *clientStream) sendStreamEvent(eventType streamEventType, err error) {
 	}
 }
 
+func target(cc *grpc.ClientConn) string {
```
dmathieu, line 236: This doesn't require a method (which won't be used for anything else). We could be setting a variable.

pellared:
@pravindra Are you planning to work on this PR?

pellared:
@jacques-n  @pravindra Are you planning to work on this PR? If not then can you please close the PR? You can consider creating an issue instead for tracking.

pravindra:
Thanks - I'll likely be able to get on this only after a couple of days. I'll close this PR & raise an issue instead.

pravindra:
https://github.com/open-telemetry/opentelemetry-go-contrib/issues/3967

PR #3892, opened 2023-05-29
Fix panic in otelhttptrace

Alkorin:
Wait100Continue trace event is sent here before "http.receive" phase which results with a panic

Wait100Continue is called here : https://cs.opensource.google/go/go/+/refs/tags/go1.20.4:src/net/http/request.go;l=688 while "http.receive" phase is started in the defer here : https://cs.opensource.google/go/go/+/refs/tags/go1.20.4:src/net/http/request.go;l=559

Thus, the span grab here : https://github.com/open-telemetry/opentelemetry-go-contrib/blob/instrumentation/net/http/httptrace/otelhttptrace/example/v0.42.0/instrumentation/net/http/httptrace/otelhttptrace/clienttrace.go#L373 is nil and https://github.com/open-telemetry/opentelemetry-go-contrib/blob/instrumentation/net/http/httptrace/otelhttptrace/example/v0.42.0/instrumentation/net/http/httptrace/otelhttptrace/clienttrace.go#L375 panics

See reproducer at https://gist.github.com/Alkorin/ba749e56911d887d1acbf1de23604095

### DIFF
--- a/instrumentation/net/http/httptrace/otelhttptrace/test/clienttrace_test.go
+++ b/instrumentation/net/http/httptrace/otelhttptrace/test/clienttrace_test.go
@@ -15,11 +15,13 @@
 package test
 
 import (
+	"bytes"
 	"context"
 	"net/http"
 	"net/http/httptest"
 	"net/http/httptrace"
 	"testing"
+	"time"
 
 	"github.com/stretchr/testify/assert"
 	"github.com/stretchr/testify/require"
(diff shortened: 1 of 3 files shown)


pellared (commented):
Can you please add test(s)?

Alkorin:
Done

hanyuancheung (approved):

dmathieu (approved):

pellared (changes requested):
The test is flaky.

When I run `go test -count=100` it always fails. Personally, I would look into possible synchronization issues.

From https://pkg.go.dev/net/http/httptrace#ClientTrace:
> Functions may be called concurrently from different goroutines and **some may be called after the request has completed** or failed.

(Not directly related to this PR). If I understand this correctly it may even mean that the current implementation may contain logical race conditions as I do not see how `activeHooks` synchronizes multiple concurrent HTTP requests.

instrumentation/net/http/httptrace/otelhttptrace/test/clienttrace_test.go
```diff
@@ -504,3 +505,35 @@ func TestHTTPRequestWithTraceContext(t *testing.T) {
...
+	require.NoError(t, err)
+	require.NoError(t, resp.Body.Close())
+	span.End()
+
+	require.Len(t, fixture.SpanRecorder.Ended(), 5)
+
+	// http.send should contains "GOT 100 - Wait" event
+	send, ok := getSpanFromRecorder(fixture.SpanRecorder, "http.send")
+	require.True(t, ok)
+
```
pellared, line 530:
```suggestion
	// From https://pkg.go.dev/net/http/httptrace#ClientTrace:
	// Functions may be called concurrently from different goroutines and some may be called after the request has completed or failed.
	// This is why require.EventuallyWithT is used here.
	var send trace.ReadOnlySpan
	require.EventuallyWithT(t, func(c *assert.CollectT) {
		var ok bool
		send, ok = getSpanFromRecorder(fixture.SpanRecorder, "http.send")
		require.True(t, ok)
	}, 5*time.Second, 10*time.Millisecond)

	// http.send should contains "GOT 100 - Wait" event
```

Alkorin:
Which golang version/os do you use ? I've checked on my side and nothing fails with 10k calls.

```
$ go1.19 version && go1.19 test -race -count=10000
go version go1.19 linux/amd64
PASS
ok  	go.opentelemetry.io/contrib/instrumentation/net/http/httptrace/otelhttptrace	10.684s

$ go1.20 version && go1.20 test -race -count=10000
go version go1.20 linux/amd64
PASS
ok  	go.opentelemetry.io/contrib/instrumentation/net/http/httptrace/otelhttptrace	10.361s
```

I'm curious to see if it's OS dependent

instrumentation/net/http/httptrace/otelhttptrace/test/clienttrace_test.go
```diff
@@ -504,3 +506,40 @@ func TestHTTPRequestWithTraceContext(t *testing.T) {
...
+	// Functions may be called concurrently from different goroutines and some may be called
+	// after the request has completed
+	var httpSendSpan trace.ReadOnlySpan
+	require.Eventually(t, func() bool {
+		var ok bool
+		httpSendSpan, ok = getSpanFromRecorder(fixture.SpanRecorder, "http.send")
+		return ok
+	}, 5*time.Second, 10*time.Millisecond)
+
+	// Fount http.send span must contains "GOT 100 - Wait" event
```
pellared, line 536:
```suggestion
	// Found http.send span must contain "GOT 100 - Wait" event
```

pellared (approved):
Thank you for your contribution 👍 

I updated the changelog.

pellared:
@dmathieu @hanyuancheung Can you please double-check after the changes?

Issue #3895, opened 2023-06-01
Add Redaction to otelhttp 


fpozzobon:
### Problem Statement

Some attributes passed to attributes might contain sensitive data (example: http.url)

### Proposed Solution

As a consumer of the library, it would be great to have possibility to apply redaction (masking value) to a defined set of attributes.

#### Alternatives

In place of redaction we could:
1. Apply a filter option which would dismiss the attributes
2. Apply an option which would enable consumer to define request/response attributes

#### Prior Art

Redaction option already [exists for header](https://github.com/open-telemetry/opentelemetry-go-contrib/blob/0332bf5c0078aefb4d8284a33c1fbcf19919ae45/instrumentation/net/http/httptrace/otelhttptrace/clienttrace.go#L83) 

### Additional Context

To be compliant with GDPR, it is important for consumer of the library to have the possibility to mask attribute values

dmathieu:
Related, for mongo: https://github.com/open-telemetry/opentelemetry-go-contrib/pull/3662

I wonder if it wouldn't be good to start an otep on this, so every SDK can use similar approaches.

PR #3897, opened 2023-06-01
Add redaction to request / response

fpozzobon:
Adding redaction option to request and response so consumer of the library can hide sensitive values

References: https://github.com/open-telemetry/opentelemetry-go-contrib/issues/3895

### DIFF
--- a/instrumentation/net/http/otelhttp/config.go
+++ b/instrumentation/net/http/otelhttp/config.go
@@ -20,6 +20,7 @@ import (
 	"net/http/httptrace"
 
 	"go.opentelemetry.io/otel"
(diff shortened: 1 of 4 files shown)


MrAlias (changes requested):
Having a one-off solution for this instrumentation does not seem appropriate for sensitive information redaction.

1. It requires the instrumenter to know before they use the instrumentation what attributes will be used.
2. It does not solve the issue for other instrumentation libraries a user may use

I think the solution mentioned [here](https://github.com/open-telemetry/opentelemetry-go/discussions/2794#discussioncomment-2590829) is more appropriate. That or something like [`redact`](https://github.com/MrAlias/redact).

Issue #3902, opened 2023-05-25
Non standard 4xx sets span status to codes.Error


scorpionknifes:
### Description

A http request for a non standard http status code 4xx error (e.g. 499), sets the span status to `codes.Errors`.
`validateHTTPStatusCode` causes this issue in http middlewares (e.g. [otelecho](https://github.com/open-telemetry/opentelemetry-go-contrib/blob/instrumentation/github.com/labstack/echo/otelecho/v0.42.0/instrumentation/github.com/labstack/echo/otelecho/echo.go#L100)).

`validateHTTPStatusCode` returns `codes.Errors` instead of `codes.Unset` although the http status is in the 4xx range.
Since 499 is not part of the [validRangesPerCategory](https://github.com/open-telemetry/opentelemetry-go/blob/main/semconv/internal/v2/http.go#L356)
It returns `codes.Errors` and `Invalid HTTP status code`

validateHTTPStatusCode - https://github.com/open-telemetry/opentelemetry-go/blob/main/semconv/internal/v2/http.go#L386-L399

### Steps To Reproduce

1. Use any middleware relying on  `validateHTTPStatusCode` e.g. [otelecho](https://github.com/open-telemetry/opentelemetry-go-contrib/blob/instrumentation/github.com/labstack/echo/otelecho/v0.42.0/instrumentation/github.com/labstack/echo/otelecho/echo.go#L100)
2. Call a endpoint that results in a non standard http status code e.g 499
3. span status = `codes.Error`

### Expected behavior

Error codes such as 499 is in range of 4xx and error should be `codes.Unset` instead of `codes.Error`

Alternatively configure it to allow user defined 4xx error codes such as 499?

PR #3966, opened 2023-06-07
Fix handling of non-standard HTTP status codes

pellared:
Fixes https://github.com/open-telemetry/opentelemetry-go-contrib/issues/3902

Also from https://greenbytes.de/tech/webdav/draft-ietf-httpbis-p2-semantics-25.html#rfc.section.6.p.2:

> HTTP status codes are extensible. HTTP clients are not required to understand the meaning of all registered status codes, though such understanding is obviously desirable. However, a client must understand the class of any status code, as indicated by the first digit, and treat an unrecognized status code as being equivalent to the x00 status code of that class, with the exception that a recipient must not cache a response with an unrecognized status code.


dmathieu (approved):

pellared:
FYI @scorpionknifes

Aneurysm9 (approved):
> Also from https://greenbytes.de/tech/webdav/draft-ietf-httpbis-p2-semantics-25.html#rfc.section.6.p.2:

The current [STD97](https://www.rfc-editor.org/info/std97) is embodied in [RFC9110](https://www.rfc-editor.org/rfc/rfc9110.html#name-status-codes).  Don't think it changes the analysis here, but always good to reference current standards over old drafts.

codyaray:
Nice, just what we were looking for! Can you cut a new minor release with this? Do you have an ETA?

PR #3968, opened 2023-06-09
[chore] updating google.golang.org/genproto

codeboten:
This is to avoid issues caused by https://github.com/googleapis/go-genproto/issues/1015. I ran into the ambiguous import in this PR: https://github.com/open-telemetry/opentelemetry-collector/pull/7863


MadVikingGod (approved):

propagators/opencensus/examples/go.mod
```diff
@@ -24,11 +24,17 @@ require (
...
 replace (
 	go.opentelemetry.io/contrib/instrumentation/google.golang.org/grpc/otelgrpc => ../../../instrumentation/google.golang.org/grpc/otelgrpc
 	go.opentelemetry.io/contrib/propagators/opencensus => ../
 )
+
+// cloud.google.com/go/compute/metadata: ambiguous import: found package cloud.google.com/go/compute/metadata in multiple modules:
+//   cloud.google.com/go v0.26.0
+//   cloud.google.com/go/compute/metadata v0.2.3
+
+replace cloud.google.com/go => cloud.google.com/go v0.110.2
```
MrAlias, line 40: `instrumentation/google.golang.org/grpc/otelgrpc/go.mod` uses `cloud.google.com/go/compute/metadata v0.2.3`:

https://github.com/open-telemetry/opentelemetry-go-contrib/blob/b038920f0c25dd64c682e11774f80e9badb308ab/instrumentation/google.golang.org/grpc/otelgrpc/go.mod#L16

Why use `cloud.google.com/go v0.110.2` here?

Also, will pining the version of the replace make dependabot not update this dependency?

codeboten: This is using the replace to ensure the old module (`cloud.google.com/go v0.26.0`) is replaced, this module comes in as a transitive dependency of `go.opencensus.io v0.24.0`

I used that version because it is the current latest version of the module, which doesn't include a `metadata` pacakge. 

> Also, will pining the version of the replace make dependabot not update this dependency?

I think you're correct about this pinning will cause dependabot not to update the dependency. I don't have a good alternative here to avoid the ambiguous import and figured it would be ok since this is an example. I'm open to suggestions though!

MadVikingGod: As an alternative we could not use a replace, and upgrade metadata to the latest main branch.

```
cloud.google.com/go/compute/metadata v0.2.4-0.20230602132616-4d8f0906c19e // indirect
```

This then imports the latest `cloud.google.com/go/compute`, which should resolve this issue.

To do this I removed that replace and ran `go get cloud.google.com/go/compute/metadata@4d8f090`

codeboten: I've removed the replace and updated metadata as suggested. Thanks @MadVikingGod!

MrAlias (approved):

pellared (approved):

exporters/autoexport/go.mod
```diff
@@ -27,7 +27,9 @@ require (
 	golang.org/x/net v0.10.0 // indirect
 	golang.org/x/sys v0.8.0 // indirect
 	golang.org/x/text v0.9.0 // indirect
-	google.golang.org/genproto v0.0.0-20230306155012-7f2fa6fef1f4 // indirect
+	google.golang.org/genproto v0.0.0-20230530153820-e85fd2cbaebc // indirect
+	google.golang.org/genproto/googleapis/api v0.0.0-20230526203410-71b5a4ffd15e // indirect
```
pellared, line 31: Why not `20230530153820-e85fd2cbaebc` like in the line above?

pellared:
Can you please resolve the conflicts after https://github.com/open-telemetry/opentelemetry-go-contrib/pull/3976 is merged?

exporters/autoexport/go.mod
codeboten: 🤷🏻 this is what `go get` had generated

codeboten: Updated it to match

Issue #3765, opened 2023-05-01
http.server.duration metric includes high cardinality attr: http.client_ip


vmihailenco:
Meaning that there are as many timeseries as you have visitors. Other http.server.* metrics have the same problem.

Those metrics also include such attrs as http.user_agent, net.sock.peer.addr, net.sock.peer.port, which should also be removed.

rangzen:
For reference, https://grafana.com/blog/2022/02/15/what-are-cardinality-spikes-and-why-do-they-matter/

rangzen:
https://github.com/open-telemetry/opentelemetry-go-contrib/blob/main/instrumentation/net/http/otelhttp/handler.go#L102

rangzen:
Wrong attributes: https://github.com/open-telemetry/opentelemetry-go-contrib/blob/main/instrumentation/net/http/otelhttp/handler.go#L217

dmathieu:
We should split the `httpconv` methods to handle traces and metrics differently. Metrics need low cardinality, while the entire point of traces is to have high cardinality.
If we just remove the attributes, that will cause trace issues such as #3743.

rangzen:
The correct attributes already exist, cf. #3182, and the last comment by @mlaflamm.

Before this commit, attributes were:
```
attributes := append(labeler.Get(), semconv.HTTPServerMetricAttributesFromHTTPRequest(h.operation, r)...)		￼
```

youngbupark:
+1 This results in the high cardinality of request metrics and memory leak in my project. Currently, I am adding the workaround to set req.RemoteAddr to empty before calling otelhttp middleware. But, I hope it will be fixed in the next release.

zsgilber:
Hi all, can we get an acknowledgement that this a real bug, and specifically that this current implementation of the metrics does not match the specification, which restricts those high cardinality attributes to just HTTP clients:

![image](https://github.com/open-telemetry/opentelemetry-go-contrib/assets/16533438/9ab10d49-65ef-4549-99c7-104b78b0c882)

hasheddan:
[...]

PR #3980, opened 2023-06-13
Fix metrics with high cardinality in otelhttp (#3765)

davendu:
Change template to add a new method `semconvutil.HTTPServerRequestMetric` and use it for metrics. This should reduce attributes with high cardinality being introduced into the metrics.


davendu:
> The solution being discussed in [open-telemetry/opentelemetry-specification#3061](https://github.com/open-telemetry/opentelemetry-specification/issues/3061) seems much better.

Didn't knew the View & Advice API before. You are correct, this solution is more elegant.

PR #4014, opened 2023-06-20
AWS-SDK-V2 instrumentation SNS Publish operation span attributes

starbeast:
Followed the way SNS publishing is handled in the Ruby SDK: https://github.com/open-telemetry/opentelemetry-ruby-contrib/blob/1990634acece4b32f56663a720e18b27a14a2b25/instrumentation/aws_sdk/lib/opentelemetry/instrumentation/aws_sdk/messaging_helper.rb#L40. The attribute setter can be later extended to support other operations if necessary

### DIFF
--- a/instrumentation/github.com/aws/aws-sdk-go-v2/otelaws/config.go
+++ b/instrumentation/github.com/aws/aws-sdk-go-v2/otelaws/config.go
@@ -19,30 +19,31 @@ import (
 	"go.opentelemetry.io/otel/trace"
 )
 
-type config struct {
-	TracerProvider         trace.TracerProvider
-	TextMapPropagator      propagation.TextMapPropagator
-	AttributeSetter        []AttributeSetter
-	AttributeSettersConfig AttributeSettersConfig
+// Config holds configuration details for the SDK.
+type Config struct {
+	TracerProvider       trace.TracerProvider
+	TextMapPropagator    propagation.TextMapPropagator
+	RecordSNSPhoneNumber bool
+	AttributeSetter      []AttributeSetter
 }
 
 // Option applies an option value.
 type Option interface {
-	apply(*config)
+	apply(*Config)
 }
 
 // optionFunc provides a convenience wrapper for simple Options
 // that can be represented as functions.
-type optionFunc func(*config)
+type optionFunc func(*Config)
 
-func (o optionFunc) apply(c *config) {
+func (o optionFunc) apply(c *Config) {
 	o(c)
 }
 
 // WithTracerProvider specifies a tracer provider to use for creating a tracer.
 // If none is specified, the global TracerProvider is used.
 func WithTracerProvider(provider trace.TracerProvider) Option {
-	return optionFunc(func(cfg *config) {
+	return optionFunc(func(cfg *Config) {
 		if provider != nil {
 			cfg.TracerProvider = provider
 		}
@@ -52,7 +53,7 @@ func WithTracerProvider(provider trace.TracerProvider) Option {
 // WithTextMapPropagator specifies a Text Map Propagator to use when propagating context.
 // If none is specified, the global TextMapPropagator is used.
 func WithTextMapPropagator(propagator propagation.TextMapPropagator) Option {
-	return optionFunc(func(cfg *config) {
+	return optionFunc(func(cfg *Config) {
 		if propagator != nil {
 			cfg.TextMapPropagator = propagator
 		}
@@ -62,14 +63,14 @@ func WithTextMapPropagator(propagator propagation.TextMapPropagator) Option {
 // WithAttributeSetter specifies an attribute setter function for setting service specific attributes.
 // If none is specified, the service will be determined by the DefaultAttributeSetter function and the corresponding attributes will be included.
 func WithAttributeSetter(attributesetters ...AttributeSetter) Option {
-	return optionFunc(func(cfg *config) {
+	return optionFunc(func(cfg *Config) {
 		cfg.AttributeSetter = append(cfg.AttributeSetter, attributesetters...)
 	})
 }
 
 // WithSNSPhoneNumber specifies if phone number for SNS Inputs should be recorded.
 func WithSNSPhoneNumber(recordPhoneNumber bool) Option {
-	return optionFunc(func(cfg *config) {
-		cfg.AttributeSettersConfig.RecordSNSPhoneNumber = recordPhoneNumber
+	return optionFunc(func(cfg *Config) {
+		cfg.RecordSNSPhoneNumber = recordPhoneNumber
 	})
 }
--- a/instrumentation/github.com/aws/aws-sdk-go-v2/otelaws/config.go
+++ b/instrumentation/github.com/aws/aws-sdk-go-v2/otelaws/config.go
@@ -15,12 +15,14 @@
 package otelaws // import "go.opentelemetry.io/contrib/instrumentation/github.com/aws/aws-sdk-go-v2/otelaws"
 
 import (
+	"context"
+
 	"go.opentelemetry.io/otel/propagation"
 	"go.opentelemetry.io/otel/trace"
 )
 
-// Config holds configuration details for the SDK.
-type Config struct {
+// config holds configuration details for the SDK.
+type config struct {
 	TracerProvider       trace.TracerProvider
 	TextMapPropagator    propagation.TextMapPropagator
 	RecordSNSPhoneNumber bool
@@ -29,21 +31,37 @@ type Config struct {
 
 // Option applies an option value.
 type Option interface {
-	apply(*Config)
+	apply(*config)
 }
 
 // optionFunc provides a convenience wrapper for simple Options
 // that can be represented as functions.
-type optionFunc func(*Config)
+type optionFunc func(*config)
 
-func (o optionFunc) apply(c *Config) {
+type contextKey string
+
+const configContextKey contextKey = "config"
+
+func injectConfig(ctx context.Context, cfg *config) context.Context {
+	return context.WithValue(ctx, configContextKey, cfg)
+}
+
+func configFromContext(ctx context.Context) (*config, bool) {
+	cfg, ok := ctx.Value(configContextKey).(*config)
+	if !ok {
+		cfg = &config{}
+	}
+	return cfg, ok
+}
+
+func (o optionFunc) apply(c *config) {
 	o(c)
 }
 
 // WithTracerProvider specifies a tracer provider to use for creating a tracer.
 // If none is specified, the global TracerProvider is used.
 func WithTracerProvider(provider trace.TracerProvider) Option {
-	return optionFunc(func(cfg *Config) {
+	return optionFunc(func(cfg *config) {
 		if provider != nil {
 			cfg.TracerProvider = provider
 		}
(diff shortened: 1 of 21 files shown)


instrumentation/github.com/aws/aws-sdk-go-v2/otelaws/snsattributes_test.go
```diff
@@ -0,0 +1,74 @@
...
+
+	"github.com/aws/aws-sdk-go-v2/aws"
+	"github.com/aws/aws-sdk-go-v2/service/sns"
+	"github.com/aws/smithy-go/middleware"
+	"github.com/stretchr/testify/assert"
+
+	semconv "go.opentelemetry.io/otel/semconv/v1.17.0"
+)
+
+func TestSNSPublishInputTargetARN(t *testing.T) {
```
dmathieu, line 29: All those tests are extremely similar. How about DRYing them up with subtests?

starbeast: valid point, thank you, I simply followed the tests written for SQS so it was not dry

instrumentation/github.com/aws/aws-sdk-go-v2/otelaws/snsattributes_test.go
```diff
@@ -0,0 +1,75 @@
...
+
+	"github.com/aws/aws-sdk-go-v2/aws"
+	"github.com/aws/aws-sdk-go-v2/service/sns"
+	"github.com/aws/smithy-go/middleware"
+	"github.com/stretchr/testify/assert"
+
+	semconv "go.opentelemetry.io/otel/semconv/v1.17.0"
+)
+
+func TestSNSPublish(t *testing.T) {
```
dmathieu, line 29:
```suggestion
func TestSNSAttributeSetter(t *testing.T) {
```

instrumentation/github.com/aws/aws-sdk-go-v2/otelaws/snsattributes.go
```diff
@@ -0,0 +1,61 @@
...
+	switch v := in.Parameters.(type) {
+	case *sns.PublishInput:
+		var value string
+		valuePointer := v.TargetArn
+		if valuePointer == nil {
+			valuePointer = v.TopicArn
+		}
+		if valuePointer == nil {
+			if v.PhoneNumber != nil {
+				value = *v.PhoneNumber
```
dmathieu, line 41: A phone number is PII data. Could we have an option to disable setting this value?

instrumentation/github.com/aws/aws-sdk-go-v2/otelaws/snsattributes.go
```diff
@@ -0,0 +1,61 @@
...
+	snsAttributes := []attribute.KeyValue{semconv.MessagingSystem("AmazonSNS")}
+
+	switch v := in.Parameters.(type) {
+	case *sns.PublishInput:
+		var value string
+		valuePointer := v.TargetArn
+		if valuePointer == nil {
+			valuePointer = v.TopicArn
+		}
+		if valuePointer == nil {
```
dmathieu, line 39: Rather than a double if and an else, it seems this may be more readable as a switch/case and methods.
Something along the lines of:

```golang
var value string
switch true {
	case v.TargetArn != nil:
		value = parseArn(v.TargetArn)
	case v.TopicArn != nil:
		value = parseArn(v.TopicArn)
	case v.PhoneNumber != nil:
		value = *v.PhoneNumber
}
```

instrumentation/github.com/aws/aws-sdk-go-v2/otelaws/snsattributes_test.go
starbeast: I think although right now `Publish` is the only operation supported by the setter, potentially in the future we may have a support for other operations, which may require a different setup for the tests and thus would rather be tested by a dedicated `Test<Operation>` function. I don't mind the renaming for how things are right now though

dmathieu: Yeah, we should match what we currently have. If things change, the test should change too.

instrumentation/github.com/aws/aws-sdk-go-v2/otelaws/snsattributes.go
dmathieu: I guess a close equivalent could be the mongo command attribute, which can be disabled:
https://github.com/open-telemetry/opentelemetry-go-contrib/blob/main/instrumentation/go.mongodb.org/mongo-driver/mongo/otelmongo/config.go#L71-L78

instrumentation/github.com/aws/aws-sdk-go-v2/otelaws/config.go
```diff
@@ -65,3 +66,10 @@ func WithAttributeSetter(attributesetters ...AttributeSetter) Option {
 		cfg.AttributeSetter = append(cfg.AttributeSetter, attributesetters...)
 	})
 }
+
+// WithAttributeSettersConfig specifies a config used by the attribute setter functions.
+func WithAttributeSettersConfig(settersConfig *AttributeSettersConfig) Option {
```
dmathieu, line 71: I think this option is way too generic. The whole purpose of functional attributes is to provide explicit options. So providing one with a config structs defeats the purpose (and can be quite confusing).

We should have an option such as `WithSNSPhoneNumber()` which enables the phone number.

starbeast: I left the config as is, just changed the option and gave the field a more specialized name.

dmathieu: What's the reason for keeping a dedicated struct which holds a single attribute? We could be passing the original config struct.

starbeast: I thought about separating concerns and only passing what's actually related to the setters, but now that I look at the other packages I don't even know why in this one the config record is not present as a field in the otelMiddlewares struct. I'm thinking about simply replacing
```
type otelMiddlewares struct {
	tracer                 trace.Tracer
	propagator             propagation.TextMapPropagator
	attributeSetter        []AttributeSetter
	attributeSettersConfig AttributeSettersConfig
}
```
with
```
type otelMiddlewares struct {
	tracer trace.Tracer
	config Config
}
```
and also passing this `Config` record to the setters

instrumentation/github.com/aws/aws-sdk-go-v2/otelaws/config.go
```diff
@@ -19,29 +19,31 @@ import (
 	"go.opentelemetry.io/otel/trace"
 )
 
-type config struct {
-	TracerProvider    trace.TracerProvider
-	TextMapPropagator propagation.TextMapPropagator
-	AttributeSetter   []AttributeSetter
+// Config holds configuration details for the SDK.
+type Config struct {
```
dmathieu, line 23: Why is config becoming public?

instrumentation/github.com/aws/aws-sdk-go-v2/otelaws/sqsattributes_test.go
```diff
@@ -26,170 +26,86 @@ import (
 	semconv "go.opentelemetry.io/otel/semconv/v1.17.0"
 )
 
-func TestSQSDeleteMessageBatchInput(t *testing.T) {
-	input := middleware.InitializeInput{
-		Parameters: &sqs.DeleteMessageBatchInput{
-			QueueUrl: aws.String("test-queue-url"),
+func TestSQSAttributeSetter(t *testing.T) {
```
dmathieu, line 29: While this test refactoring makes sense, it's unrelated to the PR's intent. Splitting it into its own PR would make reviews much faster.

instrumentation/github.com/aws/aws-sdk-go-v2/otelaws/config.go
starbeast: `AttributeSetter` is a public type with a shape of `func(context.Context, middleware.InitializeInput, *Config)`.

instrumentation/github.com/aws/aws-sdk-go-v2/otelaws/sqsattributes_test.go
starbeast: ok, makes sense, I refactored it since the shape of the setter changed and this code was updated anyway. I'll move it to another PR

instrumentation/github.com/aws/aws-sdk-go-v2/otelaws/config.go
dmathieu: That seems like a smell, as it allows anybody to update the Config pointer with `WithAttributeSetter`...

starbeast: But why would anyone do that in the first place? I mean, we can pass the value instead of course, but why would end users compromise the config?

dmathieu: It's not just about end users being able to compromise the config. It's about providing them with an elegant and coherent interface.

starbeast: Here I agree. I don't really like the idea of changing the interface (and making it not backward compatible) myself, yet I would not like to pollute the context too, but that seems like the only easy way to not expose unnecessary details or sacrifice exportable setters. Do you maybe have a vision for this?

starbeast: @dmathieu I committed the solution with the config passed via the context, which kinda makes the config not a part of  the public API, which on the other hand makes using stuff like `RecordSNSPhoneNumber` completely internal. I guess that's more or less a valid compromise as in case of custom attribute setters the users can control the behavior of the setters themselves and they don't need to access the config to do that.

dmathieu: You may have forgotten to push. But if you were passing the config through context's values, it wouldn't be a good solution either.

I'd be interested to hear other people's ideas here. But one solution could be to have two types of attribute setters. Private ones, which do get the config and are defined by this module.
And public ones, which can be set externally by users, but don't get access to the config.

instrumentation/github.com/aws/aws-sdk-go-v2/otelaws/config.go
```diff
@@ -34,6 +38,22 @@ type Option interface {
 // that can be represented as functions.
 type optionFunc func(*config)
 
+type contextKey string
+
+const configContextKey contextKey = "config"
+
+func injectConfig(ctx context.Context, cfg *config) context.Context {
```
dashpole: Instead of passing `config` through `context.WithValue` on every request, you can define `defaultAttributeSetter` and `snsAttributeSetter` as methods on `config` (which still match the `AttributeSetter` signature) and have the exported `DefaultAttributeSetter` and `SNSAttributeSetter` functions delegate to `config{}`. Then `AppendMiddlewares` can default to `cfg.defaultAttributeSetter` without needing context injection or unexporting the existing setters.

PR #4017, opened 2023-06-22
Add WithRouteTagMiddleware function to otelhttp package

terev:
This new function simplifies usage of this utility by users of third party router packages.


CHANGELOG.md
```diff
@@ -13,6 +13,7 @@ This project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0.htm
 - Add `NewMiddleware` function in `go.opentelemetry.io/contrib/instrumentation/net/http/otelhttp`. (#2964)
 - Add the new `go.opentelemetry.io/contrib/instrgen` package to provide auto-generated source code instrumentation. (#3068, #3108)
 - The `go.opentelemetry.io/contrib/exporters/autoexport` package to provide configuration of trace exporters with useful defaults and envar support. (#2753)
+- Add `WithRouteTagMiddleware` function in `go.opentelemetry.io/contrib/instrumentation/net/http/otelhttp`.
```
```suggestion
- Add `WithRouteTagMiddleware` function in `go.opentelemetry.io/contrib/instrumentation/net/http/otelhttp`. (#4017)
```

instrumentation/net/http/otelhttp/handler.go
```diff
@@ -261,9 +261,17 @@ func setAfterServeAttributes(span trace.Span, read, wrote int64, statusCode int,
...
+// WithRouteTagMiddleware annotates a span with the provided route name using the
+// RouteKey Tag.
+func WithRouteTagMiddleware(route string) func(http.Handler) http.Handler {
+	return func(next http.Handler) http.Handler {
+		return http.HandlerFunc(func(w http.ResponseWriter, r *http.Request) {
+			span := trace.SpanFromContext(r.Context())
+			span.SetAttributes(semconv.HTTPRoute(route))
+			next.ServeHTTP(w, r)
+		})
+	}
```
dmathieu, line 276: Could you add a test?

pellared:
> This new function simplifies usage of this utility within third party router packages.

Could you please provide examples? You can even consider adding "test examples".

instrumentation/net/http/otelhttp/handler.go
```diff
@@ -261,9 +261,17 @@ func setAfterServeAttributes(span trace.Span, read, wrote int64, statusCode int,
...
+	return WithRouteTagMiddleware(route)(h)
+}
+
+// WithRouteTagMiddleware annotates a span with the provided route name using the
+// RouteKey Tag.
+func WithRouteTagMiddleware(route string) func(http.Handler) http.Handler {
+	return func(next http.Handler) http.Handler {
+		return http.HandlerFunc(func(w http.ResponseWriter, r *http.Request) {
+			span := trace.SpanFromContext(r.Context())
+			span.SetAttributes(semconv.HTTPRoute(route))
```
pellared, line 273: Side note:

Would it not be good to also update the span name?

See: https://github.com/open-telemetry/opentelemetry-specification/blob/main/specification/trace/semantic_conventions/http.md#name

instrumentation/net/http/otelhttp/handler.go
terev: Added a test for this.

instrumentation/net/http/otelhttp/handler.go
terev: That seems to make sense based on the documented convention.

instrumentation/net/http/otelhttp/go.sum
```diff
@@ -16,8 +16,12 @@ go.opentelemetry.io/otel v1.16.0 h1:Z7GVAX/UkAXPKsy94IU+i6thsQS4nb7LviLpnaNeW8s=
 go.opentelemetry.io/otel v1.16.0/go.mod h1:vl0h9NUa1D5s1nv3A5vZOYWn8av4K8Ml6JDeHrT/bx4=
 go.opentelemetry.io/otel/metric v1.16.0 h1:RbrpwVG1Hfv85LgnZ7+txXioPDoh6EdbZHo26Q3hqOo=
 go.opentelemetry.io/otel/metric v1.16.0/go.mod h1:QE47cpOmkwipPiefDwo2wDzwJrlfxxNYodqc4xnGCo4=
+go.opentelemetry.io/otel/sdk v1.16.0 h1:Z1Ok1YsijYL0CSJpHt4cS3wDDh7p572grzNrBMiMWgE=
+go.opentelemetry.io/otel/sdk v1.16.0/go.mod h1:tMsIuKXuuIWPBAOrH+eHtvhTL+SntFtXF9QD68aP6p4=
```
pellared, line 20: You have to move the test to `/test` subdirectory as the instrumentation libraries must not depend on the SDK.

terev:
I've added one example so far and can add more in the morning. I added it to instrumentation/net/http/otelhttp/example/server/github.com/labstack/echo . It semanticly makes sense but may be tough to find. What do you think?

instrumentation/net/http/otelhttp/handler.go
terev: @pellared Did you want me to implement this on this pr?

terev:
@pellared I've added a bunch of examples to instrumentation/net/http/otelhttp/example/server. Probably not the best place so lmk if there's somewhere else they should be.

dmathieu:
I think simpler, and [testable examples](https://go.dev/blog/examples) would be better than this. What we need is a code sample, not a full-blown runnable app.

instrumentation/net/http/otelhttp/example/server/github.com/labstack/echo/server.go
```diff
@@ -0,0 +1,84 @@
...
+		id := c.Param("id")
+		name := getUser(c.Request().Context(), id)
+		return c.JSON(http.StatusOK, struct {
+			ID   string
+			Name string
+		}{
+			ID:   id,
+			Name: name,
+		})
+	}, echo.WrapMiddleware(otelhttp.WithRouteTagMiddleware("/users/:id")))
```
pellared, line 59: Would it not be possible to get rid of the `"/users/:id"` duplication?

I do not like requiring to add `echo.WrapMiddleware(otelhttp.WithRouteTagMiddleware()` for each registering handler.

I would prefer to have something like


```go
	r.Use(echo.WrapMiddleware(otelhttp.NewMiddleware("my-server")))
        r.Use(OTelRouteTaggerMiddleware()) // or: r.Use(otelecho.RouteTaggerMiddleware())

	r.GET("/users/:id", func(c echo.Context) error {
		id := c.Param("id")
		name := getUser(c.Request().Context(), id)
		return c.JSON(http.StatusOK, struct {
			ID   string
			Name string
		}{
			ID:   id,
			Name: name,
		})
	})
```

pellared (commented):
I am really low time until Wednesday, but I did my best to put any feedback that can hopefully help.

instrumentation/net/http/otelhttp/example/server/github.com/labstack/echo/server.go
terev: @pellared I can't come up with a way we can create a generic middleware for this since the route tag is usually router specific and not possible to retrieve in our middleware. If users want a catch all middleware I think each library would need to implement this. For example echo would need a middleware that uses the result of `c.Path()` as the route tag. 

We could provide a utility function that adds a passed route tag as an attribute and changes the span name as discussed previously.

pellared: > We could provide a utility function that adds a passed route tag as an attribute and changes the span name as discussed previously.

I think this may be more flexible and powerful. Writing a little middleware that would use this function should not be a big deal.

terev:
> I think simpler, and [testable examples](https://go.dev/blog/examples) would be better than this. What we need is a code sample, not a full-blown runnable app.

Not sure how to do this and actually add examples for the `NewMiddlware` function. If I add this into the root of the otelhttp module it'll add a bunch of extra dependencies. Should these examples be put in the example directory as testable examples? Even though they'll be examples for the example module?

instrumentation/net/http/otelhttp/example/server/github.com/labstack/echo/server.go
terev: Added this utility function to otelhttp and route tagger middleware in 4 of the http router modules. Added a question about where to put the examples.

dmathieu:
> Not sure how to do this and actually add examples for the NewMiddlware function. If I add this into the root of the otelhttp module it'll add a bunch of extra dependencies. Should these examples be put in the example directory as testable examples? Even though they'll be examples for the example module?

They can be put in  a `test` submodule. gRPC does that for example: https://github.com/open-telemetry/opentelemetry-go-contrib/tree/main/instrumentation/google.golang.org/grpc/otelgrpc/test

terev:
> > Not sure how to do this and actually add examples for the NewMiddlware function. If I add this into the root of the otelhttp module it'll add a bunch of extra dependencies. Should these examples be put in the example directory as testable examples? Even though they'll be examples for the example module?
> 
> They can be put in the same directory, but with a different module name. See this example from gRPC: https://github.com/open-telemetry/opentelemetry-go-contrib/blob/main/instrumentation/google.golang.org/grpc/otelgrpc/example_interceptor_test.go

Won't that still require dependencies in the go.mod? I created a file called `instrumentation/net/http/otelhttp/example_echo_middleware_test.go` that is defined within the otelhttp_test package. Then running `go mod tidy -v` in that directory adds the dependencies required by the example to go.mod.

terev:
It's still unclear to me where I should be putting the example tests. Maybe there is not really a good answer for this at the moment?

instrumentation/net/http/otelhttp/example/server/github.com/labstack/echo/server.go
pellared: Can you update the examples? The existing example structure is good enough. We can refine it later once we have a consensus on the implementation part.

instrumentation/github.com/emicklei/go-restful/otelrestful/restful.go
```diff
@@ -84,3 +86,11 @@ func OTelFilter(service string, opts ...Option) restful.FilterFunction {
...
 	}
 }
+
+// RouteTaggerMiddleware annotates the current span with the handler's route path.
+func RouteTaggerMiddleware() restful.FilterFunction {
+	return func(request *restful.Request, response *restful.Response, chain *restful.FilterChain) {
+		otelhttp.AnnotateSpanWithHTTPRoute(request.Request.Context(), request.SelectedRoutePath())
+		chain.ProcessFilter(request, response)
+	}
+}
```
pellared, line 96: We should not depend on `otelhttp` in other instrumentation libraries. These libraries are not stable and bumping in `otelhttp` can break `otelgin` users. The `AnnotateSpanWithHTTPRoute` implementation is so small that I would rather inline it in existing instrumentation packages.

I would prefer if this PR is only about `otelhttp`.

I would move the middleware implementations using `AnnotateSpanWithHTTPRoute` to examples under `otelhttp`. This would demonstrate how someone could use `otelhttp` e.g. instead of `otelmux`.

terev:
Does anyone have opinions on whether this should be closed? It seems like the third party instrumentation is due to be removed at some point so maybe this is obsolete?

PR #4031, opened 2023-06-27
Add issue templates for remaining packages

MrAlias:


pellared (approved):

hanyuancheung (approved):

dmathieu (approved):
Should we auto-generate those template to ensure every package has a one, and there's no removed package with a remaining template?

MrAlias:
> Should we auto-generate those template to ensure every package has a one, and there's no removed package with a remaining template?

There are subtle differences in the labeling, naming, and assignee that I'm not sure we want to build out automation for this. It will likely require more time to build than we will use it.

If you feel like it is something you want to do though, I'm happy to accept such a script in an additional PR.

PR #4033, opened 2023-06-30
fix: Do not modify the origin request in RoundTripper

### LINKED ISSUE #3969: otelhttp.Transport should not mutate the incoming request.
See https://pkg.go.dev/net/http#RoundTripper:

```golang
	// RoundTrip should not modify the request, except for
	// consuming and closing the Request's Body. RoundTrip may
	// read fields of the request in a separate goroutine. Callers
	// should not mutate or reuse the request until the Response's
	// Body has been closed.
```

gastonqiu:
According to http RoundTripper [spec](https://pkg.go.dev/net/http#RoundTripper), we shouldn't modify the incoming request. I believe the correct way is cloning the request.
ref: https://github.com/golang/oauth2/blob/989acb1bfed17be45134185bd228d89675a68f19/transport.go#L50 

See: 

close #3969


instrumentation/net/http/otelhttp/transport.go
```diff
@@ -109,7 +109,7 @@ func (t *Transport) RoundTrip(r *http.Request) (*http.Response, error) {
 		ctx = httptrace.WithClientTrace(ctx, t.clientTrace(ctx))
 	}
 
-	r = r.WithContext(ctx)
+	r = r.Clone(ctx) // According to RoundTripper spec we shouldn't modify the origin request.
```
Aneurysm9, line 112: Can a unit test be added to verify that the original request is not modified so that we can avoid regression in the future?

CHANGELOG.md
```diff
@@ -26,6 +26,7 @@ This project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0.htm
   - `go.opentelemetry.io/contrib/instrumentation/gopkg.in/macaron.v1/otelmacaron`
   - `go.opentelemetry.io/contrib/instrumentation/net/http/httptrace/otelhttptrace`
   - `go.opentelemetry.io/contrib/instrumentation/net/http/otelhttp`
+- Shouldn't modify the origin request in RoundTripper (#4033)
```
dashpole, line 29:
```suggestion
- Do not modify the origin request in RoundTripper in `go.opentelemetry.io/contrib/instrumentation/net/http/otelhttp`. (#4033)
```

dmathieu (approved):

MrAlias (approved):

pellared (approved):

PR #4044, opened 2023-07-03
otelkit: Allow specifying a custom span type

jackwilsdon:
This is useful for when an endpoint may be wrapping a call to another service, or when an endpoint is used as a consumer (e.g. with https://github.com/alebabai/go-kit-kafka). [The go-kit opentracing provider supports this via use of `WithTags`](https://github.com/go-kit/kit/blob/dfe43fa6a8d72c23e2205d0b80e762346e203f78/tracing/opentracing/endpoint.go#L100-L118).

### DIFF
--- a/instrumentation/github.com/go-kit/kit/otelkit/config.go
+++ b/instrumentation/github.com/go-kit/kit/otelkit/config.go
@@ -47,6 +47,9 @@ type config struct {
 	// GetAttributes is an optional function that can extract trace attributes
 	// from the context and add them to the span.
 	GetAttributes func(ctx context.Context) []attribute.KeyValue
+
+	// SpanKind is the kind of span to be created by this middleware.
+	SpanKind trace.SpanKind
 }
 
 // Option configures an EndpointMiddleware.
@@ -104,5 +107,13 @@ func WithAttributes(attrs ...attribute.KeyValue) Option {
 func WithAttributeGetter(fn func(ctx context.Context) []attribute.KeyValue) Option {
 	return optionFunc(func(o *config) {
 		o.GetAttributes = fn
+	})
+}
+
+// WithSpanKind sets the span kind for spans created by this middleware.
+// If this option is not provided, then trace.SpanKindServer is used.
+func WithSpanKind(kind trace.SpanKind) Option {
+	return optionFunc(func(o *config) {
+		o.SpanKind = kind
 	})
 }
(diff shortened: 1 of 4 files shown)


pellared:
The span kind is strictly defined by the HTTP semantic conventions. See: https://github.com/open-telemetry/opentelemetry-specification/blob/v1.21.0/specification/trace/semantic_conventions/http.md#http-server-semantic-conventions

If you want to change the semantic conventions then you should create an issue (and maybe PR) in https://github.com/open-telemetry/semantic-conventions (new repo for semantic conventions).

I think this PR should be closed or at least moved to draft.

jackwilsdon:
go-kit supports transports other than HTTP (e.g. it comes with a AMQP transport) - https://github.com/open-telemetry/opentelemetry-specification/blob/4229197f45f0da0642bc7c0127e84c8aa1439489/specification/trace/semantic_conventions/messaging.md#span-kind mentions using the consumer/producer span kinds.

instrumentation/github.com/go-kit/kit/otelkit/config.go
```diff
@@ -106,3 +109,10 @@ func WithAttributeGetter(fn func(ctx context.Context) []attribute.KeyValue) Opti
 		o.GetAttributes = fn
 	})
 }
+
+// WithSpanKind sets the span kind for spans created by this middleware.
```
pellared, line 113: Please document the default value.

```suggestion

// WithSpanKind sets the span kind for spans created by this middleware.
// If this option is not provided, then trace.SpanKindServer is used.

**Review by pellared — APPROVED** [review]

File: instrumentation/github.com/go-kit/kit/otelkit/config.go (hunk 1 above)
Comment (jackwilsdon) [reply instrumentation/github.com/go-kit/kit/otelkit/config.go:113]: Done :+1:

File: instrumentation/github.com/go-kit/kit/otelkit/config.go [hunk 2]
```diff
@@ -106,3 +109,11 @@ func WithAttributeGetter(fn func(ctx context.Context) []attribute.KeyValue) Opti
...
+// If this option is not provided, then trace.SpanKindServer is used.
+func WithSpanKind(kind trace.SpanKind) Option {
```
pellared, line 115: I think that only two values of `trace.SpanKind` are valid:

- `trace.SpanKindServer`
- `trace.SpanKindConsumer`


Can we change and add `WithSpanKindServer()` and `WithSpanKindConsumer()` options instead?

Credits to @Aneurysm9

pellared (changes requested):
Second review after getting more feedback.

instrumentation/github.com/go-kit/kit/otelkit/config.go
jackwilsdon: I've been using `SpanKindInternal` on my endpoints and instead putting server/consumer on the transport, as it's up to the transport whether the endpoint is called as part of a server or consumer. Changing this to just those two options somewhat limits the flexibility here.

pellared: I am not following. Why would you set to `Internal`? 

>  as it's up to the transport whether the endpoint is called as part of a server or consumer

How would you modify the span kind? If it would be possible then you would not even need this functionality.

However, now I see that go-kit transport/endpoint can also be a "consumer" or "client" so probably it is better to keep as it is. Approving again  😉

pellared (approved):

pellared:
Closing as during SIG meeting we agreed that we are dropping support for `otelkit`.
See https://github.com/open-telemetry/opentelemetry-go-contrib/issues/4085

PR #4045, opened 2023-07-04
jaegerremote: Add WithSamplingStrategyFetcher option

morus12:
* Fixed an issue where the fetcher would panic if the default transport is not of type *http.Transport due to the cast:\
\
`customTransport := http.DefaultTransport.(*http.Transport).Clone()`
\
To address this, we've now set the `http.Client.Timeout` rather than `customTransport.ResponseHeaderTimeout` to prevent panic.

* This PR exposes SamplingStrategyFetcher and WithSamplingStrategyFetcher, enabling the definition of a custom fetcher that utilizes a custom HTTP client with a transport other than *http.Transport. We have integrated [httpmock](https://github.com/jarcoal/httpmock), which clashes with the default http.Client. To address this, we employ custom HTTP clients in our fetcher implementation.


dmathieu:
Please add a unit test.

pellared (approved):

pellared (dismissed):
PS. Only "Client Sampling" is deprecated https://www.jaegertracing.io/docs/1.47/sampling/

hanyuancheung (approved):

yurishkuro (dismissed):

samplers/jaegerremote/sampler_remote_test.go
```diff
@@ -586,3 +587,16 @@ func TestSamplingStrategyParserImpl_Error(t *testing.T) {
...
+
+func TestDefaultSamplingStrategyFetcher_Panic(t *testing.T) {
+	defaultTransport := http.DefaultTransport
+	http.DefaultTransport = http.NewFileTransport(http.Dir("/"))
+
+	t.Cleanup(func() {
+		http.DefaultTransport = defaultTransport
+	})
+
+	require.Panics(t, func() {
```
yurishkuro, line 599: I don't think this test should be added, it tests an internal implementations detail as if that is a requirement for default fetcher to fail on a different transport.

yurishkuro: The opposite test makes more sense, that the sampler does not fail when used with a custom fetcher and the default transport is overwritten.

morus12: > the sampler does not fail when used with a custom fetcher

Whether a sampler fails or not depends on its implementation. 

The test I added exposes an issue with the current implementation, so we can all agree that `WithSamplingStrategyFetcher` should be public to at least be able to avoid the issue.

yurishkuro: > so we can all agree that `WithSamplingStrategyFetcher` should be public to at least be able to avoid the issue.

Yes and the way to demonstrate that is with a test that does NOT fail when used with custom sampler. The fact that the existing sampler panics is completely irrelevant (and is frankly a bug), not something that should be codified as expected behavior with a test.

samplers/jaegerremote/sampler_remote_options.go
```diff
@@ -124,7 +124,7 @@ func WithLogger(logger logr.Logger) Option {
 }
 
 // samplingStrategyFetcher creates a Option that initializes sampling strategy fetcher.
```
pellared, line 126: 1. The comment should start with `WithSamplingStrategyFetcher` instead of `samplingStrategyFetcher`.
2. Can you add information when this option should be used?

samplers/jaegerremote/sampler_remote_test.go
pellared: @morus12 I think it is a bug as well.

1. I suggest to create an issue for it. There should be on panic.
2. You can still have the test but you should add a comment that this test demonstrates a bug which should be fixed and reference the issue created in point 1.

pellared (changes requested):
Few things needs to be addressed.

pellared:
@morus12 Do you plan to update the PR and address the comments?

morus12:
1. I added comments.
2. Set the `defaultRemoteSamplingTimeout` only when `http.DefaultClient.Transport` is `*http.Transport`.
3. Test if `defaultRemoteSamplingTimeout` is set when conditions are met.
4. Test if there is no panic with custom `http.DefaultClient.Transport`.

Let me know if you need any other changes.

CHANGELOG.md
```diff
@@ -19,6 +19,7 @@ This project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0.htm
 - Add `NewMiddleware` function in `go.opentelemetry.io/contrib/instrumentation/net/http/otelhttp`. (#2964)
 - The `go.opentelemetry.io/contrib/exporters/autoexport` package to provide configuration of trace exporters with useful defaults and environment variable support. (#2753, #4100, #4130, #4132, #4134)
 - `WithRouteTag` in `go.opentelemetry.io/contrib/instrumentation/net/http/otelhttp` adds HTTP route attribute to metrics. (#615)
+- Add `WithSamplingStrategyFetcher` which sets custom fetcher implementation. Fixed panic when default HTTP round-tripper is not `*http.Transport` (#4045)
```
pellared, line 22: Can you move the entries to `Unreleased`?

Can you split it into two entries?

In `Added`:

```
- Add `WithSamplingStrategyFetcher` option in `go.opentelemetry.io/contrib/samplers/jaegerremote` which sets a custom fetcher implementation. (#4045)
```

In `Fixed`

```
- Fixed panic in `go.opentelemetry.io/contrib/samplers/jaegerremote` when default HTTP round-tripper `http.DefaultTransport` is not `*http.Transport`. (#4045)
```

samplers/jaegerremote/sampler_remote.go
```diff
@@ -273,15 +273,20 @@ type httpSamplingStrategyFetcher struct {
...
-
-	return &httpSamplingStrategyFetcher{
+	fetcher := &httpSamplingStrategyFetcher{
 		serverURL: serverURL,
-		httpClient: http.Client{
-			Transport: customTransport,
-		},
 	}
+
+	if customTransport, ok := http.DefaultTransport.(*http.Transport); ok {
```
pellared, line 280: If I understand correctly. If `http.DefaultTransport` is NOT `*http.Transport` then the request send by the fetcher would NOT have any timeout. If I am correct, then I do not think it is acceptable.

Maybe we should always set `http.Client.Timeout` instead of setting `customTransport.ResponseHeaderTimeout`? If my thinking is correct then would we even need a custom transport?

CC @yurishkuro

morus12: > Maybe we should always set http.Client.Timeout instead of setting customTransport.ResponseHeaderTimeout?

I like this idea, updated.

CHANGELOG.md
```diff
@@ -11,6 +11,11 @@ This project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0.htm
 ### Added
 
 - Add the new `go.opentelemetry.io/contrib/instrgen` package to provide auto-generated source code instrumentation. (#3068, #3108)
+- Add `WithSamplingStrategyFetcher` which sets custom fetcher implementation. (#4045)
+
+### Fixed
+
+- Fixed panic when default HTTP round-tripper is not `*http.Transport` (#4045)
```
```suggestion
- Do not panic when the default HTTP round-tripper is not `*http.Transport`. (#4045)
```

samplers/jaegerremote/sampler_remote.go
pellared: @yurishkuro please double-check my idea. I do not trust myself 😉

CHANGELOG.md
```diff
@@ -11,6 +11,11 @@ This project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0.htm
 ### Added
 
 - Add the new `go.opentelemetry.io/contrib/instrgen` package to provide auto-generated source code instrumentation. (#3068, #3108)
+- Add `WithSamplingStrategyFetcher` which sets custom fetcher implementation. (#4045)
+
+### Fixed
+
+- Do not panic when the default HTTP round-tripper is not `*http.Transport` (#4045)
```
dmathieu, line 18: Nit, but we finish our changelog entries with a dot.

```suggestion
- Do not panic when the default HTTP round-tripper is not `*http.Transport`. (#4045)
```

dmathieu (approved):

CHANGELOG.md
```diff
@@ -11,6 +11,11 @@ This project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0.htm
 ### Added
 
 - Add the new `go.opentelemetry.io/contrib/instrgen` package to provide auto-generated source code instrumentation. (#3068, #3108)
+- Add `WithSamplingStrategyFetcher` which sets custom fetcher implementation. (#4045)
```
pellared, line 14: I am curious if you still need this functionality. You mentioned only the problem with panic in the PR description. If you do not need this functionality I would rather not add it and just have the fix for the panic.

morus12: We use [httpmock](https://github.com/jarcoal/httpmock) which interferes with a default `http.Client`. To fix it we use custom HTTP Clients, e.g. in fetcher implementation. If that's no problem for you, let's leave it public.

pellared (approved):
Waiting for codeowner (@yurishkuro) review.

samplers/jaegerremote/sampler_remote.go
```diff
@@ -273,13 +273,16 @@ type httpSamplingStrategyFetcher struct {
 }
 
 func newHTTPSamplingStrategyFetcher(serverURL string) *httpSamplingStrategyFetcher {
-	customTransport := http.DefaultTransport.(*http.Transport).Clone()
-	customTransport.ResponseHeaderTimeout = defaultRemoteSamplingTimeout
+	var rt http.RoundTripper
+	if t, ok := http.DefaultTransport.(*http.Transport); ok {
+		rt = t.Clone()
+	}
```
yurishkuro, line 279: Since the timeout is no longer set on the transport, can we not just delete this part? Default transport will be used automatically by the client, would it not? What's the purpose of cloning it?

yurishkuro (approved):
Lgtm, although I don't follow the need to clone the transport.

pellared:
@morus12 Please address https://github.com/open-telemetry/opentelemetry-go-contrib/pull/4045#discussion_r1339234126 and update the branch and I will merge the PR 🎉

samplers/jaegerremote/sampler_remote.go
morus12: This avoids races on `http.DefaultTransport` in tests.

We can either:
* remove this and the test that causes the race https://github.com/open-telemetry/opentelemetry-go-contrib/pull/4045/files#diff-643eabddd4dc6c7bed5f17028eae0306783f2346732f1ab135e3d8becb5aa766R596
* leave it as is

pellared: @morus12 Remove the test. It does not really add any value.

morus12:
I appreciate your help! 
It was an exciting challenge, maybe I'll pick up another issue ;)

PR #4090, opened 2023-07-17
Migration from `github.com/Shopify/sarama` to `github.com/IBM/sarama`

batazor:
Need to change the URL to sarama.
Changelog: https://github.com/IBM/sarama/releases/tag/v1.40.0

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -27,7 +27,7 @@ This project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0.htm
   - `go.opentelemetry.io/contrib/instrumentation/net/http/httptrace/otelhttptrace`
   - `go.opentelemetry.io/contrib/instrumentation/net/http/otelhttp`
 - Do not modify the origin request in RoundTripper in `go.opentelemetry.io/contrib/instrumentation/net/http/otelhttp`. (#4033)
-- Migration from `github.com/Shopify/sarama` to `github.com/IBM/sarama`
+- Updated dependency of `github.com/Shopify/sarama` in `go.opentelemetry.io/contrib/instrumentation/github.com/IBM/sarama/otelsarama` to `github.com/IBM/sarama` so as to match the upstream package rename. (#4090)
 
 ### Removed
 
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -27,6 +27,7 @@ This project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0.htm
   - `go.opentelemetry.io/contrib/instrumentation/net/http/httptrace/otelhttptrace`
   - `go.opentelemetry.io/contrib/instrumentation/net/http/otelhttp`
 - Do not modify the origin request in RoundTripper in `go.opentelemetry.io/contrib/instrumentation/net/http/otelhttp`. (#4033)
+- Migration from `github.com/Shopify/sarama` to `github.com/IBM/sarama`
 
 ### Removed
 
--- a/CODEOWNERS
+++ b/CODEOWNERS
@@ -37,7 +37,7 @@ instrumentation/github.com/go-kit/kit/otelkit/                          @open-te
 instrumentation/github.com/gocql/gocql/otelgocql/                       @open-telemetry/go-approvers
 instrumentation/github.com/gorilla/mux/otelmux/                         @open-telemetry/go-approvers
 instrumentation/github.com/labstack/echo/otelecho/                      @open-telemetry/go-approvers
-instrumentation/github.com/IBM/sarama/otelsarama/                       @open-telemetry/go-approvers @pellared @hanyuancheung
+instrumentation/github.com/Shopify/sarama/otelsarama/                   @open-telemetry/go-approvers @pellared @hanyuancheung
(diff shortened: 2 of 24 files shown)


batazor:
I think I've updated all the necessary places.

MrAlias (changes requested):
It is unclear we would want to change our package name to match the upstream redirect. If we did end up wanting to do so, it will require a deprecation of the existing package.

pellared:
@batazor Can the new instrumentation library be https://github.com/IBM/sarama or new repository e.g. https://github.com/IBM/otelsarama?

CODEOWNERS
```diff
@@ -37,7 +37,7 @@ instrumentation/github.com/go-kit/kit/otelkit/                          @open-te
 instrumentation/github.com/gocql/gocql/otelgocql/                       @open-telemetry/go-approvers
 instrumentation/github.com/gorilla/mux/otelmux/                         @open-telemetry/go-approvers
 instrumentation/github.com/labstack/echo/otelecho/                      @open-telemetry/go-approvers
-instrumentation/github.com/Shopify/sarama/otelsarama/                   @open-telemetry/go-approvers @pellared @hanyuancheung
+instrumentation/github.com/IBM/sarama/otelsarama/                       @open-telemetry/go-approvers @pellared @hanyuancheung
```
MrAlias, line 40:
```suggestion
instrumentation/github.com/Shopify/sarama/otelsarama/                   @open-telemetry/go-approvers @pellared @hanyuancheung
```

CHANGELOG.md
```diff
@@ -27,6 +27,7 @@ This project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0.htm
   - `go.opentelemetry.io/contrib/instrumentation/net/http/httptrace/otelhttptrace`
   - `go.opentelemetry.io/contrib/instrumentation/net/http/otelhttp`
 - Do not modify the origin request in RoundTripper in `go.opentelemetry.io/contrib/instrumentation/net/http/otelhttp`. (#4033)
+- Migration from `github.com/Shopify/sarama` to `github.com/IBM/sarama`
```
MrAlias, line 30:
```suggestion
- Updated dependency of `github.com/Shopify/sarama` in `go.opentelemetry.io/contrib/instrumentation/github.com/IBM/sarama/otelsarama` to `github.com/IBM/sarama` so as to match the upstream package rename. (#4090)
```

MrAlias (approved):

XSAM (approved):

pellared:
I created https://github.com/IBM/sarama/issues/2510.

hanyuancheung (approved):

pellared:
> It is unclear we would want to change our package name to match the upstream redirect. If we did end up wanting to do so, it will require a deprecation of the existing package.

> I left only changing the path to the dependency, don't change the path in this package, this will save the working state of observability

I would rather change the package name to match upstream redirect as the users already need to change the import paths for the library 😉

dmathieu:
I agree that if we end up keeping this instrumentation, we should rename the package as well.
Ideally, now would be a good time to move it into a better location though.

pellared (changes requested):
Blocking mainly because of https://github.com/open-telemetry/opentelemetry-go-contrib/pull/4090#issuecomment-1641843514 but also awaiting feedback under https://github.com/IBM/sarama/issues/2510

MrAlias:
I don't think the question of renaming this package should block the update of the dependency.

instrumentation/README.md
```diff
@@ -40,25 +40,25 @@ It will include packages outside of this project.
...
+|                [github.com/gorilla/mux](./github.com/gorilla/mux/otelmux)                |         |   ✓    |
+|             [github.com/labstack/echo](./github.com/labstack/echo/otelecho)              |         |   ✓    |
+|             [github.com/IBM/sarama](./github.com/Shopify/sarama/otelsarama)              |         |   ✓    |
+|       [go.mongodb.org/mongo-driver](./go.mongodb.org/mongo-driver/mongo/otelmongo)       |         |   ✓    |
+|               [google.golang.org/grpc](./google.golang.org/grpc/otelgrpc)                |    ✓    |   ✓    |
+|                 [gopkg.in/macaron.v1](./gopkg.in/macaron.v1/otelmacaron)                 |         |   ✓    |
+|                                      [host](./host)                                      |    ✓    |        |
+|                             [net/http](./net/http/otelhttp)                              |    ✓    |   ✓    |
+|                 [net/http/httptrace](./net/http/httptrace/otelhttptrace)                 |         |   ✓    |
+|                                   [runtime](./runtime)                                   |    ✓    |        |
```
Aneurysm9, line 61: Please, no.

Aneurysm9 (commented):
I'm not sure it's appropriate to simply change the dependency here.  Won't that break anyone still using the prior import paths who updates this instrumentation?  Should we deprecate this instrumentation?  New instrumentation for the new import path can be created, ideally owned by the sarama maintainers, but anywhere really.

MrAlias:
> I'm not sure it's appropriate to simply change the dependency here. Won't that break anyone still using the prior import paths who updates this instrumentation?

Good point.

pellared:
Closing as there is no consensus to ONLY update the dependency using the new import path.

PR #4101, opened 2023-07-26
autoprop: Handle empty env var as unset

pellared:
## Why

https://github.com/open-telemetry/opentelemetry-specification/blob/v1.23.0/specification/configuration/sdk-environment-variables.md?plain=1#L19

## What

Handle empty value of `OTEL_PROPAGATORS` environment variable the same way as when the variable is unset

## Testing

The returned result of `autoprop.NewTextMapPropagator` has not changed. The only difference is that we do not log an unnecessary error via

https://github.com/open-telemetry/opentelemetry-go-contrib/blob/3061fc5ecc2eee007916fcc29bbdd756bdee5c75/propagators/autoprop/propagator.go#L53-L54

I am not sure it it is worth to add a unit test for it.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -27,7 +27,7 @@ This project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0.htm
   - `go.opentelemetry.io/contrib/instrumentation/net/http/httptrace/otelhttptrace`
   - `go.opentelemetry.io/contrib/instrumentation/net/http/otelhttp`
 - Do not modify the origin request in RoundTripper in `go.opentelemetry.io/contrib/instrumentation/net/http/otelhttp`. (#4033)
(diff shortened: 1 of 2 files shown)


dashpole (approved):

MrAlias:
Doesn't this mean that if a user does something like `OTEL_PROPAGATORS="$MY_RESOLVED_PROPS"`, but `$MY_RESOLVED_PROPS` is undefined the code will silently drop the error and just choose something for the user?

pellared:
> Doesn't this mean that if a user does something like `OTEL_PROPAGATORS="$MY_RESOLVED_PROPS"`, but `$MY_RESOLVED_PROPS` is undefined the code will silently drop the error and just choose something for the user?

Correct. This is the expected behavior. More context: https://github.com/open-telemetry/opentelemetry-specification/pull/2045

MrAlias (approved):

Issue #4013, opened 2023-06-20
autoexport: Support `none` as exporter


pellared:
Is there a no-op exporter we can use with autoexport?

_Originally posted by @james-callahan in https://github.com/open-telemetry/opentelemetry-go-contrib/issues/2753#issuecomment-1592139673_

pellared:
@MikeGoldsmith Do you want to work on it?

pellared:
How do you prefer `func NewSpanExporter(ctx context.Context, opts ...Option) (trace.SpanExporter, error) {` to behave when user selects `none`?

I see following possibilities

## Option A - return a NOOP exporter

Pros:

1. Ease of use by the user - no special handling
1. This would align with the behavior of `autoprop.NewTextMapPropagator`

Cons:

1. Performance downgrade. If `none` is selected then there may be no need to even setup a metrics provider. How the user would know that a "None" (noop) exporter was returned. MITIGATION: We cn add a `IsNone` function so that the user could handle the special case when `none` is selected.

## Option B - `return nil, nil`

Pros:

1. Easy to implement

Cons:

1. It is bug prone for the user. It may be easy for the user to configure a `nil` exporter as  `return nil, nil` is not very popular (but it happens). Users may not read the documentation and not handle this case.
2. It requires handling a `if exp == nil` case

## Option C - `return nil, ErrNone`

Pros:

1. Looks idiomatic
2. If user does not handle the exporter `err == ErrNone` he will handle it like an error. It is less bug prone than option A.

Cons:

1. This is not really an error.
2. The error handling code may look worse than for option A. In most cases setting a NOOP exporter should be good enought.


I am leaning towards Option A. @open-telemetry/go-approvers, WDYT?

dashpole:
I prefer A as well.

PR #4130, opened 2023-07-31
autoexport: Handle OTEL_TRACES_EXPORTER=none

pellared:
Fixes https://github.com/open-telemetry/opentelemetry-go-contrib/issues/4013 (option A)


exporters/autoexport/noop.go
```diff
@@ -0,0 +1,43 @@
...
+}
+
+// Shutdown is part of trace.SpanExporter interface.
+func (e noop) Shutdown(ctx context.Context) error {
+	return nil
+}
+
+// IsNone returns true for the exporter returned by [NewSpanExporter]
+// when OTEL_TRACES_EXPORTER environment variable is set to "none".
+func IsNone(e trace.SpanExporter) bool {
```
dashpole, line 40: Should this be defined in a test file, since it appears to only be used for testing?

dashpole: Got it, thanks.  I'm OK with this for now, as we don't have an alternative mechanism to disable the meterprovider.  But in the future, we hopefully can control that directly with config.

hanyuancheung (approved):

Issue #2618, opened 2022-08-01
Request to Add Metrics for otelgin


aaqaishtyaq:
## Background

Hey, I am trying to use OpenTelemetry at work. We are using `gin` and would also want to use `otel` for metrics. Currently, the `otelgin` package only provides `traces`.  Would it be possible to add support for metrics as well? 

**Package Link**: [otelgin](https://github.com/open-telemetry/opentelemetry-go-contrib/tree/main/instrumentation/github.com/gin-gonic/gin/otelgin)

## Proposed Solution

There is a `gin` middleware package [go-gin-prometheus](https://github.com/zsais/go-gin-prometheus) which adds metrics for every HTTP request. With adding tracing, metrics can probably also be added.

PR #4202, opened 2023-08-17
Record http.server.duration metric in gin instrumentation

anuraaga:
For #2618 (didn't mark as fixing since it's only one metric)

I noticed gRPC instrumentation records duration metric, which was great, but unfortunately gin doesn't yet. If it's possible to add it, that would be nice.

Note, this doesn't bump the semconv version from 1.17 so it is this version of the spec

https://github.com/open-telemetry/opentelemetry-specification/blob/v1.17.0/specification/metrics/semantic_conventions/http-metrics.md

I know that there is significant churn coming in the conventions so if it's not a good time to add this metric, it's fine to close this and I'll just add it directly in the apps, but first I wanted to see if it's possible to do it here where all gin users would be able to benefit.

Many patterns follow that of the gRPC instrumentation's metrics support.


anuraaga:
@hanyuancheung or anyone would you be able to review this PR? Thanks!

MrAlias:
> I know that there is significant churn coming in the conventions so if it's not a good time to add this metric, it's fine to close this and I'll just add it directly in the apps

I would hold off on this for this reason.

utr1903:
First of all, @anuraaga thanks for the great work! I was willing to have a similar functionality as in [`otelhttp`](https://github.com/open-telemetry/opentelemetry-go-contrib/blob/main/instrumentation/net/http/otelhttp/handler.go) just like your implementation :)

@MrAlias A general question: When would we able to review and merge this PR and according to which semantic conventions version?

PR #4228, opened 2023-08-21
add config parsing module

codeboten:
This code is moved from the OpenTelemetry Collector to attempt sharing the same code across both the collector and the go otel implementation.

TODO:
- [x] removing the `confmap` import
- [x] add make target to generate config code from json schema


config/config.go
```diff
@@ -0,0 +1,143 @@
+// Copyright The OpenTelemetry Authors
+// SPDX-License-Identifier: Apache-2.0
+
+package config // import "go.opentelemetry.io/contrib/config"
+
+import (
+	"fmt"
+
+	"go.opentelemetry.io/collector/confmap"
```
mx-psi, line 9: Looks like both `Unmarshal` functions could instead be validate functions, right? And we can implement the `Validate` methods without `confmap`, so that should be enough to remove the `confmap` dependency

codeboten: Removed the unmarshal methods/

config/generated_config.go
```diff
@@ -0,0 +1,661 @@
+// Code generated by github.com/atombender/go-jsonschema, DO NOT EDIT.
```
pellared, line 1: I made a high level look. Here are some questions which come to my mind.

1. How and where do we want to use this module? What is it responsible for?
2. What we will do once there would be a breaking change in the JSON schema? If it will make a breakin change in the exported API then users may encounter problems at build time when bumping modules.
3. Is this really work properly with JSON unmarshal? Does it not miss `json` tags?
4. Do we want to support `yaml.v3`?
5. Are we able e.g. to fix some naming nit issues like: Otlp -> OTLP?
6. Do we think that it is a good idea to have the generated code exported? Maybe we should have it unexported or internal and then map it some more Go idomatic types (e.g. get rid of pointers etc)?

I have a feeling that the module should provide functions like `func FromJSON(r io.Reader) (*Config, error)` and `func FromJSON(r io.Reader) (*Config, error)` that would read a config in given format from a stream. The function would unmarshal -> validate -> map to a hand-made `Config` struct.

Probably adding package docs and testable example would also help in reviewing and validating the design.

pellared: I think that the scope of this PR could be the part defined as `Configurator.Parse` in https://github.com/open-telemetry/opentelemetry-specification/pull/3437.

mx-psi: I am not a part of the Configuration WG, but this is my best attempt at answering the questions.

> How and where do we want to use this module? What is it responsible for?

This code comes from the Collector. For example, the generated config file current location is this: https://github.com/open-telemetry/opentelemetry-collector/blob/main/service/telemetry/generated_config.go. It is part of the work from the OpenTelemetry Configuration WG and it would have the same owners here as it has there. 

The intent in the Collector is to provide the configuration structs to be used to configure the Collector's internal telemetry, but it makes more sense to have it here since these would also be used as part of the OpenTelemetry Go support for the general configuration schema being worked on by the Config WG.

> What we will do once there would be a breaking change in the JSON schema? If it will make a breakin change in the exported API then users may encounter problems at build time when bumping modules.

I think that is to be expected at this point, and the way to signal this is to release this module with 0.X versioning.

> Is this really work properly with JSON unmarshal? Does it not miss json tags?

Json tags are not necessary in general (https://go.dev/play/p/5ZH3i283Ufi) and this uses mapstructure tags in any case to provide JSON support.

> Do we want to support yaml.v3?

`mapstructure` tags should provide support for YAML, it is what we use in the Collector.

> Are we able e.g. to fix some naming nit issues like: Otlp -> OTLP?

I think that would be an issue on https://github.com/open-telemetry/opentelemetry-configuration

> Do we think that it is a good idea to have the generated code exported? Maybe we should have it unexported or internal and then map it some more Go idomatic types (e.g. get rid of pointers etc)?

This is necessary today due to the Collector code structure. We can think about a refactor later before making this module 1.0.

codeboten: Thanks @pellared & @mx-psi for the comments.

> Are we able e.g. to fix some naming nit issues like: Otlp -> OTLP?

Yes this is configurable via the go-jsonschema which should be ok to do. I suppose we could also propose the change upstream as @mx-psi suggested, but if all we care about are the generated go structs, it's probably enough to move forward w/ the changes in the go-jsonschema generation.

> Do we think that it is a good idea to have the generated code exported? Maybe we should have it unexported or internal and then map it some more Go idomatic types (e.g. get rid of pointers etc)?

I don't have a good alternative at this point. We may want to follow the same model as semantic conventions and release the module behind a version that matches released versions of the configuration. WDYT?

> Probably adding package docs and testable example would also help in reviewing and validating the design.

Yup... 100%. I was mostly just trying to move the minimum amount of code over from the Collector into a separate package to start untangling the dependencies from that side.

I'm ok with adding the functionality to do the parsing and configure the SDK as part of this initial PR, or we can break it up. Whatever makes the most sense here, i'm open to suggestions

pellared: First I want to understand how people would like to use it. From my perspective the OpenTelemetry Go users would primarily need a configuration model (POJO like, without any tags) and something which creates the providers based on the model. Something like:

```go
package config

// Config represents the SDK configuration model.
type Config struct {}

// NewSDK creates SDK providers based on the configuration model.
func NewSDK(ctx, context.Context, c Config) (*SDK, error) {}


type SDK struct {}
func (sdk *SDK) Shutdown() error {}
func (sdk *SDK) TraceProvider() *trace.TraceProvider {}
func (sdk *SDK) MeterProvider() *meter.MeterProvider {}
```

The config parsers could be in child modules to reduce the number of dependencies (not everyone wants to take e.g. yaml.v3 as a dependency). Still we can have `config/configyaml` module with exporter API like following:

```go
package configyaml

func Parse(r io.Reader) (*Config, error){}
```

Then in future we could create more modules handling other "protocols" (json, protobuf etc).

Example usage:

```go
package main

import (
	"context"
	"errors"
	"log"
	"os"
	"os/signal"

	"go.opentelemetry.io/contrib/config"
	"go.opentelemetry.io/contrib/config/configyaml"
)

func main() {
	if err := run(); err != nil {
		log.Fatalln(err)
	}
}


func run() (err error) {
	// Handle SIGINT (CTRL+C) gracefully.
	ctx, stop := signal.NotifyContext(context.Background(), os.Interrupt)
	defer stop()

	otelSDK, err := setupOTelSDK(ctx, serviceName, serviceVersion)
	if err != nil {
		return
	}
	// Handle shutdown properly so nothing leaks.
	defer func() {
		err = errors.Join(err, otelSDK.Shutdown(context.Background()))
	}()

	// ... other application bootstrapping code ...

	return
}

func setupOTelSDK() (*config.SDK, error) {
	f, err := os.ReadFile("otel.yaml")
	if err != nil {
		return nil, err
	}
	defer f.Close()

	cfg, err := configyaml.Parse(f)
	if err != nil {
		return nil, err
	}

	return config.NewSDK(ctx, cfg)
}
```

Would such design be usable for the collector?

As a side not, I can predict that they may be OTel Go users who would even prefer using such `config` package (without any file parsing) by creating `config.Config` in code than using the SDK.
Pros: it may be less code to write, more declarative
Cons: not strongly typed, lower control.

pellared:
I think this PR helps validating the https://github.com/open-telemetry/opentelemetry-specification/pull/3437

CC @jack-berg

config/generated_config.go
mx-psi: > Would such design be usable for the collector?

Some requirements for this package to be usable by the Collector are:

1. Ability to use this configuration as part of a larger configuration. The configuration will be under the `service::telemetry` section, so we need to be able to embed this configuration inside a larger config structure (or have a parser that accounts for this).
2. During the transition towards this configuration, ability to mix this configuration with our legacy schema and only partially expose this configuration. For example, for metrics we now expose just the metric readers and not the views, and we do so under [a specific struct in our configuration](https://github.com/open-telemetry/opentelemetry-collector/blob/589488839f582632f53d84526207e733475ccc21/service/telemetry/config.go#L100-L114) that also has the legacy way of configuring metrics via Prometheus.
3. Support for the Collector `${..}` syntax for loading environment variables and merging different configurations. If we don't expose the configuration, this could be accomplished by depending on the Collector's [`confmap` module](https://pkg.go.dev/go.opentelemetry.io/collector/confmap) but it would imply  having the `service::telemetry` section as a special case.

I think it is hard to support these with the proposed design, specially (2) seems to be hard without exposing the individual configuration structs in some capacity.

mx-psi: One way to hide the structs while remaining compatible with the Collector way of doing things is using lots of interfaces as shown here https://go.dev/play/p/vZWLwzV0B5z. IMO it leads to an awkward UX, but it's a possibility.

pellared: This looks like very specific parsing logic for the collector. Why not having the parsing code in collector repository?

Could the collector still benefit from a `config` package which would bootstrap the SDK based on the passed config?

mx-psi: > This looks like very specific parsing logic for the collector. Why not having the parsing code in collector repository?

But then we are back at square one (which is roughly what this PR proposes): we need to have the structs with the struct tags to be part of the public API for that to be possible, otherwise how would the Collector implement parsing?

pellared: Collector could do the parsing like proposed in this PR and map its result to `config.Config`.

> we need to have the structs with the struct tags to be part of the public API

This would be a leakage of internal detail. Why not `json` tags? Why not `yaml` tags? Why not `xml` tags?

mx-psi: > This would be a leakage of internal details of the clients using the config package.

I disagree, this would mean having more public API but would not mean leaking anything; we are just making it available for users because it is (AFAICT, ignoring the interfaces approach mentioned above) the only way external consumers of the package can actually implement advanced use cases for parsing. The fact that the tags are called `mapstructure` does not mean an actual dependency on the mapstructure package.

In the Collector, we explicitly call out that struct tags are part of the public API https://github.com/open-telemetry/opentelemetry-collector/blob/main/VERSIONING.md#policy and we make struct tags publicly available for all configuration structs.

>  Why not json tags? Why not yaml tags? Why not xml tags?

We could add those if necessary in the future, although note that mapstructure can be used to map to JSON, XML and YAML, and that in principle that struct tag name is arbitrary and not tied to a specific package.

pellared: From https://www.reddit.com/r/golang/comments/rtne10/good_antipattern_blog_post/ (posting via reddit as it also has some comments)

> There’s one feature in Go that introduces strong coupling and makes you think you’re following DRY. It’s using multiple tags in a single struct.

I agree with it based on my experience. The way you define the struct and with tags is coupled the the parsing library that you are using. If we add these tags then we should also check if these are really working. We should include tests to verify if they work with the libraries that support them. Otherwise, how would we know that the tags are good?

codeboten: > Could the collector still benefit from a config package which would bootstrap the SDK based on the passed config?

The collector could still benefit from the package for configuring the SDK. This would at least remove the need to have this live in multiple places. 

> Why not having the parsing code in collector repository?

I do worry that having to parse the configuration in a separate repo, then mapping the values to the `config.Config` object will end up causing some unnecessary duplicate work for both repositories but maybe there's no way around it if the decision in this repo is not to make struct tags available for consumers.

@pellared Just to be clear, the idea of `config.Config` would be to make structs generated by jsonschema available without the struct tags?

pellared: Having that said (written 😉) I think that using `jsonschema` for generation with not perfect names and pointers would be an excellent initial PR. 

We can then create follow-up issues (and PRs) to handle other concerns (step by step) e.g.
- config: Refine model naming
- config: Do not use pointers in model
- config: Add a function which creates SDK from model

MadVikingGod: I'm going to push back on the use of a custom `Optional` type.  For as long as we had json and yaml marshalers, a pointer with `omitEmpty` has been how to represent this kind of Optional.  We shouldn't make a helper type just to avoid some helper functions.  

I am a bit concerned about the names of the schema and how they might be looking to change over time.  After the first release with this module there needs to be little churn as it can make different versions incompatible with each other.

config/metric_view.go
```diff
@@ -0,0 +1,96 @@
...
+
+package config // import "go.opentelemetry.io/contrib/config"
+
+import (
+	"go.opentelemetry.io/otel/attribute"
+	"go.opentelemetry.io/otel/sdk/instrumentation"
+	sdkmetric "go.opentelemetry.io/otel/sdk/metric"
+)
+
+func viewOptionsFromConfig(views []View) []sdkmetric.Option {
```
MadVikingGod, line 12: Is there a reason we have these private functions and not building the SDKs from these options?

MadVikingGod (commented):
I think this is an excellent start to this approach.  

A README.md with how this will be used, what constraints there are, and what features are needed would go a long way to help evaluate if this is good enough for the first draft.

config/generated_config.go
pellared: I created https://github.com/open-telemetry/opentelemetry-go-contrib/labels/area%3A%20config for easier tracking. This PR could be only about solving https://github.com/open-telemetry/opentelemetry-go-contrib/issues/4369.

codeboten:
@pellared  @mx-psi @MadVikingGod i've produced another PR that includes a subset of the changes here. It is trimmed down to the code to generate models from the jsonschema, the generated code, and a readme... i figured this would be a better way to get the changes in. See https://github.com/open-telemetry/opentelemetry-go-contrib/pull/4376

Moving just that code over would also allow us to pull code out of the collector contrib package right away, see the draft pr here: https://github.com/open-telemetry/opentelemetry-collector/pull/8620

codeboten (commented):
Closing this PR as it will be broken up into several PRs

PR #4229, opened 2023-08-22
autoexport: Add support for metrics

### LINKED ISSUE #4131: autoexport: Add metrics support
Make sure to support `OTEL_METRICS_EXPORTER` and handle `none`, `otlp`, ~~`prometheus`~~.

Reference https://github.com/open-telemetry/opentelemetry-specification/blob/main/specification/configuration/sdk-environment-variables.md#exporter-selection

punya:
Fixes #4131.

Reorganizes autoexport into

*   generic registry and env lookup independent of signal type
*   metric- and trace-specific wrappers around the generic part, to preserve a simple API
*   pre-defined metric and trace factories

Also reorganizes tests to line up with these layers.


exporters/autoexport/exporter.go
```diff
@@ -102,16 +115,75 @@ func NewSpanExporter(ctx context.Context, opts ...Option) (trace.SpanExporter, e
...
+//
+// An error is returned if an environment value is set to an unhandled value.
+//
+// Use [RegisterMetricReader] to handle more values of OTEL_METRICS_EXPORTER.
+//
+// Use [WithFallbackMetricReader] option to change the returned exporter
+// when OTEL_TRACES_EXPORTER is unset or empty.
+//
+// Use [IsNoneSpanExporter] to check if the retured exporter is a "no operation" exporter.
+func NewMetricReader(ctx context.Context, opts ...Option) (metric.Reader, error) {
```
pellared, line 143: Can we define a `MetricOption` instead of reusing existing `Option`.

We could also consider renaming `Option` to `SpanOption`, but this would have to be as a separate PR.

exporters/autoexport/exporter.go
```diff
@@ -102,16 +115,75 @@ func NewSpanExporter(ctx context.Context, opts ...Option) (trace.SpanExporter, e
...
 		return nil, err
 	}
-	return config.fallbackExporter, nil
+	return config.fallbackSpanExporter, nil
+}
+
+// NewMetricReader returns a configured [go.opentelemetry.io/otel/sdk/metric.Reader]
+// defined using the environment variables described below.
+//
+// OTEL_METRICS_EXPORTER defines the metrics exporter; supported values:
```
pellared, line 124: Can you also add support for `prometheus`?

pellared (commented):
I just made a very quick look. I would need to look deeper in the following days. Looks nice so far👍

exporters/autoexport/exporter.go
punya: I'll rename `Option` to `SpanOption` but leave `Option` as an alias for backwards compatibility. Later we can decide if/when we want to remove it.

exporters/autoexport/exporter.go
pellared: This can be in a separate PR is you prefer.

exporters/autoexport/noop.go
```diff
@@ -41,3 +42,11 @@ func IsNoneSpanExporter(e trace.SpanExporter) bool {
 	_, ok := e.(noop)
 	return ok
 }
+
+var noopMetricReader = metric.NewManualReader()
```
pellared, line 46: There could be one corner case where returning a global object could produce a bug from the user perspective. Let me try to illustrate it:

```go
ctx := context.Background()
os.Setenv("OTEL_METRICS_EXPORTER", "none")

exp, _ := autoexport.NewMetricReader(ctx)
fmt.Println(exp.Shutdown(ctx))
// Output: <nil>

exp, _ = autoexport.NewMetricReader(ctx)
fmt.Println(exp.Shutdown(ctx))
// Output: reader is shutdown
```

See: https://github.com/open-telemetry/opentelemetry-go/blob/60666c554065ac4da502fe28943eea4b938ab479/sdk/metric/manual_reader.go#L76-L94

I think we should implement something more or less like (not tested):

```go
type noopMetricReader struct {
	*metric.ManualReader // this would have to be initialized with metric.NewManualReader() when creating noopMetrics 
}

// MarshalLog returns logging data.
func (*noopMetricReader) MarshalLog() interface{} {
	return struct{ Type string }{Type: "noop metrics exporter"}
}

// Shutdown is noop.
func (*noopMetricReader) Shutdown(context.Context) error {
	return nil
}

// Collect is noop.
func (*noopMetricReader) Collect(context.Context, metricdata.ResourceMetrics) error {
	return nil
}
```

pellared (commented):
Looks very good 👍 I just found one more issue.

exporters/autoexport/exporter.go
MrAlias: Given this is an unstable package, we could also just make the name change directly as long as we document it in the CHANGELOG. Otherwise, we should deprecate the alias now so we can remove it in the next release.

exporters/autoexport/noop.go
punya: Thanks for the catch!

exporters/autoexport/exporter.go
punya: I marked it as deprecated following the commenting convention. Do I also need to add it to the changelog?

punya:
@pellared @MrAlias this is ready for another round of review.

exporters/autoexport/metrics.go
```diff
@@ -0,0 +1,102 @@
...
+			if err != nil {
+				return nil, err
+			}
+			return metric.NewPeriodicReader(r), nil
+		default:
+			return nil, errInvalidOTLPProtocol
+		}
+	})
+	RegisterMetricReader("prometheus", func(ctx context.Context) (metric.Reader, error) {
+		return prometheus.New()
```
dashpole, line 97: **Non-blocking for this PR**:  Note that this does not actually serve the prometheus metric endpoint.  A user would still need to create the prometheus handler and listen on a port.

This could be a reasonable place to implement the prometheus exporter environment variables: https://github.com/open-telemetry/opentelemetry-specification/blob/563958cb2bd8529990f19fdce7a5f3643bf63091/specification/configuration/sdk-environment-variables.md?plain=1#L201

dashpole (approved):

dashpole:
From https://github.com/open-telemetry/opentelemetry-specification/blob/563958cb2bd8529990f19fdce7a5f3643bf63091/specification/configuration/sdk-environment-variables.md?plain=1#L222:

> The SDK MAY accept a comma-separated list to enable setting multiple exporters.

This would be useful, but doesn't really fit with how the package is currently designed.

exporters/autoexport/metrics.go
punya: Should they be implemented in this package or in the Prometheus package?

exporters/autoexport/spans.go
```diff
@@ -18,52 +18,23 @@ import (
...
-		exp, err := spanExporter(context.Background(), "otlp")
-		if err != nil {
-			return cfg, err
-		}
-		cfg.fallbackExporter = exp
-	}
-	return cfg, nil
-}
+// SpanOption applies an autoexport configuration option.
+type SpanOption = option[trace.SpanExporter]
```
MrAlias, line 27: Should this be `TraceOption`?

I know we have the `SpanExporter`, but I wonder if this needs to match that or the corollary to metric: trace.

MrAlias (approved):

pellared:
> From https://github.com/open-telemetry/opentelemetry-specification/blob/563958cb2bd8529990f19fdce7a5f3643bf63091/specification/configuration/sdk-environment-variables.md?plain=1#L222:
> 
> > The SDK MAY accept a comma-separated list to enable setting multiple exporters.
> 
> This would be useful, but doesn't really fit with how the package is currently designed.

I created https://github.com/open-telemetry/opentelemetry-go-contrib/issues/4471

exporters/autoexport/metrics.go
pellared: I would rather implement it in this package.

punya: Sounds good, I have created #4472 to track that work.

pellared (approved):

CHANGELOG.md
```diff
@@ -74,6 +74,7 @@ This project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0.htm
 - Add `WithSpanOptions` option in `go.opentelemetry.io/contrib/instrumentation/google.golang.org/grpc/otelgrpc`. (#3768)
 - Add testing support for Go 1.21. (#4233)
 - Add `WithFilter` option to `go.opentelemetry.io/contrib/instrumentation/github.com/gorilla/mux/otelmux`. (#4230)
+- Add metrics support to `go.opentelemetry.io/contrib/exporters/autoexport`. (#4229)
```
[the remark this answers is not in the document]
punya: Thanks for the catch - the line was originally in "unreleased" but must have moved during merges.

exporters/autoexport/spans.go
pellared: We can address it in a separate PR if needed.

Personally, I was thinking about renaming to `SpanExporterOption`.

Let's track it in a separate issue.

pellared: https://github.com/open-telemetry/opentelemetry-go-contrib/issues/4474

PR #4258, opened 2023-08-29
Add method CloseIdleConnections for otelhttp.Transport

lesha888:
Add method CloseIdleConnections for otelhttp.Transport

CloseIdleConnections closes any connections on its Transport which
were previously connected from previous requests but are now
sitting idle in a "keep-alive" state. It does not interrupt any
connections currently in use.
If the Client's Transport does not have a CloseIdleConnections method
then this method does nothing.


dmathieu:
You will need to sign the CLA. And this will need a test.

PR #4322, opened 2023-09-21
otelgrpc: Remove high cardinality metric attributes

### LINKED ISSUE #3071: otelgrpc high metric cardinality from rpc.server.duration

MadVikingGod:
Related to #3536 
Closes #3071

This removes Peer attributes from grpc.


dashpole (approved):

CHANGELOG.md
```diff
@@ -14,6 +14,10 @@ This project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0.htm
 - Set the description for the `rpc.server.duration` metric in `go.opentelemetry.io/contrib/instrumentation/google.golang.org/grpc/otelgrpc`. (#4302)
 - Add `NewServerHandler` and `NewClientHandler` that return a `grpc.StatsHandler` used for gRPC instrumentation in `go.opentelemetry.io/contrib/instrumentation/google.golang.org/grpc/otelgrpc`. (#3002)
 
+### Removed
+
+- The The high cardinality attributes `net.sock.peer.*` and `net.peer.*` were removed from the metrics generated by `go.opentelemetry.io/contrib/instrumentation/google.golang.org/grpc/otelgrpc`. (#????)
```
dmathieu, line 19:
```suggestion
- The The high cardinality attributes `net.sock.peer.*` and `net.peer.*` were removed from the metrics generated by `go.opentelemetry.io/contrib/instrumentation/google.golang.org/grpc/otelgrpc`. (#4322 )
```

dmathieu (commented):
Should the attributes being emitted be tested?

CHANGELOG.md
```diff
@@ -14,6 +14,10 @@ This project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0.htm
 - Set the description for the `rpc.server.duration` metric in `go.opentelemetry.io/contrib/instrumentation/google.golang.org/grpc/otelgrpc`. (#4302)
 - Add `NewServerHandler` and `NewClientHandler` that return a `grpc.StatsHandler` used for gRPC instrumentation in `go.opentelemetry.io/contrib/instrumentation/google.golang.org/grpc/otelgrpc`. (#3002)
 
+### Removed
+
+- The The high cardinality attributes `net.sock.peer.*` and `net.peer.*` were removed from the metrics generated by `go.opentelemetry.io/contrib/instrumentation/google.golang.org/grpc/otelgrpc`. (#4322)
```
pellared, line 19:
```suggestion
- The `net.sock.peer.*` and `net.peer.*` high cardinality attributes are removed from the metrics generated by `go.opentelemetry.io/contrib/instrumentation/google.golang.org/grpc/otelgrpc`. (#4322)
```

pellared (approved):

pellared:
> Should the attributes being emitted be tested?

IMO they should. Example assertion: https://github.com/open-telemetry/opentelemetry-go-contrib/blob/bf86b26a1d265ad04a07ae2abadcd08e9d18e174/instrumentation/net/http/otelhttp/test/handler_test.go#L44

PS. Now we also have [`metricdatatest.IgnoreValue`](https://pkg.go.dev/go.opentelemetry.io/otel/sdk/metric/metricdata/metricdatatest#IgnoreValue) 😉

pellared:
> But we also wouldn't want these tests to break just because we updated to the next semantic convention, that would be toil.

On the other hand, it is double-checking (validating) the changes.

pellared (approved):
Approving as https://github.com/open-telemetry/opentelemetry-go-contrib/issues/4323 is created and added to the next release milestone.

instrumentation/google.golang.org/grpc/otelgrpc/interceptor.go
```diff
@@ -498,6 +498,11 @@ func StreamServerInterceptor(opts ...Option) grpc.StreamServerInterceptor {
 	}
 }
 
+func metricAttributes(fullMethod string) []attribute.KeyValue {
+	_, attrs := internal.ParseFullMethod(fullMethod)
```
MrAlias, line 502: This is now being called twice. Once in `spanInfo` and now here. Why not just update `spanInfo` to returns two sets of attributes, one for spans and the other for metrics. The latter being the equivalent of what is returned here?

instrumentation/google.golang.org/grpc/otelgrpc/interceptor.go
```diff
@@ -368,7 +368,7 @@ func UnaryServerInterceptor(opts ...Option) grpc.UnaryServerInterceptor {
 		var statusCode grpc_codes.Code
 		defer func(t time.Time) {
 			elapsedTime := time.Since(t) / time.Millisecond
-			attr = append(attr, semconv.RPCGRPCStatusCodeKey.Int64(int64(statusCode)))
+			attr := append(metricAttrs, semconv.RPCGRPCStatusCodeKey.Int64(int64(statusCode)))
```
MrAlias, line 371: Given `metricAttrs` and `attrs` share the same backing array, this will mutate `attrs` as well.

```suggestion
			attr := append([]attribute.KeyValue{semconv.RPCGRPCStatusCodeKey.Int64(int64(statusCode))}, metricAttrs...)
```

MadVikingGod: This doesn't modify the underlying slice when I tested by forcing peerInfo to return data.  Either way, I switched to the suggestion.

puckpuck:
Any updates on when this is going to get merged and released?

dashpole (approved):

pellared (approved):

pellared:
@MadVikingGod @MrAlias are there any blockers?

pellared (changes requested):
After dealing with gRPC code reviews recently I have a strong opinion that tests should be included as part of this PR.

I will do my best to help and take https://github.com/open-telemetry/opentelemetry-go-contrib/issues/4483

pellared:
You can change the `doCalls` in `gprc_test.go` to the following code:

```go
func doCalls(t *testing.T, cOpt []grpc.DialOption, sOpt []grpc.ServerOption) error {
	l := newGRPCTestServer(t, func(s *grpc.Server) {
		pb.RegisterTestServiceServer(s, interop.NewTestServer())
	}, sOpt...)

	ctx := context.Background()
	conn, err := grpc.DialContext(
		ctx,
		l.Addr().String(),
		append([]grpc.DialOption{
			grpc.WithTransportCredentials(insecure.NewCredentials()),
		}, cOpt...)...,
	)
	if err != nil {
		return err
	}
	defer conn.Close()
	client := pb.NewTestServiceClient(conn)

	interop.DoEmptyUnaryCall(client)
	interop.DoLargeUnaryCall(client)
	interop.DoClientStreaming(client)
	interop.DoServerStreaming(client)
	interop.DoPingPong(client)

	return nil
}

// newGRPCTestServer starts a gRPC server for testing purposes.
// The gRPC server is stopped at the end of the test during Cleanup.
// Use the register callback to setup custom gRPC method handlers.
func newGRPCTestServer(t *testing.T, register func(*grpc.Server), opt ...grpc.ServerOption) net.Listener {
	t.Helper()

	listener, err := net.Listen("tcp", "127.0.0.1:0")
	require.NoError(t, err, "failed to open port")

	grpcServer := grpc.NewServer(opt...)
	register(grpcServer)
	errCh := make(chan error)
	go func() {
		errCh <- grpcServer.Serve(listener)
	}()
	t.Cleanup(func() {
		grpcServer.Stop()
		assert.NoError(t, <-errCh, "grpcServer.Serve errored")
	})
	return listener
}
```

This will produce a gRPC test server similar to `httptest` and therefore the instrumentation will create the `peer` attributes. The `newGRPCTestServer` implementantion is taken from: https://gist.github.com/pellared/1be421c4a7bee3406fbadb04c12bda26 which I created a few years ago and I never had any issues with it.

MadVikingGod:
Let me get this straight: for me to remove problematic attributes from the metrics, like the ones in https://github.com/open-telemetry/opentelemetry-go-contrib/security/advisories/GHSA-5r5m-65gx-7vrh, I have to:
1. Change the test setup that was previously there.
2. Figure out a way to smuggle the peer port into the individual tests
3. Fix all the span tests that now have peer attributes.
4. Verify that we **still** don't have the attributes we are removing.

I'm all for having better tests, but this seems to be scope creep for dealing with an issue seen in the wild.

pellared:
> Let me get this straight: for me to remove problematic attributes from the metrics, like the ones in [GHSA-5r5m-65gx-7vrh](https://github.com/open-telemetry/opentelemetry-go-contrib/security/advisories/GHSA-5r5m-65gx-7vrh), I have to:
> 
> 1. Change the test setup that was previously there.
> 2. Figure out a way to smuggle the peer port into the individual tests
> 3. Fix all the span tests that now have peer attributes.
> 4. Verify that we **still** don't have the attributes we are removing.
> 
> I'm all for having better tests, but this seems to be scope creep for dealing with an issue seen in the wild.

I already did the 1. and 2. for you. Given that the issue is important then we should have test coverage to avoid regression bugs.

RmStorm:
Uhm, not be out of line or anything but everybody that updates their otel dependencies and pulls in these high cardinality metrics basically ends up completely poisoning their Prometheus instance with a giant cardinality explosion. This is very difficult to recover from.. This PR should be merged and released ASAP. @pellared I don't think you're giving the negative effects of this problem due consideration. It would be more prudent to extend the test coverage in a follow up PR..

pellared:
> Uhm, not be out of line or anything but everybody that updates their otel dependencies and pulls in these high cardinality metrics basically ends up completely poisoning their Prometheus instance with a giant cardinality explosion. This is very difficult to recover from.. This PR should be merged and released ASAP. @pellared I don't think you're giving the negative effects of this problem due consideration. It would be more prudent to extend the test coverage in a follow up PR..

Tests are vital to keep the project maintainable. If this is such an urgent issue then it deserves test coverage. I find tests and documentation as a fundamental part of a complete contribution. I already presented how tests can be added in this PR and @MadVikingGod almost done it here: https://github.com/open-telemetry/opentelemetry-go-contrib/pull/4503. I would not feel comfortable releasing without having tests, especially that I verified how tests can be added (and provided feedback on it).

RmStorm:
> > Uhm, not be out of line or anything but everybody that updates their otel dependencies and pulls in these high cardinality metrics basically ends up completely poisoning their Prometheus instance with a giant cardinality explosion. This is very difficult to recover from.. This PR should be merged and released ASAP. @pellared I don't think you're giving the negative effects of this problem due consideration. It would be more prudent to extend the test coverage in a follow up PR..
> 
> Tests are vital to keep the project maintainable. If this is such an urgent issue then it deserves test coverage. I find tests and documentation as a fundamental part of a complete contribution. I already presented how tests can be added in this PR and @MadVikingGod almost done it here: #4503. I would not feel comfortable releasing without having tests, especially that I verified how tests can be added (and provided feedback on it).

I see my edit was poorly timed :sweat_smile: I completely agree with you that tests are vital, I really do. But as I pointed out in my edit, checking that these particular attributes are not present does not actually prevent a regression of the core problem. Simply reverting the original change (which was adding the labels that cause the explosion) and then in parallel working on a PR that would actually prevent regressions seems like a much more prudent approach to me. Again, everybody updating their otel dependencies is going to get their Prometheus instance poisoned. We had to nuke ours.

pellared (approved):
Code is covered by tests thanks to https://github.com/open-telemetry/opentelemetry-go-contrib/pull/4503

Issue #4339, opened 2023-09-26
bridges/prometheus: Refine documentation


pellared:
I just left some comments for the documentation. Reference: https://github.com/open-telemetry/opentelemetry-go/blob/main/CONTRIBUTING.md#documentation

@dashpole let me know if you want me to merge this PR as it is.

_Originally posted by @pellared in https://github.com/open-telemetry/opentelemetry-go-contrib/pull/4227#pullrequestreview-1643674489_

PR #4348, opened 2023-09-26
Convert prometheus bridge readme to godocs

dashpole:
fixes https://github.com/open-telemetry/opentelemetry-go-contrib/issues/4339


dashpole:
@pellared I wanted to make the example print something out to the console, but that requires importing the stdoutmetric package.  Is there a way to add dependencies only for the example?

pellared:
> @pellared I wanted to make the example print something out to the console, but that requires importing the stdoutmetric package. Is there a way to add dependencies only for the example?

No 😢 We had the same problem in the sdk/metric example. We decided to simply use a ManualReader instead. See https://github.com/open-telemetry/opentelemetry-go/blob/9bbefc6cc31995c7bdd37c69eb153fc2201c3562/sdk/metric/example_test.go#L47-L50

MrAlias (approved):

bridges/prometheus/example_test.go
```diff
@@ -0,0 +1,36 @@
...
+package prometheus_test
+
+import (
+	"go.opentelemetry.io/contrib/bridges/prometheus"
+	"go.opentelemetry.io/otel/sdk/metric"
+)
+
+func ExampleNewMetricProducer() {
+	// Create a Promethes bridge "Metric Producer" which adds metrics from the
+	// [prometheus.DefaultGatherer]. Add the [WithGatherer] option to add metrics
```
pellared, line 24:
```suggestion
	// prometheus.DefaultGatherer. Add the WithGatherer option to add metrics
```

These are just regular code comments. The "Go Doc" syntax does not work here.

pellared (approved):
One suggestion. Awaiting for a non-Splunk approval.

pellared:
Should we also add "Bridges" here: https://github.com/open-telemetry/opentelemetry-go-contrib#contents?

dashpole:
> Should we also add "Bridges" here: https://github.com/open-telemetry/opentelemetry-go-contrib#contents?

Done https://github.com/open-telemetry/opentelemetry-go-contrib/pull/4348/commits/35c2be67700cc0cde69626415c57de704b03e7f0

pellared (approved):

hanyuancheung (approved):

PR #4356, opened 2023-10-01
otelgrpc: Add metrics support to NewServerHandler and NewClientHandler

### LINKED ISSUE #4316: otelgrpc: Add metrics support for stats handler
> > > I noticed that the stats handler misses the functionality you added here: #2700. Any reasons for that?
> > 
> > 
> > I was plan to implement the stats.handler step by step, cover the trace first, and then cover the metrics. Which is also convenient for review.
> > Do you think I should implement it in this pr or create another pr for metrics?
> > I'm open for both way. @pellared
> 
> During the SIG meetings we discussed that we would like to have a feature-parity before we deprecate interceptors as part of [#3002 (review)](https://github.com/open-telemetry/opentelemetry-go-contrib/pull/3002#pullrequestreview-1613182957).
> 
> Still I think we can tackle it in a separate PR.
> 
> Are there any other features which are missing when comparing with the interceptors?

I think metric is the only missing feature, since I wrote a simple e2e test `TestStatsHandler` just like interceptors to make sure the trace related function is complete.
@pellared

_Originally posted by @fatsheep9146 in https://github.com/open-telemetry/opentelemetry-go-contrib/issues/3002#issuecomment-1723227007_

fatsheep9146:
fix #4316


instrumentation/google.golang.org/grpc/otelgrpc/test/grpc_stats_handler_test.go
```diff
@@ -579,3 +588,49 @@ func checkServerSpans(t *testing.T, spans []trace.ReadOnlySpan) {
...
+			method := getRPCMethod(attr)
+			assert.NotEmpty(t, method)
+			assert.ElementsMatch(t, []attribute.KeyValue{
+				semconv.RPCMethod(method),
+				semconv.RPCService("grpc.testing.TestService"),
+				otelgrpc.RPCSystemGRPC,
+				otelgrpc.GRPCStatusCodeKey.Int64(int64(codes.OK)),
+			}, attr)
+		}
+	}
```
pellared, line 612: Can you try using https://pkg.go.dev/go.opentelemetry.io/otel/sdk/metric/metricdata/metricdatatest#AssertEqual? 

You can find example usages even in this repository.

instrumentation/google.golang.org/grpc/otelgrpc/stats_handler.go
```diff
@@ -173,14 +284,27 @@ func handleRPC(ctx context.Context, rs stats.RPCStats) {
...
+			rpcStatusAttr = semconv.RPCGRPCStatusCodeKey.Int(int(s.Code()))
 		} else {
-			span.SetAttributes(statusCodeAttr(grpc_codes.OK))
+			rpcStatusAttr = semconv.RPCGRPCStatusCodeKey.Int(int(grpc_codes.OK))
 		}
+		span.SetAttributes(rpcStatusAttr)
 		span.End()
+
+		metricAttrs = append(metricAttrs, rpcStatusAttr)
+		r.rpcDuration.Record(context.TODO(), int64(rs.EndTime.Sub(rs.BeginTime)), metric.WithAttributes(metricAttrs...))
```
pellared, line 302: Shouldn't we use the `ctx` passed to `handleRPC`?

fatsheep9146: I have tried to use `ctx`, but it won't work, I'm still digging the reason.

instrumentation/google.golang.org/grpc/otelgrpc/test/grpc_stats_handler_test.go
fatsheep9146: Ok, I will try

instrumentation/google.golang.org/grpc/otelgrpc/stats_handler.go
```diff
@@ -33,98 +36,204 @@ type gRPCContextKey struct{}
...
+	}
+
+	h.r.rpcResponseSize, err = h.r.meter.Int64Histogram("rpc.server.response.size",
+		metric.WithDescription("Measures size of RPC response messages (uncompressed)."),
+		metric.WithUnit("By"))
+	if err != nil {
+		otel.Handle(err)
+	}
+
+	h.r.rpcRequestsPerRPC, err = h.r.meter.Int64Histogram("rpc.server.requests_per_rpc",
```
ktong, line 82: Curious, why does here use Histogram instead of Counter?

dashpole: It is following the specification: https://opentelemetry.io/docs/specs/otel/metrics/semantic_conventions/rpc-metrics/#rpc-server

instrumentation/google.golang.org/grpc/otelgrpc/stats_handler.go
```diff
@@ -33,98 +36,204 @@ type gRPCContextKey struct{}
...
+	h.r.tracer = h.config.TracerProvider.Tracer(
 		instrumentationName,
 		trace.WithInstrumentationVersion(SemVersion()),
 	)
+	h.r.meter = h.config.MeterProvider.Meter(
+		instrumentationName,
+		metric.WithInstrumentationVersion(Version()),
+		metric.WithSchemaURL(semconv.SchemaURL),
+	)
+	var err error
```
dashpole, line 60: Should we return the error here so users can handle it?

instrumentation/google.golang.org/grpc/otelgrpc/stats_handler.go
```diff
@@ -33,98 +36,204 @@ type gRPCContextKey struct{}
...
+
 type serverHandler struct {
 	*config
-	tracer trace.Tracer
+	r *handler
 }
 
 // TagRPC can attach some information to the given context.
 func (h *serverHandler) TagRPC(ctx context.Context, info *stats.RPCTagInfo) context.Context {
+	// fmt.Printf("server TagRPC, ctx.Err %v\n", ctx.Err())
```
dashpole, line 116: nit: remove commented out code

dashpole: Below as well

instrumentation/google.golang.org/grpc/otelgrpc/stats_handler.go
Sovietaced: IIRC the interceptor had the same behavior. IMO it feels like an error should be returned here and the logger should be used for steady state failures.

instrumentation/google.golang.org/grpc/otelgrpc/stats_handler.go
```diff
@@ -33,98 +36,204 @@ type gRPCContextKey struct{}
...
 		instrumentationName,
 		trace.WithInstrumentationVersion(SemVersion()),
 	)
+	h.r.meter = h.config.MeterProvider.Meter(
+		instrumentationName,
+		metric.WithInstrumentationVersion(Version()),
+		metric.WithSchemaURL(semconv.SchemaURL),
+	)
+	var err error
+	h.r.rpcDuration, err = h.r.meter.Int64Histogram("rpc.server.duration",
```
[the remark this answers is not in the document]
Sovietaced: I like this idea, record data with the highest accuracy possible.

instrumentation/google.golang.org/grpc/otelgrpc/test/grpc_stats_handler_test.go
```diff
@@ -579,3 +588,49 @@ func checkServerSpans(t *testing.T, spans []trace.ReadOnlySpan) {
...
+func checkServerRecords(t *testing.T, reader metric.Reader) {
+	rm := metricdata.ResourceMetrics{}
+	err := reader.Collect(context.Background(), &rm)
+	assert.NoError(t, err)
+	require.Len(t, rm.ScopeMetrics, 1)
+	require.Len(t, rm.ScopeMetrics[0].Metrics, 5)
+	for _, m := range rm.ScopeMetrics[0].Metrics {
+		require.IsType(t, m.Data, metricdata.Histogram[int64]{})
+		data := m.Data.(metricdata.Histogram[int64])
+		for _, dpt := range data.DataPoints {
```
Sovietaced, line 601: Should buckets be validated to ensure the correct delta of time has been calculated?

Sovietaced: Oh nevermind, I guess we don't have control of the timestamps or don't have pluggable clocks for testing.

instrumentation/google.golang.org/grpc/otelgrpc/stats_handler.go
```diff
@@ -173,14 +284,27 @@ func handleRPC(ctx context.Context, rs stats.RPCStats) {
...
-			span.SetAttributes(statusCodeAttr(s.Code()))
+			rpcStatusAttr = semconv.RPCGRPCStatusCodeKey.Int(int(s.Code()))
 		} else {
-			span.SetAttributes(statusCodeAttr(grpc_codes.OK))
+			rpcStatusAttr = semconv.RPCGRPCStatusCodeKey.Int(int(grpc_codes.OK))
 		}
+		span.SetAttributes(rpcStatusAttr)
 		span.End()
+
+		metricAttrs = append(metricAttrs, rpcStatusAttr)
```
Sovietaced, line 301: Glad to see this make it in

instrumentation/google.golang.org/grpc/otelgrpc/stats_handler.go
fatsheep9146: I found the cause. 

The reason is that, when we add instrument in `HandleRPC` function, when the `rs stats.RPCStats` is type of `*stats.End` and `*stats.OutTrailer`, the context we get is already cancelled by in grpc-go sdk. 

https://github.com/grpc/grpc-go/blob/6e9c88b0acf15de2256c3d3a257869e0669572d9/internal/transport/http2_server.go#L1277

```go
func (t *http2Server) finishStream(s *Stream, rst bool, rstCode http2.ErrCode, hdr *headerFrame, eosReceived bool) {
	// In case stream sending and receiving are invoked in separate
	// goroutines (e.g., bi-directional streaming), cancel needs to be
	// called to interrupt the potential blocking on other goroutines.
	s.cancel()

	oldState := s.swapState(streamDone)
```

And the metric sdk will check if ctx.Err() is nil when record the measurement
https://github.com/open-telemetry/opentelemetry-go/blob/c047088605b454f172765621d53107f80b1e6417/sdk/metric/instrument.go#L205
```
func (i *int64Inst) aggregate(ctx context.Context, val int64, s attribute.Set) { // nolint:revive  // okay to shadow pkg with method.
	if err := ctx.Err(); err != nil {
		return
	}
	for _, in := range i.measures {
		in(ctx, val, s)
	}
}
```

So the aggregate function will fail, the measurement will also be dropped.

So we should 1. use new context object. 2. add the measurement before `*stats.End` and `*stats.OutTrailer`

But some metric can only be measured when it comes to `*stats.End` and `*stats.OutTrailer`, for example duration, requests per rpc and responses per rpc. (Since in streaming model, we can only know the final count of requests and response when it comes to `*stats.End`). So I vote for the solution 1.

pellared: > And the metric sdk will check if ctx.Err() is nil when record the measurement
https://github.com/open-telemetry/opentelemetry-go/blob/c047088605b454f172765621d53107f80b1e6417/sdk/metric/instrument.go#L205

I am not convinced that this what SDK should be doing. @MrAlias WDYT?

pellared:
@fatsheep9146, will you still be able to work on this PR?

instrumentation/google.golang.org/grpc/otelgrpc/stats_handler.go
pellared: I created https://github.com/open-telemetry/opentelemetry-go/pull/4671

fatsheep9146:
> @fatsheep9146, will you still be able to work on this PR?

Sorry, last week I'm busy my wedding. I will continue to work on this issue today. @pellared

instrumentation/google.golang.org/grpc/otelgrpc/stats_handler.go
pellared: This would be a better solution: https://github.com/open-telemetry/opentelemetry-go/pull/4671#issuecomment-1783534940

instrumentation/google.golang.org/grpc/otelgrpc/test/grpc_stats_handler_test.go
pellared: You may also look at https://github.com/open-telemetry/opentelemetry-go-contrib/issues/4483

instrumentation/google.golang.org/grpc/otelgrpc/stats_handler.go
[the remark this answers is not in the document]
fatsheep9146: Done

instrumentation/google.golang.org/grpc/otelgrpc/test/grpc_stats_handler_test.go
fatsheep9146: I already change unit test by using `metricdatatest` package.

I use `AssertEqual` to compare each `metricdata.Metrics` object not `metricdata.ScopeMetrics`, since the we could not know the value of `rpc.xxx.duration` in advance. 

So for `rpc.xxx.duration`, we should use both `metricdatatest.IgnoreTimestamp()`, `metricdatatest.IgnoreValue()`. But for other metric, we could only use `metricdatatest.IgnoreTimestamp()`.

@pellared Could you help look at the code to make sure I use this package correctly?

instrumentation/google.golang.org/grpc/otelgrpc/stats_handler.go
```diff
@@ -33,24 +36,90 @@ type gRPCContextKey struct{}
...
+
+type handler struct {
+	tracer             trace.Tracer
+	meter              metric.Meter
+	rpcDuration        metric.Float64Histogram
+	rpcRequestSize     metric.Int64Histogram
+	rpcResponseSize    metric.Int64Histogram
+	rpcRequestsPerRPC  metric.Int64Histogram
+	rpcResponsesPerRPC metric.Int64Histogram
+}
```
pellared, line 50: Why do we even need this handler?

Can we put all the instruments, providers, meter and tracer in `config` and initialize everything in  `newConfig`?

instrumentation/google.golang.org/grpc/otelgrpc/stats_handler.go
```diff
@@ -172,15 +280,25 @@ func handleRPC(ctx context.Context, rs stats.RPCStats) {
...
-			span.SetAttributes(statusCodeAttr(grpc_codes.OK))
+			rpcStatusAttr = semconv.RPCGRPCStatusCodeKey.Int(int(grpc_codes.OK))
 		}
+		span.SetAttributes(rpcStatusAttr)
 		span.End()
+
+		metricAttrs = append(metricAttrs, rpcStatusAttr)
+		r.rpcDuration.Record(context.TODO(), float64(rs.EndTime.Sub(rs.BeginTime)), metric.WithAttributes(metricAttrs...))
+		r.rpcRequestsPerRPC.Record(context.TODO(), gctx.messagesReceived, metric.WithAttributes(metricAttrs...))
+		r.rpcResponsesPerRPC.Record(context.TODO(), gctx.messagesSent, metric.WithAttributes(metricAttrs...))
```
pellared, line 300: This would be a proper solution: https://github.com/open-telemetry/opentelemetry-go/pull/4671#issuecomment-1783534940

instrumentation/google.golang.org/grpc/otelgrpc/test/grpc_stats_handler_test.go
pellared: I simply suggest to assert against `metricdata.ScopeMetrics` and use `metricdatatest.IgnoreValue()`.

1. I do not see a lot of value in asserting the exact metrics values.
2. Checking the histograms' "values" (buckets) produces a lot of boilerplate which IMO is hard to decipher. 
3. Asserting against whole `metricdata.ScopeMetrics` makes sure that we have not forgotten to assert any metric
4. The most important things to assert are the metrics and their attributes so that we can easily spot cardinality metric attributes.

pellared (commented):
I think we are very close 😉

instrumentation/google.golang.org/grpc/otelgrpc/test/grpc_stats_handler_test.go
fatsheep9146: Ok, this makes sense.

fatsheep9146: Done

instrumentation/google.golang.org/grpc/otelgrpc/stats_handler.go
fatsheep9146: Ok, I will change like this.

dashpole (approved):

instrumentation/google.golang.org/grpc/otelgrpc/stats_handler.go
fatsheep9146: Yes, we can. 

To be honest, I also consider putting every into config at first. But, then I found there already exists a metric `rpcServerDuration` which is already used in the implementation of `interceptor` and conflicts with the name `rpcDuration` we use for `stats.Handler`.  If we put them together, I thought maybe confused for users. So I add a new `handler` struct.

So how about we put all those five metric `rpcDuration`... into config, and keep the original `rpcServerDuration`, just like following

```
type config struct {
	Filter           Filter
	Propagators      propagation.TextMapPropagator
	TracerProvider   trace.TracerProvider
	MeterProvider    metric.MeterProvider
	SpanStartOptions []trace.SpanStartOption

	ReceivedEvent bool
	SentEvent     bool

	meter             metric.Meter
	rpcServerDuration metric.Int64Histogram

	rpcDuration        metric.Float64Histogram
	rpcRequestSize     metric.Int64Histogram
	rpcResponseSize    metric.Int64Histogram
	rpcRequestsPerRPC  metric.Int64Histogram
	rpcResponsesPerRPC metric.Int64Histogram
}
```

When the implementation of interceptor is deprecated, the we remove the old `rpcServerDuration` metric?  WDYT?
@pellared

pellared: How about updating the `interceptor` to use the new `rpcDuration`?

fatsheep9146: > How about updating the `interceptor` to use the new `rpcDuration`?

That works for me, I will try to work like that

fatsheep9146: Done

instrumentation/google.golang.org/grpc/otelgrpc/stats_handler.go
fatsheep9146: Done @pellared

instrumentation/google.golang.org/grpc/otelgrpc/stats_handler.go
```diff
@@ -33,24 +36,33 @@ type gRPCContextKey struct{}
...
 
-type serverHandler struct {
-	*config
-	tracer trace.Tracer
+// TagConn can attach some information to the given context.
+func (h *serverHandler) TagConn(ctx context.Context, info *stats.ConnTagInfo) context.Context {
+	span := trace.SpanFromContext(ctx)
+	attrs := peerAttr(peerFromCtx(ctx))
+	span.SetAttributes(attrs...)
+
```
pellared, line 60:
```suggestion
```

pellared (approved):
Thank you again for your contribution 🎉

Sovietaced:
@pellared when do you think a release will be cut? 

I've been waiting on this or https://github.com/open-telemetry/opentelemetry-go-contrib/pull/4356 for a couple months now

pellared:
> @pellared when do you think a release will be cut?
> 
> I've been waiting on this or #4356 for a couple months now

I hope that it is a matter of at most two weeks. See
- https://github.com/open-telemetry/opentelemetry-go-contrib/pulls?q=is%3Aopen+is%3Apr+milestone%3Av0.46.0
- https://github.com/open-telemetry/opentelemetry-go/pulls?q=is%3Aopen+is%3Apr+milestone%3Av1.20.0

Issue #4373, opened 2023-10-02
configyaml: Parse model from YAML


pellared:
The config parsers could be located in child modules to reduce the number of dependencies (not everyone wants to take e.g. yaml.v3 as a dependency). We can have `config/configyaml` module with exported API like following:

```go
package configyaml

func Parse(r io.Reader) (Config, error) {}
```

_Based on discussion in https://github.com/open-telemetry/opentelemetry-go-contrib/pull/4228#discussion_r1339824254_

pellared:
PTAL @codeboten

carrbs:
@codeboten , @pellared should I implement this, (and remove the comment about JSON?), since we are likely just implementing for YAML atm?

https://github.com/open-telemetry/opentelemetry-go-contrib/blob/96790b3addf68bdceb23a44d8f080085bfb82936/config/config.go#L116-L118

pellared:
Works for me but I AM leaving the decision to @codeboten

codeboten:
i think thats a sensible way forward

carrbs:
Spent some more time looking into this and I have questions on what the scope of this issue is. I'm new to otel, so hopefully you can give me some grace if there are obvious oversights or overly simple questions.

I can see that this repo is leveraging the "_official_" [opentelemetry specification schema](https://github.com/open-telemetry/opentelemetry-configuration/blob/main/schema/opentelemetry_configuration.json) to render the [generated_config.go](https://github.com/open-telemetry/opentelemetry-go-contrib/blob/main/config/generated_config.go) file.

1. Should the `Parse(config_file)` function utilize this generated config for validation?
2. Is the purpose to ensure that _all_ known properties from the generated config map from the input YAML to the expected structs in the generated config file, or just to focus on the environment variable handling?
3. If the YAML input file contains mapping unknown to the schema, how is that handled? 
    - The examples describing the nuance around environment variables are not part of the schema, and I also saw the `"additionalProperties": true` on the schema ☝️.
    
Any thoughts on this would be awesome, I'd be happy to help on this but if it's not high priority we can table it :)

MadVikingGod:
The current generated config uses the `mapstructure` tags.  This means that parsing yaml (and json) is done by marshaling raw yaml into a `map[string]interface{}` and then using the `mapstructure` package to `Decode()` the map into the actual config.

I think it might be best if we have an example of this.

pellared:
[...]

PR #4376, opened 2023-10-03
[config] introduce config package

codeboten:
This change introduces the config package with the following changes:

- add go-jsonschema tool to tools.go
- add `make genjsonschema` target to Makefile
- run the genjsonschema target to generate the code from the opentelemetry configuration schema
- add a readme for the package
- run dependabot-generate

This is a scoped down version of https://github.com/open-telemetry/opentelemetry-go-contrib/pull/4228

Note that the generated model code lives in a package that includes the version of the schema in the package name. I'm not sure if that's the right thing to do, but it is similar to what is done for semconv.


mx-psi (approved):

MadVikingGod (approved):

config/README.md
```diff
@@ -0,0 +1,31 @@
+# Configuration Library
+
+This package can be used to parse a configuration file that follows the JSON
+Schema defined by the [OpenTelemetry Configuration] schema.
+
+The package contains:
+
+- models generated via the JSON schema using the [go-jsonschema] library
+- a `Create` function that interprets [configuration model] and return SDK components (TODO)
+- a `Parse` function that parses and validates a [configuration file] (TODO)
```
pellared, line 10: I am not sure if we need this. We could simply provide examples e.g. how we could parse using packages like `json`, `yaml`.

config/README.md
```diff
@@ -0,0 +1,31 @@
+# Configuration Library
+
+This package can be used to parse a configuration file that follows the JSON
+Schema defined by the [OpenTelemetry Configuration] schema.
+
+The package contains:
+
+- models generated via the JSON schema using the [go-jsonschema] library
```
pellared, line 8: Are we not we concerned that we are coupling the model to the JSON, YAML parsing libraries?
Reference: https://www.reddit.com/r/golang/comments/rtne10/good_antipattern_blog_post/

Makefile
```diff
@@ -283,3 +286,33 @@ COMMIT ?= "HEAD"
...
+
+# Generate structs for configuration from opentelemetry-configuration schema
+genjsonschema: genjsonschema-cleanup $(GOJSONSCHEMA)
+	mkdir -p ${OPENTELEMETRY_CONFIGURATION_JSONSCHEMA_SRC_DIR}
+	curl -sSL https://api.github.com/repos/open-telemetry/opentelemetry-configuration/tarball/${OPENTELEMETRY_CONFIGURATION_JSONSCHEMA_VERSION} | tar xz --strip 1 -C ${OPENTELEMETRY_CONFIGURATION_JSONSCHEMA_SRC_DIR}
+	$(GOJSONSCHEMA) \
+		--capitalization ID \
+		--capitalization OTLP \
+		--struct-name-from-title \
+		--package config \
```
pellared, line 310: We could name the package `configmapstructure` if this would be a package for that would help parsing the model using `mapstructure`. Reference: https://github.com/open-telemetry/opentelemetry-go-contrib/issues/4373

tools/go.mod
```diff
@@ -1,10 +1,13 @@
 module go.opentelemetry.io/contrib/tools
 
-go 1.20
+go 1.21
+
+toolchain go1.21.1
```
pellared, line 5: Is the intention of this line to use the newest compiler? 

If so then we could

```suggestion
toolchain go1.21.2
```

pellared (dismissed):
> Note that the generated model code lives in a package that includes the version of the schema in the package name.
> I'm not sure if that's the right thing to do, but it is similar to what is done for semconv.

With semconv we intented to support all kind of versions. For the configuration, I suppose it is OK to support the latest configuration schema version. Therefore, I do not think we should include a version in the package name. Similar PR: https://github.com/open-telemetry/opentelemetry-go/pull/4538

Makefile
```diff
@@ -283,3 +286,33 @@ COMMIT ?= "HEAD"
...
+OPENTELEMETRY_CONFIGURATION_JSONSCHEMA_SRC_DIR=tmp/opentelememetry-configuration
+
+# The SHA matching the current version of the opentelemetry-configuration schema to use
+OPENTELEMETRY_CONFIGURATION_JSONSCHEMA_VERSION=v0.1.0
+
+# Cleanup temporary directory
+genjsonschema-cleanup:
+	rm -Rf ${OPENTELEMETRY_CONFIGURATION_JSONSCHEMA_SRC_DIR}
+
+GENERATED_CONFIG=./config/pkg/${OPENTELEMETRY_CONFIGURATION_JSONSCHEMA_VERSION}/generated_config.go
```
pellared, line 300: > Note that the generated model code lives in a package that includes the version of the schema in the package name.
I'm not sure if that's the right thing to do, but it is similar to what is done for semconv.

With semconv we indented to support all kind of versions. For the configuration, I suppose it is OK to support the latest configuration schema version. Therefore, I do not think we should include a version in the package name. Similar PR: https://github.com/open-telemetry/opentelemetry-go/pull/4538

```suggestion
GENERATED_CONFIG=./config/model.go
```

config/README.md
pellared: I think we will simply couple this package to some concrete library to support parsing.

Makefile
pellared: I think we will simply couple this package to some concrete library.

pellared (approved):

codeboten:
@pellared well i tried removing the `toolchain` directive and it caused a lint failure: https://github.com/open-telemetry/opentelemetry-go-contrib/actions/runs/6472155249/job/17572209023?pr=4376

I've re-added it, though since it's automatically added I didn't update it to 1.21.2. Appreciate any thoughts/advice here

PR #4401, opened 2023-10-09
feat(handler): support custom carriers

omar25h:
This allows overriding the carrier where the incoming HTTP request does not contain the trace context in the header, but rather in the request body. This is especially relevant for environments where the request received is not under the control of the applications such as Pub/Sub push subscription requests.

### DIFF
--- a/instrumentation/net/http/otelhttp/config.go
+++ b/instrumentation/net/http/otelhttp/config.go
@@ -36,6 +36,7 @@ type config struct {
 	Tracer            trace.Tracer
 	Meter             metric.Meter
 	Propagators       propagation.TextMapPropagator
+	CarrierFn         func(*http.Request) propagation.TextMapCarrier
 	SpanStartOptions  []trace.SpanStartOption
 	PublicEndpoint    bool
 	PublicEndpointFn  func(*http.Request) bool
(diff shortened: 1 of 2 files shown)


pellared:
The PR lacks description (or even an issue). I am not sure if allowing custom carriers (other than the idiomatic HTTP Header) is something that we should support in OTel HTTP instrumentation.

pellared:
>  request received is not under the control of the applications such as Pub/Sub push subscription requests.

I think the description lacks information. The HTTP header is part of the HTTP request and response. Maybe it is a limitation of some library that just operates on the HTTP request/response body and the libraries should be refined instead?

omar25h:
> > request received is not under the control of the applications such as Pub/Sub push subscription requests.
> 
> I think the description lacks information. The HTTP header is part of the HTTP request and response. Maybe it is a limitation of some library that just operates on the HTTP request/response body and the libraries should be refined instead?

That is correct. The problem we are trying to solve is that Pub/Sub does not allow one to pass in extra headers, so the way to propagate trace information is via message attributes  or message payload, where they exist in the request body

pellared:
>  The problem we are trying to solve is that Pub/Sub does not allow one to pass in extra headers,

I suggest tackling this issue 😉 This is in HTTP design to use headers for adding additional information with an HTTP request or response.

> HTTP headers let the client and the server pass additional information with an HTTP request or response.

From https://developer.mozilla.org/en-US/docs/Web/HTTP/Headers

omar25h:
Pub/Sub has a solution for this apparently in the subscription configuration, so a solution is already available to the problem. Feel free to close the merge request if you think it is not something the library should allow.

Thanks!

Issue #4371, opened 2023-10-02
config: Create SDK from model


pellared:
Add a function which creates SDK from model.

E.g. 

```go
package config

// NewSDK creates SDK providers based on the configuration model.
func NewSDK(ctx, context.Context, c *OpenTelemetryConfiguration) (*SDK, error) {}

type SDK struct {}
func (sdk *SDK) Shutdown() error {}
func (sdk *SDK) TraceProvider() *trace.TraceProvider {}
func (sdk *SDK) MeterProvider() *meter.MeterProvider {}
```

pellared:
PTAL @codeboten

codeboten:
@pellared already started on this, please assign it to me when you have a moment

PR #4414, opened 2023-10-11
[config] add NewSDK func

codeboten:
For now it's returning a noop tracer/meter provider if no configuration is found. Otherwise it returns an unconfigured SDK meter/tracer provider.

This PR is mostly to start putting the pieces together for the interface. For reviewers: I decided to go w/ passing in options rather than specific parameters into NewSDK.

Related to https://github.com/open-telemetry/opentelemetry-go-contrib/issues/4371


config/config_test.go
```diff
@@ -0,0 +1,50 @@
...
+			name:           "no-configuration",
+			tracerProvider: trace.NewNoopTracerProvider(),
+			meterProvider:  noop.NewMeterProvider(),
+		},
+		{
+			name: "with-configuration",
+			cfg: []ConfigurationOption{
+				WithContext(context.Background()),
+				WithOpenTelemetryConfiguration(OpenTelemetryConfiguration{
+					TracerProvider: &TracerProvider{},
```
pellared, line 36: Side note:

The OTel Go SDK misses "introspection" API 😞 
How do you plan to test more complex setup (e.g. with that an exporter)?
Should we add "accessors" (getters) to the Go SDK?  Do you plan to use `reflect`? Other ideas?

Reference: https://blog.ploeh.dk/2013/04/04/structural-inspection/

config/config_test.go
```diff
@@ -0,0 +1,50 @@
...
+	"go.opentelemetry.io/otel/trace"
+)
+
+func TestNewSDK(t *testing.T) {
+	tests := []struct {
+		name           string
+		cfg            []ConfigurationOption
+		tracerProvider any
+		meterProvider  any
+		err            error
```
pellared, line 24:
```suggestion
		wantTracerProvider any
		wantMeterProvider  any
		wantErr            error
```

config/config_test.go
```diff
@@ -0,0 +1,50 @@
...
+			},
+			tracerProvider: &sdktrace.TracerProvider{},
+			meterProvider:  &sdkmetric.MeterProvider{},
+		},
+	}
+	for _, tt := range tests {
+		sdk, err := NewSDK(tt.cfg...)
+		require.Equal(t, tt.err, err)
+		require.IsType(t, tt.tracerProvider, sdk.TracerProvider())
+		require.IsType(t, tt.meterProvider, sdk.MeterProvider())
```
pellared, line 48: I think we can do all assertions (no need to fail fast)

```suggestion
		assert.IsType(t, tt.tracerProvider, sdk.TracerProvider())
		assert.IsType(t, tt.meterProvider, sdk.MeterProvider())
```

config/config.go
```diff
@@ -0,0 +1,74 @@
...
+}
+
+// WithOpenTelemetryConfiguration sets the OpenTelemetryConfiguration used
+// to produce the SDK.
+func WithOpenTelemetryConfiguration(cfg OpenTelemetryConfiguration) ConfigurationOption {
+	return configurationOptionFunc(func(c configOptions) configOptions {
+		c.opentelemetryConfig = cfg
+		return c
+	})
+}
```
pellared, line 74: Shouldn't `cfg OpenTelemetryConfiguration` be a required parameter?

pellared (commented):
Overall looks good 👍

config/config_test.go
codeboten: I was going to ask the same thing. Another option would be to test providers with full end-to-end setup, not super easy but at least then we'll know for sure the components work as intended. This could get somewhat tricky for testing all the permutations of options though

config/config.go
codeboten: It could definitely be, i suppose not setting it would be useful for someone who wants a noop SDK... not sure if that's a realistic use-case, but i figured it would be easier to be consistent w/ all options.

pellared: >  i suppose not setting it would be useful for someone who wants a noop SDK... 

Interesting use case. For me you can leave it as it is. You can resolve the comment or ask/wait for other opinions.

config/config_test.go
pellared: Personally, I think it would be good to add  "introspection" API  (add getters and export some types) in OTel Go SDK (if the SIG agrees that it would be beneficial).
Before it is done, the probably most efficient way to avoid rewriting all tests in future would be to use `reflect` to access private fields and types. Later the `reflect` could be changed to use the newly exported API (getters and types).

We can also discuss it further during SIG meeting.

pellared: I created https://github.com/open-telemetry/opentelemetry-go/issues/4606. Resolving this conversation.

config/config_test.go
codeboten: renamed 👍🏻

pellared:
Please add a changelog entry 😉

pellared (approved):

codeboten:
The failing tests appear unrelated...

pellared:
> The failing tests appear unrelated...

True. I created https://github.com/open-telemetry/opentelemetry-go-contrib/issues/4416

CHANGELOG.md
```diff
@@ -12,7 +12,8 @@ This project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0.htm
 
 - Add the new `go.opentelemetry.io/contrib/instrgen` package to provide auto-generated source code instrumentation. (#3068, #3108)
 - Add `"go.opentelemetry.io/contrib/samplers/jaegerremote".WithSamplingStrategyFetcher` which sets custom fetcher implementation. (#4045)
-- Add `"go.opentelemetry.io/contrib/config"` package that includes configuration models generated via go-jsonschema (#4376)
+- Add `"go.opentelemetry.io/contrib/config"` package that includes configuration models generated via go-jsonschema. (#4376)
+- Add `NewSDK` method to `"go.opentelemetry.io/contrib/config"`. The initial implementation only returns noop providers. (#4414)
```
pellared, line 16: nit 😉 

```suggestion
- Add `NewSDK` function to `"go.opentelemetry.io/contrib/config"`. The initial implementation only returns noop providers. (#4414)
```

pellared (approved):

codeboten:
Anything else needed for this PR before merging?

Issue #4419, opened 2023-10-12
[otelecho] Add option to skip calling global handler


rattuscz:
### Problem Statement

I want to be able to control whether otelecho middleware calls the global error handler or not.

Using multiple middlewares that cannot be configured this way the global error handler gets called multiple times for the exact same error. Handling this multiple processing of same error on global error handler seems to me as wrong way to go.

The only way to prevent this is to eat the error before this middleware is run after request is handled, which then does not append the error to span as attribute.

I believe this also goes against best practice of either handling the error or propagating, not both.

### Proposed Solution

Add new option to middleware config to control this behavior, with default value resulting in same behavior as it is now to be BC.

#### Prior Art

In echo request logger middleware there is the exact same setting that controls calling of the global handler:
https://github.com/labstack/echo/blob/98a523756d875bc13475bcb6237f09e771cbe321/middleware/request_logger.go#L115

erkanzileli:
I'm having the same issue. You're right. 
The middleware calls the error handler and returns the error. When the middleware returns an error, echo calls the error handler again. so in this way, the error handler gets called twice. This results in a duplicated response body like below.

```json
{
    "message": "content"
}
{
    "message": "content"
}
```

MrAlias:
Instead of modifying the instrumentation, I would recommend you register a rate-limiting error handler. You can customize the behavior of how errors are handle however you want by using your own error handler. Adding to the API surface error of the instrumentation does not seem like a prudent way to solve an issue that can already be solved with a custom error handler.

pellared:
@rattuscz See https://github.com/open-telemetry/opentelemetry-go-contrib/pull/4420#issuecomment-1813097682 by @MrAlias:

> I'm not opposed to this.

Also we agree that it is acceptable (breaking) change as the Go module is not stable. I think that the users would like this change.

rattuscz:
As long as it's addressed somehow I'm happy :-)

shonigbaum:
Hey,
any news on this? We are facing the same issue. Will this be fixed soon? Or is there a workaround recommended?

dmathieu:
otelecho is lacking an owner, is deprecated and will be removed in a couple months.
https://github.com/open-telemetry/opentelemetry-go-contrib/issues/5550

Resolving this would be a good way for someone to step up before becoming an owner.

shonigbaum:
[...]

PR #4420, opened 2023-10-12
[WIP] otelecho: config option to skip global error handler call

rattuscz:
First try at implementing #4419

### DIFF
--- a/instrumentation/github.com/labstack/echo/otelecho/config.go
+++ b/instrumentation/github.com/labstack/echo/otelecho/config.go
@@ -26,6 +26,7 @@ type config struct {
 	TracerProvider oteltrace.TracerProvider
 	Propagators    propagation.TextMapPropagator
 	Skipper        middleware.Skipper
+	HandleError    bool
 }
 
 // Option specifies instrumentation configuration options.
@@ -64,5 +65,12 @@ func WithTracerProvider(provider oteltrace.TracerProvider) Option {
 func WithSkipper(skipper middleware.Skipper) Option {
 	return optionFunc(func(cfg *config) {
 		cfg.Skipper = skipper
+	})
+}
+
+// WithoutHandleError instructs middleware to not call global error handler when next middleware/handler returns an error.
+func WithoutHandleError() Option {
+	return optionFunc(func(cfg *config) {
+		cfg.HandleError = false
 	})
 }
--- a/instrumentation/github.com/labstack/echo/otelecho/echo.go
+++ b/instrumentation/github.com/labstack/echo/otelecho/echo.go
@@ -36,7 +36,9 @@ const (
 
 // Middleware returns echo middleware which will trace incoming requests.
 func Middleware(service string, opts ...Option) echo.MiddlewareFunc {
-	cfg := config{}
+	cfg := config{
+		HandleError: true,
+	}
 	for _, opt := range opts {
 		opt.apply(&cfg)
 	}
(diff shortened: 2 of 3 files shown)


MrAlias:
https://github.com/open-telemetry/opentelemetry-go-contrib/issues/4419#issuecomment-1813039942

> Instead of modifying the instrumentation, I would recommend you register a rate-limiting error handler. You can customize the behavior of how errors are handle however you want by using your own error handler. Adding to the API surface error of the instrumentation does not seem like a prudent way to solve an issue that can already be solved with a custom error handler.

pellared:
To be honest, I do not understand why `c.Error(err)` is even called given that the `err` is returned. Shouldn't we simply remove this line?

PS. I never used echo.

MrAlias:
> To be honest, I do not understand why `c.Error(err)` is even called given that the `err` is returned by the middleware. Shouldn't we simply remove this line?
> 
> PS. I never used echo.

I'm not opposed to this.

hanyuancheung (approved):
This PR SGTM

hanyuancheung:
@rattuscz Please do not forget to add a changelog.

pellared:
> Anyway it will be BC break 

It is not a stable Go module so breaking change are acceptable if there are good reasons to do so.

rattuscz:
Okie, changed this to implement the removal.
I've once again today run into issue with this so I hope this is merged & released soon :-)

CHANGELOG.md
```diff
@@ -12,6 +12,9 @@ This project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0.htm
 
 - Add the new `go.opentelemetry.io/contrib/instrgen` package to provide auto-generated source code instrumentation. (#3068, #3108)
 
+### Changed
+- **BC break** `go.opentelemetry.io/contrib/instrumentation/github.com/labstack/echo/otelecho` remove calling global error handler (#4419, #4420)
```
```suggestion
- Remove call to global error handler in `go.opentelemetry.io/contrib/instrumentation/github.com/labstack/echo/otelecho` (#4420)
```

CHANGELOG.md
```diff
@@ -38,6 +38,9 @@ This project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0.htm
 
 - Upgrade dependencies of OpenTelemetry Go to use the new [`v1.21.0`/`v0.44.0` release](https://github.com/open-telemetry/opentelemetry-go/releases/tag/v1.21.0). (#4582)
 
+### Changed
+- Remove call to global error handler in `go.opentelemetry.io/contrib/instrumentation/github.com/labstack/echo/otelecho` (#4420)
```
dmathieu, line 42: This is also improperly placed in the changelog. It needs to be in the `unreleased` section.

dmathieu (commented):
Tests are failing.

dmathieu:
Well, it looks like the `TestStatusError` test needs to be updated to properly reflect those changes.

rattuscz:
> Well, it looks like the `TestStatusError` test needs to be updated to properly reflect those changes.

Sure I can fix the tests, if that's the way to go.
Not really sure if there is some better way how to avoid those problems, because it will still be quite easy to have it wrongly setup just because of order of middlewares/logger/error handler.

In our projects we tend to have logger as the last middleware and eating the error returned from handler functions so the response is already set, error logged/handled and all is great.
Others may have different order and suffer breaks when we change this.

Will leave it here for a few days before commiting

asp3cto:
Still got this error, and well, it's quite difficult to catch because of middleware debugging issues. In echo style, the handlers and middlewares should not call c.Error explicitly, and refactoring whole project just to fit the handling logic in this middleware is not great. Any plans to fix?

dmathieu:
This PR has had no response by its owner for 2 years. I think it's safe to say that you can take its commits, merge with the main branch and open a new one.

asp3cto:
@dmathieu can this be done as a breaking change or just an option with current implementation by default?

dmathieu:
The package isn't stable. So it can be a breaking change.

asp3cto:
@dmathieu Decided to make it an option with current logic by default, see #8025

PR #4448, opened 2023-10-20
Expose instrumentation scope name

utezduyar:
It is good to know the scope to apply view filters. For example if we would like to drop all or some of the meters from this library, below code is how the filtering is applied. The scope name is used explicitly and a change in the scope name by the runtime instrumentation library will break applications. 

```
metricsdk.Instrument{Name: "*",
Scope: instrumentation.Scope{
	Name:    "go.opentelemetry.io/contrib/instrumentation/runtime",
}},
metricsdk.Stream{Aggregation: metricsdk.AggregationDrop{}},
```

Instead of this, get the scope name from the library itself

```
metricsdk.Instrument{Name: "*",
Scope: instrumentation.Scope{
	Name:    runtime.Name(),
}},
metricsdk.Stream{Aggregation: metricsdk.AggregationDrop{}},
```


utezduyar:
@pellared How does it look now?

instrumentation/runtime/runtime.go
```diff
@@ -95,6 +95,11 @@ func newConfig(opts ...Option) config {
 	return c
 }
 
+// Name returns the scope name of the library
```
pellared, line 98:
```suggestion
// Name returns the instrumentation scope name.
```

pellared (commented):
> How does it look now?

Now, I find it clear 👍 

I will first change the the title and ask others if they like the proposal.

If the proposal would be accepted then:

1. We would have to do this for all instrumentation libraries.
2. Changelog would have to be updated.

pellared:
@open-telemetry/go-approvers PTAL

Give 👍 if you like the proposal.
Give 👎 if you are against.

Additionally more detailed (text) feedback is welcome.

instrumentation/runtime/runtime.go
```diff
@@ -95,6 +95,11 @@ func newConfig(opts ...Option) config {
 	return c
 }
 
+// Name returns the instrumentation scope name.
+func Name() string {
+	return "go.opentelemetry.io/contrib/instrumentation/runtime"
```
dashpole, line 100: Would a `const` be better/simpler?

utezduyar: Sure. If the proposal passes, I will make the change.

Aneurysm9: If we're going to do this I agree it should be a `const` rather than a function and it needs to be done for all instrumentation.  The name should also be more descriptive, such as `ScopeName`.

pellared: @utezduyar Please update this PR. I will try to bring this PR during next SIG meeting.

instrumentation/runtime/runtime.go
```diff
@@ -24,6 +24,9 @@ import (
 	"go.opentelemetry.io/otel/metric"
 )
 
+// ScopeName the instrumentation scope name for identifying meters.
```
pellared, line 27:
```suggestion
// ScopeName is the instrumentation scope name.
```

utezduyar:
> 1. We would have to do this for all instrumentation libraries.

What other libraries are you aware of? I can get to them too.

pellared:
> > 1. We would have to do this for all instrumentation libraries.
> 
> What other libraries are you aware of? I can get to them too.

All under https://github.com/open-telemetry/opentelemetry-go-contrib/tree/main/instrumentation

utezduyar:
I think I found all that are using MeterProvider.

pellared:
> I think I found all that are using MeterProvider.

The same should be done for `TraceProvider.Tracer` (like you did for `otelhttp`).

utezduyar:
> > I think I found all that are using MeterProvider.
> 
> The same has to be done for all `TraceProvider.Tracer` usages (like you did in `otelhttp`).

otelhttp has a meterprovider that is why I did it. Do we have a use case for exposing the scope names for tracers? Of course I don't mind making the change. I am just curious.

pellared:
> Do we have a use case for exposing the scope names for tracers?

API consistency.

For now, leave it as it is. I added your proposal as a topic for today's SIG meeting

pellared:
@utezduyar During SIG meeting we agreed that

1. we are fine with the proposal
2. we want to have `ScopeName` for all instrumentation libraries and reuse it when creating meters and tracers
3. `CHANGELOG.md` entry must be added

I am changing the PR to draft. Please make change it to "Ready for Review" when you update the PR.

pellared:
What about `ScopeName` for `otelgrpc`?

utezduyar:
> What about `ScopeName` for `otelgrpc`?

It is back again.

pellared (commented):
I noticed that we are also missing `ScopeName` in `otelmacaron` and `otelhttptrace`.

utezduyar:
> otelhttptrace

Can you point me the file name for this? I still couldn't find it.

pellared:
> > otelhttptrace
> 
> Can you point me the file name for this? I still couldn't find it.

`instrumentation/net/http/httptrace/otelhttptrace/clienttrace.go`

pellared (approved):

instrumentation/README.md
```diff
@@ -92,5 +92,5 @@ Additionally the following guidelines for package composition need to be followe
 - All instrumentation packages MUST provide an option to accept a `TracerProvider` if it uses a Tracer, a `MeterProvider` if it uses a Meter, and `Propagators` if it handles any context propagation.
   Also, packages MUST use the default `TracerProvider`, `MeterProvider`, and `Propagators` supplied by the `global` package if no optional one is provided.
 - All instrumentation packages MUST NOT provide an option to accept a `Tracer` or `Meter`.
-- All instrumentation packages MUST create any used `Tracer` or `Meter` with a name matching the instrumentation package name.
-- All instrumentation packages MUST create any used `Tracer` or `Meter` with a semantic version corresponding to the version of the module containing the instrumentation.
+- All instrumentation packages MUST define `ScopeName` constant with a value matching the instrumentation package and use it when creating `Tracer` or `Meter`.
+- All instrumentation packages MUST define `Version` function returning the version of the module containing the instrumentation and use it when creating `Tracer` or `Meter` .
```
dashpole, line 96:
```suggestion
- All instrumentation packages MUST define a `ScopeName` constant with a value matching the instrumentation package and use it when creating a `Tracer` or `Meter`.
- All instrumentation packages MUST define a `Version` function returning the version of the module containing the instrumentation and use it when creating a `Tracer` or `Meter` .
```

dashpole (approved):

pellared:
@utezduyar Thank you for your contribution 🎉

utezduyar:
> @utezduyar Thank you for your contribution 🎉 

Thanks for your assistance.

PR #4475, opened 2023-10-24
[WIP]hostmetrics: add additional metrics for host

### LINKED ISSUE #4443: hostmetrics: Request for additional metrics or custom host metrics
## Problem Statement

The repository currently provides essential host metrics but stands to benefit from greater flexibility, as there is often a requirement to implement additional metrics to meet specific monitoring and analysis needs. Enhancing it would be a valuable addition and one that would more closely resemble the receiver's behavior.

## Proposed Solution

This will involve expanding the existing set of metrics to include those that the host receiver already provides. Whether they are default or optional, we can naturally map them to metrics considered in official documentation that already adhere to the semantic convention specifications.

### Alternatives

To meet the unique requirements of individual users, we could implement a custom metric feature. This feature would empower users to introduce additional metrics that align with their monitoring needs. By doing so, users can avoid the need to patch together different solutions in an unintuitive manner, resulting in more streamlined and efficient code.

With this proposal, the community could eventually contribute further metrics if they find their custom metrics useful and wish to share them with others. This approach will also help keep the community more engaged through a single package.

### Prior Art
The only example I could find that somewhat resembles the alternative. [Example](https://github.com/GoogleCloudPlatform/golang-samples/blob/c772370d70284f716f60bbf4e5d49c97134fcd24/monitoring/snippets/metric_create.go#L31). 

## Additional Context

The following could be taken as an example for the first proposal.

- New Metric: system.cpu.utilization that maps to -> [system.cpu.utilization](https://github.com/open-telemetry/opentelemetry-collector-contrib/blob/main/receiver/hostmetricsreceiver/internal/scraper/cpuscraper/documentation.md#systemcpuutilization)
- Description: *Difference in system.cpu.time since the last measurement per logical CPU, divided by the elapsed time (value in interval [0,1]).*

This can be implemented using: [gopsutil:psutil for golang](https://github.com/shirou/gopsutil/blob/master/cpu/cpu.go), a package already in use.

```go
// Percent calculates the percentage of cpu used either per CPU or combined.
// If an interval of 0 is given it will compare the current cpu times against the last call.
// Returns one value per cpu, or a single value if percpu is set to false.
func Percent(interval time.Duration, percpu bool) ([]float64, error) {
	return PercentWithContext(context.Background(), interval, percpu)
}
``` 

```go
func (h *host) register() error {

...

if hostCPUutilization, err = h.meter.Float64ObservableGauge(
		"system.cpu.utilization",
		metric.WithUnit("1"),
		metric.WithDescription(
			"CPU utilization of this process.",
		),
	); err != nil {
		return err
	}

...

cpuPercent, err := cpu.Percent(0, false)

...

// Host cpu utilization
opt = metric.WithAttributeSet(AttributeCPUutilization)
o.ObserveFloat64(hostCPUutilization, cpuPercent[0] / 100, opt)

...

}
``` 

***Note***: Im not proposing all metrics to  be added, but i think already present metrics could really benefit from complementary ones to heighten comprehension.

## Closure

I think either of the proposed enhancements would provide more valuable insights into the system's performance, as well as contribute to the growth and adaptability of the host metrics instrumentation.

Thank you for considering this proposal!

TelmoMtzLarrinaga:
This PR adds the following additional metrics for the host instrumentation:
    - system.cpu.utilization
    - system.cpu.logical.count
    - system.cpu.physical.count
 
Closes #4443

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -15,7 +15,8 @@ This project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0.htm
 - Add `"go.opentelemetry.io/contrib/config"` package that includes configuration models generated via go-jsonschema. (#4376)
 - Add `NewSDK` function to `"go.opentelemetry.io/contrib/config"`. The initial implementation only returns noop providers. (#4414)
 - Add metrics support to `go.opentelemetry.io/contrib/exporters/autoexport`. (#4229)
-- Add three additional cpu metrics to `go.opentelemetry.io/contrib/instrumentation/host`. It now also tracks: `system.cpu.utilization, system.cpu.logical.count and system.cpu.physical.count`.(#4475)
+- Add additional CPU metrics to `go.opentelemetry.io/contrib/instrumentation/host`: `system.cpu.utilization`, `system.cpu.logical.count`, `system.cpu.physical.count`. (#4475)
+
 ### Changed
 
 - Dropped compatibility testing for [Go 1.19].
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -15,7 +15,7 @@ This project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0.htm
 - Add `"go.opentelemetry.io/contrib/config"` package that includes configuration models generated via go-jsonschema. (#4376)
 - Add `NewSDK` function to `"go.opentelemetry.io/contrib/config"`. The initial implementation only returns noop providers. (#4414)
 - Add metrics support to `go.opentelemetry.io/contrib/exporters/autoexport`. (#4229)
-
+- Add three additional cpu metrics to `go.opentelemetry.io/contrib/instrumentation/host`. It now also tracks: `system.cpu.utilization, system.cpu.logical.count and system.cpu.physical.count`.(#4475)
 ### Changed
 
 - Dropped compatibility testing for [Go 1.19].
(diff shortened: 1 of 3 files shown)


pellared (commented):
Are you able to add tests?

See https://github.com/open-telemetry/opentelemetry-go-contrib/blob/main/instrumentation/host/host_test.go#L17-L19.

Please look at https://pkg.go.dev/go.opentelemetry.io/otel/sdk/metric/metricdata/metricdatatest and how other instrumentation libraries are tested using this package.

pellared:
Please also update `CHANGELOG.md` and `doc.go` in the package.

TelmoMtzLarrinaga:
I have seen all the feedback. Thanks for taking the time; I greatly appreciate it, really.

> Are you able to add tests?

I'll look into it. I will try to go through it carefully so it doesn't requiere multiple iterations.

pellared:
> I have seen all the feedback. Thanks for taking the time; I greatly appreciate it, really.
> 
> > Are you able to add tests?
> 
> I'll look into it. I will try to go through it carefully so it doesn't requiere multiple iterations if i end up adding them.

I would be awesome if you would be able to add tests first for the existing code (not from this PR) as a separate PR.

herrBez:
Hi there, are there any update on this Pull Request?

TelmoMtzLarrinaga:
Hi @herrBez! I still have to look into adding test for the instrumentation. Right now i kind of a bit caught up with other work so unfortunately not updates from my part. Sorry for the inconvenience, really.

Issue #4476, opened 2023-10-24
otelgin.Middleware("service name")  doesn't work


andotorg:
### Description

otelgin.Middleware("service name")  doesn't work

<img width="620" alt="image" src="https://github.com/open-telemetry/opentelemetry-go-contrib/assets/25116294/86005a0e-f8e9-4690-9cd6-70a943ec9425">
It is not the service name specified by me, but obtained by myself. The name is not standardized and not optimized.

### Environment

- OS: Mac Linux
- Architecture: [e.g. x86, i386]
- Go Version: [e.g. 1.20]
- `go.opentelemetry.io/contrib` version: [e.g. v0.19.0, 3c7face]
- `go.opentelemetry.io/contrib/instrumentation/github.com/gin-gonic/gin/otelgin` version: [e.g. v0.45.0, 3c7face]

### Steps To Reproduce

1. Using this code  `otelgin.Middleware("service name")`
2. Run server
3. See error  
<img width="620" alt="image" src="https://github.com/open-telemetry/opentelemetry-go-contrib/assets/25116294/86005a0e-f8e9-4690-9cd6-70a943ec9425">

### Expected behavior

jaeger display service name : `service name`

pellared:
`service.name` should be set in a resource and not by an instrumentation library,.

See:
- https://opentelemetry.io/docs/instrumentation/go/getting-started/
- https://opentelemetry.io/docs/instrumentation/go/resources/
- https://opentelemetry.io/docs/instrumentation/go/manual/
- https://opentelemetry.io/docs/specs/semconv/resource/#service

andotorg:
The incoming service that is adapted by the gin has no effect, so what is the meaning of the incoming service?
`engine.Use(otelgin.Middleware(what‘s the service？？？？？？ "))`

pellared:
It sets the `net.host.name` attribute (tag).

andotorg:
`net.host.name` this not work, not  display

pellared:
> `net.host.name` this not work, not display

Can you give me a screenshot of tags of the span created by otelgin?

andotorg:
![image](https://github.com/open-telemetry/opentelemetry-go-contrib/assets/25116294/eb1beb54-264d-4b35-9dee-b6c720786ee3)
see , thanks

pellared:
The bug description is not accurate. There are no version. There are no repro steps.
Maybe you are not testing against the latest version?
[...]

PR #4477, opened 2023-10-24
update gin services name key is service.name

andotorg:
fix  [#4476](https://github.com/open-telemetry/opentelemetry-go-contrib/issues/4476)

### DIFF
--- a/instrumentation/github.com/gin-gonic/gin/otelgin/internal/semconvutil/netconv.go
+++ b/instrumentation/github.com/gin-gonic/gin/otelgin/internal/semconvutil/netconv.go
@@ -69,7 +69,7 @@ type netConv struct {
 }
 
 var nc = &netConv{
-	NetHostNameKey:     semconv.NetHostNameKey,
+	NetHostNameKey:     semconv.ServiceNameKey,
 	NetHostPortKey:     semconv.NetHostPortKey,
 	NetPeerNameKey:     semconv.NetPeerNameKey,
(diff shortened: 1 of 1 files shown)


pellared:
This change is not compliant with the OpenTelemetry Specification. See https://github.com/open-telemetry/opentelemetry-go-contrib/issues/4476#issuecomment-1777870968 for more info.

MrAlias:
> This change is not compliant with the OpenTelemetry Specification. See [#4476 (comment)](https://github.com/open-telemetry/opentelemetry-go-contrib/issues/4476#issuecomment-1777870968) for more info.

Agreed. Thank you @andotorg for taking the time to contribute. However, we are not able to accept these changes as they do not comply with the OpenTelemetry semantic conventions. Closing.

andotorg:
@MrAlias 
Sorry, I cannot accept your opinion. Are you saying that `jaeger` was not developed according to the specifications?
router.Use(otelgin.Middleware("foobar", otelgin.WithTracerProvider(provider)))   
`foobar` will not be displayed in the jaeger service list, then you think it does not comply with the specification, but at the display level the service list is `service.name`, a service cannot be displayed to users as a service using `net.host.name` Check? Do you think this is reasonable?

andotorg:
Sorry, It is not in compliance with the standard. I think it can be changed to follow the standard. I hope that the value of this parameter can be displayed normally in `jaeger`. Should I deal with `otelgin` again?

PR #4479, opened 2023-10-24
Add support for Prometheus exporter

### LINKED ISSUE #4472: autoexport: Add support for Prometheus
* Support `OTLP_METRICS_EXPORTER=prometheus`
* Serve a Prometheus metric endpoint based on the values of `OTEL_EXPORTER_PROMETHEUS_HOST` and `OTEL_EXPORTER_PROMETHEUS_PORT` [[spec](https://github.com/open-telemetry/opentelemetry-specification/blob/563958cb2bd8529990f19fdce7a5f3643bf63091/specification/configuration/sdk-environment-variables.md?plain=1#L201)]

This was partially implemented in https://github.com/open-telemetry/opentelemetry-go-contrib/pull/4229 but then removed in favor of a full implementation in a separate PR.

Previous [discussion](https://github.com/open-telemetry/opentelemetry-go-contrib/pull/4229#discussion_r1364353029) [discussion](https://github.com/open-telemetry/opentelemetry-go-contrib/pull/4229#discussion_r1367350098)

punya:
Fixes #4472


exporters/autoexport/metrics.go
```diff
@@ -94,4 +98,43 @@ func init() {
...
+			ReadTimeout:  5 * time.Second,
+			WriteTimeout: 10 * time.Second,
+			IdleTimeout:  120 * time.Second,
+		})
+
+		return prometheus.New()
+	})
+}
+
+func serve(ctx context.Context, server *http.Server) {
```
dashpole, line 123: Have you tested this?  I don't think you are serving prometheus metrics on the server:
https://github.com/open-telemetry/opentelemetry-go/blob/d18277e3c63f64cb1d94608d561786a632187d0f/example/prometheus/main.go#L111

exporters/autoexport/metrics.go
```diff
@@ -94,4 +98,43 @@ func init() {
...
+		}
+
+		go serve(ctx, &http.Server{
+			Addr:         host + ":" + port,
+			ReadTimeout:  5 * time.Second,
+			WriteTimeout: 10 * time.Second,
+			IdleTimeout:  120 * time.Second,
+		})
+
+		return prometheus.New()
```
pellared, line 119: We have a goroutine leak. The HTTP server should be closed when the user closes the returned `metric.Reader`. I think we should create a dedicated implementation of `metric.Reader` to handle shutdown which will take care of cleaning up of both the HTTP server and Prometheus exporter.

exporters/autoexport/metrics.go
punya: I deleted the key part while refactoring, and my tests didn't save me :disappointed: .

Please take another look.

exporters/autoexport/metrics.go
punya: Agreed. I'll work on that.

exporters/autoexport/metrics.go
punya: I need to make the test more robust (wait for the server to be ready).

exporters/autoexport/metrics.go
```diff
@@ -94,4 +99,47 @@ func init() {
...
+			host = "localhost"
+		}
+
+		port := os.Getenv("OTEL_EXPORTER_PROMETHEUS_PORT")
+		if port == "" {
+			port = "9464"
+		}
+
+		mux := http.NewServeMux()
+		mux.Handle("/metrics", promhttp.Handler())
```
pellared, line 114: What do you think of creating an explicit registerer that we can pass using `WithRegisterer` option)? I think it could be safer than using the globals offered by prometheus package if for any reason the user would already use the prometheus globals.

See: https://pkg.go.dev/github.com/prometheus/client_golang/prometheus#hdr-A_Basic_Example

@dashpole WDYT?

dashpole: That makes sense to me.  It means you won't get the "default" prometheus instrumentation experience, like go runtime metrics, but that is probably OK, since OTel will have that kind of instrumentation available.

It does make it harder to combine prometheus instrumentation with otel instrumentation in a single endpoint.  But if someone needs to do that, autoexport might not be the best solution for them...

exporters/autoexport/metrics.go
```diff
@@ -94,4 +99,47 @@ func init() {
...
 	})
+	RegisterMetricReader("prometheus", func(ctx context.Context) (metric.Reader, error) {
+		host := os.Getenv("OTEL_EXPORTER_PROMETHEUS_HOST")
+		if host == "" {
+			host = "localhost"
+		}
+
+		port := os.Getenv("OTEL_EXPORTER_PROMETHEUS_PORT")
+		if port == "" {
+			port = "9464"
```
dashpole, line 110: Where does this port come from?

exporters/autoexport/metrics.go
dashpole: Looks better, thanks!

exporters/autoexport/metrics.go
```diff
@@ -94,4 +99,47 @@ func init() {
...
+		}
+
+		mux := http.NewServeMux()
+		mux.Handle("/metrics", promhttp.Handler())
+
+		go serve(ctx, &http.Server{
+			Addr:         host + ":" + port,
```
dashpole, line 117: Where do these values come from?

exporters/autoexport/metrics.go
pellared: https://github.com/open-telemetry/opentelemetry-specification/blob/082207b76839e81b13e111588c4b2279c41af1a9/specification/configuration/sdk-environment-variables.md?plain=1#L208

exporters/autoexport/metrics.go
punya: Done.

exporters/autoexport/metrics.go
punya: Added a comment to explain.

exporters/autoexport/metrics.go
```diff
@@ -94,4 +100,53 @@ func init() {
...
+
+		mux := http.NewServeMux()
+		reg := prometheus.NewRegistry()
+		mux.Handle("/metrics", promhttp.HandlerFor(reg, promhttp.HandlerOpts{Registry: reg}))
+
+		// Timeouts are necessary to make a server resilent to attacks, but
+		// ListenAndServe doesn't set any. We use timeout values from this example: https://blog.cloudflare.com/exposing-go-on-the-internet/#:~:text=There%20are%20three%20main%20timeouts
+		// TODO: find defaults with a better justification, and/or make these configurable
+		go serve(ctx, &http.Server{
+			Addr:         host + ":" + port,
```
pellared, line 116: Can we use [`net.Listen`](https://pkg.go.dev/net#Listen) before starting the goroutine so that we would return earlier if the provided address or port is invalid?

The we will  use `server.Listen` instead of `server.ListenAndServe` and we would not need to set `server.Addr`.

exporters/autoexport/metrics_test.go
```diff
@@ -63,3 +72,24 @@ func TestMetricExporterOTLPOverInvalidProtocol(t *testing.T) {
 	_, err := NewMetricReader(context.Background())
 	assert.Error(t, err)
 }
+
+func TestMetricExporterPrometheus(t *testing.T) {
+	t.Setenv("OTEL_METRICS_EXPORTER", "prometheus")
```
pellared, line 77: Can we add `defer goleak.VerifyNone(t)` to this test to make sure that we are not leaking any goroutine?

exporters/autoexport/metrics.go
```diff
@@ -94,4 +100,53 @@ func init() {
...
+
+	var err error
+	select {
+	case <-ctx.Done():
+		err = server.Close()
+	case err = <-serverErr:
+	}
+
+	if err != nil && err != http.ErrServerClosed {
+		log.Printf("error closing Prometheus HTTP server: %v", err)
```
pellared, line 141: AFAIK we should use [`otel.Handle`](https://pkg.go.dev/go.opentelemetry.io/otel#Handle) to report errors.

I suggest something like

`otel.Handle(fmt.Errorf("close Prometheus HTTP server: %w", err))`

punya: Done.

exporters/autoexport/metrics.go
punya: Done.

exporters/autoexport/metrics_test.go
punya: Done.

exporters/autoexport/metrics.go
punya: Done.

dashpole (approved):

exporters/autoexport/metrics.go
```diff
@@ -94,4 +104,62 @@ func init() {
...
+			otel.Handle(fmt.Errorf("unable to bind address %s for Prometheus exporter: %w", addr, err))
+		}
+
+		go func() {
+			otel.Handle(server.Serve(lis))
+		}()
+
+		reader, err := promexporter.New(promexporter.WithRegisterer(reg))
+		if err != nil {
+			return nil, err
```
pellared, line 140: I think if this one returns then the `server` will leak (it won't be closed).

How about moving `reader, err := promexporter.New(promexporter.WithRegisterer(reg))` before `go func() { otel.Handle(server.Serve(lis))`? I guess it should work. Also there is no need to listen if we would fail to create the exporter.

exporters/autoexport/metrics.go
```diff
@@ -94,4 +104,62 @@ func init() {
...
+		port := getenv("OTEL_EXPORTER_PROMETHEUS_PORT", "9464")
+		addr := host + ":" + port
+		lis, err := net.Listen("tcp", addr)
+		if err != nil {
+			// TODO: retry binding, to handle process restarts
+			otel.Handle(fmt.Errorf("unable to bind address %s for Prometheus exporter: %w", addr, err))
+		}
+
+		go func() {
+			otel.Handle(server.Serve(lis))
```
pellared, line 135: From: https://pkg.go.dev/net/http#Server.Serve

> Serve always returns a non-nil error and closes l. After Shutdown or Close, the returned error is ErrServerClosed.

I do not think we want to report `ErrServerClosed`.

I think may be good to verify in the tests that no errors was reported via `otel.Handle`. You can do it by using `otel.SetErrorHandler`.

exporters/autoexport/metrics.go
```diff
@@ -94,4 +104,62 @@ func init() {
...
+		}
+
+		// environment variable names and defaults specified at https://opentelemetry.io/docs/specs/otel/configuration/sdk-environment-variables/#prometheus-exporter
+		host := getenv("OTEL_EXPORTER_PROMETHEUS_HOST", "localhost")
+		port := getenv("OTEL_EXPORTER_PROMETHEUS_PORT", "9464")
+		addr := host + ":" + port
+		lis, err := net.Listen("tcp", addr)
+		if err != nil {
+			// TODO: retry binding, to handle process restarts
+			otel.Handle(fmt.Errorf("unable to bind address %s for Prometheus exporter: %w", addr, err))
```
pellared, line 131: I think here we should simply:

```suggestion
			return nil, fmt.Errorf("bind address %s for Prometheus exporter: %w", addr, err)
```

Can you add a test for this scenario (probably setting some invalid port via env var would be good enough to reproduce this)?

exporters/autoexport/metrics.go
```diff
@@ -94,4 +104,62 @@ func init() {
...
+		// create an isolated registry instead of using the global registry --
+		// the user might not want to mix OTel with non-OTel metrics
+		reg := prometheus.NewRegistry()
+
+		mux := http.NewServeMux()
+		mux.Handle("/metrics", promhttp.HandlerFor(reg, promhttp.HandlerOpts{Registry: reg}))
+		server := http.Server{
+			// Timeouts are necessary to make a server resilent to attacks, but ListenAndServe doesn't set any.
+			// We use values from this example: https://blog.cloudflare.com/exposing-go-on-the-internet/#:~:text=There%20are%20three%20main%20timeouts
+			// TODO: find defaults with a better justification, and/or make these configurable
```
pellared, line 117: For me these values are reasonable and there is no need for a TODO comment. Anyone not satisfied with the values can create an issue.

```suggestion
```

pellared (commented):
Probably the last comments from my side.

exporters/autoexport/metrics_test.go
```diff
@@ -63,3 +73,26 @@ func TestMetricExporterOTLPOverInvalidProtocol(t *testing.T) {
 	_, err := NewMetricReader(context.Background())
 	assert.Error(t, err)
 }
+
+func TestMetricExporterPrometheus(t *testing.T) {
+	defer goleak.VerifyNone(t)
+
+	t.Setenv("OTEL_METRICS_EXPORTER", "prometheus")
+	t.Setenv("OTEL_EXPORTER_PROMETHEUS_PORT", "12345")
```
pellared, line 81:
```suggestion
	t.Setenv("OTEL_EXPORTER_PROMETHEUS_PORT", "0")
```

This should make the test less flaky 😉 

AFAIK port 0 tells to find any available port.

exporters/autoexport/metrics.go
punya: Done.

exporters/autoexport/metrics_test.go
punya: Done.

exporters/autoexport/metrics.go
punya: Done.

exporters/autoexport/metrics.go
punya: Done.

pellared (commented):
I just noticed that doc comment for `NewMetricReader` needs to be updated.

exporters/autoexport/metrics_test.go
```diff
@@ -63,3 +75,56 @@ func TestMetricExporterOTLPOverInvalidProtocol(t *testing.T) {
...
+
+	switch os.Geteuid() {
+	case -1:
+		t.Skip("skipping invalid port check on Windows")
+	case 0:
+		t.Skip("skipping invalid port check because the test is running as root (!!)")
+	}
+
+	// an unprivileged user on Unix should fail to bind a port < 1024
+	t.Setenv("OTEL_EXPORTER_PROMETHEUS_PORT", "10")
```
pellared, line 126: This can be simply

```suggestion
	t.Setenv("OTEL_EXPORTER_PROMETHEUS_PORT", "invalid-port")
```

pellared (approved):

pellared:
Nice work 👍 I plan to merge it on Monday to give others chance to review.

punya:
Thanks for the detailed review @pellared.

Issue #4480, opened 2023-10-25
otelgrpc's UnaryServerInterceptor cannot send metrics with correct grpc status code


liufuyang:
### Description

In this part of the code, as it is in a `defer` call, the `statusCode` seems to be always 0 and cannot be set to the actual status returned by the handler call.

https://github.com/open-telemetry/opentelemetry-go-contrib/blob/5adc27110c6f8edff55e07c668aeb140a166dcfd/instrumentation/google.golang.org/grpc/otelgrpc/interceptor.go#L370-L375)

I guess instead of using `defer` and moving that piece of code after the handler call should make it work?

Issue #4416, opened 2023-10-11
Flaky test in otelgrpc TestStatsHandler/ServerSpans


pellared:
https://github.com/open-telemetry/opentelemetry-go-contrib/actions/runs/6486518230/job/17614855539?pr=4414

```
go test -timeout 60s -short ./instrumentation/google.golang.org/grpc/otelgrpc/test/...
--- FAIL: TestStatsHandler (0.01s)
    --- FAIL: TestStatsHandler/ServerSpans (0.00s)
        grpc_stats_handler_test.go:321: 
            	Error Trace:	/home/runner/work/opentelemetry-go-contrib/opentelemetry-go-contrib/instrumentation/google.golang.org/grpc/otelgrpc/test/grpc_stats_handler_test.go:321
            	            				/home/runner/work/opentelemetry-go-contrib/opentelemetry-go-contrib/instrumentation/google.golang.org/grpc/otelgrpc/test/grpc_stats_handler_test.go:53
[...]

PR #4481, opened 2023-10-25
fix(otelgrpc): correctly assign grpc status code

liufuyang:
closes https://github.com/open-telemetry/opentelemetry-go-contrib/issues/4480

I tried to write a test as well to capture the issue above, also made a fix in `interceptor.go`

### DIFF
--- a/instrumentation/google.golang.org/grpc/otelgrpc/interceptor.go
+++ b/instrumentation/google.golang.org/grpc/otelgrpc/interceptor.go
@@ -375,13 +375,13 @@ func UnaryServerInterceptor(opts ...Option) grpc.UnaryServerInterceptor {
 			grpcStatusCode = s.Code()
 			statusCode, msg := serverStatus(s)
 			span.SetStatus(statusCode, msg)
-			span.SetAttributes(statusCodeAttr(s.Code()))
+			span.SetAttributes(statusCodeAttr(grpcStatusCode))
 			if cfg.SentEvent {
 				messageSent.Event(ctx, 1, s.Proto())
 			}
 		} else {
 			grpcStatusCode = grpc_codes.OK
-			span.SetAttributes(statusCodeAttr(grpc_codes.OK))
+			span.SetAttributes(statusCodeAttr(grpcStatusCode))
 			if cfg.SentEvent {
 				messageSent.Event(ctx, 1, resp)
 			}
--- a/instrumentation/google.golang.org/grpc/otelgrpc/interceptor.go
+++ b/instrumentation/google.golang.org/grpc/otelgrpc/interceptor.go
@@ -367,28 +367,26 @@ func UnaryServerInterceptor(opts ...Option) grpc.UnaryServerInterceptor {
 		}
 
 		before := time.Now()
-		var grpcStatusCode grpc_codes.Code
 
 		resp, err := handler(ctx, req)
+
+		s, _ := status.FromError(err)
 		if err != nil {
-			s, _ := status.FromError(err)
-			grpcStatusCode = s.Code()
 			statusCode, msg := serverStatus(s)
 			span.SetStatus(statusCode, msg)
-			span.SetAttributes(statusCodeAttr(grpcStatusCode))
 			if cfg.SentEvent {
 				messageSent.Event(ctx, 1, s.Proto())
 			}
 		} else {
-			grpcStatusCode = grpc_codes.OK
-			span.SetAttributes(statusCodeAttr(grpcStatusCode))
 			if cfg.SentEvent {
 				messageSent.Event(ctx, 1, resp)
 			}
 		}
+		grpcStatusCodeAttr := statusCodeAttr(s.Code())
+		span.SetAttributes(grpcStatusCodeAttr)
 
 		elapsedTime := time.Since(before).Milliseconds()
-		attr = append(attr, statusCodeAttr(grpcStatusCode))
+		attr = append(attr, grpcStatusCodeAttr)
 		cfg.rpcServerDuration.Record(ctx, elapsedTime, metric.WithAttributes(attr...))
 
 		return resp, err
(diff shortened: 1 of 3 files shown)


dashpole (approved):

liufuyang:
@dashpole Thanks for the fast review. I pushed in again to fix some formatting issues.

liufuyang:
I think the failed test is related to https://github.com/open-telemetry/opentelemetry-go-contrib/issues/4416?

instrumentation/google.golang.org/grpc/otelgrpc/test/interceptor_test.go
```diff
@@ -1069,3 +1094,22 @@ func TestStreamServerInterceptorEvents(t *testing.T) {
...
+	data := rm.ScopeMetrics[0].Metrics[0].Data.(metricdata.Histogram[int64])
+	dpt, ok := getMetricFromData(data, name)
+	assert.True(t, ok)
+	attr := dpt.Attributes.ToSlice()
+	assert.ElementsMatch(t, []attribute.KeyValue{
+		semconv.RPCMethod(name),
+		semconv.RPCService(serviceName),
+		otelgrpc.RPCSystemGRPC,
+		otelgrpc.GRPCStatusCodeKey.Int64(int64(code)),
+	}, attr)
```
pellared, line 1114: I suggest using `metricdatatest` instead. 

```suggestion
	want := metricdata.ScopeMetrics{
		Scope: instrumentation.Scope{
			Name:      "go.opentelemetry.io/contrib/instrumentation/google.golang.org/grpc/otelgrpc",
			SchemaURL: "https://opentelemetry.io/schemas/1.17.0",
			Version:   otelgrpc.Version(),
		},
		Metrics: []metricdata.Metrics{
			{
				Name:        "rpc.server.duration",
				Description: "Measures the duration of inbound RPC.",
				Unit:        "ms",
				Data: metricdata.Histogram[int64]{
					Temporality: metricdata.CumulativeTemporality,
					DataPoints: []metricdata.HistogramDataPoint[int64]{
						{
							Attributes: attribute.NewSet(
								semconv.RPCMethod(name),
								semconv.RPCService(serviceName),
								otelgrpc.RPCSystemGRPC,
								otelgrpc.GRPCStatusCodeKey.Int64(int64(code)),
							),
						},
					},
				},
			},
		},
	}

	rm := metricdata.ResourceMetrics{}
	err := reader.Collect(context.Background(), &rm)
	assert.NoError(t, err)
	require.Len(t, rm.ScopeMetrics, 1)
	metricdatatest.AssertEqual(t, want, rm.ScopeMetrics[0], metricdatatest.IgnoreTimestamp(), metricdatatest.IgnoreValue())
```

To make this working you would also need to change the `TestUnaryServerInterceptor` to update to setup providers, recorders etc for each test case. I also noticed that the providers are not being shutdown.

I think it is better to address it in a separate PR as I think this refactoring could be done for most of the gRPC tests.

pellared: I created https://github.com/open-telemetry/opentelemetry-go-contrib/issues/4483

pellared:
> I think the failed test is related to #4416?

90% that you are correct.

I just left 2 minor comments that I would love to have addressed and one bigger for which I created a separate issue.

Thanks for your contribution 👍

instrumentation/google.golang.org/grpc/otelgrpc/test/interceptor_test.go
liufuyang: Aha, I see, thanks, as you probably noticed I basically copied such comparison code from other places.

So it would be nice if we could leave this as a following up refactory task.

However, for now, I cannot promise I have more free time to do such following-up fixes. Hopefully, that is okay?

liufuyang:
Thanks, I fixed up the 2 minor comments as you suggested :)

pellared (approved):

instrumentation/google.golang.org/grpc/otelgrpc/interceptor.go
```diff
@@ -366,31 +366,31 @@ func UnaryServerInterceptor(opts ...Option) grpc.UnaryServerInterceptor {
...
+		before := time.Now()
+		var grpcStatusCode grpc_codes.Code
 
 		resp, err := handler(ctx, req)
 		if err != nil {
 			s, _ := status.FromError(err)
+			grpcStatusCode = s.Code()
 			statusCode, msg := serverStatus(s)
 			span.SetStatus(statusCode, msg)
 			span.SetAttributes(statusCodeAttr(s.Code()))
```
MrAlias, line 378: This is going to set the status code. Why duplicate this?

liufuyang: So you mean to change it to `span.SetAttributes(statusCodeAttr(grpcStatusCode))`?

I can do that.

If you mean other things, please explain more :)

MrAlias: Yeah, that change sounds good to me. It should be unified outside of this conditional though (i.e. on line 391 or near by).

liufuyang: I see. Noticing line 390 to 392 are for that `rpc.server.duration` metric only. They have nothing to do with tracing or the `span` here.

liufuyang: I made a simple change here to use `grpcStatusCode` instead `s.Code()` or `grpc_codes.OK` below.

Otherwise, one could also put this `s, _ := status.FromError(err)`
outside on the top of the `if` block, as `status.FromError(err)` returns an effective `OK` status.

But as there are multiple other places in this file that are written with a similar pattern, I will just keep the changes in this PR relatively small and not change too much here.

Hope that makes sense :)

MrAlias: If we are unifying the code into a var I am in favor of unifying its use outside of the conditional. Please update with the simplifications you mentioned.

liufuyang: Oh thanks, do you mean I do it for the whole file or just this part of code?

liufuyang: I made a small update on the code part I changed cff17fa, not touching on other similar places in this file.

MrAlias (approved):

pellared:
@liufuyang Thank you for your contribution 👍

PR #4488, opened 2023-10-28
Add error as a span event

a-menshchikov:
This PR adds recording an error info as a span event in a stats handler on stats.End (if an error occured).


pellared (commented):
What is the reason to add the error additionally as an event? Are we missing any information passed  (`s.Message()`) to status description?

I found out that we are usually not getting any valuable data via `RecordError`.

a-menshchikov:
@pellared hi.
While I was looking for an answer to your question, I realized that span tags really have all the same information which contained in the error event. The event timestamp cannot be useful in that case because it points on stats.End moment.

It looks like the PR can be closed.

pellared:
Closing. Still thanks for your contribution 👍

PR #4500, opened 2023-10-31
otelhttp: set unit on all instruments

ash2k:
Currently the histogram metric doesn't have units specified. It's good to have them explicitly specified.

The recorded value is actually milliseconds, not microseconds:

https://github.com/open-telemetry/opentelemetry-go-contrib/blob/bead7e47c4d7e080c8ea6c838b4bbd11cb27eccf/instrumentation/net/http/otelhttp/handler.go#L230-L233

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -15,6 +15,7 @@ This project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0.htm
 - Add `"go.opentelemetry.io/contrib/config"` package that includes configuration models generated via go-jsonschema. (#4376)
 - Add `NewSDK` function to `"go.opentelemetry.io/contrib/config"`. The initial implementation only returns noop providers. (#4414)
 - Add metrics support (No-op, OTLP and Prometheus) to `go.opentelemetry.io/contrib/exporters/autoexport`. (#4229, #4479)
+- otelhttp: set unit on the server latency histogram (#4500)
 
 ### Changed
 
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -15,7 +15,7 @@ This project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0.htm
 - Add `"go.opentelemetry.io/contrib/config"` package that includes configuration models generated via go-jsonschema. (#4376)
 - Add `NewSDK` function to `"go.opentelemetry.io/contrib/config"`. The initial implementation only returns noop providers. (#4414)
 - Add metrics support (No-op, OTLP and Prometheus) to `go.opentelemetry.io/contrib/exporters/autoexport`. (#4229, #4479)
-- otelhttp: set unit on the server latency histogram (#4500)
+- otelhttp: set units on all instruments (#4500)
 
 ### Changed
 
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -15,7 +15,7 @@ This project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0.htm
 - Add `"go.opentelemetry.io/contrib/config"` package that includes configuration models generated via go-jsonschema. (#4376)
 - Add `NewSDK` function to `"go.opentelemetry.io/contrib/config"`. The initial implementation only returns noop providers. (#4414)
 - Add metrics support (No-op, OTLP and Prometheus) to `go.opentelemetry.io/contrib/exporters/autoexport`. (#4229, #4479)
-- otelhttp: set units on all instruments (#4500)
+- Set unit and description on all instruments in `go.opentelemetry.io/contrib/instrumentation/net/http/otelhttp`. (#4500)
 
 ### Changed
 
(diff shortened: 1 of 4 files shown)


pellared:
Can you set units for all of the metric instruments?
Can you update the tests?

ash2k:
@pellared Done. I think the spec and what is actually implemented is not the same thing because the spec talks about compressed payload but the implementation measures uncompressed data read/written by the app. This is probably fine, but perhaps the spec needs to be augmented with metrics for uncompressed data. Also, some description language in the spec seem to be not written grammatically correct (from what I can tell, not a native speaker).

ash2k:
Side note/question: it's really weird that maps are used when just normal fields could be. What is the reason? Map lookup involves computing the hash of the string key, etc and is definitely more work for the CPU vs just a pointer dereference.

CHANGELOG.md
```diff
@@ -15,6 +15,7 @@ This project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0.htm
 - Add `"go.opentelemetry.io/contrib/config"` package that includes configuration models generated via go-jsonschema. (#4376)
 - Add `NewSDK` function to `"go.opentelemetry.io/contrib/config"`. The initial implementation only returns noop providers. (#4414)
 - Add metrics support (No-op, OTLP and Prometheus) to `go.opentelemetry.io/contrib/exporters/autoexport`. (#4229, #4479)
+- otelhttp: set units on all instruments (#4500)
```
```suggestion
- Set unit and description on all instruments in `go.opentelemetry.io/contrib/instrumentation/net/http/otelhttp`. (#4500)
```

pellared (approved):

pellared:
> @pellared Done. I think the spec and what is actually implemented is not the same thing because the spec talks about compressed payload but the implementation measures uncompressed data read/written by the app. This is probably fine, but perhaps the spec needs to be augmented with metrics for uncompressed data. Also, some description language in the spec seem to be not written grammatically correct (from what I can tell, not a native speaker).

I created https://github.com/open-telemetry/opentelemetry-go-contrib/issues/4502

Yeah, I saw this as well and I agree 👍

pellared:
> @pellared Done.

Thank you ❤️ 

>  I think the spec and what is actually implemented is not the same thing because the spec talks about compressed payload but the implementation measures uncompressed data read/written by the app. This is probably fine, but perhaps the spec needs to be augmented with metrics for uncompressed data. Also, some description language in the spec seem to be not written grammatically correct (from what I can tell, not a native speaker).

I think the current `main` version is a lot better: https://github.com/open-telemetry/semantic-conventions/blob/main/docs/http/http-metrics.md. Thank you for all the remarks 👍

instrumentation/net/http/otelhttp/handler.go
```diff
@@ -107,13 +107,16 @@ func (h *middleware) createMeasures() {
 	h.counters = make(map[string]metric.Int64Counter)
 	h.valueRecorders = make(map[string]metric.Float64Histogram)
 
-	requestBytesCounter, err := h.meter.Int64Counter(RequestContentLength)
+	requestBytesCounter, err := h.meter.Int64Counter(RequestContentLength, metric.WithUnit("By"),
+		metric.WithDescription("Measures the size of HTTP request content length (uncompressed)"))
```
MrAlias, line 111: nit

```suggestion
	requestBytesCounter, err := h.meter.Int64Counter(
		RequestContentLength,
		metric.WithUnit("By"),
		metric.WithDescription("Measures the size of HTTP request content length (uncompressed)"),
	)
```

instrumentation/net/http/otelhttp/handler.go
```diff
@@ -107,13 +107,16 @@ func (h *middleware) createMeasures() {
...
 	handleErr(err)
 
-	responseBytesCounter, err := h.meter.Int64Counter(ResponseContentLength)
+	responseBytesCounter, err := h.meter.Int64Counter(ResponseContentLength, metric.WithUnit("By"),
+		metric.WithDescription("Measures the size of HTTP response content length (uncompressed)"))
```
MrAlias, line 115: nit

```suggestion
	responseBytesCounter, err := h.meter.Int64Counter(
		ResponseContentLength,
		metric.WithUnit("By"),
		metric.WithDescription("Measures the size of HTTP response content length (uncompressed)"),
	)
```

instrumentation/net/http/otelhttp/handler.go
```diff
@@ -107,13 +107,16 @@ func (h *middleware) createMeasures() {
...
 	handleErr(err)
 
-	serverLatencyMeasure, err := h.meter.Float64Histogram(ServerLatency)
+	serverLatencyMeasure, err := h.meter.Float64Histogram(ServerLatency, metric.WithUnit("ms"),
+		metric.WithDescription("Measures the duration of HTTP request handling"))
```
MrAlias, line 119: nit

```suggestion
	serverLatencyMeasure, err := h.meter.Float64Histogram(
		ServerLatency,
		metric.WithUnit("ms"),
		metric.WithDescription("Measures the duration of HTTP request handling"),
	)
```

MrAlias (approved):

dmathieu (approved):

Issue #1352, opened 2021-10-12
Flaky test: TestInterceptors/StreamClientSpans


codeboten:
As per the output available [here](https://github.com/open-telemetry/opentelemetry-go-contrib/runs/3874070980?check_suite_focus=true), a flaky test was causing CI failures with the following message:

```
=== RUN   TestInterceptors/StreamClientSpans
    grpc_test.go:181: 
        	Error Trace:	grpc_test.go:181
        	            				grpc_test.go:106
        	Error:      	"[%!s(*trace.snapshot=&{grpc.testing.TestService/StreamingInputCall {[155 73 36 223 179 3 93 25 
22 238 66 50 110 75 56 179] [234 158 25 193 181 253 113 135] 1 {[]} false} {[0 0 0 0 0 0 0 0 0 0 0 0 0 0 0 0] [0 0 0 0 0 0 0 
0] 0 {[]} false} 3 {13858018644475018353 7069793 0x8860e20} {13858018644475405458 7456898 0x8860e20} 
[{rpc.system {4 0 grpc <nil>}} {rpc.service {4 0 grpc.testing.TestService <nil>}} {rpc.method {4 0 StreamingInputCall 
<nil>}} {rpc.grpc.status_code {2 0  <nil>}}] [{message [{message.type {4 0 SENT <nil>}} {message.id {2 1  <nil>}} 
{message.uncompressed_size {2 27190  <nil>}}] 0 {13858018644475097554 7148994 0x8860e20}} {message 
[{message.type {4 0 SENT <nil>}} {message.id {2 2  <nil>}} {message.uncompressed_size {2 12  <nil>}}] 0 
{13858018644475100754 7152194 0x8860e20}} {message [{message.type {4 0 SENT <nil>}} {message.id {2 3  <nil>}} 
{message.uncompressed_size {2 1834  <nil>}}] 0 {13858018644475105254 7156794 0x8860e20}} {message 
[{message.type {4 0 SENT <nil>}} {message.id {2 4  <nil>}} {message.uncompressed_size {2 45912  <nil>}}] 0 
{13858018644475116554 7168094 0x8860e20}}] [] {0 } 0 0 0 0 0xa80c9d0 {go.opentelemetry.io/contrib/instrumentation
/google.golang.org/grpc/otelgrpc semver:0.25.0 }}) %!s(*trace.snapshot=&{grpc.testing.TestService/StreamingOutputCall 
[...]

PR #4503, opened 2023-11-02
otelgrpc: Refine tests to use a net socket instead of a buffer

MadVikingGod:
These are the changes proposed by @pellared in https://github.com/open-telemetry/opentelemetry-go-contrib/pull/4322.

Currently, I haven't addressed the metric failure because when #4322 lands, those attributes will be removed.


instrumentation/google.golang.org/grpc/otelgrpc/test/grpc_test.go
```diff
@@ -411,7 +437,10 @@ func checkStreamServerSpans(t *testing.T, spans []trace.ReadOnlySpan) {
 		semconv.RPCService("grpc.testing.TestService"),
 		otelgrpc.RPCSystemGRPC,
 		otelgrpc.GRPCStatusCodeKey.Int64(int64(codes.OK)),
-	}, streamInput.Attributes())
+		semconv.NetSockPeerAddr("127.0.0.1"),
+	}, filterKVs(streamInput.Attributes(), semconv.NetSockPeerPortKey))
+	// Client port is random
+	assert.True(t, existKVs(streamInput.Attributes(), semconv.NetSockPeerAddrKey))
```
pellared, line 443: How about creating a helper like

```go
func findAttribute(attr []attribute.KeyValue, key attribute.Key) attribute.KeyValue
	for _, a := range attr {
		if a.Key == key {
			return a
		}
	}
	return attribute.KeyValue{}
}
```

and use it like this:

```go
netSockPeerPort := findAttribute(streamInput.Attributes(), semconv.NetSockPeerPortKey)
require.NoZero(t, netSockPeerPort)
assert.ElementsMatch(t, []attribute.KeyValue{
		semconv.RPCMethod("StreamingInputCall"),
		semconv.RPCService("grpc.testing.TestService"),
		otelgrpc.RPCSystemGRPC,
		otelgrpc.GRPCStatusCodeKey.Int64(int64(codes.OK)),
		semconv.NetSockPeerAddr("127.0.0.1"),
		netSockPeerPort 
	}, streamInput.Attributes())
```

pellared: Even this should work

```go
ssert.ElementsMatch(t, []attribute.KeyValue{
		semconv.RPCMethod("StreamingInputCall"),
		semconv.RPCService("grpc.testing.TestService"),
		otelgrpc.RPCSystemGRPC,
		otelgrpc.GRPCStatusCodeKey.Int64(int64(codes.OK)),
		semconv.NetSockPeerAddr("127.0.0.1"),
		findAttribute(streamInput.Attributes(), semconv.NetSockPeerPortKey),
	}, streamInput.Attributes())
```

MadVikingGod: I don't think we would want to rely on the empty attribute.  There is a chance both the port attribute doesn't exist, returning the empty attribute, and the instrumentation puts an empty attribute in the list.  Together this test would pass.

If the call was `findAttribute() (attribute.KeyValue, bool)` that would work, but doesn't really simplify the tests.

pellared: Then the first code snippet: https://github.com/open-telemetry/opentelemetry-go-contrib/pull/4503#discussion_r1380830445 would work for you thanks to 

```go
require.NoZero(t, netSockPeerPort)
```

Personally, I think that a case when the instrumentation puts an empty attribute in the list is such an edge case that I would not worry about it.

pellared:
I am fine with the tests in this PR and I would love to see them as a part of https://github.com/open-telemetry/opentelemetry-go-contrib/pull/4322

As for me https://github.com/open-telemetry/opentelemetry-go-contrib/pull/4503#discussion_r1380830445 can be addressed in a separate PR.

MrAlias:
Given this is less than 300 lines and we've talked a decent amount about this already, I'm in favor of combining this with the PR this is intended to test. 👍

instrumentation/google.golang.org/grpc/otelgrpc/test/grpc_test.go
```diff
@@ -47,42 +48,47 @@ var wantInstrumentationScope = instrumentation.Scope{
...
 const bufSize = 2048
 
-func doCalls(cOpt []grpc.DialOption, sOpt []grpc.ServerOption) error {
-	l := bufconn.Listen(bufSize)
-	defer l.Close()
-
-	s := grpc.NewServer(sOpt...)
-	pb.RegisterTestServiceServer(s, interop.NewTestServer())
+// newGrpcTest creats a grpc server, starts it, and executes all the calls, closes everything down.
+func newGrpcTest(listener net.Listener, cOpt []grpc.DialOption, sOpt []grpc.ServerOption) (err error) {
```
pellared, line 52:
```suggestion
func newGrpcTest(listener net.Listener, cOpt []grpc.DialOption, sOpt []grpc.ServerOption) error {
```

pellared (approved):

dashpole (approved):

instrumentation/google.golang.org/grpc/otelgrpc/test/grpc_stats_handler_test.go
```diff
@@ -46,14 +47,18 @@ func TestStatsHandler(t *testing.T) {
 	serverMetricReader := metric.NewManualReader()
 	serverMP := metric.NewMeterProvider(metric.WithReader(serverMetricReader))
 
-	assert.NoError(t, doCalls(
+	listener := bufconn.Listen(bufSize)
+	defer listener.Close()
```
pellared, line 51: I guess that as a follow-up PR we would like to use `net.Listen("tcp", "127.0.0.1:0")` here instead.

pellared: https://github.com/open-telemetry/opentelemetry-go-contrib/issues/4532

pellared:
This https://github.com/open-telemetry/opentelemetry-go-contrib/pull/4503/commits/be618e6fbac4d63290f69f23e957480317b4b123 attempts to fix https://github.com/open-telemetry/opentelemetry-go-contrib/issues/1352 and https://github.com/open-telemetry/opentelemetry-go-contrib/issues/4416

Issue #4449, opened 2023-10-20
propagator/aws/xray: G404: Use of weak random number generator 


pellared:
### Description

Issue detected by `gosec`

```
golangci-lint ./propagators/aws
xray/idgenerator.go:72:19: G404: Use of weak random number generator (math/rand instead of crypto/rand) (gosec)
        gen.randSource = rand.New(rand.NewSource(rngSeed))
```

### Expected behavior

Use a better random number generator.

pellared:
CC @Aneurysm9

pellared:
By @Aneurysm9 

> This doesn't require the use of a cryptographically-secure PRNG because guessing what the next ID will be doesn't provide any advantage to an attacker. Even so, it is seeded with random data from a (presumabbly) high entropy crypto-secure PRNG. This is the same approach as taken by the [default ID generator](https://github.com/open-telemetry/opentelemetry-go/blob/main/sdk/trace/id_generator.go).

PR #4504, opened 2023-11-04
Fix use of weak random number generator

takp:
Resolves https://github.com/open-telemetry/opentelemetry-go-contrib/issues/4449.


pellared (approved):
@Aneurysm9 PTAL

takp:
@Aneurysm9 Could you give a review?

Aneurysm9 (changes requested):
This doesn't require the use of a cryptographically-secure PRNG because guessing what the next ID will be doesn't provide any advantage to an attacker.  Even so, it is seeded with random data from a (presumabbly) high entropy crypto-secure PRNG.  This is the same approach as taken by the [default ID generator](https://github.com/open-telemetry/opentelemetry-go/blob/main/sdk/trace/id_generator.go).

If you do insist on making this change, it will require benchmarks to quantify the impact and some way to assess whether that tradeoff is worthwhile.  I'm not sure it can be.

pellared:
@Aneurysm9 Would there be a problem if the IDs would conflict? May worry is not about the security but the possible insufficient entropy of the current implementation.

However, I agree that the current implementation can be acceptable at this point of time given the precedence of Trace and Span ID generation logic.

@takp Can you make some benchmarks to compare the implementations?

takp:
Sure! I'll prepare benchmark.

takp:
```go
package xray

import (
	"context"
	"testing"

)

// Benchmark for NewSpanID
func BenchmarkIDGenerator_NewSpanID(b *testing.B) {
	gen := NewIDGenerator()
	ctx := context.Background()
	traceID := trace.TraceID{}

	b.ResetTimer()
	for i := 0; i < b.N; i++ {
		_ = gen.NewSpanID(ctx, traceID)
	}
}

// Benchmark for NewIDs
func BenchmarkIDGenerator_NewIDs(b *testing.B) {
	gen := NewIDGenerator()
	ctx := context.Background()

	b.ResetTimer()
	for i := 0; i < b.N; i++ {
		_, _ = gen.NewIDs(ctx)
	}
}
````

**Comment by takp** [comment]
@Aneurysm9 @pellared I executed the benchmark and the original implementation is much faster with my local mac.

- Original implementation is approx. 45 times faster than the PR for `NewSpanID`.
- Original implementation is approx. 15 times faster than the PR for `NewIDs`.

I think this is not a trivial difference. If this module doesn't require cryptographical security, I think we don't have to change current implementation 👍 What do you think?

---

#### Original implementation:
```sh
$ go test -bench=.
goos: darwin
goarch: amd64
pkg: go.opentelemetry.io/contrib/propagators/aws/xray
cpu: Intel(R) Core(TM) i7-9750H CPU @ 2.60GHz
BenchmarkIDGenerator_NewSpanID-12           	45713550	        26.06 ns/op
BenchmarkIDGenerator_NewIDs-12              	 6522184	       182.1 ns/op
BenchmarkStartAndEndSampledSpan-12          	 1804806	       666.0 ns/op
BenchmarkStartAndEndNestedSampledSpan-12    	 2208139	       542.4 ns/op
BenchmarkPropagatorExtract-12               	31596346	        37.83 ns/op
BenchmarkPropagatorInject-12                	51045776	        23.42 ns/op
PASS
ok  	go.opentelemetry.io/contrib/propagators/aws/xray	9.945s
```

#### This PR:
```sh
$ go test -bench=.
goos: darwin
goarch: amd64
pkg: go.opentelemetry.io/contrib/propagators/aws/xray
cpu: Intel(R) Core(TM) i7-9750H CPU @ 2.60GHz
BenchmarkIDGenerator_NewSpanID-12           	 1027538	      1181 ns/op
BenchmarkIDGenerator_NewIDs-12              	  444454	      2690 ns/op
BenchmarkStartAndEndSampledSpan-12          	  363070	      3224 ns/op
BenchmarkStartAndEndNestedSampledSpan-12    	  666268	      1734 ns/op
BenchmarkPropagatorExtract-12               	30904088	        38.24 ns/op
BenchmarkPropagatorInject-12                	51253266	        23.43 ns/op
PASS
ok  	go.opentelemetry.io/contrib/propagators/aws/xray	8.525s
```


#### Benchmark

```go
package xray

import (
	"context"
	"testing"

)

// Benchmark for NewSpanID
func BenchmarkIDGenerator_NewSpanID(b *testing.B) {
	gen := NewIDGenerator()
	ctx := context.Background()
	traceID := trace.TraceID{}

	b.ResetTimer()
	for i := 0; i < b.N; i++ {
		_ = gen.NewSpanID(ctx, traceID)
	}
}

// Benchmark for NewIDs
func BenchmarkIDGenerator_NewIDs(b *testing.B) {
	gen := NewIDGenerator()
	ctx := context.Background()

	b.ResetTimer()
	for i := 0; i < b.N; i++ {
		_, _ = gen.NewIDs(ctx)
	}
}
````

pellared:
@takp I agree. If the collisions would ever be a concern then the API may be extended to accept a custom ID generator. 

Closing. I will make a PR to get rid of the TODO comments next week. Thank you.

Aneurysm9:
> the API may be extended to accept a custom ID generator.

This already is a custom ID generator that was needed only for compatibility with AWS X-Ray's former requirement that trace IDs start with a recent timestamp.  That requirement [has been eliminated](https://aws.amazon.com/about-aws/whats-new/2023/10/aws-x-ray-w3c-format-trace-ids-distributed-tracing/), but we need to keep this implementation here as it's in a stable module and it may still be used by existing systems for some time to come.  Not sure it's worth a major version bump to remove it but we can probably mark it as deprecated at this point.

pellared:
I created https://github.com/open-telemetry/opentelemetry-go-contrib/pull/4635

Issue #4318, opened 2023-09-18
otelgrpc: Deprecate the interceptors 


pellared:
> I think we should also update the existing examples to use the new instrumentation.
> 
> I propose to:
> 
> 1. Merge this PR
> 2. Create an issue for https://github.com/open-telemetry/opentelemetry-go-contrib/pull/3002/files#r1209293763
> 3. Update the existing examples to use the new functionality.
> 4. Deprecate the interceptors so that they are available in next release.
> 5. After the next release remove the interceptors. 

@open-telemetry/go-approvers please upvote/downvote if you agree or not. Feel free to leave any additional feedback.

_Originally posted by @pellared in https://github.com/open-telemetry/opentelemetry-go-contrib/pull/3002#pullrequestreview-1613182957_

1. Blocked by https://github.com/open-telemetry/opentelemetry-go-contrib/issues/4316
2. Deprecate the interceptors 
3. Update the docs so that grpc.StatsHandler is the first-class citizen

dashpole:
I think we also need to add metrics to the StatsHandler implementation.

pellared:
FYI @fatsheep9146

pellared:
> I think we also need to add metrics to the StatsHandler implementation.

Covered by https://github.com/open-telemetry/opentelemetry-go-contrib/issues/4316

PR #4534, opened 2023-11-07
otelgrpc: Deprecate interceptors in favor of stats handlers

pellared:
Fixes https://github.com/open-telemetry/opentelemetry-go-contrib/issues/4318

We strongly still recommend you to use [stats.Handler], mainly for two reasons.

Functional advantages: [stats.Handler] has more information for user to build more flexible and granular metric, for example:
  - multiple different types of represent "data length": In [stats.InPayload], there exists "Length", "CompressedLength", "WireLength" to denote the size of uncompressed, compressed payload data, with or without framing data. But in interceptors, we can only got uncompressed data, and this feature is also removed due to performance problem.
  - more accurate timestamp: [stats.InPayload]'s "RecvTime" and [stats.OutPayload]'s "SentTime" records more accurate timestamp that server got and sent the message, the timestamp recorded by interceptors depends on the location of this interceptors in the total interceptor chain.
  - some other use cases: for example, catch failure of decoding message.

Performance advantages: If too many interceptors are registered in a service, the interceptor chain can become too long, which increases the latency and processing time of the entire RPC call.

[stats.Handler]: https://pkg.go.dev/google.golang.org/grpc/stats#Handler
[grpc.UnaryClientInterceptor]: https://pkg.go.dev/google.golang.org/grpc#UnaryClientInterceptor
[grpc.UnaryServerInterceptor]: https://pkg.go.dev/google.golang.org/grpc#UnaryServerInterceptor
[grpc.StreamClientInterceptor]: https://pkg.go.dev/google.golang.org/grpc#StreamClientInterceptor
[grpc.StreamServerInterceptor]: https://pkg.go.dev/google.golang.org/grpc#StreamServerInterceptor
[stats.OutPayload]: https://pkg.go.dev/google.golang.org/grpc/stats#OutPayload
[stats.InPayload]: https://pkg.go.dev/google.golang.org/grpc/stats#InPayload

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -17,7 +17,7 @@ This project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0.htm
 - Add metrics support (No-op, OTLP and Prometheus) to `go.opentelemetry.io/contrib/exporters/autoexport`. (#4229, #4479)
 - Add support for `console` span exporter and metrics exporter in `go.opentelemetry.io/contrib/exporters/autoexport`. (#4486)
 - Set unit and description on all instruments in `go.opentelemetry.io/contrib/instrumentation/net/http/otelhttp`. (#4500)
-- Add metric support for `grpc.StatsHandler` in `go.opentelemetry.io/contrib/instrumentation/google.golang.org/grpc/otelgrpc`. (#4356) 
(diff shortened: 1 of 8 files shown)


pellared:
@fatsheep9146 Could you please review this PR?

dashpole (approved):

MrAlias (approved):

PR #4320, opened 2023-09-20
Update http Semconv to v1.20.0

MadVikingGod:
This updates all http instrumentation (anything using `internal/shared/semconvutil`) to v1.20.0.

In addition, I did an audit of the currently recommended semantic conventions for HTTP and put a comment of what is included and what is not to hopefully be easier to reference on the next update.

This doesn't go to 1.21 as that included a new set of logic around the `OTEL_SEMCONV_STABILITY_OPT_IN` environment variable.  That will require a release using 1.20.0 for this to work correctly.

PR #4543, opened 2023-11-10
[net/http] add http.server.active_requests metric

hadrienk:
Follow up of https://github.com/open-telemetry/opentelemetry-go-contrib/pull/2617. As @Aneurysm9 pointed out the `Labeler` can only be used after the next handler has been called. This means that for active requests there's no way to inject attributes this way.


hadrienk:
I suppose we could add a comment somewhere about the fact that only metrics recorded after the handler is called are modifiable using a `Labeler`.

harsh-vipps:
➕

pellared (commented):
1. Please cover the new metric in tests.
2. Please add a changelog entry.

pellared:
Please resolve the conflicts. Otherwise I cannot even run the GitHub workflows.

instrumentation/net/http/otelhttp/go.mod
```diff
@@ -7,6 +7,7 @@ require (
 	github.com/stretchr/testify v1.8.4
 	go.opentelemetry.io/otel v1.20.0
 	go.opentelemetry.io/otel/metric v1.20.0
+	go.opentelemetry.io/otel/sdk/metric v1.20.0
```
pellared, line 10: The instrumentation library MUST not depend on any SDK implementation. I think you need this only for testing. Then please move the test code under `test` subdirectory to solve the issue.

hadrienk: I see. Will do.

hadrienk:
Tried hard to add the route tag to the `active_request` but since it has to be started _before_ the tag is injected using the `Labeler` there's way to fix that without changing the API a lot.

hadrienk:
Anything else you need me to do on this PR?

instrumentation/net/http/otelhttp/handler.go
```diff
@@ -226,6 +244,10 @@ func (h *middleware) serveHTTP(w http.ResponseWriter, r *http.Request, next http
 	labeler := &Labeler{}
 	ctx = injectLabeler(ctx, labeler)
 
+	attrs := metric.WithAttributes(httpSchemeFromRequest(r), semconv.HTTPMethod(r.Method))
```
pellared, line 247: This line would bring back the vulnerability caused by unbound cardinality metric attribute (HTTP method) https://github.com/open-telemetry/opentelemetry-go-contrib/security/advisories/GHSA-rcjv-mgp8-qvmr

pellared (changes requested):
1. There is a vulnerability in the code
2. The build fails
3. I am not sure if we should add this metric at the point as it is marked as experimental in stable HTTP conventions: https://opentelemetry.io/docs/specs/semconv/http/http-metrics/#metric-httpserveractive_requests. For sure I would prefer having https://github.com/open-telemetry/opentelemetry-go-contrib/pull/4320 first,

pellared:
> Tried hard to add the route tag to the `active_request` but since it has to be started _before_ the tag is injected using the `Labeler` there's way to fix that without changing the API a lot.

Take notice that this attribute is not specified in https://github.com/open-telemetry/opentelemetry-specification/blob/v1.20.0/specification/metrics/semantic_conventions/http-metrics.md nor https://opentelemetry.io/docs/specs/semconv/http/http-metrics/#metric-httpserveractive_requests

piotrkowalczuk:
What's the future of this PR?

dmathieu:
Based on its staleness, this PR would probably need to be restarted from scratch.
Also, the [http.server.active_requests](https://github.com/open-telemetry/semantic-conventions/blob/main/docs/http/http-metrics.md#metric-httpserveractive_requests) semantic conventions is still in development.
In order to avoid having breaking changes with semantic conventions again, we're not currently accepting unstable metrics in otelhttp.
So we would also need a good proposal to explicitly opt into those metrics (such as the [SDK x package](https://github.com/open-telemetry/opentelemetry-go/tree/main/sdk/internal/x)).

Issue #4547, opened 2023-11-10
gRPC metrics record duration in nanoseconds instead of milliseconds


Sovietaced:
### Description

RPC metrics are recording float64 of nanoseconds instead of milliseconds as they are intended to. 

### Expected behavior

If a float is going to be used to record RPC durations (similar to what is done for HTTP durations) we should be recording milliseconds and not nanoseconds.

Sovietaced:
I will file a pull request for this momentarily.

MrAlias:
Please link to appropriate semantic conventions stating these values should be recorded as nanoseconds prior to opening a PR.

As far as I can tell, they are defined to be measured milliseconds: https://github.com/open-telemetry/semantic-conventions/blob/main/docs/rpc/rpc-metrics.md#metric-rpcserverduration

MrAlias:
> Please link to appropriate semantic conventions stating these values should be recorded as nanoseconds prior to opening a PR.
> 
> As far as I can tell, they are defined to be measured milliseconds: https://github.com/open-telemetry/semantic-conventions/blob/main/docs/rpc/rpc-metrics.md#metric-rpcserverduration

Oh, I miss-read :facepalm:. You are going the other direction

ash2k:
Shouldn't all duration metrics be in seconds? I think I saw this requirement/recommendation somewhere in the spec. For histograms this would mean using custom buckets (which is now possible!) as the default ones would not be granular enough for "seconds" resolution.

MrAlias:
> Shouldn't all duration metrics be in seconds? I think I saw this requirement/recommendation somewhere in the spec. For histograms this would mean using custom buckets (which is now possible!) as the default ones would not be granular enough for "seconds" resolution.

@ash2k please open an issue to track this independently.

For what it's worth, the change from `ms` to `s` happened in later versions of the semantic conventions that we do not currently support. We should track this so we can make sure we include it when we do update.

ash2k:
@MrAlias Opened https://github.com/open-telemetry/opentelemetry-go-contrib/issues/4590. Thank you.

PR #4548, opened 2023-11-10
otelgrpc: stats handlers record durations in ms instead of ns

Sovietaced:
This pull request fixes an issue recently introduced in https://github.com/open-telemetry/opentelemetry-go-contrib/pull/4356. 

I believe it was unintentional that durations would be recorded using nanoseconds when the histogram is a float64. The code is now aligned with how http metric durations are recorded. 

Fixes issue: https://github.com/open-telemetry/opentelemetry-go-contrib/issues/4547


CHANGELOG.md
```diff
@@ -12,6 +12,10 @@ This project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0.htm
 
 - Add the new `go.opentelemetry.io/contrib/instrgen` package to provide auto-generated source code instrumentation. (#3068, #3108)
 
+### Fixed
+
+- The `grpc.StatsHandler` now records RPC durations in ms instead of ns. (#4547)
```
MrAlias, line 17:
```suggestion
- The `go.opentelemetry.io/contrib/instrumentation/google.golang.org/grpc/otelgrpc` instrumentation now records RPC durations in `ms` instead of `ns`. (#4547, #4548)
```

MrAlias (approved):

Sovietaced:
Is it too much overhead for this to make it into a patch? We had been eagerly waiting the latest release for the stats handler metrics.

MrAlias:
> Is it too much overhead for this to make it into a patch? We had been eagerly waiting the latest release for the stats handler metrics.

Are you not able to take a dependency on the git hash when this merges?

Sovietaced:
> Are you not able to take a dependency on the git hash when this merges?

Yeah we should! Sorry I'm new to Go

CHANGELOG.md
```diff
@@ -12,6 +12,10 @@ This project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0.htm
 
 - Add the new `go.opentelemetry.io/contrib/instrgen` package to provide auto-generated source code instrumentation. (#3068, #3108)
 
+### Fixed
+
+- The `go.opentelemetry.io/contrib/instrumentation/google.golang.org/grpc/otelgrpc` instrumentation now records RPC durations in `ms` instead of `ns`. (#4547, #4548)
```
pellared, line 17:
```suggestion
- The `go.opentelemetry.io/contrib/instrumentation/google.golang.org/grpc/otelgrpc` instrumentation now records RPC durations in `ms` instead of `ns`.  (#4548)
```

instrumentation/google.golang.org/grpc/otelgrpc/stats_handler.go
```diff
@@ -194,7 +194,11 @@ func (c *config) handleRPC(ctx context.Context, rs stats.RPCStats) {
 		span.End()
 
 		metricAttrs = append(metricAttrs, rpcStatusAttr)
-		c.rpcDuration.Record(wctx, float64(rs.EndTime.Sub(rs.BeginTime)), metric.WithAttributes(metricAttrs...))
+
+		// Use floating point division here for higher precision (instead of Millisecond method).
+		elapsedTime := float64(rs.EndTime.Sub(rs.BeginTime)) / float64(time.Millisecond)
```
pellared, line 199: How about doing the same floating point division in `interceptor.go`?

pellared (approved):

instrumentation/google.golang.org/grpc/otelgrpc/stats_handler.go
Sovietaced: Addressed in 4b3e369d676a7bcd113b393df36d62a9215fe177

Sovietaced: Please suggest a change to the CHANGELOG if you feel it is necessary.

vroldanbet:
Excuse the drive-by comment 👋🏻 

Security scanning tools are flagging `go.opentelemetry.io/contrib/instrumentation/google.golang.org/grpc/otelgrpc@v0.45.0` with CVE, so folks are being prompted to update to `0.46.0`. The new release also introduces the deprecation of a bunch of interceptor factory methods, and because docs didn't clarify how to move to `stats.Handler`, I started looking into the code, which led me to this issue. For reference the CVE:

```
┌──────────────────────────────────────────────────────────────┬────────────────┬──────────┬────────┬───────────────────┬───────────────┬───────────────────────────────────────────────────────┐
│                           Library                            │ Vulnerability  │ Severity │ Status │ Installed Version │ Fixed Version │                         Title                         │
├──────────────────────────────────────────────────────────────┼────────────────┼──────────┼────────┼───────────────────┼───────────────┼───────────────────────────────────────────────────────┤
│ go.opentelemetry.io/contrib/instrumentation/google.golang.o- │ CVE-2023-47108 │ HIGH     │ fixed  │ 0.45.0            │ 0.46.0        │ otelgrpc DoS vulnerability due to unbound cardinality │
│ rg/grpc/otelgrpc                                             │                │          │        │                   │               │ metrics                                               │
│                                                              │                │          │        │                   │               │ https://avd.aquasec.com/nvd/cve-2023-47108            │
└──────────────────────────────────────────────────────────────┴────────────────┴──────────┴────────┴───────────────────┴───────────────┴───────────────────────────────────────────────────────┘
```

The title of this PR and its accompanying bug report seemed _scary_ as it would mean starting to report several orders of magnitude more milliseconds. I wrote a test that compared the previous code and the proposed code, and I couldn't see a difference other than the resolution of the float, but the unit seems correct. Am I missing something? If the goal is to increase the resolution of the float, then the bug and PR description should be updated.

Here is a small snippet showing it: https://go.dev/play/p/oF4t3c9wBDZ

dashpole (approved):
I do think this justifies a patch release given the recent CVE

Sovietaced:
> The title of this PR and its accompanying bug report seemed _scary_ as it would mean starting to report several orders of magnitude more milliseconds. I wrote a test that compared the previous code and the proposed code, and I couldn't see a difference other than the resolution of the float, but the unit seems correct. Am I missing something? If the goal is to increase the resolution of the float, then the bug and PR description should be updated.

If you look at the original commit you'll see a `time.Duration` being casted to a `float64`: https://github.com/open-telemetry/opentelemetry-go-contrib/pull/4548/commits/d36f18973c571dbf4141a0f3bf52312a8348d10a

The other changes were just done as follow ups to comments about improving the resolution of the float.

vroldanbet:
>If you look at the original commit you'll see a time.Duration being casted to a float64: https://github.com/open-telemetry/opentelemetry-go-contrib/commit/d36f18973c571dbf4141a0f3bf52312a8348d10a

🤦🏻 I missed that the problem was indeed in the stats handler, and I had only looked into the interceptor. All good!

pellared:
@Sovietaced Thank you for your contribution 👍

PR #4583, opened 2023-11-16
config: add Shutdown method

codeboten:
This allows users to call Shutdown on the tracer/meter providers configured in the SDK.


config/config.go
```diff
@@ -41,9 +52,18 @@ func NewSDK(opts ...ConfigurationOption) (SDK, error) {
...
 	return SDK{
-		meterProvider:  initMeterProvider(o),
-		tracerProvider: initTracerProvider(o),
+		meterProvider:  mp,
+		tracerProvider: tp,
+		shutdown: func(ctx context.Context) error {
+			// TODO: handle errors
+			mpShutdown(ctx)
+			tpShutdown(ctx)
+			return nil
```
pellared, line 65: how about

```suggestion
			err : = mp.Shutdown(ctx)
			return errors.Join(tp.Shutdown(ctx))
```

config/config.go
```diff
@@ -41,9 +52,18 @@ func NewSDK(opts ...ConfigurationOption) (SDK, error) {
 		o = opt.apply(o)
 	}
 
+	mp, mpShutdown := initMeterProvider(o)
```
pellared, line 55: I think you can considering looking at this code https://github.com/open-telemetry/opentelemetry-go/blob/main/example/dice/otel.go as reference for `NewSDK`.

config/config.go
codeboten: updated

pellared (approved):
It might be good to add a changelog entry.

codeboten:
Thanks for the review @pellared, changelog added

hanyuancheung (approved):

PR #4591, opened 2023-11-19
net/http/otelhttp mutate request to enable context retrieval from parent middlewares

dillonstreator:
I have a middleware chain that looks like:
panic recovery -> otelhttp -> logger
The logger middleware injects my logger into the request context in a mutating fashion and is used in the panic recovery middleware to enable better tracing.

In order to enable this use case, `otelhttp` would need to mutate the request rather than forwarding a shallow copy of the request.

I'll admit that the middleware chain is fairly brittle as the introduction of another middleware in the chain could break the implied contract but I'm not sure I see another approach.

Here's an example of the use case https://github.com/dillonstreator/template-go-chi/blob/15ca73625798180fb8df7a3c8d16634b6f8c76df/main.go#L40-L82

This example is leveraging my fork of `otelhttp` which includes the request mutating and works pretty nicely https://github.com/dillonstreator/opentelemetry-go-contrib/commit/1e3363d236ad721ee9a61b6008809b918897f79c#diff-70a93fdc86adfd2f20a035bb596f6d9320d6fed7008eaefc6c7d7cde2b1fc97aR229-R230

### DIFF
--- a/instrumentation/net/http/otelhttp/handler.go
+++ b/instrumentation/net/http/otelhttp/handler.go
@@ -226,7 +226,8 @@ func (h *middleware) serveHTTP(w http.ResponseWriter, r *http.Request, next http
 	labeler := &Labeler{}
 	ctx = injectLabeler(ctx, labeler)
 
-	next.ServeHTTP(w, r.WithContext(ctx))
+	*r = *r.WithContext(ctx)
+	next.ServeHTTP(w, r)
 
 	setAfterServeAttributes(span, bw.read, rww.written, rww.statusCode, bw.err, rww.err)
 
--- a/instrumentation/net/http/otelhttp/handler_test.go
+++ b/instrumentation/net/http/otelhttp/handler_test.go
@@ -96,3 +96,15 @@ func TestHandler(t *testing.T) {
 		})
 	}
 }
+
+func TestHandler_request_propagation(t *testing.T) {
+	r, err := http.NewRequest(http.MethodGet, "http://localhost/", nil)
+	require.NoError(t, err)
+
+	rr := httptest.NewRecorder()
+	otelhttp.NewHandler(
+		http.HandlerFunc(func(w http.ResponseWriter, r2 *http.Request) {
+			assert.True(t, r == r2)
+		}), "test_handler",
+	).ServeHTTP(rr, r)
+}


pellared (changes requested):
I am not supporting this change.

> enable context retrieval from parent middlewares

I would say that this breaks the context (as well as middlewares) design. The parent should not have access to context generated by the children.

> mutate request

Mutating a request created by the user is a very bad idea.

Still, thanks for the PR.

pellared:
Here is an example of a middleware that is retrieving the context generated by `otelhttp`: https://github.com/signalfx/splunk-otel-go/tree/main/instrumentation/net/http/splunkhttp

Maybe it will help you to achieve what you need.

dillonstreator:
Hi @pellared, appreciate your input and the example you provided. I've taken a look at the splunk-otel-go middleware and unfortunately it doesn't directly address the specific use case I'm trying to solve for. In the specific scenario I'm working with, the requirement is for the parent middleware (panic recovery) to access context information injected by child middleware(s) (otelhttp & a logger middleware) for better tracing and logging during panics.

One alternative I came across is to reorder my middleware chain such that the panic recovery is a child middleware of the otelhttp and the logger middleware. My only hesitation is that it feels more correct to have recovery first on the off chance otelhttp or the logger middleware panics. However, from the standpoint of the immutable nature of the context interface, I understand the push to have the middleware chain flow in "dependency-based" order.

> >enable context retrieval from parent middlewares
>
> I would say that this breaks the context (as well as middlewares) design. The parent should not have access to context generated by the children.
>
> >mutate request
>
> Mutating a request created by the user is a very bad idea.

I'm wondering if you can provide more color to this response. It would be valuable for myself as well as potential future readers. I understand that the methods exposed by `http.Request` and `context.Context` are immutable by design but there are use cases (such as the one being described here) that fall outside the guardrails imposed by this immutability. If the stance being taken is that these use cases aren't going to be supported by `opentelemetry-go-contrib`, that's okay and I can proceed with the fork.

Looking forward to your insights.

MadVikingGod:
Closing because this is against the contact of the interface this is implementing. Specifically in the [`http.Handler`](https://pkg.go.dev/net/http#Handler) Docs:

> Except for reading the body, handlers should not modify the provided Request. 

Why you shouldn't modify the Request is a question for the go HTTP team.

Also Noted from the Handler is that it already supplies a panic recovery mechanism:

> If ServeHTTP panics, the server (the caller of ServeHTTP) assumes that the effect of the panic was isolated to the active request. It recovers the panic, logs a stack trace to the server error log, and either closes the network connection or sends an HTTP/2 RST_STREAM, depending on the HTTP protocol. To abort a handler so the client sees an interrupted response but the server doesn't log an error, panic with the value ErrAbortHandler. 

If that is the case, I wonder what, if anything, your panic middleware adds.

Issue #4542, opened 2023-11-09
otelhttp: Record metrics on timed out requests


pellared:
On an HTTP client implementation, I have a meter using the request context and I would like to keep track of HTTP timeouts, however, the metrics are ignored because of this timeout.

_Originally posted by @ravanscafi in https://github.com/open-telemetry/opentelemetry-go/issues/3618#issuecomment-1804264546_

We should do something like here: https://github.com/open-telemetry/opentelemetry-go-contrib/blob/fe68fe96a73a70a27afad0772618d7b15b3aa124/instrumentation/google.golang.org/grpc/otelgrpc/stats_handler.go#L206-L235

carrbs:
Would it be possible to include some additional context for someone who is looking to pick up a "good first issue"? Potentially a way to reproduce the bug/scenario?

It appears we can likely copy (or generalize to share between the two) the `withoutCancel` flow and call it ([here](https://github.com/open-telemetry/opentelemetry-go-contrib/blob/main/instrumentation/net/http/otelhttp/handler.go#L229)) before calling the next middleware/handler in the chain, so it doesn't cancel the context prematurely, something like:
```go
// instrumentation/net/http/otelhttp/handler.go:232
ctxWithoutCancel := withoutCancel(ctx)
next.ServeHTTP(w, r.WithContext(ctxWithoutCancel))
```
But I'm not certain that's the requested fix 🤔... (likely bc most of this is very new to me 😁).

pellared:
> Would it be possible to include some additional context for someone who is looking to pick up a "good first issue"? Potentially a way to reproduce the bug/scenario?

I see two ways

First one, to cancel using an HTTP Client. See: https://pkg.go.dev/net/http#Request.Context

> For incoming server requests, the context is canceled when the client's connection closes,

Second one, to cancel through HTTP Server itself.  See `Server.BaseContext` : https://pkg.go.dev/net/http#Server

> BaseContext optionally specifies a function that returns the base context for incoming requests on this server.



> It appears we can likely copy (or generalize to share between the two) the withoutCancel flow and call it ([here](https://github.com/open-telemetry/opentelemetry-go-contrib/blob/main/instrumentation/net/http/otelhttp/handler.go#L229)) before calling the next middleware/handler in the chain, so it doesn't cancel the context prematurely, something like:

[...]

PR #4634, opened 2023-11-27
otelhttp: Record metrics on timed out requests

carrbs:
fixes: #4542 

Copied verbatim the `withoutCancel` flow (referenced in #4542) and used it to ensure the metrics are recorded even if the parent `ctx` is closed. This "patch" (as well as the one in: `instrumentation/google.golang.org/grpc/otelgrpc/stats_handler.go`) can be updated to use the `context.WithoutContext` in Go 1.21 once we are no longer supporting Go 1.20.


pellared (commented):
Implementation looks OK.

I tsuggest adding a test in `instrumentation/net/http/otelhttp/test/handler_test.go`.

I would try adding a server that would always have a canceled request. I guess that something like this should work:

```go
	// here setup a handler instrumented with otelhttp

	ctx, cancel := context.WithCancel(context.Background())
	cancel()
	l, err := net.Listen("tcp", "localhost:0")
	require.NoError(t, err)
	srv := &http.Server{
		BaseContext:  func(_ net.Listener) context.Context { return ctx },
		ReadTimeout:  time.Second,
		WriteTimeout: 10 * time.Second,
		Handler:      otelHandler,
	}
	go func() {
		// When Shutdown is called, Serve immediately returns ErrServerClosed.
		assert.Equal(t, http.ErrServerClosed, srv.Serve(l))
	}()
	t.Cleanup(func() {
		assert.NoError(t, srv.Shutdown(context.Background()) 
	})

	// here make a call using HTTP Client. The address of the server is l.Addr().String()
```

carrbs:
Hi @pellared, I attempted to test `ctxWithoutCancel` with your suggestion, but I think something is amiss either with my request or how I've instrumented the handler. The test passes whether or not `ctxWithoutCancel` is used. Likely (I hope) it's something obvious you can redirect me with on how I set up the test.

instrumentation/net/http/otelhttp/test/handler_test.go
```diff
@@ -539,3 +541,64 @@ func TestWithRouteTag(t *testing.T) {
...
+
+	t.Cleanup(func() {
+		assert.NoError(t, srv.Shutdown(context.Background()))
+	})
+
+	req, err := http.NewRequest("GET", "http://"+l.Addr().String(), nil)
+	require.NoError(t, err)
+
+	rr := httptest.NewRecorder()
+	otelHandler.ServeHTTP(rr, req)
```
pellared, line 583: This is not making a real HTTP request.

You should probably do something like this instead\

```suggestion
	resp, err := client.Do(req)
	require.NoError(t, err)
```

pellared, line 583: This is not making a real HTTP request.

You should probably do something like this instead:

```suggestion
	_, err := client.Get("http://"+l.Addr().String())
	require.NoError(t, err)
```

instrumentation/net/http/otelhttp/handler.go
```diff
@@ -281,3 +283,34 @@ func WithRouteTag(route string, h http.Handler) http.Handler {
 		h.ServeHTTP(w, r)
 	})
 }
+
+func withoutCancel(ctx context.Context) context.Context {
+	if ctx.Done() == nil {
+		return ctx
+	}
+	return withoutCancelCtx{ctx}
```
pellared, line 291: Why not?

```suggestion
	if parent == nil {
		panic("cannot create context from nil parent")
	}
	return withoutCancelCtx{parent}
```

It is copied from https://cs.opensource.google/go/go/+/refs/tags/go1.21.4:src/context/context.go;l=563-571 therefore I would keep it.

pellared:
@carrbs Has mu suggestion helped?

instrumentation/net/http/otelhttp/handler.go
carrbs: Sounds good, I didn't know this had been copied from this. The thinking was that, if a context that wasn't cancellable was passed to `withoutCancel` it seemed like we could return/no-op (and, if it was `nil`, we'd get a panic).

carrbs:
@pellared, yes, thank you! I'd overlooked that when trying to track down how to test this. It's passing now and I confirmed 
here: 
https://github.com/open-telemetry/opentelemetry-go/blob/6cee2b4a4c76b581115d0d0ca150ad8b2e683db6/sdk/metric/instrument.go#L208-L210
and here:
https://github.com/open-telemetry/opentelemetry-go/blob/6cee2b4a4c76b581115d0d0ca150ad8b2e683db6/sdk/metric/instrument.go#L208-L210
The expected behavior of the `withoutCancelCtx`. I'm gonna clean the PR up and take it out of draft.

instrumentation/net/http/otelhttp/handler.go
```diff
@@ -232,13 +233,14 @@ func (h *middleware) serveHTTP(w http.ResponseWriter, r *http.Request, next http
 		attributes = append(attributes, semconv.HTTPStatusCode(rww.statusCode))
 	}
 	o := metric.WithAttributes(attributes...)
-	h.requestBytesCounter.Add(ctx, bw.read, o)
-	h.responseBytesCounter.Add(ctx, rww.written, o)
+	ctxWithoutCancel := withoutCancel(ctx)
```
carrbs, line 236: Now that you've shared that [snippet from the context source](https://cs.opensource.google/go/go/+/refs/tags/go1.21.4:src/context/context.go;l=563-571) @pellared, is there a reason we can't use that directly? I.e.
```suggestion
	ctxWithoutCancel := context.WithoutCancel(ctx)
```

(and then remove the local implementation)?

pellared: Because `context.WithoutCancel` was introduced in Go 1.21 and we also support Go 1.20.

pellared:
> I'm gonna clean the PR up and take it out of draft.

Please do not forget to add an entry in "Fixed" section in `CHANGELOG.md` 😉

CHANGELOG.md
```diff
@@ -15,6 +15,7 @@ This project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0.htm
 
 ### Fixed
 
+- Mimic the `context.WithoutCancel` behavior introduced in Go 1.21 to fix metric recording on timed out requests in `go.opentelemetry.io/contrib/instrumentation/net/http/otelhttp`. (#4542)
```
```suggestion
- Fix `http.Handler` metric instrumentation in `go.opentelemetry.io/contrib/instrumentation/net/http/otelhttp` to record metrics on timed out requests. (#4542)
```

pellared (approved):

pellared:
@carrbs Thank you for your contribution. Do not hesitate to leave suggestions that could make future contributions easier.

pellared:
@Aneurysm9 @dmathieu PTAL as code owners.

carrbs:
what's up with the `codecov/patch` failed test? Is it expecting some unit testing?

pellared:
> what's up with the `codecov/patch` failed test? Is it expecting some unit testing?

It is fine. It just notifies that some new code is not covered by tests which is fine. See: https://app.codecov.io/gh/open-telemetry/opentelemetry-go-contrib/pull/4634?src=pr&el=h1&utm_medium=referral&utm_source=github&utm_content=checks&utm_campaign=pr+comments&utm_term=open-telemetry

pellared:
The fix would not be needed if https://github.com/open-telemetry/opentelemetry-go/pull/4671 would be accepted (and merged).

instrumentation/net/http/otelhttp/handler.go
```diff
@@ -281,3 +283,34 @@ func WithRouteTag(route string, h http.Handler) http.Handler {
 		h.ServeHTTP(w, r)
 	})
 }
+
+func withoutCancel(parent context.Context) context.Context {
```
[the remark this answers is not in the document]
carrbs: Happy to do this, but it sounding like the better option might be the SDK native one (so long as we would not need the flexibility of calling those `.Add` and `.Record` methods in such a way where we want _different_ behavior based on the context's cancelled state)?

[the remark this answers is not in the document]
carrbs: Added, in case we opt to merge this PR. Thanks @MrAlias!

pellared:
I am changing to draft because of https://github.com/open-telemetry/opentelemetry-go/pull/4671

@carrbs Feel free to review the referenced PR.

pellared:
Closing per https://github.com/open-telemetry/opentelemetry-go/pull/4671

PR #4656, opened 2023-12-02
Update comment about http.target on HTTPServerRequest

cartermp:
This comment is out of date and ended up confusing a customer, who was expecting http.target to be added to their traces (as it's a spec-required thing). Later on in the guts of this is the comment that clarifies that it's not added.


pellared (commented):
The changes need to be done in `internal/shared/semconvutil/httpconv.go.tmpl` and then the code has to be regenerated (e.g. by running `make`).

PR #4677, opened 2023-12-06
Add support to record errors from lambda handlers

austindrenski:
Currently, `otellambda` doesn't update the instrumentation-provided span based on the result of the wrapped handler.

I was surprised by this, as most of my OTel instrumentation work involves [open-telemetry/opentelemetry-dotnet](https://github.com/open-telemetry/opentelemetry-dotnet), so this may well be a sensible default for golang, but given the breadth and scale of the OTel projects, I figured this might also just be an oversight, so wanted to open a PR and contribute some code.

This PR adds two config methods:
1. `WithRecordError(bool)` which instructs the `wrappedHandler` to call `span.RecordError(...)` if `wrappedHandler.Handler.Invoke(...)` returns an error.
2. `WithSetStatus(bool)` which instructs the `wrappedHandler` to call `span.SetStatus(...)` if `wrappedHandler.Handler.Invoke(...)` returns an error.

___

_Off-topic, but in the [dotnet OTel libraries, it's also common to provide enrichment delegates](https://github.com/open-telemetry/opentelemetry-dotnet/tree/main/src/OpenTelemetry.Instrumentation.Http#enrich-httpclient-api) which can be used to filter/manipulate the instrumentation-provided span in other ways (e.g. setting custom attributes/tags based on the exception being thrown). I considered adding similar functionality here, but decided to hold off on doing anything more until the code owners have had an opportunity to weigh in._

## Background

_based on `go.opentelemetry.io/contrib/instrumentation/github.com/aws/aws-lambda-go/otellambda v0.46.1`_

sanitized screencap of Datadog showing a trace's flame graph and error list

sanitized screencap of Datadog showing an APM error graph showing a 0% error rate before and a non-zero error rate after

### Sample code based on version `0.0.112` in the above screencap

```go
func main() {
	lambda.Start(otellambda.InstrumentHandler(handle))
}

func handle(ctx context.Context, data []byte) error {
	ctx, span := trace.SpanFromContext(ctx).TracerProvider().Tracer("...").Start(ctx, "...")

	// lots of internal logic
	// ...
	// err := ...

	if err != nil {
		span.RecordError(err)
		return err
	}

	return nil
}
```

### Sample code based on version `0.0.113` in the above screencap

Note that our handler is no longer creating its own internal span. We can work around this by defining our own `wrappingHandler`, but the path of least resistance is to just work with the instrumentation-provided span.

```diff
func main() {
	lambda.Start(otellambda.InstrumentHandler(handle))
}

func handle(ctx context.Context, data []byte) error {
-	ctx, span := trace.SpanFromContext(ctx).TracerProvider().Tracer("...").Start(ctx, "...")
+	span := trace.SpanFromContext(ctx)

	// lots of internal logic
	// ...
	// err := ...

	if err != nil {
		span.RecordError(err)
		span.SetStatus(codes.Error, err.Error())
		return err
	}

	return nil
}
```

### Sample code based on this patch

Now we can create an internal span in our handler _and_ choose whether errors recorded in this layer should _also_ be passed back up to the instrumentation-provided span to be recorded there as well. 

_(Example here might be overly simplified, but underlying all of the required obfuscation there is a real-world business case for this. I can provide more details upon request if required.)_

```diff
func main() {
-	lambda.Start(otellambda.InstrumentHandler(handle))
+	lambda.Start(otellambda.InstrumentHandler(handle, otellambda.WithRecordError(true), otellambda.WithSetStatus(true)))
}

func handle(ctx context.Context, data []byte) error {
-	span := trace.SpanFromContext(ctx)
+	ctx, span := trace.SpanFromContext(ctx).TracerProvider().Tracer("...").Start(ctx, "...")

	// lots of internal logic
	// ...
	// err := ...

	if err != nil {
		span.RecordError(err)
		span.SetStatus(codes.Error, err.Error())
		return err
	}

	return nil
}

```

### DIFF
--- a/instrumentation/github.com/aws/aws-lambda-go/otellambda/config.go
+++ b/instrumentation/github.com/aws/aws-lambda-go/otellambda/config.go
@@ -96,10 +96,9 @@ type config struct {
 	// handler should be recorded on the instrumentation-provided span
 	RecordError bool
 
-	// SetError is used to determine if an error returned from the wrapped
-	// handler should cause the instrumentation-provided span to set the
-	// span status to codes.Error
-	SetError bool
+	// RecordStackTrace is used to determine if the stack trace should be
+	// included when RecordError is true
+	RecordStackTrace bool
 }
 
 // WithTracerProvider configures the TracerProvider used by the
@@ -139,19 +138,9 @@ func WithPropagator(propagator propagation.TextMapPropagator) Option {
 // error returned by the wrapped lambda on the instrumentation-provided span.
 //
 // By default, returned errors are not recorded.
-func WithRecordError(recordError bool) Option {
+func WithRecordError(recordError bool, recordStackTrace bool) Option {
 	return optionFunc(func(c *config) {
 		c.RecordError = recordError
-	})
-}
-
-// WithSetStatus configures whether the instrumentation should set the span
-// status to codes.Error on the instrumentation-provided span when the
-// wrapped handler returns an error.
-//
-// By default, returned the status is not set.
-func WithSetStatus(setError bool) Option {
-	return optionFunc(func(c *config) {
-		c.SetError = setError
+		c.RecordStackTrace = recordError && recordStackTrace
 	})
 }
--- a/instrumentation/github.com/aws/aws-lambda-go/otellambda/config.go
+++ b/instrumentation/github.com/aws/aws-lambda-go/otellambda/config.go
@@ -91,6 +91,15 @@ type config struct {
 	// The default value of Propagator the global otel Propagator
 	// returned by otel.GetTextMapPropagator()
 	Propagator propagation.TextMapPropagator
+
+	// RecordError is used to determine if an error returned from the wrapped
+	// handler should be recorded on the instrumentation-provided span
+	RecordError bool
+
+	// SetError is used to determine if an error returned from the wrapped
+	// handler should cause the instrumentation-provided span to set the
+	// span status to codes.Error
+	SetError bool
 }
 
 // WithTracerProvider configures the TracerProvider used by the
@@ -123,5 +132,26 @@ func WithEventToCarrier(eventToCarrier EventToCarrier) Option {
 func WithPropagator(propagator propagation.TextMapPropagator) Option {
 	return optionFunc(func(c *config) {
 		c.Propagator = propagator
+	})
+}
+
+// WithRecordError configures whether the instrumentation should record an
+// error returned by the wrapped lambda on the instrumentation-provided span.
+//
+// By default, returned errors are not recorded.
+func WithRecordError(recordError bool) Option {
+	return optionFunc(func(c *config) {
+		c.RecordError = recordError
+	})
+}
+
+// WithSetStatus configures whether the instrumentation should set the span
+// status to codes.Error on the instrumentation-provided span when the
+// wrapped handler returns an error.
+//
+// By default, returned the status is not set.
+func WithSetStatus(setError bool) Option {
+	return optionFunc(func(c *config) {
+		c.SetError = setError
 	})
 }
--- a/instrumentation/github.com/aws/aws-lambda-go/otellambda/wrapHandler.go
+++ b/instrumentation/github.com/aws/aws-lambda-go/otellambda/wrapHandler.go
@@ -17,6 +17,7 @@ package otellambda // import "go.opentelemetry.io/contrib/instrumentation/github
 import (
 	"context"
 	"go.opentelemetry.io/otel/codes"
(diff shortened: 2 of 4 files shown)


dmathieu:
Why would someone explicitly not want to record errors? (do we need the option? How about recording the error all the time?)

austindrenski:
> Why would someone explicitly not want to record errors? (do we need the option? How about recording the error all the time?)

Great question. I think there are two parts to the answer:

1. IMO we could elide `WithSetStatus(...)` and take the stance that `otellambda` will ___always___ call `span.SetStatus(codes.Error, err.Error())` if the wrapped handler returns an error.

   That's actually what I wrote initially, treating it was an outright bug to not propagate the error status to the parent span (_it certainly caused me a headache in production!_), but I eventually pulled back and hid it behind a config option to avoid breaking any legitimate workflows (_doubtful as I am that they even exist_).

   That said, I'm more than happy to elide `WitSetStatus(...)` if everyone's amenable. Would indeed offer a better happy path out-of-the-box.

2. However, it's a slightly different story for `WithRecordError(...)`. At least in the dotnet world, there's an effort to avoid adding exception events to a span without user opt-in.

   There are better explanations out there, but IIRC they mostly boil down to **(A)** more span events == more data to export, **(B)** recording an exception == serialization == allocations.

   (_Should note that dotnet exceptions by default include the message + the stack trace, so recording an exception as a span event can have a significant impact on span size. This might matter less in golang where `err.Error()` seems to (often? always?) return just the error message._)

pellared:
Does [this comment](https://github.com/open-telemetry/opentelemetry-go-contrib/pull/4488#pullrequestreview-1703589242) applies to this PR ?

How about simply adding `span.SetStatus(codes.Error, err.Error())` when an error is returned?

I think this is also inline with https://github.com/open-telemetry/opentelemetry-go-contrib/pull/4677#issuecomment-1843250118.

instrumentation/github.com/aws/aws-lambda-go/otellambda/config.go
```diff
@@ -125,3 +133,14 @@ func WithPropagator(propagator propagation.TextMapPropagator) Option {
 		c.Propagator = propagator
 	})
 }
+
+// WithRecordError configures whether the instrumentation should record an
+// error returned by the wrapped lambda on the instrumentation-provided span.
+//
+// By default, returned errors are not recorded.
+func WithRecordError(recordError bool, recordStackTrace bool) Option {
```
Aneurysm9, line 141: Should this be two separate options?  We don't typically accept multiple distinct configuration values in options and I don't think there's any need to check whether `c.RecordError` is set when setting `c.RecordStackTrace` as the stack trace config is only ever used when the instrumentation is configured to record errors.

instrumentation/github.com/aws/aws-lambda-go/otellambda/wrapHandler.go
```diff
@@ -38,6 +40,10 @@ func (h wrappedHandler) Invoke(ctx context.Context, payload []byte) ([]byte, err
 
 	response, err := h.handler.Invoke(ctx, payload)
 	if err != nil {
+		if h.instrumentor.configuration.RecordError {
+			span.RecordError(err, trace.WithStackTrace(h.instrumentor.configuration.RecordStackTrace))
+		}
+		span.SetStatus(codes.Error, err.Error())
```
Aneurysm9, line 46: I'm not sure this is correct.  There are situations where an error may be returned from the handler but that should not result in the span status being set to error.  For instance, if the handler is supporting an HTTP-based service and an error would indicate a `4xx` status code, the span status [MUST](https://github.com/open-telemetry/semantic-conventions/blob/main/docs/http/http-spans.md#status) be left unset.  Similarly, the status message shouldn't be set when it can be inferred from other data such as the HTTP status code.

Given that the instrumented handler has access to this span and can set the status as appropriate I would omit setting it here.

instrumentation/github.com/aws/aws-lambda-go/otellambda/wrapHandler.go
```diff
@@ -38,6 +40,10 @@ func (h wrappedHandler) Invoke(ctx context.Context, payload []byte) ([]byte, err
 
 	response, err := h.handler.Invoke(ctx, payload)
 	if err != nil {
+		if h.instrumentor.configuration.RecordError {
```
dashpole: It looks like `WithRecordError` is only handled here in `WrapHandler`, so callers using `InstrumentHandler` (in `wrapLambdaHandler.go`) won't have errors recorded on the span. Could we wire this into `wrappedHandlerFunction` (or `instrumentor.tracingEnd`) as well and add a corresponding `InstrumentHandler` test case?

PR #4701, opened 2023-12-14
Add test to otelhttp to validate semantic conventions

MadVikingGod:
I want to start a conversation about semantic conventions and how we can better validate instrumentation outputs.

I'm looking for some feedback on this approach and what I need to do to get this as the pattern for validating most/all of our instrumentation.

This PR here adds a test in the otelhttp package that uses testcontainers to run `https://github.com/MadVikingGod/otel-semconv-checker` and capture the output from a wrapped HTTP client and server.

This has a hard dependency on docker, which is why I put the test behind a `-docker` flag.

How I would see this used in other instrumentation is a similar testcontiner setup, but different config, gprc wouldn't have the same attribute groups, and a different server/client wrap setup.


instrumentation/net/http/otelhttp/test/semantic_convention_test.go
```diff
@@ -0,0 +1,142 @@
...
+var docker = flag.Bool("docker", false, "Enable tests that use docker.")
+
+func TestSemanticConventions(t *testing.T) {
+	if !*docker {
+		t.Skip("Docker not enabled.")
+	}
+	wd, err := os.Getwd()
+	ctx := context.Background()
+	req := testcontainers.ContainerRequest{
+		Image:        "ghcr.io/madvikinggod/semantic-convention-checker:0.0.1",
```
pellared, line 43: I see that https://github.com/MadVikingGod/otel-semconv-checker is a Go application.
How about simply running the HTTP server in the same process (reuse Go code) instead of running as a docker container?

instrumentation/net/http/otelhttp/test/semantic_convention_test.go
```diff
@@ -0,0 +1,158 @@
...
+	defer ts.Close()
+
+	client := ts.Client()
+	client.Transport = otelhttp.NewTransport(client.Transport, otelhttp.WithSpanNameFormatter(formatter), otelhttp.WithTracerProvider(tp))
+
+	resp, err := client.Post(ts.URL+"/", "application/text", strings.NewReader("Hello, Server!"))
+	require.NoError(t, err)
+
+	_, _ = io.Copy(io.Discard, resp.Body)
+	_ = resp.Body.Close()
```
dmathieu, line 123: Shouldn't we assert no error there?

instrumentation/net/http/otelhttp/test/semantic_convention_test.go
```diff
@@ -0,0 +1,158 @@
...
+
+type lc struct {
+	t testing.TB
+}
+
+func (l *lc) Accept(log testcontainers.Log) {
+	l.t.Log(string(log.Content))
+}
+
+var docker = flag.Bool("docker", false, "Enable tests that use docker.")
```
dmathieu, line 49: Should there be a specific test suite running this in CI? Or do you plan this for later?
Also, an alternative here rather than use the `flag` package could be to use a `// go:build docker` build flag so the file doesn't run by default, but can be enabled.

pellared: > // go:build docker

Let's not do that. Many tools (IDE, golangci-lint) might stop working.

However, I think we could (should) enable it by default.

We can consider skipping when `-short` is provided using this pattern: https://pkg.go.dev/testing#hdr-Skipping.

MadVikingGod: I made it part of the non-`-short` tests.  I do want to warn that this makes docker required for running all tests.  including `make test` and `make precommit`

instrumentation/net/http/otelhttp/test/semantic_convention_test.go
```diff
@@ -0,0 +1,157 @@
...
+	})
+	require.NoError(t, err)
+	defer func() {
+		require.NoError(t, scc.Terminate(ctx))
+	}()
+	scc.FollowOutput(&lc{t: t})
+	err = scc.StartLogProducer(ctx)
+	require.NoError(t, err)
+
+	otel.SetErrorHandler(otel.ErrorHandlerFunc(func(err error) {
```
dashpole: We should restore the previous global `otel.ErrorHandler` via `t.Cleanup` so subsequent tests in this package don't call `t.Errorf` on a finished test (also worth closing `rc` and passing `string(stdout)` to `t.Log` below).
```suggestion
	origHandler := otel.GetErrorHandler()
	t.Cleanup(func() { otel.SetErrorHandler(origHandler) })
	otel.SetErrorHandler(otel.ErrorHandlerFunc(func(err error) {
```

instrumentation/net/http/otelhttp/test/semantic_convention_test.go
```diff
@@ -0,0 +1,157 @@
...
+
+	_, err = io.Copy(io.Discard, resp.Body)
+	assert.NoError(t, err)
+	err = resp.Body.Close()
+	assert.NoError(t, err)
+
+	err = tp.ForceFlush(context.Background())
+	require.NoError(t, err)
+
+	time.Sleep(5 * time.Second)
```
dashpole: Calling `tp.Shutdown(ctx)` here instead of sleeping will synchronously flush any remaining spans and cleanly stop the background `BatchSpanProcessor` and `otlptracegrpc` goroutines before `scc.Terminate(ctx)` tears down the container.
```suggestion
	require.NoError(t, tp.Shutdown(ctx))
```

Issue #4678, opened 2023-12-07
Remove otelmongo


pellared:
### Problem Statement

OTel maintainers and approvers are on low capacity and the `otelmongo` module has no codeowners.

### Proposed Solution

Move the ownership to MongoDB team. First deprecate and later remove `otelmongo` from opentelemetry-go-contrib.

#### Alternatives

Find codeowners among OTel Go approvers.

### Additional Context

There is an informal agreement that the MongoDB (Go driver) team will take over the instrumentation library.
Reference: https://jira.mongodb.org/browse/GODRIVER-2938

pellared:
Closing as duplicate of https://github.com/open-telemetry/opentelemetry-go-contrib/issues/5551

PR #4702, opened 2023-12-15
Created statement marshaller

fgarciau:


pellared:
FYI https://github.com/open-telemetry/opentelemetry-go-contrib/issues/4678

IvanOfThings:
> FYI #4678

Thank you @pellared for the information we'll keep watching about the progress on the migration

instrumentation/go.mongodb.org/mongo-driver/mongo/otelmongo/statement_marshaller.go
```diff
@@ -0,0 +1,62 @@
...
+// and could lead to leaks of sensitive information.
+type limitedStatementMarshaller struct {
+	limit int
+}
+
+// Marshal marshales the whole db statement in Json format to include it in the db statement attribute.
+func (m *limitedStatementMarshaller) Marshal(command bson.Raw) []attribute.KeyValue {
+	b, _ := bson.MarshalExtJSON(command, false, false)
+	statement := string(b)
+	if m.limit > 0 {
```
dashpole: If `len(statement) < m.limit`, `statement[:m.limit]` will panic with a slice bounds out of range error. Please also add unit tests for `NewLimitedStatementMarshaller` covering statements both shorter and longer than `limit`.
```suggestion
	if m.limit > 0 && len(statement) > m.limit {
```

instrumentation/go.mongodb.org/mongo-driver/mongo/otelmongo/config.go
```diff
@@ -77,3 +79,11 @@ func WithCommandAttributeDisabled(disabled bool) Option {
 		cfg.CommandAttributeDisabled = disabled
 	})
 }
+
+// WithMarshaller sets the marshalled in chargee of managing the bson command that will be marshalled into the
+// db.statement attribute
+func WithMarshaller(marshaller StatementMarshaller) Option {
+	return optionFunc(func(cfg *config) {
+		cfg.statementMarshaller = marshaller
```
dashpole: If `marshaller` is `nil`, `m.cfg.statementMarshaller.Marshal` will panic when command attributes are enabled. We should guard against `nil` here like `WithTracerProvider` does:
```suggestion
		if marshaller != nil {
			cfg.statementMarshaller = marshaller
		}
```

instrumentation/go.mongodb.org/mongo-driver/mongo/otelmongo/example_test.go
```diff
@@ -50,3 +50,32 @@ func Example() {
 		panic(err)
 	}
 }
+
+func ExampleLimitedStatementMarshaller() {
```
dashpole: `go test` fails here with `ExampleLimitedStatementMarshaller refers to unknown identifier: LimitedStatementMarshaller` because `LimitedStatementMarshaller` is not exported. Renaming this to `ExampleNewLimitedStatementMarshaller` fixes the build:
```suggestion
func ExampleNewLimitedStatementMarshaller() {
```

instrumentation/go.mongodb.org/mongo-driver/mongo/otelmongo/example_test.go
```diff
@@ -50,3 +50,32 @@ func Example() {
...
+	// connect to MongoDB
+	opts := options.Client()
+	opts.Monitor = otelmongo.NewMonitor(
+		otelmongo.WithMarshaller(otelmongo.NewLimitedStatementMarshaller(50)),
```
dashpole: Since `CommandAttributeDisabled` defaults to `true`, the marshaller configured here won't be invoked unless `WithCommandAttributeDisabled(false)` is also passed (or `WithMarshaller` enables it automatically):
```suggestion
		otelmongo.WithCommandAttributeDisabled(false),
		otelmongo.WithMarshaller(otelmongo.NewLimitedStatementMarshaller(50)),
```

Issue #3134, opened 2023-01-05
Add metrics for transport layer in instrumentation net/http


kw510:
Metrics are provided in the handler of net/http, however they are not provided in the transport layer. This would be super useful to monitor outgoing requests.

https://github.com/open-telemetry/opentelemetry-go-contrib/blob/main/instrumentation/net/http/otelhttp/transport.go

rangzen:
I invite you to follow #3765 to avoid duplicating the actual problem when resolving this issue.

Sovietaced:
@MrAlias Is this an issue you would be open to having a community contribution for?

jlordiales:
@Sovietaced there's already an open PR for this: https://github.com/open-telemetry/opentelemetry-go-contrib/pull/3769

Sovietaced:
Thanks @ jlordiales

I will focus my efforts there.

Issue #1336, opened 2021-10-08
Add Instrumentation (metric) for net/http


victoraugustolls:
## Background

**Package Link**: `net/http`

stdlib http client.

## Proposed Solution

Collect `http.client.duration` metric from the stdlib http client. Currently I'm using a custom `RoundTripper` that calls the `otelhttp` tracing `RoundTripper`, and I think it would be nice if this was inside the contrib. If agreed, I can open a PR for it!

I see something similar was done in #427 , but it was unfortunately closed! That makes me think that this might not be desired? This is why I'm asking in advance. Thanks!

### Metrics

Instruments

- http.client.duration: measure the duration of the outbound HTTP request
  - type: Histogram
  - unit: milliseconds
  - description: measure the duration of the outbound HTTP request

### Prior Art

- This is a metric defined in the [specs](https://github.com/open-telemetry/opentelemetry-specification/blob/main/specification/metrics/semantic_conventions/http-metrics.md#http-client). Previous PR here: #427 .

## Tasks

- Code complete:
  - [ ] Comprehensive unit tests.
  - [ ] End-to-end integration tests.
  - [ ] Tests all passing.
  - [ ] Instrumentation functionality verified.
- Documented
  - [ ] Added to the [OpenTelemetry Registry](https://opentelemetry.io/registry/)
  - [ ] README included for the module describing high-level purpose.
  - [ ] Complete documentation of all public API including package documentation.
  - [ ] [Instrumentation documentation](https://github.com/open-telemetry/opentelemetry-go-contrib/blob/main/instrumentation/README.md#instrumentation-packages) updated.
- Examples
  - [ ] `Dockerfile` file to build example application.
  - [ ] `docker-compose.yml` to run example in a docker environment to demonstrate instrumentation.

tonglil:
Personally I feel like that kind of information is critical to the instrumentation. As far as I can tell, the otelhttp instrumentation is mostly barebones at the moment and only records `HTTP GET/POST/etc`, like without parameterized http.url and http.target attributes that I feel like are needed.

Would love to see this data @victoraugustolls!

Issue #4501, opened 2023-11-01
Metrics are non conformant to semantic conventions


sk-:
### Description

Two of the metrics provided by the `otelhttp` package don't follow the semantic conventions. These metrics:
- `http.server.response_content_length` which should be called `http.server.request.body.size` and be an histogram (see [spec](https://opentelemetry.io/docs/specs/semconv/http/http-metrics/#metric-httpserverrequestbodysize))
 `http.server.request_content_length` which should be called `http.server.response.body.size` and be an histogram (see [spec](https://opentelemetry.io/docs/specs/semconv/http/http-metrics/#metric-httpserverresponsebodysize)

### Expected behavior

The `otelhttp` package (and all packages in this repo) should follow the semantic conventions.

MadVikingGod:
That was changed in the 1.21 and later version of the semantic convention.  Currently, otelhttp is on 1.20, which uses [`http.request_content_length`](https://github.com/open-telemetry/opentelemetry-specification/blob/v1.20.0/semantic_conventions/trace/http.yaml#L11)

There is work that needs to be done to be able to transition from 1.20 to later, because of the introduced breaking changes in the semantic convention, but should resolve this issue with it.

dashpole:
@MadVikingGod is there an issue tracking the work that needs to be done to get past the breaking changes in semconv?  It looks like we do have v1.23.1 generated: https://github.com/open-telemetry/opentelemetry-go/pull/4749

danilobuerger:
The metric names have been changed, but haven't been migrated to histogram

PR #3769, opened 2023-05-03
[otelhttp] transport metrics

RangelReale:
Add HTTP client metrics following the [semantic conventions](https://github.com/open-telemetry/opentelemetry-specification/blob/v1.17.0/specification/metrics/semantic_conventions/http-metrics.md#http-client).

As users of `Transport` are not expected to have a handler where they can customize the attributes using a `Labeler`, callback function options were added to add attributes from the request and the reponse objects.

Fixes #3134 
Fixes #1336

PR #4707, opened 2023-12-17
otelhttp: client metrics

Sovietaced:
This pull request attempts to follow up on the work that @rangelReale did in https://github.com/open-telemetry/opentelemetry-go-contrib/pull/3769 and close out the work. This pull request adds support for http client metrics that mostly adhere to the [v1.20 semantic conventions](https://github.com/open-telemetry/opentelemetry-specification/blob/v1.20.0/specification/metrics/semantic_conventions/http-metrics.md#http-client). In many ways this code attempts to mirror the http server metric code in implementation. 

As such, it reuses the HTTP request body wrapper logic and introduces HTTP response body wrapper logic with callbacks to record metrics once the response body has been read.

Open Questions: 
* As written these client metrics follow the convention of the server metrics, but don't actually follow the convention of the OTEL spec as noted in https://github.com/open-telemetry/opentelemetry-go-contrib/issues/4501. I guess this migration can be done later?
* I don't believe the code will report HTTP response metrics unless the response body is read, which is the responsibility of the client. Are we ok with this behavior? I tried to follow the the pattern where reading bodies is done with wrapping to minimize allocations. 
* It is not clear to me how the body wrappers work with long lived connections?

Fixes https://github.com/open-telemetry/opentelemetry-go-contrib/issues/3134
Fixes https://github.com/open-telemetry/opentelemetry-go-contrib/issues/1336


instrumentation/net/http/otelhttp/common.go
```diff
@@ -31,10 +31,17 @@ const (
...
 // Server HTTP metrics.
 const (
-	RequestCount          = "http.server.request_count"           // Incoming request count total
-	RequestContentLength  = "http.server.request_content_length"  // Incoming request bytes total
-	ResponseContentLength = "http.server.response_content_length" // Incoming response bytes total
-	ServerLatency         = "http.server.duration"                // Incoming end to end duration, milliseconds
+	ServerRequestCount          = "http.server.request_count"           // Incoming request count total
+	ServerRequestContentLength  = "http.server.request_content_length"  // Incoming request bytes total
+	ServerResponseContentLength = "http.server.response_content_length" // Incoming response bytes total
+	ServerLatency               = "http.server.duration"                // Incoming end to end duration, milliseconds
```
ash2k, line 37: I wonder if these constants need to be public? Technically renaming them is a breaking change. I think the smaller the exposed API surface it, the better it is. WDYT?

instrumentation/net/http/otelhttp/internal/semconvutil/httpconv.go
```diff
@@ -286,6 +294,39 @@ func (c *httpConv) ClientRequest(req *http.Request) []attribute.KeyValue {
...
+	}
+	peer, p := firstHostPort(h, req.Header.Get("Host"))
+	port := requiredHTTPPort(req.URL != nil && req.URL.Scheme == "https", p)
+	if port > 0 {
+		n++
+	}
+
+	attrs := make([]attribute.KeyValue, 0, n)
+	attrs = append(attrs, c.method(req.Method))
+	attrs = append(attrs, c.NetConv.PeerName(peer))
```
ash2k, line 321: This could be a single append call.

instrumentation/net/http/otelhttp/transport.go
```diff
@@ -16,34 +16,45 @@ package otelhttp // import "go.opentelemetry.io/contrib/instrumentation/net/http
 
 import (
 	"context"
+	"go.opentelemetry.io/otel/metric"
```
ash2k, line 19: I wonder if this project has a formatting check in CI...

instrumentation/net/http/otelhttp/transport.go
```diff
@@ -121,27 +174,46 @@ func (t *Transport) RoundTrip(r *http.Request) (*http.Response, error) {
...
+
+	// traces
 	span.SetAttributes(semconvutil.HTTPClientResponse(res)...)
 	span.SetStatus(semconvutil.HTTPClientStatus(res.StatusCode))
-	res.Body = newWrappedBody(span, res.Body)
+
+	res.Body = newWrappedBody(span, readRecordFunc, res.Body)
+
+	// Use floating point division here for higher precision (instead of Millisecond method).
+	elapsedTime := float64(time.Since(requestStartTime)) / float64(time.Millisecond)
```
ash2k, line 196: https://github.com/open-telemetry/opentelemetry-go-contrib/issues/4590

instrumentation/net/http/otelhttp/transport_test.go
```diff
@@ -18,9 +18,16 @@ import (
 	"bytes"
 	"context"
 	"errors"
+	"fmt"
+	"go.opentelemetry.io/otel/attribute"
+	"go.opentelemetry.io/otel/sdk/instrumentation"
+	"go.opentelemetry.io/otel/sdk/metric/metricdata/metricdatatest"
+	semconv "go.opentelemetry.io/otel/semconv/v1.20.0"
```
ash2k, line 25: Inconsistent grouping here too.

ash2k (changes requested):

instrumentation/net/http/otelhttp/transport.go
Sovietaced: I agree with the general desire for this codebase to follow the OTEL specification, but I also see some value in following existing convention and then having a more thought out approach to migrating towards seconds across the entire codebase. 

This is the same reason I stuck with the incorrectly named metrics, so that this code would help reach feature parity with service side metrics

instrumentation/net/http/otelhttp/transport.go
Sovietaced: It might be, I did most of this work on a plane with a 3 kb/s connection so unfortunately I wasn't able to download `golangci-lint` and reproduce the issues locally, I'll try and sort it out

instrumentation/net/http/otelhttp/internal/semconvutil/httpconv.go
Sovietaced: Good point

instrumentation/net/http/otelhttp/common.go
Sovietaced: Oh that is a good point, I forget that other go modules could be relying on these public constants. I don't see why they need to be public though. I'm only familiar with the gRPC metrics and those definitely don't follow this pattern.

instrumentation/net/http/otelhttp/transport_test.go
```diff
@@ -19,17 +19,27 @@ import (
...
+
 	"github.com/stretchr/testify/assert"
 	"github.com/stretchr/testify/require"
 
 	"go.opentelemetry.io/otel/codes"
 	"go.opentelemetry.io/otel/propagation"
 	"go.opentelemetry.io/otel/trace"
+
+	"go.opentelemetry.io/otel/sdk/metric"
+	"go.opentelemetry.io/otel/sdk/metric/metricdata"
```
MadVikingGod, line 42: Whichever test uses the SDK needs to be in the `test` directory.  To prevent this from depending on the SDK.

instrumentation/net/http/otelhttp/transport.go
ash2k: Thank you. I'll leave this up to maintainers to decide.

instrumentation/net/http/otelhttp/common.go
ash2k: Another data point here - this module (`go.opentelemetry.io/contrib/instrumentation/net/http/otelhttp`) is at `v0.46.1` so making breaking changes is still allowed. So we better get it in shape before it reaches 1.0. This includes hiding/removing everything that doesn't need to be exposed, renaming metrics/method/etc, changing units, etc. So now might be a good time as any to make these private, while it's not too late.

instrumentation/net/http/otelhttp/transport_test.go
Sovietaced: Fixed in 1a4c8cb51c3ecbc82a27bbe897f74e545b051252

instrumentation/net/http/otelhttp/common.go
lenta-platform: At the moment there is global instrumentration vs semconv naming mismath, e.g. "http.client.duration" vs "http.client.request.duration", "http.server.duration" vs "http.server.request.duration" and so on.

Spec: https://opentelemetry.io/docs/specs/semconv/http/http-metrics/#metric-httpclientrequestduration

Will we drop semconv, is it outdated?

pellared:
> I don't believe the code will report HTTP response metrics unless the response body is read, which is the responsibility of the client. Are we ok with this behavior? I tried to follow the the pattern where reading bodies is done with wrapping to minimize allocations.

We are fine. From https://pkg.go.dev/net/http:

> The caller must close the response body when finished with it:
> ...
> Caller should close resp.Body when done reading from it.

Also we already have this requirement for trace instrumentation.

Side note: probably we should call it out in package documentation.

pellared:
> As written these client metrics follow the convention of the server metrics, but don't actually follow the convention of the OTEL spec as noted in https://github.com/open-telemetry/opentelemetry-go-contrib/issues/4501. I guess this migration can be done later?

We want to first be be compliant with the [v1.20 semantic conventions](https://github.com/open-telemetry/opentelemetry-specification/blob/v1.20.0/specification/metrics/semantic_conventions/http-metrics.md#http-client).  _Next semantic convention versions have many breaking changes and we want to help a little with the migration._

instrumentation/net/http/otelhttp/common.go
pellared: These constants should live in `semconv` packages. See: https://github.com/open-telemetry/opentelemetry-go/issues/4528

I would rather not export it in `otelhttp` as I see no big value and it would increase the API surface.

instrumentation/net/http/otelhttp/transport.go
pellared: It should be `ms` like defiened in https://github.com/open-telemetry/opentelemetry-specification/blob/v1.20.0/specification/metrics/semantic_conventions/http-metrics.md#http-client. Resolving.

instrumentation/net/http/otelhttp/common.go
```diff
@@ -31,10 +31,17 @@ const (
...
+)
+
+// Client HTTP metrics.
+const (
+	ClientRequestContentLength  = "http.client.request_content_length"  // Outgoing request bytes total
+	ClientResponseContentLength = "http.client.response_content_length" // Outgoing response bytes total
```
pellared, line 43: These do not match https://github.com/open-telemetry/opentelemetry-specification/blob/v1.20.0/specification/metrics/semantic_conventions/http-metrics.md#http-client

Can we also fix the metrics for HTTP server instrumentation?

instrumentation/net/http/otelhttp/transport.go
```diff
@@ -63,19 +75,45 @@ func NewTransport(base http.RoundTripper, opts ...Option) *Transport {
...
+	)
+	handleErr(err)
+
+	t.latencyMeasure, err = t.meter.Float64Histogram(
+		ClientLatency,
+		metric.WithUnit("ms"),
+		metric.WithDescription("Measures the duration of HTTP request handling"),
+	)
+	handleErr(err)
+}
```
pellared, line 115: nit: Can we use the descriptions from https://github.com/open-telemetry/opentelemetry-specification/blob/v1.20.0/specification/metrics/semantic_conventions/http-metrics.md#http-client? Just I would just get rid of "(compressed)".

instrumentation/net/http/otelhttp/common.go
```diff
@@ -31,10 +31,17 @@ const (
 
 // Server HTTP metrics.
 const (
-	RequestCount          = "http.server.request_count"           // Incoming request count total
-	RequestContentLength  = "http.server.request_content_length"  // Incoming request bytes total
-	ResponseContentLength = "http.server.response_content_length" // Incoming response bytes total
-	ServerLatency         = "http.server.duration"                // Incoming end to end duration, milliseconds
+	ServerRequestCount          = "http.server.request_count"           // Incoming request count total
```
pellared, line 34: Could we update the `otelhttp` package documentation to include the information about the emited metrics? Something like https://pkg.go.dev/go.opentelemetry.io/contrib/instrumentation/runtime

This can be addressed as a separate issue and PR.

pellared:
Please add a changelog entry.

Sovietaced:
> Please add a changelog entry.

Will do, I didn't add one since you seemed a bit picky about the changelog entries on my last PR :)

instrumentation/net/http/otelhttp/common.go
Sovietaced: I will defer that to a separate issue/PR

instrumentation/net/http/otelhttp/test/transport_test.go
```diff
@@ -238,3 +248,117 @@ func TestWithHTTPTrace(t *testing.T) {
...
+	hist := dur.Data.(metricdata.Histogram[float64])
+	assert.Equal(t, metricdata.CumulativeTemporality, hist.Temporality)
+	require.Len(t, hist.DataPoints, 1)
+	dPt := hist.DataPoints[0]
+	assert.Equal(t, attrs, dPt.Attributes, "attributes")
+	assert.Equal(t, uint64(1), dPt.Count, "count")
+	assert.Equal(t, []float64{0, 5, 10, 25, 50, 75, 100, 250, 500, 750, 1000, 2500, 5000, 7500, 10000}, dPt.Bounds, "bounds")
+
+	// Duration value is not deterministic because the code does not support a pluggable clock.
+	// So just value that one of the buckets has been incremented.
```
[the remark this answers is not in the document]
Sovietaced: 0c097201a95fe872a4b19ba097f095cdae4353a7

instrumentation/net/http/otelhttp/common.go
Sovietaced: 0c097201a95fe872a4b19ba097f095cdae4353a7

instrumentation/net/http/otelhttp/transport.go
Sovietaced: 0c097201a95fe872a4b19ba097f095cdae4353a7

instrumentation/net/http/otelhttp/common.go
Sovietaced: https://github.com/open-telemetry/opentelemetry-go-contrib/issues/4774

instrumentation/net/http/otelhttp/common.go
```diff
@@ -31,10 +31,16 @@ const (
...
+	ServerRequestSize  = "http.server.request.size"  // Incoming request bytes total
+	ServerResponseSize = "http.server.response.size" // Incoming response bytes total
+	ServerDuration     = "http.server.duration"      // Incoming end to end duration, milliseconds
+)
+
+// Client HTTP metrics.
+const (
+	ClientRequestSize  = "http.client.request.size"  // Outgoing request bytes total
+	ClientResponseSize = "http.client.response.size" // Outgoing response bytes total
+	ClientDuration     = "http.client.duration"      // Outgoing end to end duration, milliseconds
```
pellared, line 43: Can we unexport these constants and add a changelog entry in "Removed" section?

```
- Remove `RequestCount`, `RequestContentLength`, `ResponseContentLength`, `ServerLatency` constants from `go.opentelemetry.io/contrib/instrumentation/net/http/otelhttp`. (#4707)
```

pellared (dismissed):

instrumentation/net/http/otelhttp/common.go
SodaDev, line 43: I guess it is not aligned with semconv from OTEL: https://opentelemetry.io/docs/specs/semconv/http/http-metrics/#http-client

instrumentation/net/http/otelhttp/internal/semconvutil/httpconv.go
```diff
@@ -286,6 +294,38 @@ func (c *httpConv) ClientRequest(req *http.Request) []attribute.KeyValue {
 	return attrs
 }
 
+// ClientRequestMetrics returns metric attributes for an HTTP request made by a client. The
+// following attributes are always returned: "http.method", "net.peer.name".
+// The following attributes are returned if the related values
+// are defined in req: "net.peer.port".
+func (c *httpConv) ClientRequestMetrics(req *http.Request) []attribute.KeyValue {
```
SodaDev, line 301: Maybe it could be replaced with
semconv.HTTPClientAttributesFromHTTPRequest(req) ?

pellared: The `semconv` does not have such function.

SodaDev: yes, it was deprecated at version 12. It was replaced with httpconv.ClientRequest(req) from "go.opentelemetry.io/otel/semconv/v1.20.0/httpconv"

instrumentation/net/http/otelhttp/common.go
Sovietaced: 233049a5211e6aebcbdd90d6fd4fe4d50e114f49

pellared: Server* constans have to be unexported as well.

Sovietaced: My bad, I misunderstood. Will take care of that when I get back to the computer.

pellared (approved):

Sovietaced:
@dmathieu @Aneurysm9 @dashpole @hanyuancheung Please take a look when you have a chance. TY

instrumentation/google.golang.org/grpc/otelgrpc/example/api/hello-service.pb.go
```diff
@@ -5,9 +5,11 @@
 Package api is a generated protocol buffer package.
 
 It is generated from these files:
+
```
dmathieu, line 8: Why are those new lines here? They seem unrelated to this PR.

Sovietaced: Hmm. Not sure. Let me back those out

Sovietaced: Fixed in f54fc6c7805342222601f1edfc27d94d642fc9a0

dmathieu (approved):

instrumentation/net/http/otelhttp/transport.go
```diff
@@ -63,19 +75,45 @@ func NewTransport(base http.RoundTripper, opts ...Option) *Transport {
...
 	t.spanNameFormatter = c.SpanNameFormatter
 	t.clientTrace = c.ClientTrace
 }
 
+func (t *Transport) createMeasures() {
+	var err error
+	t.requestBytesCounter, err = t.meter.Int64Counter(
+		clientRequestSize,
+		metric.WithUnit("By"),
+		metric.WithDescription("Measures the size of HTTP request messages."),
```
dashpole, line 98: Should this be `Measures the size of HTTP request messages (compressed).`?

https://github.com/open-telemetry/opentelemetry-specification/blob/855d80c93ffc11ee7c96177be8082989afb020f5/semantic_conventions/metrics/http.yaml#L56C13-L56C69

dashpole: Same for response below, and for server request/response.

pellared: I noticed that it was `(uncompressed)` before. Isn't `http` package handling the transport compression and therefore we get the uncompressed message size value?

dashpole: Ah.  Makes sense.  We aren't really correctly implementing this metric then, are we?

pellared: This is how I see it. I was not checking if it is possible to implement it correctly.

dashpole (approved):

Sovietaced:
@pellared is there anything else blocking this from being merged?

pellared:
@Sovietaced No blockers. I am just giving time for @MadVikingGod until end of week to review it.

instrumentation/net/http/otelhttp/transport.go
```diff
@@ -171,11 +246,15 @@ func (wb *wrappedBody) Write(p []byte) (int, error) {
...
 	n, err := wb.body.Read(b)
+	// Locally record the number of bytes read
+	wb.read.Add(int64(n))
 
 	switch err {
 	case nil:
 		// nothing to do here but fall through to the return
 	case io.EOF:
+		// Record the total number of bytes read
+		wb.record(wb.read.Load())
```
MadVikingGod, line 257: Currently, this will only record on an EOF.  Is there a way to record on an EOF or a `close()` but only record once?

The use case I'm thinking of is if a client reads part-way and is canceled.

instrumentation/net/http/otelhttp/common.go
```diff
@@ -31,10 +31,16 @@ const (
 
 // Server HTTP metrics.
 const (
-	RequestCount          = "http.server.request_count"           // Incoming request count total
-	RequestContentLength  = "http.server.request_content_length"  // Incoming request bytes total
-	ResponseContentLength = "http.server.response_content_length" // Incoming response bytes total
-	ServerLatency         = "http.server.duration"                // Incoming end to end duration, milliseconds
+	serverRequestSize  = "http.server.request.size"  // Incoming request bytes total
+	serverResponseSize = "http.server.response.size" // Incoming response bytes total
+	serverDuration     = "http.server.duration"      // Incoming end to end duration, milliseconds
```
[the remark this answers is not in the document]
ash2k: `go.opentelemetry.io/contrib/instrumentation/net/http/otelhttp` is at `v0.46.1`. If now is not the right moment to make breaking changes, then... when is? I may be wrong, but I thought breaking changes are ok and had happened in `v0.x.x` modules here before. 🤔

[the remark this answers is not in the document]
Sovietaced: I originally tried not to include any breaking changes but got guidance from @pellared to update these: https://github.com/open-telemetry/opentelemetry-go-contrib/pull/4707#issuecomment-1871335266

instrumentation/net/http/otelhttp/transport.go
Sovietaced: Good point, I will add a test case for that and update the code.

Sovietaced: Addressed in 0411674549f801f9aa02f2c91a46dc6967854a03. Let me know how that looks :)

instrumentation/net/http/otelhttp/common.go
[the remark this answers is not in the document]
pellared: @MadVikingGod

instrumentation/net/http/otelhttp/transport.go
Sovietaced: ping @MadVikingGod

hanyuancheung (approved):

instrumentation/net/http/otelhttp/transport.go
```diff
@@ -109,7 +148,22 @@ func (t *Transport) RoundTrip(r *http.Request) (*http.Response, error) {
...
+	writeRecordFunc := func(int64) {}
+	var bw bodyWrapper
+	// if request body is nil or NoBody, we don't want to mutate the body as it
+	// will affect the identity of it in an unforeseeable way because we assert
+	// ReadCloser fulfills a certain interface and it is indeed nil or NoBody.
+	if r.Body != nil && r.Body != http.NoBody {
+		bw.ReadCloser = r.Body
+		bw.record = writeRecordFunc
+		r.Body = &bw
+	}
```
MadVikingGod, line 165: Why are we wrapping the body, with a recorder that does nothing?

instrumentation/net/http/otelhttp/transport.go
```diff
@@ -121,27 +175,46 @@ func (t *Transport) RoundTrip(r *http.Request) (*http.Response, error) {
...
+	readRecordFunc := func(n int64) {
+		t.responseBytesCounter.Add(ctx, n, o)
+	}
+
+	// traces
 	span.SetAttributes(semconvutil.HTTPClientResponse(res)...)
 	span.SetStatus(semconvutil.HTTPClientStatus(res.StatusCode))
-	res.Body = newWrappedBody(span, res.Body)
+
+	res.Body = newWrappedBody(span, readRecordFunc, res.Body)
```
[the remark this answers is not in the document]
Sovietaced: There was a comment about not using the content length in the previous pull request for this feature here: https://github.com/open-telemetry/opentelemetry-go-contrib/pull/3769#discussion_r1333577399

instrumentation/net/http/otelhttp/transport.go
MadVikingGod: I'm saying that if we don't do anything in the record method, we shouldn't wrap it.  
That is to allow us to record a metric of how much was read.

instrumentation/net/http/otelhttp/common.go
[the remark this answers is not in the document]
MadVikingGod: Yes, we can make breaking changes, but no, we shouldn't do this lightly.

We need to deprecate these constants before we pull them out.  Pulling them out can make some user's code incompatible with newer versions of this package (and potentially the broader ecosystem).  While I do think that there is a low chance this will break any of the [827](https://pkg.go.dev/go.opentelemetry.io/contrib/instrumentation/net/http/otelhttp) that import this library, we should act carefully around breaking changes.

instrumentation/net/http/otelhttp/transport.go
Sovietaced: apologies, you're absolutely right. this is some left over cruft from the old PR where it was being used to record events. Will remove this shortly.

Sovietaced: We aren't doing anything in the record _callback_, however we are still using the wrapper to pull the number of bytes that were read.

instrumentation/net/http/otelhttp/common.go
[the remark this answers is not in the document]
Sovietaced: Added back and added docs about deprecation here: 70c1791e8dbdcd7324c5b1c78645b3e4a2b1d695

MadVikingGod (approved):

Sovietaced:
@pellared should be good to go

instrumentation/net/http/otelhttp/transport_test.go
```diff
@@ -294,7 +294,7 @@ func TestWrappedBodyClose(t *testing.T) {
 	wb := &wrappedBody{span: trace.Span(s), record: record, body: readCloser{}}
 	assert.NoError(t, wb.Close())
 	s.assert(t, true, nil, codes.Unset, "")
-	assert.False(t, called, "record should not have been called")
+	assert.True(t, called, "record should not have been called")
```
pellared, line 297:
```suggestion
	assert.True(t, called, "record should have been called")
```

instrumentation/net/http/otelhttp/transport_test.go
```diff
@@ -312,7 +312,7 @@ func TestWrappedBodyCloseError(t *testing.T) {
 	wb := &wrappedBody{span: trace.Span(s), record: record, body: readCloser{closeErr: expectedErr}}
 	assert.Equal(t, expectedErr, wb.Close())
 	s.assert(t, true, nil, codes.Unset, "")
-	assert.False(t, called, "record should not have been called")
+	assert.True(t, called, "record should not have been called")
```
pellared, line 315:
```suggestion
	assert.True(t, called, "record should have been called")
```

instrumentation/net/http/otelhttp/transport.go
```diff
@@ -263,7 +264,16 @@ func (wb *wrappedBody) Read(b []byte) (int, error) {
 	return n, err
 }
 
+// recordBytesRead is a function that ensures the number of bytes read is recorded once and only once.
+func (wb *wrappedBody) recordBytesRead() {
+	if wb.recorded.CompareAndSwap(false, true) {
```
pellared, line 269: Maybe add a comment here that we do not need to use `sync.Once` as it is OK that have "eventual consistency"? Basically if there are two goroutines running in parallel that call `recordBytesRead`, one can return before the second called `wb.record(wb.read.Load())`. It is OK and using `CompareAndSwap` instead of `sync.Once` is a performance optimization, but I think it is worth adding a comment.

pellared (commented):
Little comments that I would like to have addressed. Besides looks good 👍

instrumentation/net/http/otelhttp/transport.go
Sovietaced: Interesting, I'm relatively new to go so I had not known about `sync.Once`. I just wanted semantics close to Java's atomic boolean so that is why I chose `atomic.bool`. Anyway, added comments in 38861442146dafb02345a3cd06d6bd96030c77f6

gaiaz-iusipov:
It would be nice to have something similar to `otelhttp.WithRouteTag` for the transport, thus allowing the separation of different requests in metrics. For example: `otelhttp.WithRequestTag`.

pellared:
@Sovietaced Thank you for your contribution

Sovietaced:
> It would be nice to have something similar to `otelhttp.WithRouteTag` for the transport, thus allowing the separation of different requests in metrics. For example: `otelhttp.WithRequestTag`.

Maybe create a separate issue for this?

StarpTech:
Hi, it would be great if new features like metrics are built behind a feature flag. For example, we use the middleware for everything expect metrics. As a workaround, we pass a Noop meter.

instrumentation/net/http/otelhttp/test/handler_test.go
```diff
@@ -50,8 +50,8 @@ func assertScopeMetrics(t *testing.T, sm metricdata.ScopeMetrics, attrs attribut
 	require.Len(t, sm.Metrics, 3)
 
 	want := metricdata.Metrics{
-		Name:        "http.server.request_content_length",
-		Description: "Measures the size of HTTP request content length (uncompressed)",
+		Name:        "http.server.request.size",
```
StarpTech, line 53: This looks to me like a breaking change. Why wasn't it communicated as such?

dashpole: @SharpTech sorry about that.  I've opened https://github.com/open-telemetry/opentelemetry-go-contrib/pull/4898 to add it to the changelog.

StarpTech: Hi @dashpole wouldn't it be better to rollback so the deprecation statement remain true?

dashpole: The deprecation statements refer to these constants: https://github.com/open-telemetry/opentelemetry-go-contrib/pull/4707/files#diff-be9e2412d1919396470e196d2be4bddf6997059892167fd8381a9cfa478933faR34

> I also think breaking stuff with a feature release isn't the right approach.

The otelhttp module is not stable, and is still released under a v0.* version, so breaking changes are expected.  You can see the versioning policy here: https://github.com/open-telemetry/opentelemetry-go/blob/main/VERSIONING.md.  The telemetry in otelhttp follows the semantic conventions for http metrics: https://opentelemetry.io/docs/specs/semconv/http/http-metrics/, so breaking changes to conventions require us to make breaking changes here.

StarpTech: I understand, but users like me running this in production. We need to deprecate fields before we can upgrade. If I would want to use the new metrics I couldn't. I'm saying the feature and the breaking change could have been shipped in two PR  (releases).

PR #4617, opened 2023-11-24
config: add meter/tracer provider config

codeboten:
This pulls in the code that was in the collector to configure the MeterProvider and TracerProvider.

Fixes #4371

PR #4741, opened 2023-12-21
config: NewSDK can return valid TracerProvider

codeboten:
This PR is another attempt at filling in the SDK details for NewSDK. It breaks the #4617 per signal to make it more manageable to review.

Part of #4371


config/config_test.go
```diff
@@ -51,3 +51,7 @@ func TestNewSDK(t *testing.T) {
 		require.Equal(t, tt.wantShutdownErr, sdk.Shutdown(context.Background()))
 	}
 }
+
+func toPtr[V int | string](in V) *V {
```
pellared, line 55: Can we rename it to `ptr`? Simiarly in Go we have `String` and not e.g. `ToString`.

Is there any reason why we constrain the type parameters to only `int` or `string`?

```suggestion
func ptr[T any](v T) *T {
	return &v
}
```
```

File: config/trace.go [hunk 2]
```diff
@@ -4,15 +4,174 @@
...
+	}
+
+	return otlptracegrpc.New(ctx, opts...)
+}
+
+func initOTLPHTTPSpanExporter(ctx context.Context, otlpConfig *OTLP) (sdktrace.SpanExporter, error) {
+	opts := []otlptracehttp.Option{}
+
+	if len(otlpConfig.Endpoint) > 0 {
+		u, err := url.ParseRequestURI(normalizeEndpoint(otlpConfig.Endpoint))
```
pellared, line 117: `normalizeEndpoint` should not be used here.

It would break https://github.com/open-telemetry/opentelemetry-specification/blob/main/specification/protocol/exporter.md#endpoint-urls-for-otlphttp:

> An SDK MUST NOT modify the URL in ways other than specified above.

CHANGELOG.md
```diff
@@ -12,6 +12,7 @@ This project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0.htm
 
 - Add the new `go.opentelemetry.io/contrib/instrgen` package to provide auto-generated source code instrumentation. (#3068, #3108)
 - Add `SDK.Shutdown` method in `"go.opentelemetry.io/contrib/config"`. (#4583)
+- The `config.NewSDK` function now returns a configured SDK with a valid `TracerProvider`. (#4741)
```
pellared, line 15:
```suggestion
- `NewSDK` in `go.opentelemetry.io/contrib/config` now returns a configured SDK with a valid `TracerProvider`. (#4741)
```

pellared (commented):
Side note: I have not managed to find time to make a complete review, but I think it is worth to give some feedback.

config/trace.go
```diff
@@ -4,15 +4,174 @@
...
+		}
+		opts = append(opts, otlptracegrpc.WithEndpoint(u.Host))
+		if u.Scheme == "http" {
+			opts = append(opts, otlptracegrpc.WithInsecure())
+		}
+	}
+
+	if otlpConfig.Compression != nil {
+		switch *otlpConfig.Compression {
+		case "gzip":
```
pellared, line 95: Should we add consts like for `protocolProtobufHTTP`? Take notice it would be reused in `initOTLPHTTPSpanExporter`.

config/trace.go
```diff
@@ -4,15 +4,174 @@
...
+			opts = append(opts, otlptracegrpc.WithCompressor(*otlpConfig.Compression))
+		case "none":
```
pellared, line 97: I think we can have a const for `none` which will probably be used in more places.

config/trace_test.go
```diff
@@ -25,8 +35,413 @@ func TestInitTracerPovider(t *testing.T) {
...
+}
+
+func TestSpanProcessor(t *testing.T) {
+	consoleExporter, err := stdouttrace.New(
+		stdouttrace.WithPrettyPrint(),
+	)
+	require.NoError(t, err)
+	otlpGRPCExporter, err := otlptracegrpc.New(context.TODO())
+	require.NoError(t, err)
+	otlpHTTPExporter, err := otlptracehttp.New(context.TODO())
```
pellared, line 52:
```suggestion
	ctx := context.Background()
	otlpGRPCExporter, err := otlptracegrpc.New(ctx)
	require.NoError(t, err)
	otlpHTTPExporter, err := otlptracehttp.New(ctx)
```

config/trace.go
```diff
@@ -4,15 +4,174 @@
...
+	"go.opentelemetry.io/otel/sdk/resource"
 	sdktrace "go.opentelemetry.io/otel/sdk/trace"
 	"go.opentelemetry.io/otel/trace"
 	"go.opentelemetry.io/otel/trace/noop"
 )
 
-func initTracerProvider(cfg configOptions) (trace.TracerProvider, shutdownFunc) {
+var errNoValidSpanExporter = errors.New("no valid span exporter")
+
+func initTracerProvider(cfg configOptions, res *resource.Resource) (trace.TracerProvider, shutdownFunc, error) {
```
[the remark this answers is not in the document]
codeboten: done

config/trace.go
```diff
@@ -4,15 +4,174 @@
...
+		opts = append(opts, otlptracegrpc.WithTimeout(time.Millisecond*time.Duration(*otlpConfig.Timeout)))
+	}
+	if len(otlpConfig.Headers) > 0 {
+		opts = append(opts, otlptracegrpc.WithHeaders(otlpConfig.Headers))
+	}
+
+	return otlptracegrpc.New(ctx, opts...)
+}
+
+func initOTLPHTTPSpanExporter(ctx context.Context, otlpConfig *OTLP) (sdktrace.SpanExporter, error) {
```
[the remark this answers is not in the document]
codeboten: done

config/trace.go
```diff
@@ -4,15 +4,174 @@
...
+		if err != nil {
+			return nil, err
+		}
+		return sdktrace.NewSimpleSpanProcessor(exp), nil
+	}
+	return nil, fmt.Errorf("unsupported span processor type %v", processor)
+}
+
+func initOTLPgRPCSpanExporter(ctx context.Context, otlpConfig *OTLP) (sdktrace.SpanExporter, error) {
+	opts := []otlptracegrpc.Option{}
```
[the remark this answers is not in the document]
codeboten: fixed 👍🏻

config/trace.go
codeboten: added

config/trace.go
codeboten: added

config/trace.go
```diff
@@ -4,15 +4,174 @@
...
+	"go.opentelemetry.io/otel/exporters/otlp/otlptrace/otlptracehttp"
+	"go.opentelemetry.io/otel/exporters/stdout/stdouttrace"
+	"go.opentelemetry.io/otel/sdk/resource"
 	sdktrace "go.opentelemetry.io/otel/sdk/trace"
 	"go.opentelemetry.io/otel/trace"
 	"go.opentelemetry.io/otel/trace/noop"
 )
 
-func initTracerProvider(cfg configOptions) (trace.TracerProvider, shutdownFunc) {
+var errNoValidSpanExporter = errors.New("no valid span exporter")
```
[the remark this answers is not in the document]
codeboten: i thought it would be convenient for testing, but since its only used in one place, not a big benefit. removed

config/trace.go
```diff
@@ -4,15 +4,174 @@
...
+		}
+		opts = append(opts, sdktrace.WithSpanProcessor(sp))
+	}
+	tp := sdktrace.NewTracerProvider(opts...)
+	return tp, tp.Shutdown, nil
+}
+
+func spanExporter(ctx context.Context, exporter SpanExporter) (sdktrace.SpanExporter, error) {
+	if exporter.Console != nil {
+		return stdouttrace.New(
```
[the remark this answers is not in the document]
codeboten: json schema validation for configuration that sets multiple exporters would fail, but it's true that it would be possible for someone to set multiple values through the model. i could add a check here for multiple exporters configured? WDYT?

config/trace.go
```diff
@@ -4,15 +4,172 @@
...
+		return noop.NewTracerProvider(), noopShutdown, nil
+	}
+	opts := []sdktrace.TracerProviderOption{
+		sdktrace.WithResource(res),
+	}
+	for _, processor := range cfg.opentelemetryConfig.TracerProvider.Processors {
+		sp, err := spanProcessor(cfg.ctx, processor)
+		if err != nil {
+			return noop.NewTracerProvider(), noopShutdown, err
+		}
```
MadVikingGod, line 33: Suggestion: Nothing is worse than fixing one error for a new one to appear.  Could we move the return the error case to be something that combines all the errors?

config/trace.go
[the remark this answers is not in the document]
MadVikingGod: I think this would be a good check to add.

config/trace.go
```diff
@@ -4,15 +4,172 @@
...
+			return otlpGRPCSpanExporter(ctx, exporter.OTLP)
+		default:
+			return nil, fmt.Errorf("unsupported protocol %q", exporter.OTLP.Protocol)
+		}
+	}
+	return nil, errors.New("no valid span exporter")
+}
+
+func spanProcessor(ctx context.Context, processor SpanProcessor) (sdktrace.SpanProcessor, error) {
+	if processor.Batch != nil {
```
MadVikingGod, line 60: Same as above:  can we check that we didn't didn't set both?

MadVikingGod (approved):

config/trace.go
codeboten: check added 👍🏻

config/trace.go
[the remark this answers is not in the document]
codeboten: added the check and a test for it

config/trace.go
codeboten: @MadVikingGod i've updated this to create a slice of errs and return using errors.Join if it's not empty, please take a look

config/trace.go
pellared: @codeboten What about this comment?

codeboten: yes I haven't addressed it yet. will do so today

codeboten: @pellared removed the function

pellared: Sorry, but I my comment was not very explicit. The function should not be used in **this** place. I think it is required for parsing the endpoint for gRPC protocol 😬

codeboten: I ended up doing updating the grpc case to try parsing the URL, and if the scheme isn't present (which helpfully `ParseRequestURI` puts the host in the place of the scheme 😆) it will use the full endpoint string instead of u.Host.

So for example if endpoint is `http://localhost:4317`, the following options are set:

```
WithEndpoint("localhost:4317")
WithInsecure()
```

If the endpoint string is `localhost:4317`, the following option is set:

```
WithEndpoint("localhost:4317")
```

I think this makes sense, because someone may want to set `localhost:4317` and have it be a secure option. WDYT?

pellared: I see. Can you just add comments to the gRPC parsing case as this is hacky and it will be hard to understand in future?

config/trace.go
```diff
@@ -94,6 +94,11 @@ func otlpGRPCSpanExporter(ctx context.Context, otlpConfig *OTLP) (sdktrace.SpanE
 		if err != nil {
 			return nil, err
 		}
+		// ParseRequestURI leaves the Host field empty when no
+		// scheme is specified (i.e. localhost:4317). This check is
+		// here to support the case where a user may not specify a
+		// scheme. The code does its best effort here by using
+		// otlpConfig.Endpoint as-is in that case
```
pellared, line 101:
```suggestion
		// otlpConfig.Endpoint as-is in that case.
```

pellared (approved):

config/config.go
```diff
@@ -101,3 +119,13 @@ func WithOpenTelemetryConfiguration(cfg OpenTelemetryConfiguration) Configuratio
...
 // - https://github.com/open-telemetry/opentelemetry-go-contrib/issues/4371
+
+func newResource(res *Resource) (*resource.Resource, error) {
+	if res == nil {
+		return resource.Default(), nil
+	}
+	return resource.Merge(resource.Default(),
+		resource.NewWithAttributes(semconv.SchemaURL,
+			semconv.ServiceName(*res.Attributes.ServiceName),
+		))
```
pellared, line 130: How about moving to `resource.go` and adding tests in `resource_test.go`?

Can we do it in a separate PR?

pellared (approved):
I just have one comment before merging.

config/config.go
codeboten: Sure, I'll move this next before the meter provider PR

pellared: https://github.com/open-telemetry/opentelemetry-go-contrib/issues/4821

PR #4804, opened 2024-01-11
config: NewSDK can return valid MeterProvider

### LINKED ISSUE #4371: config: Create SDK from model

codeboten:
Follow up to #4741, does the same but for metric signal.
    
Fixes #4371


config/metric.go
```diff
@@ -4,15 +4,257 @@
...
+		opts = append(opts, otlpmetricgrpc.WithTimeout(time.Millisecond*time.Duration(*otlpConfig.Timeout)))
+	}
+	if len(otlpConfig.Headers) > 0 {
+		opts = append(opts, otlpmetricgrpc.WithHeaders(otlpConfig.Headers))
+	}
+
+	return otlpmetricgrpc.New(ctx, opts...)
+}
+
+func prometheusReader(ctx context.Context, prometheusConfig *Prometheus) (sdkmetric.Reader, error) {
```
MadVikingGod, line 193: I would just like to confirm this, not that it is a blocker or not.

Is there a way to use the config mechanism and add a route to an existing HTTP server?  Or will this always be a separate prometheus instance and HTTP server?

codeboten: it's a good question, there's only host and port defined as parameters in the configuration today

MadVikingGod: This is tied with the prom config.  If there isn't a host, we could instead use the [Default Prometheus instance](https://pkg.go.dev/github.com/prometheus/client_golang/prometheus#pkg-variables).  This would let users start the server on their own.

MadVikingGod: I talked with Codeboten offline, and I think this is an acceptable approach.
We will start by requiring it initially; this lets us get this going quicker.  If there is a demand for using the already configured Default Prometheus then we can relax this requirement.

MadVikingGod (approved):

pellared:
@codeboten Can you please update the branch? I will do my best to review the PR this week.

hanyuancheung (approved):

config/metric.go
```diff
@@ -4,15 +4,257 @@
...
+	if exporter.OTLP != nil {
+		var err error
+		var exp sdkmetric.Exporter
+		switch exporter.OTLP.Protocol {
+		case protocolProtobufHTTP:
+			exp, err = otlpHTTPMetricExporter(ctx, exporter.OTLP)
+		case protocolProtobufGRPC:
+			exp, err = otlpGRPCMetricExporter(ctx, exporter.OTLP)
+		default:
+			return nil, fmt.Errorf("unsupported protocol %s", exporter.OTLP.Protocol)
```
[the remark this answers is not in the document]
codeboten: @pellared will take a look at this next week.

dashpole (approved):

config/metric.go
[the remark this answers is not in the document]
codeboten: @pellared test added

pellared (approved):

config/metric.go
```diff
@@ -4,15 +4,257 @@
...
+		ReadTimeout:  5 * time.Second,
+		WriteTimeout: 10 * time.Second,
+		IdleTimeout:  120 * time.Second,
+		Handler:      mux,
+	}
+	addr := fmt.Sprintf("%s:%d", *prometheusConfig.Host, *prometheusConfig.Port)
+
+	// TODO: add support for constant label filter
+	// 	otelprom.WithResourceAsConstantLabels(attribute.NewDenyKeysFilter()),
+	// )
```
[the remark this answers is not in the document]
codeboten: Done https://github.com/open-telemetry/opentelemetry-go-contrib/issues/5381

PR #4805, opened 2024-01-12
otelmongo: Add command.duration metric

devnev:


instrumentation/go.mongodb.org/mongo-driver/mongo/otelmongo/mongo.go
```diff
@@ -21,89 +21,103 @@ import (
...
 	}
 
 	if err != nil {
-		span.SetStatus(codes.Error, err.Error())
+		request.span.SetStatus(codes.Error, err.Error())
+		request.attrs = append(request.attrs, semconv.OtelStatusCodeError)
+	} else {
+		request.attrs = append(request.attrs, semconv.OtelStatusCodeOk)
 	}
+	m.durations.Record(ctx, evt.Duration.Seconds()/1000, metric.WithAttributeSet(attribute.NewSet(request.attrs...)))
```
dashpole: `evt.Duration.Seconds()` is already in seconds, so dividing by `1000` records thousandths of a second rather than milliseconds. Multiply by `1000` for `ms` (or use `evt.Duration.Seconds()` Directly if switching the unit to `s`):
```suggestion
	m.durations.Record(ctx, evt.Duration.Seconds()*1000, metric.WithAttributeSet(attribute.NewSet(request.attrs...)))
```

instrumentation/go.mongodb.org/mongo-driver/mongo/otelmongo/mongo.go
```diff
@@ -135,9 +149,14 @@ func extractCollection(evt *event.CommandStartedEvent) (string, error) {
 // NewMonitor creates a new mongodb event CommandMonitor.
 func NewMonitor(opts ...Option) *event.CommandMonitor {
 	cfg := newConfig(opts...)
+	histogram, err := cfg.Meter.Float64Histogram("command.duration", metric.WithUnit("ms"), metric.WithDescription("Duration of finished commands"))
```
dashpole: To align with OpenTelemetry database metric semantic conventions (and `otelmongo` v2), could we name this metric `db.client.operation.duration` instead of `command.duration`?

Makefile
```diff
@@ -259,15 +259,16 @@ test-mongo-driver:
...
 	  CMD=mongo IMG_NAME=mongo-integ ./tools/wait.sh; \
-	  (cd instrumentation/go.mongodb.org/mongo-driver/mongo/otelmongo/test && \
-	    $(GO) test \
-		  -covermode=$(COVERAGE_MODE) \
-		  -coverprofile=$(COVERAGE_PROFILE) \
-		  -coverpkg=go.opentelemetry.io/contrib/instrumentation/go.mongodb.org/mongo-driver/mongo/otelmongo/...  \
-		  ./... \
-	    && $(GO) tool cover -html=$(COVERAGE_PROFILE) -o coverage.html); \
+	  trap 'docker stop mongo-integ' EXIT; \
+	  pushd instrumentation/go.mongodb.org/mongo-driver/mongo/otelmongo/test; \
```
dashpole: `pushd` and `popd` are bashisms and will fail when `make` runs recipes under `/bin/sh` (`dash` on Debian/Ubuntu). Keeping `(cd ... && ...)` with the `trap` above will stay POSIX-compatible.

PR #4806, opened 2024-01-13
config: Parse model from JSON

### LINKED ISSUE #4412: config: Parse model from JSON
Implement https://github.com/open-telemetry/opentelemetry-specification/blob/main/specification/configuration/file-configuration.md#parse for JSON.

carrbs:
resolves #4412

Implements parsing of a JSON configuration file into an OpenTelemetryConfiguration, and tests it.

### DIFF
--- a/config/config.go
+++ b/config/config.go
@@ -5,7 +5,9 @@ package config // import "go.opentelemetry.io/contrib/config"
 
 import (
 	"context"
+	"encoding/json"
 	"errors"
+	"io"
 
 	"go.opentelemetry.io/otel/metric"
 	"go.opentelemetry.io/otel/trace"
@@ -95,9 +97,18 @@ func WithOpenTelemetryConfiguration(cfg OpenTelemetryConfiguration) Configuratio
 	})
 }
 
+// ParseJSON parses a JSON configuration file into an OpenTelemetryConfiguration.
+func ParseJSON(r io.Reader) (*OpenTelemetryConfiguration, error) {
+	var cfg OpenTelemetryConfiguration
+	decoder := json.NewDecoder(r)
+	if err := decoder.Decode(&cfg); err != nil {
+		return nil, err
+	}
+	return &cfg, nil
+}
+
 // TODO: implement parsing functionality:
 // - https://github.com/open-telemetry/opentelemetry-go-contrib/issues/4373
-// - https://github.com/open-telemetry/opentelemetry-go-contrib/issues/4412
 
 // TODO: create SDK from the model:
 // - https://github.com/open-telemetry/opentelemetry-go-contrib/issues/4371
--- a/config/config.go
+++ b/config/config.go
@@ -7,7 +7,6 @@ import (
 	"context"
 	"encoding/json"
 	"errors"
-	"io"
 
 	"go.opentelemetry.io/otel/metric"
 	"go.opentelemetry.io/otel/trace"
@@ -98,10 +97,9 @@ func WithOpenTelemetryConfiguration(cfg OpenTelemetryConfiguration) Configuratio
 }
 
 // ParseJSON parses a JSON configuration file into an OpenTelemetryConfiguration.
-func ParseJSON(r io.Reader) (*OpenTelemetryConfiguration, error) {
+func ParseJSON(data []byte) (*OpenTelemetryConfiguration, error) {
 	var cfg OpenTelemetryConfiguration
-	decoder := json.NewDecoder(r)
-	if err := decoder.Decode(&cfg); err != nil {
+	if err := json.Unmarshal(data, &cfg); err != nil {
 		return nil, err
 	}
 	return &cfg, nil
--- a/config/config_test.go
+++ b/config/config_test.go
@@ -6,8 +6,6 @@ package config
 import (
 	"context"
 	"fmt"
-	"io"
-	"strings"
 	"testing"
 
 	"github.com/stretchr/testify/assert"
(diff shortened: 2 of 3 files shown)


config/config_test.go
```diff
@@ -51,3 +54,43 @@ func TestNewSDK(t *testing.T) {
...
+func TestParseJSON(t *testing.T) {
+	tests := []struct {
+		name     string
+		input    string
+		wantErr  error
+		wantType interface{}
+	}{
+		{
+			name:     "valid JSON",
+			input:    `{"file_format": "json", "disabled": false}`,
```
pellared, line 67: This would return the same result

```suggestion
			input:    `{"file_format": "json", "disabled": false}THIS IS INTENTIONALLY MALFORMED`,
```

Reference: https://dev.to/taqkarim/you-might-not-be-using-json-decoder-correctly-in-golang-12mb

Maybe it would be safer to change the API to accept `[]byte` instead of `io.Reader` and use `json.Unmarshal`? Especially that the performance difference is not big. Reference: https://dev.to/jpoly1219/to-unmarshal-or-to-decode-json-processing-in-go-explained-5870

Or maybe we should even not add such function, but simply add an an example to docs instead?

E.g. 

```go
func ExampleParseJSON() {
	input := `{"file_format": "0.1", "disabled": true}`

	var cfg config.OpenTelemetryConfiguration
	if err := json.Unmarshal(([]byte)(input), &cfg); err != nil {
		panic(err)
	}

	fmt.Printf("Disabled: %v", cfg.Disabled)

	// Output: Disabled: true
}
```

I am leaning towards adding a runnable example in docs instead of adding a function.

config/config.go
```diff
@@ -95,9 +97,18 @@ func WithOpenTelemetryConfiguration(cfg OpenTelemetryConfiguration) Configuratio
 	})
 }
 
+// ParseJSON parses a JSON configuration file into an OpenTelemetryConfiguration.
+func ParseJSON(r io.Reader) (*OpenTelemetryConfiguration, error) {
+	var cfg OpenTelemetryConfiguration
+	decoder := json.NewDecoder(r)
+	if err := decoder.Decode(&cfg); err != nil {
```
pellared, line 104: Can you make some more advanced test? I am not sure if it works "out of the box". We may be missing JSON tags. See: https://github.com/open-telemetry/opentelemetry-go-contrib/blob/ba19074a6785b2b65c86a52c0649cf040319fa96/Makefile#L322

@codeboten Am I correct?

config/config_test.go
carrbs: A runnable example in the docs sounds just fine to me. I can also rewrite this to accept the `[]byte` string so we can use the more dependable `json.Unmarshal` method in the parser if we decide to take this route.

config/config.go
carrbs: @pellared / @codeboten, I wasn't super clear on what the expected input would be, so perhaps I've overlooked something. I'd thought the `OpenTelemetryConfiguration` was specifically designed to build the parameters as `mapstructures` so the user could choose to define their config using either yaml or json. I (possibly naively) was building this to accept any valid json that had the [required property](https://github.com/open-telemetry/opentelemetry-configuration/blob/62138088f545154d15aca4e74fe6f6ae93c94454/schema/opentelemetry_configuration.json#L33) from the configuration input used in the Makefile:
https://github.com/open-telemetry/opentelemetry-go-contrib/blob/ba19074a6785b2b65c86a52c0649cf040319fa96/Makefile#L324
 Perhaps we can clarify what are the expected valid inputs for this (and #4373), and what we'd expect as output on valid/invalid cases? I feel this would help with implementation and writing more comprehensive tests. lmkwyt!

pellared: I was thinking of removing this line: 

https://github.com/open-telemetry/opentelemetry-go-contrib/blob/ba19074a6785b2b65c86a52c0649cf040319fa96/Makefile#L322

As then the `OpenTelemetryConfiguration` type will have  `mapstructure`, `json` and `yaml` tags. See: 

https://github.com/omissis/go-jsonschema/blob/6fcc83b8eebe874eeebc20474775280877cc62fe/main.go#L167-L168

Then we could simply provide examples how one can parse input to `OpenTelemetryConfiguration` using different libraries.

The parsing method do NOT need to do any validation e.g. check for required fields. The validation has to be done by `NewSDK` because the user could even create `OpenTelemetryConfiguration`  "by hand" without parsing any JSON or YAML.

config/config_test.go
pellared: I am leaning towards a runnable a example to reduce the API surface.

When I created https://github.com/open-telemetry/opentelemetry-go-contrib/issues/4412 I did NOT want `OpenTelemetryConfiguration` to have `mapstructure` tags, but after some time and disucssion agreed that it is acceptable. I will update the issue.

CC @codeboten

pellared:
Now, I see that the specification currently only requires implementing parsing of YAML: https://github.com/open-telemetry/opentelemetry-specification/blob/main/specification/configuration/file-configuration.md#parse

@codeboten should we change to draft (or even close) this PR? WDYT?

codeboten:
@pellared it might make sense to only implement YAML and leave JSON until there's a need for it. @carrbs do you have a specific need for JSON parsing to be implemented?

carrbs:
>  @carrbs do you have a specific need for JSON parsing to be implemented?

@codeboten, not at all, I was just looking for low hanging fruit to work on. Closing this sounds good to me!

config/config_test.go
```diff
@@ -51,3 +52,47 @@ func TestNewSDK(t *testing.T) {
...
+			wantType: nil,
+		},
+	}
+
+	for _, tt := range tests {
+		t.Run(tt.name, func(t *testing.T) {
+			if got, err := ParseJSON(([]byte)(tt.input)); err != nil {
+				require.Equal(t, tt.wantErr.Error(), err.Error())
+			} else {
+				assert.IsType(t, tt.wantType, got)
```
dashpole: Using `assert.IsType` here only checks the static return type `*OpenTelemetryConfiguration` rather than the parsed field values, which hides the fact that `got.FileFormat` is actually `""` in the `valid JSON` test case (since `OpenTelemetryConfiguration` only has `mapstructure` tags and no `json` tags). Consider asserting the expected struct value with `assert.Equal` instead, and guarding against `tt.wantErr` being `nil` before calling `tt.wantErr.Error()` on line 92.

PR #4825, opened 2024-01-17
[instrumentation/google.golang.org/grpc/otelgrpc] Do not assume HandleRPC receives a gRPCContext

mx-psi:
To help prevent crashes such as open-telemetry/opentelemetry-collector/issues/9296, check if the context is nil before proceeding.

This is inspired by: https://github.com/grpc/grpc-go/blob/ddd377f19841eae70862559c854d957d61b3b692/stats/opencensus/opencensus.go#L204-L207


instrumentation/google.golang.org/grpc/otelgrpc/stats_handler.go
```diff
@@ -136,6 +136,10 @@ func (h *clientHandler) HandleConn(context.Context, stats.ConnStats) {
 func (c *config) handleRPC(ctx context.Context, rs stats.RPCStats, isServer bool) { // nolint: revive  // isServer is not a control flag.
 	span := trace.SpanFromContext(ctx)
 	gctx, _ := ctx.Value(gRPCContextKey{}).(*gRPCContext)
```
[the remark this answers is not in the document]
mx-psi: That's not an error but a `bool` from the casting. I made some changes to report an error when this happens and make the check a bit more clear, PTAL

instrumentation/google.golang.org/grpc/otelgrpc/stats_handler.go
```diff
@@ -135,7 +139,11 @@ func (h *clientHandler) HandleConn(context.Context, stats.ConnStats) {
...
 	span := trace.SpanFromContext(ctx)
-	gctx, _ := ctx.Value(gRPCContextKey{}).(*gRPCContext)
+	gctx, ok := ctx.Value(gRPCContextKey{}).(*gRPCContext)
+	if !ok || gctx == nil {
+		otel.Handle(errMissinggRPCContext)
+		return
+	}
 	var messageId int64
 	metricAttrs := make([]attribute.KeyValue, 0, len(gctx.metricAttrs)+1)
 	metricAttrs = append(metricAttrs, gctx.metricAttrs...)
```
Aneurysm9, line 149:
```suggestion
	var metricAttrs []attribute.KeyValue
	var messageId int64

	gctx, ok := ctx.Value(gRPCContextKey{}).(*gRPCContext)
	if ok && gctx != nil {
		metricAttrs = make([]attribute.KeyValue, 0, len(gctx.metricAttrs)+1)
		metricAttrs = append(metricAttrs, gctx.metricAttrs...)
	}
```

Reading the rest of the method it looks like there's already an assumption that `gctx` may be `nil`, so simply guarding the use of it to set `metricAttrs` may be the better approach than bailing entirely.

MadVikingGod: This doesn't look universal.  It is nil checked for `InPayload` and `OutPayload`, but not on `End`.  We should probably make them all nil checked.

mx-psi: Woops, missed your comment @MadVikingGod, but I don't think I have the context here, what are `InPayload`, `OutPayload` and `End`?

pellared:
Is it not an effect of bug in https://github.com/grpc/grpc-go? If so could you additionally report it there?

References:
- https://github.com/grpc/grpc-go/tree/master/examples/features/stats_monitoring
- https://pkg.go.dev/google.golang.org/grpc/stats#Handler

Side note: I am not against having a "defensive" code.

instrumentation/google.golang.org/grpc/otelgrpc/stats_handler.go
pellared, line 1: Looks like `if gctx != nil` is missing for this block of code: https://github.com/open-telemetry/opentelemetry-go-contrib/blob/60fbb4ee9849df0dfe43242a84b04368677c84ce/instrumentation/google.golang.org/grpc/otelgrpc/stats_handler.go#L201-L208

instrumentation/google.golang.org/grpc/otelgrpc/stats_handler.go
MadVikingGod: @pellared got it in a different comment: https://github.com/open-telemetry/opentelemetry-go-contrib/pull/4825#discussion_r1456321716

instrumentation/google.golang.org/grpc/otelgrpc/stats_handler.go
```diff
@@ -135,10 +135,14 @@ func (h *clientHandler) HandleConn(context.Context, stats.ConnStats) {
...
 	span := trace.SpanFromContext(ctx)
-	gctx, _ := ctx.Value(gRPCContextKey{}).(*gRPCContext)
+	var metricAttrs []attribute.KeyValue
 	var messageId int64
-	metricAttrs := make([]attribute.KeyValue, 0, len(gctx.metricAttrs)+1)
-	metricAttrs = append(metricAttrs, gctx.metricAttrs...)
+
+	gctx, ok := ctx.Value(gRPCContextKey{}).(*gRPCContext)
+	if ok && gctx != nil {
+		metricAttrs = make([]attribute.KeyValue, 0, len(gctx.metricAttrs)+1)
```
seh, line 143: Why the extra one here?

instrumentation/google.golang.org/grpc/otelgrpc/stats_handler.go
```diff
@@ -135,10 +135,14 @@ func (h *clientHandler) HandleConn(context.Context, stats.ConnStats) {
...
+		metricAttrs = append(metricAttrs, gctx.metricAttrs...)
```
seh, line 144: You can use `copy` instead of you replace the zero on the preceding line with the length of the source slice. Alternately, consider using [the `slices.Clone` function](https://pkg.go.dev/golang.org/x/exp/slices#Clone) if you don't really need that extra slot in the backing array. It's implemented by `append` today, without the preceding call to `make`.

codeboten:
> Is it not an effect of bug in https://github.com/grpc/grpc-go? If so could you additionally report it there?
> 
> References:
> 
>     * https://github.com/grpc/grpc-go/tree/master/examples/features/stats_monitoring
> 
>     * https://pkg.go.dev/google.golang.org/grpc/stats#Handler
> 
> 
> It would be also great to reproduce the issue and have a test that would help avoiding regression bugs.
> 
> Side note: I am not against having a "defensive" code.

It's not clear to me why this was not encountered before, but the examples in the grpc repo does suggest that defensive code is the way to go here: https://github.com/grpc/grpc-go/blob/ddd377f19841eae70862559c854d957d61b3b692/examples/features/stats_monitoring/statshandler/handler.go#L73

I'm not sure if this expected behaviour or a new behaviour though.

instrumentation/google.golang.org/grpc/otelgrpc/stats_handler.go
MadVikingGod: That is what was there to begin with.  Its to have space for the `rpcStatusAttr` if this is a `*stats.End` case.

pellared:
> It's not clear to me why this was not encountered before, but the examples in the grpc repo does suggest that defensive code is the way to go here

I agree. Therefore, I am not against the change here.
But the [docs](https://pkg.go.dev/google.golang.org/grpc/stats#Handler) suggests that it should not happen.
I think it might be worth reporting. However, we should have repro steps...

instrumentation/google.golang.org/grpc/otelgrpc/stats_handler.go
mx-psi: Fixed

pellared (approved):
@MadVikingGod IMO we can consider adding it to the upcoming release.

@mx-psi Can you please still follow-up and try to reproduce it and eventually create an issue in https://github.com/grpc/grpc-go?

mx-psi:
Filed grpc/grpc-go/issues/6928. I still don't have a way to reproduce the issue, but I at least shared all we know so far. Feel free to comment if anything is missing there.

dashpole (approved):

PR #4826, opened 2024-01-18
Config: Parse model from YAML

carrbs:
Fixes: #4373

### DIFF
--- a/config/config.go
+++ b/config/config.go
@@ -6,11 +6,13 @@ package config // import "go.opentelemetry.io/contrib/config"
 import (
 	"context"
 	"errors"
+	"io"
 
 	"go.opentelemetry.io/otel/metric"
 	"go.opentelemetry.io/otel/sdk/resource"
 	semconv "go.opentelemetry.io/otel/semconv/v1.17.0"
 	"go.opentelemetry.io/otel/trace"
+	"gopkg.in/yaml.v3"
 )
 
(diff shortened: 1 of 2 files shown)


config/config.go
```diff
@@ -113,8 +116,34 @@ func WithOpenTelemetryConfiguration(cfg OpenTelemetryConfiguration) Configuratio
...
+			if len(match) < 2 {
+				return s
+			}
+			envVarName := string(match[1])
+			envVarValue := os.Getenv(envVarName)
+			return []byte(envVarValue)
+		})
+	}
+
+	file = replaceEnvVars(file)
```
pellared, line 135: I just want to call out one important thing from https://github.com/open-telemetry/opentelemetry-specification/blob/main/specification/configuration/file-configuration.md#environment-variable-substitution:

> It MUST NOT be possible to inject YAML structures by environment variables. For example, references to INVALID_MAP_VALUE environment variable below.

Maybe you will find some code that will help with it in https://github.com/open-telemetry/opentelemetry-collector. @codeboten: are you able to provide some guidance?

carrbs: Totally agree @pellared, also this does not solve for blocking the key replacement (i.e.`${SNEAKY}="foo"`) in the spec. I will take a look at both of these cases to see if we can come up with something more useful.

PR #4832, opened 2024-01-19
config: add support for additional properties in resource

### LINKED ISSUE #4828: config: add support for additional attributes
The Resource model only supports a ServiceName attribute.

https://github.com/open-telemetry/opentelemetry-go-contrib/blob/044e09ac8bfe0d5a60510778017bec57413243d9/config/generated_config.go#L19-L22

This should be updated to support additional attributes as pointed out by @pellared here: https://github.com/open-telemetry/opentelemetry-go-contrib/pull/4824#discussion_r1457142001

codeboten:
This allows the configuration of other attributes than service.name. ~I'm using a personal fork of go-jsonschema to test out that the implementation would work for otel-config's use-case~

Support for additional properties is now available in the go-jsonschema library, updated the PR accordingly

Fixes #4828


config/resource.go
```diff
@@ -4,16 +4,64 @@
...
+	attrs := []attribute.KeyValue{
+		semconv.ServiceName(*res.Attributes.ServiceName),
+	}
+
+	if props, ok := res.Attributes.AdditionalProperties.(map[string]any); ok {
+		for k, v := range props {
+			attrs = append(attrs, keyVal(k, v))
+		}
+	}
+
```
MadVikingGod, line 62: Is it intended that Additional Properties can override ServiceName?

MadVikingGod (approved):

config/resource.go
codeboten: Yeah it's actually the correct thing to do, because ServiceName was removed as a special attribute https://github.com/open-telemetry/opentelemetry-configuration/commit/6215d43ee00c4ebe32ca8cbaf4c2807600b73945

codeboten: when the dep on the config schema is updated next, `res.Attributes.ServiceName` will be removed altogether

themarcelor:
This would be a pretty cool change 👍🏼

MrAlias (approved):

PR #4860, opened 2024-01-26
implement cloud.{account.id,availability_zone,region} in AWS ECS detector

mmanciop:


detectors/aws/ecs/ecs.go
```diff
@@ -125,6 +129,28 @@ func (detector *resourceDetector) Detect(ctx context.Context) (*resource.Resourc
 			}
 		}
 
+		if accountId, err := detector.getAccountID(taskMetadata.TaskARN); err == nil {
```
dmathieu, line 132: If those errors aren't nil, we should report them. Either with `otel.Handle()`, or with `span.RecordError()` (I'm thinking the latter is best).

mmanciop: These error take place in the detector, which is run once on startup. Usually there won't be an active span there. In other SDKs iirc such errors are logged out, but I did not see that done in other detectors.

mmanciop: I am not familiar with `otel.Handle()`, I'll look it up :-)

mmanciop: @dmathieu I think the solution is much simpler, see c6f59b86f80c95ed926295eb419830763a3cbfc5 :-)

CHANGELOG.md
```diff
@@ -11,6 +11,7 @@ This project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0.htm
 ### Added
 
 - Add the new `go.opentelemetry.io/contrib/instrgen` package to provide auto-generated source code instrumentation. (#3068, #3108)
+- Add support for `cloud.{account.id,availability_zone,region}` in the AWS ECS detector
```
MrAlias, line 14: - please break these out into full words so users can search the changelog for them
- please add a period at the end of the sentence to follow existing formatting
- please add the PR number to follow existing formatting

detectors/aws/ecs/ecs.go
```diff
@@ -125,6 +128,26 @@ func (detector *resourceDetector) Detect(ctx context.Context) (*resource.Resourc
 			}
 		}
 
+		arnParts := strings.Split(taskMetadata.TaskARN, ":")
+		// a valid arn should have at least 6 parts
+		if len(arnParts) < 6 {
+			return empty, errCannotParseTaskArn
```
MrAlias, line 134: Needs test coverage

detectors/aws/ecs/ecs.go
```diff
@@ -209,6 +232,16 @@ func (detector *resourceDetector) getLogsAttributes(metadata *ecsmetadata.Contai
 	}, nil
 }
 
+// returns metadata v4 for the container.
+func (ecsUtils ecsDetectorUtils) getContainerMetadataV4(ctx context.Context) (*ecsmetadata.ContainerMetadataV4, error) {
+	return ecsmetadata.GetContainerV4(ctx, &http.Client{})
```
MrAlias, line 237: Needs test coverage

detectors/aws/ecs/ecs.go
```diff
@@ -209,6 +232,16 @@ func (detector *resourceDetector) getLogsAttributes(metadata *ecsmetadata.Contai
...
+}
+
+// returns metadata v4 for the task.
+func (ecsUtils ecsDetectorUtils) getTaskMetadataV4(ctx context.Context) (*ecsmetadata.TaskMetadataV4, error) {
+	return ecsmetadata.GetTaskV4(ctx, &http.Client{})
```
MrAlias, line 242: Needs test coverage

detectors/aws/ecs/ecs.go
```diff
@@ -125,6 +128,26 @@ func (detector *resourceDetector) Detect(ctx context.Context) (*resource.Resourc
 			}
 		}
 
+		arnParts := strings.Split(taskMetadata.TaskARN, ":")
+		// a valid arn should have at least 6 parts
```
MrAlias, line 132:
```suggestion
		// A valid ARN should have at least 6 parts.
```

detectors/aws/ecs/ecs.go
MrAlias, line 134: `errCannotParseTaskArn` does not effectively communicate the error. It is known why the ARN parsing failed. Include the reason in the returned errors.

detectors/aws/ecs/ecs.go
```diff
@@ -125,6 +128,26 @@ func (detector *resourceDetector) Detect(ctx context.Context) (*resource.Resourc
...
+		if len(arnParts) < 6 {
+			return empty, errCannotParseTaskArn
+		}
+
+		attributes = append(
+			attributes,
+			semconv.CloudRegion(arnParts[3]),
+			semconv.CloudAccountID(arnParts[4]),
```
MrAlias, line 140: Is this guaranteed to not change upstream?

detectors/aws/ecs/ecs.go
```diff
@@ -97,12 +100,12 @@ func (detector *resourceDetector) Detect(ctx context.Context) (*resource.Resourc
...
 
 	if len(metadataURIV4) > 0 {
-		containerMetadata, err := ecsmetadata.GetContainerV4(ctx, &http.Client{})
+		containerMetadata, err := detector.utils.getContainerMetadataV4(ctx)
 		if err != nil {
 			return empty, err
 		}
 
-		taskMetadata, err := ecsmetadata.GetTaskV4(ctx, &http.Client{})
+		taskMetadata, err := detector.utils.getTaskMetadataV4(ctx)
```
MrAlias, line 108: It seems superfluous to add a method that just makes the same direct function call. Why are `getTaskMetadataV4` and `getContainerMetadataV4` needed?

detectors/aws/ecs/ecs.go
mmanciop: The day the ARN format of AWS changes, we are all in for a funny couple years.

detectors/aws/ecs/ecs.go
mmanciop: Because mocking. Without those one liners, I could not achieve good functional coverage (as opposed to the line coverage the tooling complains about).

detectors/aws/ecs/ecs.go
MrAlias: Can we quickly (i.e. non-regex based) validate both of these values to ensure they are correct before adding them here?

detectors/aws/ecs/ecs.go
mmanciop: I got the gist from the test infrastructure that the mock facilities need them to be methods

MrAlias: Gotcha :+1:

detectors/aws/ecs/ecs.go
mmanciop: It seems to me frankly overkill. The only think it would do, is to create a more understandable error in a case that is really unlikely to ever occur: to the best of my knowledge, ARNs historically change almost never, and practically never for established services. Way too much software depends on them.

mmanciop: For reference, ECS ARNs changed back in 2020 (https://aws.amazon.com/blogs/compute/migrating-your-amazon-ecs-deployment-to-the-new-arn-and-resource-id-format-2/) and it was a Big Thing for migration periods and all :-)

detectors/aws/ecs/ecs.go
mmanciop: Done in 052fb008

CHANGELOG.md
mmanciop: Done in 3b8608d4 and 0eca55d7

dmathieu (approved):

Aneurysm9 (approved):

MrAlias (approved):

mmanciop:
Then test coverage build failure is about the automation failing to upload the result. I have seen it failing every other time or so on this PR.

PR #4874, opened 2024-01-29
AWS ECS detector should detect otel CloudRegion/CloudAccountID/CloudAvailabilityZone/CloudResourceID

andrew-rowson-lseg:
Some applications rely on the otel non-vendored cloud tags being populated in metrics/traces. For ECS workloads, we can populate these from the data that's already returned from the ECS Metadata API, so this can be used to add to the resource's attributes.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -11,6 +11,7 @@ This project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0.htm
 ### Added
 
 - Add the new `go.opentelemetry.io/contrib/instrgen` package to provide auto-generated source code instrumentation. (#3068, #3108)
+- `go.opentelemetry.io/contrib/detectors/aws/ecs` attempts to detect and add `semconv.CloudRegion`, `semconv.CloudAccountID`, `semconv.CloudAvailabilityZone` and `semconv.CloudResourceID`
 
 ## [1.22.0/0.47.0/0.16.0/0.2.0] - 2024-01-18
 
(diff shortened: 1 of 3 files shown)


andrew-rowson-lseg:
> This looks like a duplicate of #4860

Ah, that PR didn't exist when I originally forked. If it does the same thing, then we should obviously do that one instead.

mmanciop:
In #4860 I did not implement `cloud.resource.id` though. I'll add it.

mmanciop:
> > In #4860 I did not implement `cloud.resource.id` though. It's unclear to me whether it should be the Task ARN or the Container ARN. I know for a fact that there is no ECS detector out there for OTEL that has implemented `cloud.resource_id`, as I have been recently running around adding the other `cloud.*` tags. I am leaning towards the Task ARN for [these reasons](https://cloud-native.slack.com/archives/CJFCJHG4Q/p1706613367926659). Does anybody have a firm opinion on the matter?
> 
> My pitch is that it should be the containerId, because a task can contain multiple containers which may all be looking to publish otel data. ResourceId feels like it should represent the most granular ID of the metrics/trace/logs reporter. Not firmly wedded to this though!
> 
> (I'm unable to open slack in my current context, so apologies, can't see that URL).

The `#opentelemetry` channel in the CNCF Slack was indeed not the best place to have this discussion.

I opened [this issue](https://github.com/open-telemetry/semantic-conventions/issues/677) on the `open-telemetry/semantic-conventions` to track the discussion. @andrew-rowson-lseg make your voice heard over there :-)

mmanciop:
@andrew-rowson-lseg I am opening a PR for `cloud.resource_id` in a few mins, care to close this one?

MrAlias:
Closing in favor of https://github.com/open-telemetry/opentelemetry-go-contrib/pull/5091

PR #4875, opened 2024-01-29
[otelgrpc] Do not panic for nil instruments

MrAlias:
There is no guarantee when a `Meter` returns an error the instrument will not be nil. Default to a no-op implementation  of the instrument in that case so the instrumentation does not panic.

Close https://github.com/open-telemetry/opentelemetry-go/issues/4836


dmathieu (approved):

dashpole (approved):

instrumentation/google.golang.org/grpc/otelgrpc/config.go
```diff
@@ -89,39 +92,68 @@ func newConfig(opts []Option, role string) *config {
 		metric.WithUnit("ms"))
 	if err != nil {
 		otel.Handle(err)
+		if c.rpcDuration == nil {
+			c.rpcDuration = noopFloat64Hist{}
```
pellared, line 96: What is the reason that we are not simply using https://pkg.go.dev/go.opentelemetry.io/otel/metric/noop#Float64Histogram here?

MrAlias: Please see the PR description:

> This uses a custom no-op implementation of the histograms instead of the `noop` package to remove the question about what to do with the returned error from the `noop` instrument creation (i.e. `inst, /* how to handle? */ = noop.NewMeterProvider().Meter("").Int64Histogram("")`)

pellared: Clarifying. Why not the following?

```suggestion
			c.rpcDuration = noop.Float64Histogram{}
```

MrAlias: Hmm, probably because I forgot we exported those :laughing: 

SGTM. I'll update.

MrAlias: Updated. PTAL.

pellared (approved):

PR #4892, opened 2024-02-05
chore: update jaeger remote sampler example

hcelaloner:
chore: update jaeger remote sampler example using the latest otelcol-contrib extension

Previously, the example was outdated and not working properly. This commit updates the example provided using otelcol-contrib jaegerremotesampling extension.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -16,7 +16,7 @@ This project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0.htm
 
 ### Changed
 
-- Updated jaeger version in `go.opentelemetry.io/contrib/samplers/jaegerremote/example` example. (#4892)
+- Updated otel-collector from `otel/openteleemtry-collector` to `otel/opentelemetry-collector-contrib` in `go.opentelemetry.io/contrib/samplers/jaegerremote/example` example. (#4892)
 
 ### Deprecated
 
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -14,6 +14,10 @@ This project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0.htm
 - Add client metric support to `go.opentelemetry.io/contrib/instrumentation/net/http/otelhttp`. (#4707)
 - Add peer attributes to spans recorded by `NewClientHandler`, `NewServerHandler` in `go.opentelemetry.io/contrib/instrumentation/google.golang.org/grpc/otelgrpc`. (#4873)
 
+### Changed
+
+- Updated jaeger version in `go.opentelemetry.io/contrib/samplers/jaegerremote/example` example. (#4892)
+
 ### Deprecated
 
 - The `RequestCount`, `RequestContentLength`, `ResponseContentLength`, `ServerLatency` constants in `go.opentelemetry.io/contrib/instrumentation/net/http/otelhttp` are deprecated. (#4707)
--- a/samplers/jaegerremote/example/main.go
+++ b/samplers/jaegerremote/example/main.go
@@ -33,7 +33,7 @@ func main() {
 	// Optional: an implementation of logr.Logger used for demo purposes to catch potential error logs
 	logger := stdr.NewWithOptions(stdlog.New(os.Stderr, "", stdlog.LstdFlags), stdr.Options{LogCaller: stdr.All})
 
-	samplingRefreshInterval := 1 * time.Minute
+	samplingRefreshInterval := 10 * time.Second
 	jaegerRemoteSampler := jaegerremote.New(
 		"foo",
 		jaegerremote.WithSamplingServerURL("http://localhost:5778/sampling"),
--- a/samplers/jaegerremote/example/main.go
+++ b/samplers/jaegerremote/example/main.go
@@ -16,9 +16,12 @@ package main
 
 import (
 	"fmt"
+	stdlog "log"
+	"os"
 	"time"
 
 	"github.com/davecgh/go-spew/spew"
+	"github.com/go-logr/stdr"
 
 	"go.opentelemetry.io/contrib/samplers/jaegerremote"
 	"go.opentelemetry.io/otel"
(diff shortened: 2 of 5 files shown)


samplers/jaegerremote/example/main.go
```diff
@@ -42,14 +52,18 @@ func main() {
 	)
 	otel.SetTracerProvider(tp)
 
-	ticker := time.Tick(time.Second)
+	fmt.Printf("\n* Initial Jaeger Remote Sampler: %v\n\n", time.Now())
```
yurishkuro, line 55: why not use `logger`?

yurishkuro (approved):

CHANGELOG.md
```diff
@@ -21,6 +21,7 @@ This project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0.htm
 ### Fixed
 
 - Do not panic in `go.opentelemetry.io/contrib/instrumentation/google.golang.org/grpc/otelgrpc` if `MeterProvider` returns a `nil` instrument. (#4875)
+- OpenTelemetry Collector based example for `go.opentelemetry.io/contrib/samplers/jaegerremote`. (#4892)
```
dmathieu, line 24: A `Fixed` entry would be a bugfix, which this isn't. How about moving it to the `Changed` section?
Also, how about something along the lines of:

> Updated jaeger version in `go.opentelemetry.io/contrib/samplers/jaegerremote/example` example. (#4892)

hcelaloner: Thank you for the feedback. I also had some doubts about where should I add it while adding it to changelog. Of course 👍  I am going to move it in a minute.

samplers/jaegerremote/example/main.go
```diff
@@ -27,12 +30,19 @@ import (
...
+	// Optional: an implementation of logr.Logger used for demo purposes to catch potential error logs
+	logger := stdr.NewWithOptions(stdlog.New(os.Stderr, "", stdlog.LstdFlags), stdr.Options{LogCaller: stdr.All})
+
+	samplingRefreshInterval := 1 * time.Minute
 	jaegerRemoteSampler := jaegerremote.New(
 		"foo",
-		jaegerremote.WithSamplingServerURL("http://localhost:5778"),
-		jaegerremote.WithSamplingRefreshInterval(10*time.Second), // decrease polling interval to get quicker feedback
+		jaegerremote.WithSamplingServerURL("http://localhost:5778/sampling"),
+		jaegerremote.WithSamplingRefreshInterval(samplingRefreshInterval), // decrease polling interval to get quicker feedback
```
dmathieu, line 40: This here is not just an upgrade. It changes the refresh interval from 10 seconds to 1 minute, which is the default as well.

https://github.com/open-telemetry/opentelemetry-go-contrib/blob/a1b4b4d348c0ca2dc67834aba13caa2fe5c0a993/samplers/jaegerremote/sampler_remote.go#L37

So we should either:

* Keep the previous value of 10 seconds.
* Remove the option entirely, since this is the default anyway.
* Remove the comment that says we decrease the polling interval, since we don't do that anymore.

I'd tend towards option 1. An example where the refresh interval is refreshed every minute is quite long.

samplers/jaegerremote/example/main.go
```diff
@@ -27,12 +30,19 @@ import (
 )
 
 func main() {
+	// Optional: an implementation of logr.Logger used for demo purposes to catch potential error logs
+	logger := stdr.NewWithOptions(stdlog.New(os.Stderr, "", stdlog.LstdFlags), stdr.Options{LogCaller: stdr.All})
```
dmathieu, line 34: The addition of this logger is also more than just an update. Wouldn't that be adding quite a bit of noise in the output, making the example less readable?

CHANGELOG.md
```diff
@@ -14,6 +14,10 @@ This project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0.htm
 - Add client metric support to `go.opentelemetry.io/contrib/instrumentation/net/http/otelhttp`. (#4707)
 - Add peer attributes to spans recorded by `NewClientHandler`, `NewServerHandler` in `go.opentelemetry.io/contrib/instrumentation/google.golang.org/grpc/otelgrpc`. (#4873)
 
+### Changed
+
+- Updated jaeger version in `go.opentelemetry.io/contrib/samplers/jaegerremote/example` example. (#4892)
```
dmathieu, line 19: My bad, I didn't fully read the other changes when I suggested the changelog entry. You're not upgrading the jaeger version, you're updating the collector.

```suggestion
- Updated otel-collector from `otel/openteleemtry-collector` to `otel/opentelemetry-collector-contrib` in `go.opentelemetry.io/contrib/samplers/jaegerremote/example` example. (#4892)
```

samplers/jaegerremote/example/main.go
hcelaloner: When I wanted to try the example, I couldn't see any error logs from the background updater. By default sampler was using logr.Discard() if I am not wrong, I just added to a simple stdout logr implementation to be able to check error logs while trying. I didn't remove it because I thought that it could be a nice for example.

samplers/jaegerremote/example/main.go
hcelaloner: 👍  I reverted it to the previous value of 10 seconds.

samplers/jaegerremote/example/otel-collector.yaml
```diff
@@ -1,17 +1,25 @@
...
+
+processors:
+  batch:
 
 exporters:
-  logging:
+  debug:
+
+extensions:
+  jaegerremotesampling:
```
dashpole: In `otel/opentelemetry-collector-contrib:latest`, `jaegerremotesampling` binds its HTTP server to `localhost:5778` by default, so requests from the host via Docker port mapping (`5778:5778`) won't reach the container unless `0.0.0.0:5778` is configured:
```suggestion
  jaegerremotesampling:
    http:
      endpoint: "0.0.0.0:5778"
```

PR #5080, opened 2024-02-12
Use an atomic.Int64 as bodyWrapper read bytes counter

### LINKED ISSUE #4895: Data race in v0.48.0
### Description

```
WARNING: DATA RACE
Write at 0x00c000544258 by goroutine 603:
  go.opentelemetry.io/contrib/instrumentation/net/http/otelhttp.(*bodyWrapper).Read()
      external/io_opentelemetry_go_contrib_instrumentation_net_http_otelhttp/wrap.go:40 +0xa4
  io.discard.ReadFrom()
      GOROOT/src/io/io.go:658 +0x91
  io.(*discard).ReadFrom()
      <autogenerated>:1 +0x4c
  io.copyBuffer()
      GOROOT/src/io/io.go:416 +0x1be
  io.Copy()
      GOROOT/src/io/io.go:389 +0x5c
  net/http.(*transferWriter).doBodyCopy()
      GOROOT/src/net/http/transfer.go:412 +0x38
  net/http.(*transferWriter).writeBody()
      GOROOT/src/net/http/transfer.go:375 +0x82f
  net/http.(*Request).write()
      GOROOT/src/net/http/request.go:738 +0x1393
  net/http.(*persistConn).writeLoop()
      GOROOT/src/net/http/transport.go:2424 +0x372
  net/http.(*Transport).dialConn.func6()
      GOROOT/src/net/http/transport.go:1777 +0x33
Previous read at 0x00c000544258 by goroutine 598:
  go.opentelemetry.io/contrib/instrumentation/net/http/otelhttp.(*Transport).RoundTrip()
      external/io_opentelemetry_go_contrib_instrumentation_net_http_otelhttp/transport.go:185 +0xec8
  net/http.send()
      GOROOT/src/net/http/client.go:260 +0x9c6
  net/http.(*Client).send()
      GOROOT/src/net/http/client.go:181 +0x14e
  net/http.(*Client).do()
      GOROOT/src/net/http/client.go:724 +0x11a8
  net/http.(*Client).Do()
      GOROOT/src/net/http/client.go:590 +0xa65
  github.com/hashicorp/go-retryablehttp.(*Client).Do()
      external/com_github_hashicorp_go_retryablehttp/client.go:653 +0xa11
...
Goroutine 603 (running) created at:
  net/http.(*Transport).dialConn()
      GOROOT/src/net/http/transport.go:1777 +0x266a
  net/http.(*Transport).dialConnFor()
      GOROOT/src/net/http/transport.go:1467 +0x129
  net/http.(*Transport).queueForDial.func3()
      GOROOT/src/net/http/transport.go:1448 +0x44
Goroutine 598 (running) created at:
  net/http.(*Server).Serve()
      GOROOT/src/net/http/server.go:3086 +0x86c
...
```

Looks like roundtripper can return before finishing sending the request. Perhaps when server has responded faster than the request was fully sent? I think in HTTP/2 mode stdlib HTTP client handles send/receive independently and in HTTP/1 it only starts reading the response once the request has been fully sent.

https://pkg.go.dev/net/http#RoundTripper says:

> Callers should not mutate or reuse the request until the Response's Body has been closed.

Which is the case in the new code - it reads a field from the request body wrapper without waiting for the `Close` to be called.

### Environment

- OS: N/A
- Architecture: N/A
- Go Version: 1.21.7.
- `otelhttp` version: v0.48.0

### Steps To Reproduce

Run [`TestProxy_UserAccessHappyPath`](https://gitlab.com/gitlab-org/cluster-integration/gitlab-agent/-/blob/4fdc2f05024a6d95f175b7f321917232aefbfd8b/internal/module/kubernetes_api/server/proxy_test.go#L623) with `-race`. I don't have a small reproducer, sorry.

### Expected behavior

No data race.

dmathieu:
Closes #4895.

This switches the bodyWrapper struct to use an `atomic.Int64` instead of a plain `int64` integer to count read bytes.
This should solve the race condition mentioned in #4895, though unfortunately, this isn't being tested as I haven't been able to reproduce the issue within the codebase.


dashpole (approved):

MadVikingGod (approved):

MrAlias (approved):

ash2k:
I think this fix may lead to incorrect metric reported, see my comment here https://github.com/open-telemetry/opentelemetry-go-contrib/issues/4895#issuecomment-1931787280. This can be checked by having a reader for the body that is slow+a big body and a server that starts responding without consuming all of the incoming request's body.

MrAlias:
I think this is still as improvement worth including. Slightly incorrect data is better then a panic.

Please do submit a change to address the data inconsistency you mention.

ash2k:
I agree this is an improvement vs a data race. I'm not planning to work on this, but I've opened an issue so that this is not lost and someone can pick it up https://github.com/open-telemetry/opentelemetry-go-contrib/issues/5087.

Issue #5088, opened 2024-02-14
[otelgin]: Catch and record panics in Middleware


asecondo:
### Problem Statement

Today, the otelgin middleware does not handle panics. In the event that there's some bug in some service code or lower middleware layer that panics, this information is not present in the resulting span. It would be incredibly useful to leverage this information to help troubleshooting in development and production environments.

### Proposed Solution

Catch, record, and "rethrow" any panics. This solution would likely look very similar to the `otelgin.HTML(...)` handler [here](https://github.com/open-telemetry/opentelemetry-go-contrib/blob/main/instrumentation/github.com/gin-gonic/gin/otelgin/gintrace.go#L130).

#### Alternatives

- Do nothing, as is currently done today
- Add panic recovery handling to this middleware. This feels incorrect, as we'd be going beyond the expected responsibilities of this middleware.

#### Prior Art

As mentioned above, we would likely leverage a similar solution used in the same package [here](https://github.com/open-telemetry/opentelemetry-go-contrib/blob/main/instrumentation/github.com/gin-gonic/gin/otelgin/gintrace.go#L130).

### Additional Context

N/A.

pellared:
I am against adding such functionality by default as users may have their own panic recovery.

The "rethrow" would change the callstack of the panic received by the next panic recovery making it hard to find the main source of the panic in their existing panic recovery solution.

None of our instrumentation libraries handles panics. Exception is (as you mentioned) `otelgin.HTML`, which does not look right to me.

Maybe such functionality could be an opt-in option.

Similar comment here: https://github.com/open-telemetry/opentelemetry-go-contrib/pull/6366#discussion_r1856017860

flc1125:
From the code, although it handles panic, there is no situation where Status is set to Error. However, in actual application, it is very necessary.

[open-telemetry/opentelemetry-go@8b510c3/sdk/trace/span.go#L420](https://github.com/open-telemetry/opentelemetry-go/blob/8b510c36ed26fafe151df35a251f4589f5959802/sdk/trace/span.go#L420)

PR #5090, opened 2024-02-14
feat(otelgin): add support for recording panics

asecondo:
Implements as described in #5088


instrumentation/github.com/gin-gonic/gin/otelgin/gintrace.go
```diff
@@ -86,7 +86,18 @@ func Middleware(service string, opts ...Option) gin.HandlerFunc {
 			opts = append(opts, oteltrace.WithAttributes(rAttr))
 		}
 		ctx, span := tracer.Start(ctx, spanName, opts...)
-		defer span.End()
+		defer func() {
+			if cfg.RecordPanicConfig.enabled {
```
asecondo, line 90: note: added this as an opt-in behavior to preserve existing behavior

CHANGELOG.md
```diff
@@ -15,6 +15,7 @@ The next release will require at least [Go 1.21].
 
 - Add the new `go.opentelemetry.io/contrib/instrgen` package to provide auto-generated source code instrumentation. (#3068, #3108)
 - Support [Go 1.22]. (#5082)
+- Add support to record panics in `go.opentelemetry.io.contrib/instrumentation/github.com/gin-gonic/gin/otelgin` (#5090)
```
```suggestion
- Add support to record panics in `go.opentelemetry.io/contrib/instrumentation/github.com/gin-gonic/gin/otelgin`. (#5090)
```

instrumentation/github.com/gin-gonic/gin/otelgin/gintrace.go
dmathieu: I'm not sure we need this option. This is new behavior, not a breaking change.

instrumentation/github.com/gin-gonic/gin/otelgin/gintrace.go
```diff
@@ -86,7 +86,18 @@ func Middleware(service string, opts ...Option) gin.HandlerFunc {
...
+				if r := recover(); r != nil {
+					err := fmt.Errorf("error handling request: %s", r)
+					span.RecordError(err, oteltrace.WithStackTrace(cfg.RecordPanicConfig.stackTrace))
```
dmathieu, line 93: Similarly here, the config is probably superfluous.

instrumentation/github.com/gin-gonic/gin/otelgin/gintrace.go
```diff
@@ -86,7 +86,18 @@ func Middleware(service string, opts ...Option) gin.HandlerFunc {
 			opts = append(opts, oteltrace.WithAttributes(rAttr))
 		}
 		ctx, span := tracer.Start(ctx, spanName, opts...)
-		defer span.End()
+		defer func() {
+			if cfg.RecordPanicConfig.enabled {
+				if r := recover(); r != nil {
+					err := fmt.Errorf("error handling request: %s", r)
```
dmathieu, line 92: What is the added value of creating a new error here over recording the original one?

If we do want to create a new error, we should wrap the original one:
```suggestion
					err := fmt.Errorf("error handling request: %w", r)
```

instrumentation/github.com/gin-gonic/gin/otelgin/gintrace.go
```diff
@@ -86,7 +86,18 @@ func Middleware(service string, opts ...Option) gin.HandlerFunc {
...
+					span.RecordError(err, oteltrace.WithStackTrace(cfg.RecordPanicConfig.stackTrace))
+					span.SetStatus(codes.Error, "panic recovered")
```
dmathieu, line 94:
```suggestion
					span.SetStatus(codes.Error, err.Error())
```

instrumentation/github.com/gin-gonic/gin/otelgin/option.go
```diff
@@ -88,3 +129,10 @@ func WithSpanNameFormatter(f func(r *http.Request) string) Option {
 		c.SpanNameFormatter = f
 	})
 }
+
+// WithRecordPanicConfig specifies a configuration for panic recording.
+func WithRecordPanicConfig(cfg RecordPanicConfig) Option {
```
dmathieu, line 134: If we do want to keep this config, how about:

```suggestion
func WithRecordPanicConfig(enabled, stacktrace bool) Option {
```

Then, there's no need to have a config within a config.

instrumentation/github.com/gin-gonic/gin/otelgin/gintrace.go
```diff
@@ -86,7 +86,18 @@ func Middleware(service string, opts ...Option) gin.HandlerFunc {
 			opts = append(opts, oteltrace.WithAttributes(rAttr))
 		}
 		ctx, span := tracer.Start(ctx, spanName, opts...)
-		defer span.End()
```
MrAlias, line 89: This behavior is already in the default SDK's `End` method:

https://github.com/open-telemetry/opentelemetry-go/blob/e8973b75b230246545cdae072a548c83877cba09/sdk/trace/span.go#L403-L420

Why is this being added here? It seems like an SDK concern not an instrumentation one.

MrAlias: Looking at the [latest semantic conventions](https://github.com/open-telemetry/semantic-conventions/blob/8b4ad95d21b3d5e17fe97538f93f98e537013661/docs/http/http-spans.md#common-attributes), the `error.type` attribute should be used here to capture the panic information.

It is a known thing that our HTTP instrumentation is quite behind on semantic conventions, but I think this should be our target. Instead of having configuration to enable this, we should add the `error.type` attribute if a panic occurred always.

This will require this instrumentation to go through the process of upgrading semconv which will require a [migration path](https://github.com/open-telemetry/semantic-conventions/blob/8b4ad95d21b3d5e17fe97538f93f98e537013661/docs/http/migration-guide.md).

asecondo: sorry for the late reply - getting back to this now. i think that makes sense. i'm not familiar with these migrations, but i'll happily give it a shot. @MrAlias i may poke you for help with the migrations if i run into issues or have questions. i'll apply these changes, as well as the suggestions from @dmathieu.

asecondo: actually it looks like #5092 will take care of the migration plan. i might wait until that PR is merged and then make the changes to this package on top of that. how does that sound?

instrumentation/github.com/gin-gonic/gin/otelgin/gintrace.go
asecondo: i'll update to record the original error

mxpetit:
Hello,
I'm also using otelgin and would very much like this issue to be merged. Can we have a little update on this?
Thanks a lot!

asecondo:
> Hello, I'm also using otelgin and would very much like this issue to be merged. Can we have a little update on this? Thanks a lot!

hey @mxpetit - from the discussion in [this thread](https://github.com/open-telemetry/opentelemetry-go-contrib/pull/5090#discussion_r1489761938), my current plan is to wait for #5092 to complete, likely make a few changes on top of that, then look to get this merged.

flc1125:
I think the handling of this PR should refer to the method of this PR: https://github.com/open-telemetry/opentelemetry-go-contrib/pull/6366

pellared:
I added a comment here: https://github.com/open-telemetry/opentelemetry-go-contrib/issues/5088#issuecomment-2499904424

@asecondo, do you think we should change this PR to draft or close it at this point of time?

Issue #3894, opened 2023-05-31
otelgrpc: Allow adding custom attributes to otelgrpc metrics


inigohu:
### Problem Statement

The `otelgrpc` instrumentation does not allow adding custom attributes to the metrics.

### Proposed Solution

The `otelhttp` instrumentation, for example, provides a [labeler](https://pkg.go.dev/go.opentelemetry.io/contrib/instrumentation/net/http/otelhttp#Labeler) for adding custom attributes.

rehanpfmr:
@inigohu @MrAlias 
Does this need to be updated on all instrumentations?
Let me know instrumentations which can be updated, I can take it.

inigohu:
I see that the new `otelgrpc.StatsHandler` has been implemented. Would it be possible to include this feature? Maybe make [gRPCContextKey](https://github.com/open-telemetry/opentelemetry-go-contrib/blob/3cecdcf33c4701a3abd8557d38ffde090a23b696/instrumentation/google.golang.org/grpc/otelgrpc/stats_handler.go#L34C2-L34C2) and [gRPCContext](https://github.com/open-telemetry/opentelemetry-go-contrib/blob/3cecdcf33c4701a3abd8557d38ffde090a23b696/instrumentation/google.golang.org/grpc/otelgrpc/stats_handler.go#L36) public to pass custom attributes? WDYT @pellared

upalinski:
Any updates on this? Had to copy-paste stats handler and allow custom attributes to be passed outside.

pellared:
@upalinski Can you submit a PR? If so then  please make sure to include unit tests and sign EasyCLA.

dmathieu:
I believe a similar approach to what's being done in net/http (the Labeler mentioned in the issue's description) would be fine, if you want to open a PR.

duongcongtoai:
Hi, is there any update on this :D it is really useful for our usecase to tag extra labels to our grpc traffic

inigohu:
Hi @dmathieu, I've opened a [PR](https://github.com/open-telemetry/opentelemetry-go-contrib/pull/5133) if you want to take a look at it.

PR #5133, opened 2024-02-19
otelgrpc: add custom attributes to the stats handler

inigohu:
Fixes https://github.com/open-telemetry/opentelemetry-go-contrib/issues/3894


inigohu:
The StatsHandler tests did not pass because there seems to be a race condition. The problem is that I moved the span creation from `TagRPC` to `HandleRPC`, but I think the `HandleRPC` stage is the correct place to start the Span. WDYT?

inigohu:
Friendly ping @dashpole @hanyuancheung

dashpole:
This looks like it does a lot more than just add a labeler.  Would you mind doing general refactoring in its own PR?

inigohu:
> This looks like it does a lot more than just add a labeler. Would you mind doing general refactoring in its own PR?

Sure. I have simplified this PR by adding only the labeler.

instrumentation/google.golang.org/grpc/otelgrpc/stats_handler.go
```diff
@@ -30,6 +30,16 @@ type gRPCContext struct {
 	record           bool
 }
 
+// AddAttrs adds attributes to the given context.
+func AddAttrs(ctx context.Context, attrs ...attribute.KeyValue) context.Context {
```
dashpole, line 34: nit: Consider `WithAttributes` to match the naming pattern for functions that add to the context.

dashpole (commented):
can you add tests?

dashpole (commented):
Can you test the span from the stats handler as well?

inigohu:
@dashpole PTAL

dashpole (commented):
I thought we were going to use a context-based pattern (similar to the otelhttp.Labeler) to support this?  Am I misremembering?

inigohu:
>I thought we were going to use a context-based pattern (similar to the otelhttp.Labeler) to support this? Am I misremembering?

Yes, sorry, I forgot to comment this last change. My first idea was to follow the `otelhttp.Labeler` strategy using the context to pass the labels, but when I was adding the tests I realised that maybe it doesn't make much sense to do it this way, because unlike `http.Handler` and `grpc.UnaryInterceptors`, you can't chain different `grpc.StatsHandler` to add labels and pass them through the context.

That's why I changed the approach. The behaviour differs from that of `otelhttp.Labeler`. That one is request/context based tagging and this one is server based tagging.

WDYT @dashpole ?

dashpole:
A few thoughts/ideas:

* If we use constant labels as you propose here, what is the use-case?  Are constant labels actually useful?
* Should the Labeler take a stats.RPCTagInfo or stats.ConnTagInfo as an argument?
* If we still used the Labeler approach, could users set the labeler in the context using an interceptor? (Not sure if interceptors run before the stats handler?)

inigohu:
>If we use constant labels as you propose here, what is the use-case? Are constant labels actually useful?

Static attributes could be useful for tagging different servers running on the same service but on different ports, e.g. _public_, _internal, backoffice, legacy_...

>Should the Labeler take a stats.RPCTagInfo or stats.ConnTagInfo as an argument?

These attributes are already being tagged in the spans/metrics.

>If we still used the Labeler approach, could users set the labeler in the context using an interceptor? (Not sure if interceptors run before the stats handler?)

The `HandleRPC` of the `StatsHandler` is executed several times, but to summarise this would be the order:

HandleRPC(begin) -> Interceptors(pre) -> handler logic -> Interceptors(post) -> HandleRPC(end).

The problem is that despite modifying the context in an interceptor, [the interceptor does not propagate it outwards](https://github.com/grpc/grpc-go/blob/c04b085930ce33ee83cc3f92dbe7632031e127a9/server.go#L1379), so it never reaches [the end of the HandleRPC](https://github.com/grpc/grpc-go/blob/c04b085930ce33ee83cc3f92dbe7632031e127a9/server.go#L1246).

jklipp:
Hi, I've ended up here after trying to figure out how to attach custom (static) labels to otelgrpc stats.  Just wanted to add our concrete use case if it's helpful to demonstrate the need.

We've got a central cluster that talks to the same service on multiple remote clusters, and I'd like to be able to label and differentiate remote clusters so I can examine their latency metrics separately.

From what I've been able to determine poking around the codebases, attaching a different client handler with a different set of static labels/attributes on it to each of our outbound grpc client connections would let us distinguish between the connections after everything rolls up to the client latency metric.

The only other way I can see to achieve that is to have different MeterProviders with different labels, but then I believe they would conflict on the metric names if I were to try and publish the metrics through the same exporter (the otel prometheus exporter in this case).

dashpole:
I think this change is a good idea.  @hanyuancheung any thoughts?

dashpole (approved):
please add a changelog entry

CHANGELOG.md
```diff
@@ -12,6 +12,7 @@ This project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0.htm
 
 - Support for stdoutlog exporter in `go.opentelemetry.io/contrib/config`. (#5850)
 - Add macOS ARM64 platform to the compatibility testing suite. (#5868)
+- Add custom attributes to the stats handler in `go.opentelemetry.io/contrib/instrumentation/google.golang.org/grpc/otelgrpc`. (#5133)
```
dmathieu, line 15:
```suggestion
- Add the `WithAttributes` method to set custom attributes to the stats handler in `go.opentelemetry.io/contrib/instrumentation/google.golang.org/grpc/otelgrpc`. (#5133)
```

dmathieu (approved):

instrumentation/google.golang.org/grpc/otelgrpc/config.go
```diff
@@ -257,3 +258,16 @@ func (o spanStartOption) apply(c *config) {
...
+
+type attributesOption struct{ a []attribute.KeyValue }
+
+func (o attributesOption) apply(c *config) {
+	if o.a != nil {
+		c.Attributes = o.a
+	}
+}
+
+// WithAttributes returns an Option to add custom attributes to the spans and metrics.
```
pellared, line 270: I find adding attributes to both spans and metrics problematic. For spans attributes are added to give more context, but for metrics they are used also for aggregation.

pellared: @dashpole WDYT?

dashpole: right... People should probably use `WithSpanOptions(WithAttributes(...` to add span attributes, right?

pellared: I am against span options as it gives to much (changing span kind etc). See https://github.com/open-telemetry/opentelemetry-go-contrib/pull/5431. Maybe dedicated options to set attributes for each signal?

inigohu: @dashpole updated

dashpole (approved):

zchee:
@dmathieu @dashpole Sorry to interrupt, but when will this change be released (merged)?

PR #5163, opened 2024-02-23
Drop support for Go 1.20

MrAlias:


XSAM (approved):

dashpole (commented):
does make go-mod-tidy need to change?

MrAlias:
> does make go-mod-tidy need to change?

Updated

dashpole (approved):

PR #5187, opened 2024-02-27
otelhttptrace: handle missing getconn hook without panic

tonistiigi:
We have many reports that end() gets called without the span being defined in start() and causes a panic.

Ref https://github.com/moby/buildkit/issues/4377
Ref https://github.com/docker/buildx/issues/2232

I have not not fully debugged in what condition this happens in stdlib (I assume some keepalive pool case) but I don't see any other possible explanation for these panic cases.

Note that there are other httptrace hooks that also use `.root` without validation. I don't atm. have any proof that these could be called without `GetConn()` being called first as well so didn't add extra validation to these.

### DIFF
--- a/instrumentation/net/http/httptrace/otelhttptrace/clienttrace.go
+++ b/instrumentation/net/http/httptrace/otelhttptrace/clienttrace.go
@@ -215,6 +215,10 @@ func (ct *clientTracer) start(hook, spanName string, attrs ...attribute.KeyValue
 
 func (ct *clientTracer) end(hook string, err error, attrs ...attribute.KeyValue) {
 	if !ct.useSpans {
+		// sometimes end may be called without previous start
+		if ct.root == nil {
+			ct.root = trace.SpanFromContext(ct.Context)
+		}
 		if err != nil {
 			attrs = append(attrs, attribute.String(hook+".error", err.Error()))
 		}
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -22,6 +22,7 @@ This project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0.htm
 - Custom attributes targeting metrics recorded by the `go.opentelemetry.io/contrib/instrumentation/net/http/otelhttp` are not ignored anymore. (#5129)
 - Use `c.FullPath()` method to set `http.route` attribute in `go.opentelemetry.io/contrib/instrumentation/github.com/gin-gonic/gin/otelgin`. (#5734)
 - The double setup in `go.opentelemetry.io/contrib/instrumentation/net/http/httptrace/otelhttptrace/example` that caused duplicate traces. (#5564)
+- Possible nil dereference panic in `go.opentelemetry.io/contrib/instrumentation/net/http/httptrace/otelhttptrace`. (#5187)
 
 ### Deprecated
 


tonistiigi:
> This change only handles missing span in end() if useSpans is false.
It seems like that could also happen with the use of sub spans as well.

What place/variable do you have in mind?

> Also, I know this is hard to reproduce, and this package doesn't really have any tests at the moment, but would it be possible to write a test?

These calls are made from stdlib and I don't know the condition when the specific call order appears.

pellared:
> These calls are made from stdlib and I don't know the condition when the specific call order appears.

I think the main usage of `httptrace.ClientTrace` is here: https://github.com/golang/go/blob/master/src/net/http/transport.go

alessandrozucca-p:
We started seeing this as well after introducing the option `otelhttptrace.WithoutSubSpans()`

The original implementation was taking locks at the beginning on the function:

[func (ct *clientTracer) end(hook string, err error, attrs ...kv.KeyValue) {](https://github.com/open-telemetry/opentelemetry-go-contrib/commit/3fc65dc78f63bed02e186a9c22b7eb32b565fc56#diff-8bf08b754262dab08ffec9ca5fd56cb4dda0be64bb24c0c08947e615b2aa0445R114-R116)
[ct.mtx.Lock()](https://github.com/open-telemetry/opentelemetry-go-contrib/commit/3fc65dc78f63bed02e186a9c22b7eb32b565fc56#diff-8bf08b754262dab08ffec9ca5fd56cb4dda0be64bb24c0c08947e615b2aa0445R114-R116)

[defer ct.mtx.Unlock()](https://github.com/open-telemetry/opentelemetry-go-contrib/commit/3fc65dc78f63bed02e186a9c22b7eb32b565fc56#diff-8bf08b754262dab08ffec9ca5fd56cb4dda0be64bb24c0c08947e615b2aa0445R114-R116)

However [since the introduction of](https://github.com/open-telemetry/opentelemetry-go-contrib/blob/zpages/v0.52.0/instrumentation/net/http/httptrace/otelhttptrace/clienttrace.go#L225) `if !ct.useSpans {` the locks are not taken immediately. 

It made me wonder, could this have cause some race condition?

instrumentation/net/http/httptrace/otelhttptrace/clienttrace.go
```diff
@@ -226,6 +226,10 @@ func (ct *clientTracer) start(hook, spanName string, attrs ...attribute.KeyValue
 
 func (ct *clientTracer) end(hook string, err error, attrs ...attribute.KeyValue) {
 	if !ct.useSpans {
+		// sometimes end may be called without previous start
+		if ct.root == nil {
```
dmathieu, line 230: Could this be tested?

dmathieu:

tonistiigi:
Added changelog

instrumentation/net/http/httptrace/otelhttptrace/clienttrace.go
xocasdashdash: as @tonistiigi mentioned, they've gotten _some_ reports, there's very little consistency on when this happens. We've seen very similar problems with clients _very rarely_ blowing up with this same error. 

If we don't add a test

dmathieu: A unit test doesn't have to be a full reproduction. It can call `end()` with conditions that reproduce this case, just to ensure the code path doesn't fail.

krajorama:
Hi, just ran into https://github.com/moby/buildkit/issues/4377 . Any update on this PR? I understand only unit test is missing?

instrumentation/net/http/httptrace/otelhttptrace/clienttrace.go
krantideep95: I have created a new pull request https://github.com/open-telemetry/opentelemetry-go-contrib/pull/5965 which adds unit test for `end()` func

PR #5251, opened 2024-03-07
Otelmux span options

### LINKED ISSUE #5139: exception.stacktrace attribute isn't included in exception events
### Description

When a panic occurs in a request, an `exception` event is created, but the `exception.stacktrace` attribute isn't set.
According to the [otel spec on exceptions](https://opentelemetry.io/docs/specs/otel/trace/exceptions/#attributes), the `exception.stacktrace` attribute SHOULD be set. Currently it isn't, and I haven't been able to find a way to have it included.

### Environment

- OS: macOS
- Architecture: arm64
- Go Version: 1.22.0
- `otelmux` version: 0.48.0

### Steps To Reproduce

1. Use the otelmux middleware
2. Call an end-point that panics
3. See that the trace created contains an `exception` event, with no `stacktrace` attribute

### Expected behavior

I would expect the exception event to include the stacktrace. I see how this may not always be desired, so an alternative to always including it, could be a way to explicitly tell the middleware to include it.

mikk2168:
This allows setting span start and end options to the otelmux configuration, so folks can (for example), enable stack traces to record errors.

Heavily inspired by #5250.
Since `otelmux` differs a bit from the `gin` implementation, I'm not sure if adapting the `traceware` to include `SpanStartOptions` and `SpanEndOptions` is the right way to go. If not, please let me know how it should be adapted.

Closes #5139.


dmathieu (approved):

hanyuancheung (approved):

makeavish:
@dmathieu @hanyuancheung Any updates on merging this PR? Since it's approved for 2 weeks

dmathieu:
Only @open-telemetry/go-maintainers can merge PRs.

instrumentation/github.com/gorilla/mux/otelmux/config.go
```diff
@@ -86,6 +88,22 @@ func WithSpanNameFormatter(fn func(routeName string, r *http.Request) string) Op
 	})
 }
 
+// WithSpanStartOption applies options to all the HTTP span created by the
+// instrumentation.
+func WithSpanStartOption(o ...oteltrace.SpanStartOption) Option {
```
pellared, line 93: In [`otelhttp`](https://pkg.go.dev/go.opentelemetry.io/contrib/instrumentation/net/http/otelhttp#WithSpanOptions) and [`otelgrpc`](https://pkg.go.dev/go.opentelemetry.io/contrib/instrumentation/google.golang.org/grpc/otelgrpc#WithSpanOptions) this option is called `WithSpanOptions`. Can we name it in the same way for sake of consistency?

pellared (changes requested):
I am not sure if it is good to allow users to do that much.

The user can e.g. change the span kind using https://pkg.go.dev/go.opentelemetry.io/otel/trace#WithSpanKind or create attributes which are not aligned with the semantic conventions and they others may blame that the instrumentation library does not follow semantic conventions.

> so folks can (for example), enable stack traces to record errors.

I am not sure if this is a good justification. This instrumentation library is not recording any errors. 

I think we can discuss during the SIG meeting if we should make instrumentation libraries so configurable.  

I am aware that there is a precedent that `otelhttp` and `otelgrpc` already have `WithSpanOptions` but maybe we should retrospect this.

mikk2168:
@pellared Thank you for the review.
> I am not sure if it is good to allow users to do that much.
>
>The user can e.g. change the span kind using https://pkg.go.dev/go.opentelemetry.io/otel/trace#WithSpanKind or create attributes which are not aligned with the semantic conventions and they others may blame that the instrumentation library does not follow semantic conventions.

This sounds like a very valid point to me. In #5250 you suggest adding the option to capture stacktraces instead. Does that apply here as well?
If so, I'm happy to try and take a stab at it, when I have the time. Since I don't know this project too well, any pointer to where this option should be added will be much appreciated.

instrumentation/github.com/gorilla/mux/otelmux/config.go
mikk2168: Absolutely. I'll leave the code as is for now, since this may not be relevant, if it is replaced with the option to add stacktraces instead - see [this comment](https://github.com/open-telemetry/opentelemetry-go-contrib/pull/5251#issuecomment-2036544502).
I'll return to this, if we decide to keep the function.

mikk2168:
>No as this instrumentation library does not handle panics and does not call https://pkg.go.dev/go.opentelemetry.io/otel/trace#Span.RecordError.

Ah, I see. That makes sense.
Considering your initial comment, does it make sense to continue down this path, allowing users to add `SpanStartOption`s and `SpanEndOption`s? I agree with your concerns, so I see why this might not be the best solution.
If not, does it make sense to implement the error recording/stacktrace handling in a different way, or is it something that should be left out of the library (for now)?

CHANGELOG.md
```diff
@@ -11,6 +11,7 @@ This project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0.htm
 ### Added
 
 - Add the new `go.opentelemetry.io/contrib/instrgen` package to provide auto-generated source code instrumentation. (#3068, #3108)
+- Add `WithSpanStartOption` and `WithSpanEndOption` to the `go.opentelemetry.io/contrib/github.com/gorilla/mux/otelmux` package to provide options on span start and end. (#5250)
```
dashpole: The package path is missing `instrumentation/` and the PR reference points to `#5250` instead of `#5251`.
```suggestion
- Add `WithSpanStartOption` and `WithSpanEndOption` to the `go.opentelemetry.io/contrib/instrumentation/github.com/gorilla/mux/otelmux` package to provide options on span start and end. (#5251)
```

instrumentation/github.com/gorilla/mux/otelmux/mux.go
```diff
@@ -141,10 +145,10 @@ func (tw traceware) ServeHTTP(w http.ResponseWriter, r *http.Request) {
 		}
 	}
 
-	opts := []trace.SpanStartOption{
+	opts := append(tw.SpanStartOptions, []trace.SpanStartOption{
```
dashpole: Appending directly to `tw.SpanStartOptions` here can mutate its backing array across concurrent requests when `cap(tw.SpanStartOptions) > len(tw.SpanStartOptions)`, causing a data race. Consider allocating a fresh slice per request before appending:
```suggestion
	opts := make([]trace.SpanStartOption, 0, len(tw.SpanStartOptions)+4)
	opts = append(opts, tw.SpanStartOptions...)
	opts = append(opts, []trace.SpanStartOption{
```

PR #5278, opened 2024-03-13
otelgrpc: histogram buckets

saturn4er:
This allows package clients to explicitly specify bucket boundaries.


dashpole (commented):
Have you tried using a view to configure the bucket boundaries?  See https://opentelemetry.io/docs/languages/go/instrumentation/#registering-views and https://github.com/open-telemetry/opentelemetry-go/blob/6bc8314a5553ba97e4cfaf7838007085a6be2dea/sdk/metric/aggregation.go#L98

saturn4er:
Hi @dashpole, thanks, that helped

Issue #5191, opened 2024-03-01
Add zap log bridge


pellared:
Add [`zapcore.Core`](https://pkg.go.dev/go.uber.org/zap/zapcore#Core) log bridge.

pellared:
@khushijain21 is working on it.

khushijain21:
Outline of the approach

- [ ] Skeleteon of zapcore and config for the bridge
- [ ] Implement zapcore methods
- [ ] Implement zapcore encoder
- [ ] Add Benchmarks

pellared:
1. We do not need to implement zapcore.Encoder as we are providing a zapcore.Core
2. You can first implement `Write` and then `With` method as separate PRs
3. You can handle https://pkg.go.dev/go.uber.org/zap/zapcore#NamespaceType as a separate PR
4. You can handle trace correlation (handling `context.Context`) as a separate PR
5. After benchmarks are in place we can add performance improvements. I suggests adding usage of `sync.Pool` and other improvements after benchmarks are in place and we have required functionality in place.
6. Go doc comment should be updated after the implementation is done.
7. A testable example should be added demonstrating how to setup `otelzap`.
8. After the release the Go module should be added to https://opentelemetry.io/ecosystem/registry
9. After the release we could document the bridge also in https://opentelemetry.io/docs/languages/go/

Do you want to be a CODEOWNER? Can you maintain `otelzap` after your internship is finished?

PR #5279, opened 2024-03-14
Logging bridge for Zap

khushijain21:
Part of #5191

Some Notes:
Basic Structure - Zapcore implements
    - `Check` -  determines whether the supplied Entry should be logged
    - `Write`
    - `Enabled` 
    - `Sync` - we don't use buffer to write - so no code here
    - `With` - returns a child core with provided context 
    
This also implements `zapcore.ObjectEncoder ` and `zapcore.ArrayEncoder`  which decides how we would encode provided fields to log attrs - (the user can also pass their custom encoder)
Some Notes about type mapping:
   - All Uint types are converted to Int64
   - All complex types are converted to equiv string
   - Everything else which uses zap.Any to pass context and if does not match supported primitive types - is handled by `AddReflected` method which converts the field to JSON


bridges/zap/go.mod
```diff
@@ -0,0 +1,25 @@
+module zap
```
MrAlias, line 1: This module name has a high chance of conflicting with the package it is trying to bridge with (https://pkg.go.dev/go.uber.org/zap). We should try to define this so users won't need to rename the import.

```suggestion
module otelzap
```

This will need additional changes other than this to fully migrate.

bridges/zap/go.mod
```diff
@@ -0,0 +1,25 @@
...
+
+go 1.21.3
```
MrAlias, line 3:
```suggestion
go 1.21
```

bridges/zap/version.go
```diff
@@ -0,0 +1,6 @@
+package zap // import "go.opentelemetry.io/contrib/bridges/zap"
+
+// Version is the current release version of sloghandler in use.
+func Version() string {
+	return "0.0.1-alpha"
+}
```
MrAlias, line 6: Similar to what was suggested for the [slog bridge](https://github.com/open-telemetry/opentelemetry-go-contrib/blob/851b2fd97fa45bb502cc094b6e3e84464a140fc3/bridges/otelslog/version.go#L7), can you make this into an unexported const.

bridges/zap/version.go
```diff
@@ -0,0 +1,6 @@
+package zap // import "go.opentelemetry.io/contrib/bridges/zap"
```
MrAlias, line 1: All code and documentation files need to have a license header.

```suggestion
// Copyright The OpenTelemetry Authors
// SPDX-License-Identifier: Apache-2.0

package zap // import "go.opentelemetry.io/contrib/bridges/zap"
```

bridges/zap/zap_encoder.go
```diff
@@ -0,0 +1,211 @@
...
+	"fmt"
+	"strconv"
+	"time"
+
+	"go.opentelemetry.io/otel/log"
+	"go.uber.org/zap/zapcore"
+)
+
+// not optimized yet
+// this file implements object and array encoder - similar to memory encoder by zapcore
```
MrAlias, line 14: You should remove development comment here.

If you based this code on another project you need to provide a link. If it was copied in any way it's license needs to be reference in this file as well.

bridges/zap/zap_encoder.go
```diff
@@ -0,0 +1,211 @@
...
+var (
+	_ zapcore.ObjectEncoder = (*OtelObjectEncoder)(nil)
+	_ zapcore.ArrayEncoder  = (*sliceArrayEncoder)(nil)
+)
+
+type OtelObjectEncoder struct {
```
MrAlias, line 20: The short hand for OpenTelemetry is OTel.

```suggestion
type OTelObjectEncoder struct {
```

However, with the above package name suggestion, there is little need to include `OTel` in the name at all given it will be in the `otelzap` package scope.

```suggestion
type ObjectEncoder struct {
```

Would be better.

bridges/zap/zap_encoder.go
```diff
@@ -0,0 +1,211 @@
...
+	// Fields contains the entire encoded log context.
+	Fields []log.KeyValue
+	// cur is a pointer to the namespace we're currently writing to.
+	cur []log.KeyValue
+
+	reflectval log.Value
+	zapcore.Encoder
```
MrAlias, line 27: Why is this embedded? If there are methods this is providing, when they are called by a user it will cause a panic. We should implement all the needed methods instead.

bridges/zap/zap_encoder.go
```diff
@@ -0,0 +1,211 @@
...
+}
+
+// NewMapObjectEncoder creates a new map-backed ObjectEncoder.
```
MrAlias, line 30: This comment looks stale. There is no map backing this.

bridges/zap/zap_encoder.go
```diff
@@ -0,0 +1,211 @@
...
+func NewOtelObjectEncoder() *OtelObjectEncoder {
+	m := make([]log.KeyValue, 0)
+	return &OtelObjectEncoder{
+		Fields: m,
+		cur:    m,
+	}
+}
+
+// AddArray implements ObjectEncoder.
```
MrAlias, line 39: Can you replace these types of comments with ones describing what the method does. This does not provide information to the user of the API.

bridges/zap/zap_encoder.go
```diff
@@ -0,0 +1,211 @@
...
+func (m *OtelObjectEncoder) AddArray(key string, v zapcore.ArrayMarshaler) error {
+	fmt.Println(v, "inside array")
```
MrAlias, line 41:
```suggestion
```

bridges/zap/zap_encoder.go
```diff
@@ -0,0 +1,211 @@
...
+	arr := &sliceArrayEncoder{elems: make([]log.Value, 0)}
+	err := v.MarshalLogArray(arr)
+	m.cur = append(m.cur, log.Slice(key, arr.elems...))
+	return err
+}
+
+// AddObject implements ObjectEncoder.
+func (m *OtelObjectEncoder) AddObject(k string, v zapcore.ObjectMarshaler) error {
+	fmt.Println(v, "inside object")
```
MrAlias, line 50:
```suggestion
```

bridges/zap/zap_encoder.go
```diff
@@ -0,0 +1,211 @@
...
+		cur:    m,
+	}
+}
+
+// AddArray implements ObjectEncoder.
+func (m *OtelObjectEncoder) AddArray(key string, v zapcore.ArrayMarshaler) error {
+	fmt.Println(v, "inside array")
+	arr := &sliceArrayEncoder{elems: make([]log.Value, 0)}
+	err := v.MarshalLogArray(arr)
+	m.cur = append(m.cur, log.Slice(key, arr.elems...))
```
MrAlias, line 44: Appending like this is going to require heap allocations to increase the size of the slice. Is this done in the "hot path" of recording logs? If so we need to consider alternate approaches for performance.

bridges/zap/zap_encoder.go
```diff
@@ -0,0 +1,211 @@
...
+// AddDuration implements ObjectEncoder.
+func (m *OtelObjectEncoder) AddDuration(k string, v time.Duration) { m.AddInt64(k, v.Nanoseconds()) }
+
+// AddComplex128 implements ObjectEncoder.
+func (m *OtelObjectEncoder) AddComplex128(k string, v complex128) {
+	stringValue := strconv.FormatComplex(v, 'f', -1, 64)
+	m.cur = append(m.cur, log.String(k, stringValue))
+}
+
+func (m *OtelObjectEncoder) AddFloat64(k string, v float64) {
```
MrAlias, line 81: The following exported methods are missing documentation.

bridges/zap/zap_logger.go
```diff
@@ -0,0 +1,119 @@
...
+	"go.opentelemetry.io/otel/log"
+	"go.opentelemetry.io/otel/log/noop"
+	"go.uber.org/zap/zapcore"
+)
+
+const (
+	bridgeName = "go.opentelemetry.io/contrib/bridge/zapcore"
+)
+
+type OtelZapCore struct {
```
MrAlias, line 15: Similar here regarding naming.

```suggestion
type Core struct {
```

MrAlias, line 15: Missing documentation.

bridges/zap/zap_logger.go
```diff
@@ -0,0 +1,119 @@
...
+	attr   []log.KeyValue
+}
+
+var (
+	_ zapcore.Core = (*OtelZapCore)(nil)
+)
+
+// this function creates a new zapcore.Core that can be used with zap.New()
+// this instance will translate zap logs to opentelemetry logs and export them
+func NewOtelZapCore(lp log.LoggerProvider, opts ...log.LoggerOption) zapcore.Core {
```
MrAlias, line 26: This needs to accept its own package defined `Option` type.

https://github.com/open-telemetry/opentelemetry-go-contrib/blob/851b2fd97fa45bb502cc094b6e3e84464a140fc3/bridges/otelslog/version.go#L7

bridges/zap/zap_test.go
```diff
@@ -0,0 +1,237 @@
...
+	childlogger.Info(testBodyString)
+	spy.Record.WalkAttributes(func(kv log.KeyValue) bool {
+		assert.Equal(t, "workplace", string(kv.Key))
+		assert.Equal(t, "otel", kv.Value.AsString())
+		return true
+	})
+
+}
+
+// Copied from field_test.go
```
MrAlias, line 72: Please provide full URL

bridges/zap/zap_logger.go
```diff
@@ -0,0 +1,119 @@
+package zap
```
MrAlias, line 1: Missing package documentation.

MrAlias (commented):
Good start

MrAlias:
This module will need to be added to the project [`versions.yaml`](https://github.com/open-telemetry/opentelemetry-go-contrib/blob/851b2fd97fa45bb502cc094b6e3e84464a140fc3/versions.yaml#L77) under the unreleased section.

bridges/zap/zap_encoder.go
khushijain21: there is no way to know about array size here - can we agree upon a minimum buffer size?

bridges/otelzap/zap_encoder.go
```diff
@@ -0,0 +1,233 @@
...
+
+// this file implements object and array encoder - similar to memory encoder by zapcore.
+var (
+	_ zapcore.ObjectEncoder = (*ObjectEncoder)(nil)
+	_ zapcore.ArrayEncoder  = (*ArrayEncoder)(nil)
+)
+
+// Object Encoder implements zapcore.ObjectEncoder.
+// It encodes given fields to OTel attribute.
+type ObjectEncoder struct {
```
pellared, line 26: Does it need to be exported?

```suggestion
type objectEncoder struct {
```

bridges/otelzap/zap_encoder.go
```diff
@@ -0,0 +1,233 @@
...
+func (m *ObjectEncoder) AddInt8(k string, v int8)           { m.AddInt64(k, int64(v)) }
+func (m *ObjectEncoder) AddTime(k string, v time.Time)      { m.AddInt64(k, v.UnixNano()) }
+func (m *ObjectEncoder) AddUint(k string, v uint)           { m.AddUint64(k, uint64(v)) }
+func (m *ObjectEncoder) AddUint32(k string, v uint32)       { m.AddInt64(k, int64(v)) }
+func (m *ObjectEncoder) AddUint16(k string, v uint16)       { m.AddInt64(k, int64(v)) }
+func (m *ObjectEncoder) AddUint8(k string, v uint8)         { m.AddInt64(k, int64(v)) }
+func (m *ObjectEncoder) AddUintptr(k string, v uintptr)     { m.AddUint64(k, uint64(v)) }
+
+// ArrayEncoder implements zapcore.ArrayEncoder.
+type ArrayEncoder struct {
```
pellared, line 160: Does it need to be exported?

```suggestion
type arrayEncoder struct {
```

bridges/otelzap/zap_logger.go
```diff
@@ -0,0 +1,119 @@
...
+}
+
+// LevelEnabler decides whether a given logging level is enabled when logging a
+// message.
+func (o *Core) Enabled(level zapcore.Level) bool {
+	r := log.Record{}
+	r.SetSeverity(getOtelLevel(level))
+
+	// TODO:
+	// check how to propagate context
```
pellared, line 46: I am not sure if it would be possible.

```suggestion
	// TODO:
	// check how to propagate context
```

bridges/otelzap/zap_logger.go
```diff
@@ -0,0 +1,119 @@
...
+	attr := getAttr(fields)
+	// append attributes received from from parent logger
+	addattr := append(attr, o.attr...)
+
+	if len(addattr) > 0 {
+		r.AddAttributes(addattr...)
+	}
+
+	// TODO
+	// need to check how to propagate context here
```
pellared, line 89: I think that if one passed a `context.Context` via attribute using https://pkg.go.dev/go.uber.org/zap#Any e.g.

```
logger.Info("failed to fetch URL",
  zap.Any("context", ctx),
  zap.Int("attempt", 3),
  zap.Duration("backoff", time.Second),
)
```

then we can use it.

If have done something like this in logr bridge prototype https://github.com/pellared/opentelemetry-go/blob/147762ffb5c4394417b3f973274cc570723db79b/log/internal/logr.go#L49-L53

This behavior should be clearly documented in `otelzap` package documentation.

bridges/otelzap/zap_test.go
```diff
@@ -0,0 +1,353 @@
+// Copyright The OpenTelemetry Authors
+// SPDX-License-Identifier: Apache-2.0
+
+// Copyright (c) 2016-2017 Uber Technologies, Inc.
+package otelzap
```
pellared, line 5:
```suggestion
// Copyright (c) 2016-2017 Uber Technologies, Inc.

package otelzap
```

There needs to be a blank line. Otherwise it would render as package documentation.

bridges/otelzap/zap_test.go
```diff
@@ -0,0 +1,353 @@
...
+
+func (l *spyLogger) Enabled(ctx context.Context, r log.Record) bool {
+	return true
+}
+
+func NewTestOtelLogger(l log.Logger) zapcore.Core {
+	return &Core{
+		logger: l,
+	}
+}
```
pellared, line 52: This looks unnecessary.

bridges/otelzap/zap_test.go
```diff
@@ -0,0 +1,353 @@
...
+
+type addr struct {
+	IP   string
+	Port int
+}
+
+// Basic Logger Test and Child Logger test.
+func TestZapCore(t *testing.T) {
```
pellared, line 60: This test is falling.

bridges/otelzap/zap_logger.go
```diff
@@ -0,0 +1,119 @@
...
+	ctx := context.Background()
+	o.logger.Emit(ctx, r)
+	return nil
+}
+
+func (o *Core) clone() *Core {
+	return &Core{
+		logger: o.logger,
+		attr:   o.attr,
```
pellared, line 98: We should make shallow copy using https://pkg.go.dev/slices#Clone

```suggestion
		attr:   slices.Clone(o.attr),
```

Here is why: https://blog.allegro.tech/2017/07/golang-slices-gotcha.html

bridges/otelzap/zap_logger.go
```diff
@@ -0,0 +1,119 @@
...
+	}
+	return enc.cur
+}
+
+// converts zap level to Otel's log level.
+func getOtelLevel(level zapcore.Level) log.Severity {
+	// should confirm this
+	// the logic here is that
+	// zapcore.Debug = -1 & logger.Debug = 3
+	// zapcore.Info = 0   & logger.Info = 7 and so on
```
pellared, line 116: Add unit tests for all constants of https://pkg.go.dev/go.uber.org/zap/zapcore#Level

bridges/otelzap/zap_logger.go
```diff
@@ -0,0 +1,119 @@
+// Copyright The OpenTelemetry Authors
+// SPDX-License-Identifier: Apache-2.0
+
+// Package otelzap provides a bridge between the ["go.uber.org/zap"] and
```
pellared, line 4:
```suggestion
// Package otelzap provides a bridge between the [go.uber.org/zap] and
```

bridges/otelzap/zap_encoder.go
khushijain21: ObjectEncoder is an interface by zap which expects this to be exported. One example where a user may use this is https://pkg.go.dev/go.uber.org/zap#example-Object

bridges/otelzap/zap_test.go
pellared: For now you could simply inline it

```go
logger := &Core{ logger: l }
```

However, it may be better to simply copy-paste: https://github.com/open-telemetry/opentelemetry-go-contrib/blob/d15fe4f317e0376ff4a65a435b785a1c88e11b7c/bridges/otelslog/handler_test.go#L35-L76.

@MrAlias. Do you think it would be worth to extract this code to a separate package and module?

bridges/otelzap/zap_test.go
```diff
@@ -0,0 +1,261 @@
...
+		return true
+	})
+}
+
+func TestGetOtelLevel(t *testing.T) {
+	tests := []struct {
+		level       zapcore.Level
+		expectedSev log.Severity
+	}{
+		{zapcore.DebugLevel, log.SeverityDebug},   // Expected value for DebugLevel
```
pellared, line 62: Can we also add a test for an unhandled level?

bridges/zap/zap_encoder.go
khushijain21: @pellared

bridges/otelzap/zap_encoder.go
```diff
@@ -0,0 +1,233 @@
...
+
+func (a *ArrayEncoder) AppendComplex128(v complex128) {
+	stringValue := `"` + strconv.FormatComplex(v, 'f', -1, 64) + `"`
+	a.elems = append(a.elems, log.StringValue(stringValue))
+}
+
+func (a *ArrayEncoder) AppendFloat32(v float32) {
+	// preserves float32 value
+	value, _ := strconv.ParseFloat(strconv.FormatFloat(float64(v), 'f', -1, 32), 64)
+	a.AppendFloat64(value)
```
pellared, line 193: Why not

```suggestion
	a.AppendFloat64(float64(v))
```

bridges/otelzap/zap_logger.go
pellared: E.g. 



This is how the bridge can get the passed context:

```go
// converts zap fields to Otel's log attributes.
func getAttr(ctx context.Context, fields []zapcore.Field) (context.Context, []log.KeyValue) {
	enc := NewObjectEncoder(len(fields))
	for _, fld := range fields {
		if ctxFld, ok := fld.Interface.(context.Context); ok {
			ctx = ctxFld
			continue
		}
		fld.AddTo(enc)
	}
	return ctx, enc.cur
}
```

bridges/otelzap/zap_encoder.go
```diff
@@ -0,0 +1,233 @@
...
+// It encodes given fields to OTel attribute.
+type ObjectEncoder struct {
+	// cur is a pointer to the namespace we're currently writing to.
+	cur []log.KeyValue
+
+	reflectval log.Value
+}
+
+// NewObjectEncoder returns ObjectEncoder which maps zap fields to OTel attributes.
+func NewObjectEncoder(len int) *ObjectEncoder {
```
pellared, line 34: This function should be unexported.

bridges/otelzap/zap_encoder.go
khushijain21: Because float64(v) doesn't preserve the float32 value as is. This is the corresponding test that will fail 
https://github.com/open-telemetry/opentelemetry-go-contrib/blob/9579fa0c261ceeadae8e2cd499df1034c5b1f205/bridges/otelzap/zap_test.go#L103

Fo ex: 
```
a := float32(3.14)
fmt.Println(float64(a))
```
will return `3.140000104904175`

pellared: Use https://pkg.go.dev/github.com/stretchr/testify/assert#InDelta for asserting the result.

dmathieu:
Kind of nit, but do the file names need to be prepended with `zap_`?
That's not something we're doing in other packages, and this package only holds the zap bridge. So having the library name in the filename is a bit redundant.

pellared:
https://github.com/open-telemetry/opentelemetry-go-contrib/actions/runs/8595139722/job/23549314101?pr=5279

Please use `make` to check for CI build errors.

bridges/otelzap/core.go
```diff
@@ -0,0 +1,135 @@
...
+	return &Core{
+		logger: o.logger,
+		attr:   slices.Clone(o.attr),
+		ctx:    o.ctx,
+	}
+}
+
+// converts zap fields to OTel log attributes.
+func getAttr(ctx context.Context, fields []zapcore.Field) (context.Context, []log.KeyValue) {
+	enc := newObjectEncoder(len(fields))
```
pellared, line 99: This makes an allocation (on the hot path). Is there any reason why we are not using `getObjectEncoder` and `putObjectEncoder` here.

bridges/otelzap/core.go
```diff
@@ -0,0 +1,135 @@
...
+		ctx = enc.ctxfield
+	}
+	return ctx, enc.cur
+}
+
+// Method to pass context to OTel loggger as [zap.Field]
+// Ex: logger.Info("msg", otelzap.Context("key", ctx)).
+func Context(key string, val context.Context) zap.Field {
+	return zap.Reflect(key, val)
+}
```
pellared, line 113: The purpose of logging bridge is that code emitting logs do not have to depend on OTel. Why do we need it?

bridges/otelzap/encoder.go
```diff
@@ -0,0 +1,266 @@
...
+}
+
+// Converts all non-primitive types to JSON string
+// Also checks for explcit context passed
+func (m *objectEncoder) AddReflected(k string, v interface{}) error {
+	if ctx, ok := v.(context.Context); ok {
+		// assign ctx
+		m.ctxfield = ctx
+		return nil
+	}
```
pellared, line 160: Why not doing a type assertion like this in `getAttr`?

pellared: See https://github.com/open-telemetry/opentelemetry-go-contrib/pull/5279#discussion_r1544294319

bridges/otelzap/core.go
pellared: See https://github.com/open-telemetry/opentelemetry-go-contrib/pull/5279#discussion_r1544294319

khushijain21: oh, I thought this would make it clear that it comes from otelzap bridge and not from zap.  And did not add in getAttrs field as we would run the if condition on every field.

pellared: > as we would run the if condition on every field.

As long as the type assertion condition does not produce an additional heap allocation, it is not a problem.

Coupling logging code to `otelzap` defeats the idea of the bridge.

pellared: I think here is a related issue in Zap: https://github.com/uber-go/zap/issues/654

bridges/otelzap/core.go
```diff
@@ -0,0 +1,135 @@
...
+	attr   []log.KeyValue
+	ctx    context.Context
+}
+
+// Compile-time check *Core implements zapcore.Core.
+var _ zapcore.Core = (*Core)(nil)
+
+// NewOTelZapCore creates a new [zapcore.Core] that can be used with zap.New()
+// this instance will translate zap logs to opentelemetry logs and export them.
+func NewOTelZapCore(opts ...Option) zapcore.Core {
```
dmathieu, line 34: Kind of nit, but as in my previous comment, having `Otel` and `Zap` here is redundant, since that's all there will be in the package.

```suggestion
func NewCore(opts ...Option) zapcore.Core {
```

With this, the call becomes `otelzap.NewCore`, which indicates everything already.

bridges/otelzap/encoder.go
pellared, line 1: Why do we even need to implement the encoders if the only think we have to provide is a `zapcore.Core`?

bridges/otelzap/core.go
```diff
@@ -0,0 +1,135 @@
...
+}
+
+// With adds structured context to the Core.
+func (o *Core) With(fields []zapcore.Field) zapcore.Core {
+	clone := o.clone()
+	ctx, attr := getAttr(clone.ctx, fields)
+	clone.ctx = ctx
+	clone.attr = append(clone.attr, attr...)
+	return clone
+}
```
pellared, line 56: This should handle [namespaces](https://pkg.go.dev/go.uber.org/zap#Namespace). `Core.Write` also has to handle the namespaces.

The namespace handling code can be heavily inspired by groups implementation in slog bridge. See: https://github.com/open-telemetry/opentelemetry-go-contrib/blob/b343370e70b6e5a4881b7cd9915af95bd58d121f/bridges/otelslog/handler.go

bridges/otelzap/test_helper.go
```diff
@@ -0,0 +1,149 @@
...
+	"go.opentelemetry.io/otel/log/embedded"
+	"go.opentelemetry.io/otel/sdk/instrumentation"
+)
+
+// embeddedLogger is a type alias so the embedded.Logger type doesn't conflict
+// with the Logger method of the recorder when it is embedded.
+type embeddedLogger = embedded.Logger // nolint:unused  // Used below.
+
+// recorder records all [log.Record]s it is ased to emit.
+type recorder struct {
```
pellared, line 24: Can you use https://pkg.go.dev/go.opentelemetry.io/otel/log@v0.1.0-alpha.0.20240409140646-648b40eae158/logtest instead?

You can get it via `go get go.opentelemetry.io/otel/log@v0.1.0-alpha.0.20240409140646-648b40eae158`.

bridges/otelzap/core_test.go
```diff
@@ -0,0 +1,321 @@
...
+	}
+
+	assert.Equal(t, map[string]interface{}{
+		"k":     "s",
+		"users": int64(10),
+		"nested": map[string]interface{}{
+			"users": int64(11),
+		},
+	}, gotVal)
+}
```
pellared, line 173: We should test our `Core`. We do need to provide custom encoders.

pellared:
Changing this to a draft PR as I assume this is a "reference/design PR".

PR #5281, opened 2024-03-17
autoexport: Add OTEL_METRICS_PRODUCERS environment variable support

gouthamve:
This will allow using the Prometheus bridge with autoexport.

The goal is to make it easy for any Prometheus exporter to also be able to emit OTLP. See: https://prometheus.io/blog/2024/03/14/commitment-to-opentelemetry/#otlp-export-in-the-ecosystem


exporters/autoexport/metrics.go
```diff
@@ -28,10 +28,17 @@ type MetricOption = option[metric.Reader]
 
 // WithFallbackMetricReader sets the fallback exporter to use when no exporter
 // is configured through the OTEL_METRICS_EXPORTER environment variable.
-func WithFallbackMetricReader(metricReaderFactory func(ctx context.Context) (metric.Reader, error)) MetricOption {
+func WithFallbackMetricReader(metricReaderFactory func(ctx context.Context, cfg config[metric.Reader]) (metric.Reader, error)) MetricOption {
 	return withFallbackFactory[metric.Reader](metricReaderFactory)
 }
 
+// WithProducer registers producers as an external Producer of metric data for this Reader.
+func WithProducer(producer metric.Producer) MetricOption {
```
dashpole, line 36: @open-telemetry/go-approvers Are we OK mixing option-based config with env-var-based config?  The alternative would be to have a registry for metric producers, and an environment variable that can set them (e.g. `OTEL_METRIC_PRODUCERS=opencensus,prometheus,goruntime`).  I would lean towards a WithProducer option (this PR) because producers seem tightly coupled to your instrumentation (e.g. opencensus or prometheus libraries), rather than export configuration.  But it does also break from the current "all env vars" model.

dashpole (commented):
@gouthamve thanks for this!  It will need an entry in CHANGELOG.md

gouthamve:
Thanks for your review @dashpole!

Anything I can do to move this along @open-telemetry/go-approvers?

pellared:
@gouthamve Sorry for the delay. I will do my best to review it this week.

exporters/autoexport/metrics.go
pellared: Moreover, `OTEL_METRICS_PRODUCERS` is NOT defined in https://github.com/open-telemetry/opentelemetry-specification/blob/main/specification/configuration/sdk-environment-variables.md.

exporters/autoexport/metrics.go
```diff
@@ -28,10 +28,17 @@ type MetricOption = option[metric.Reader]
 
 // WithFallbackMetricReader sets the fallback exporter to use when no exporter
 // is configured through the OTEL_METRICS_EXPORTER environment variable.
-func WithFallbackMetricReader(metricReaderFactory func(ctx context.Context) (metric.Reader, error)) MetricOption {
+func WithFallbackMetricReader(metricReaderFactory func(ctx context.Context, cfg config[metric.Reader]) (metric.Reader, error)) MetricOption {
```
pellared, line 31: It will not be possible to use `WithFallbackMetricReader` as `config[metric.Reader]` is unexported.

Side note: We are missing unit tests that use `WithFallbackMetricReader` and `WithFallbackSpanExporter`

exporters/autoexport/metrics.go
```diff
@@ -67,14 +74,14 @@ func NewMetricReader(ctx context.Context, opts ...MetricOption) (metric.Reader,
 // RegisterMetricReader sets the MetricReader factory to be used when the
 // OTEL_METRICS_EXPORTERS environment variable contains the exporter name. This
 // will panic if name has already been registered.
-func RegisterMetricReader(name string, factory func(context.Context) (metric.Reader, error)) {
+func RegisterMetricReader(name string, factory func(context.Context, config[metric.Reader]) (metric.Reader, error)) {
```
pellared, line 77: It will not be possible to use `RegisterMetricReader` as `config[metric.Reader]` is unexported.

pellared (changes requested):

gouthamve:
Thanks for the review @pellared. I did the easiest thing here and exposed the `Config` object. If you see a better way, please let me know.

Also, this seems a little counter-intuitive to me. We are using the same `factory` and `Config` interface for every signal type, but the options I added are only required for the "metric" signals.

exporters/autoexport/metrics.go
pellared: SIG meeting:
Let's go for env-var-based config and concurrently propose `OTEL_METRICS_PRODUCERS` in the spec.

gouthamve:
Hi @pellared @dashpole took a stab at implementing what we discussed? Does this direction look good?

Will add unit tests once you confirm!

exporters/autoexport/metrics.go
```diff
@@ -148,6 +172,10 @@ func init() {
 
 		return readerWithServer{lis.Addr(), reader, &server}, nil
 	})
+
+	RegisterMetricProducer("prometheus", func(ctx context.Context) (metric.Producer, error) {
+		return prometheusbridge.NewMetricProducer(), nil
```
dashpole, line 177: @gouthamve is it going to be a problem for any of your use-cases that this relies on the global prometheus default gatherer?

dashpole (commented):
implementation looks correct to me.

exporters/autoexport/metrics.go
```diff
@@ -148,6 +172,10 @@ func init() {
 
 		return readerWithServer{lis.Addr(), reader, &server}, nil
 	})
+
+	RegisterMetricProducer("prometheus", func(ctx context.Context) (metric.Producer, error) {
```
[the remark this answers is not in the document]
pellared: I support adding Prometheus.

exporters/autoexport/metrics.go
gouthamve: That is a good point. I added a fallbackfactory for that.

gouthamve:
@pellared @dashpole This is now ready for review!

I added the tests and also a `WithFallbackMetricsProducer` option. Thanks in advance for your review!

exporters/autoexport/metrics.go
```diff
@@ -148,6 +195,10 @@ func init() {
...
```
dashpole, line 199: Should we register a "none" producer?  It would be used if someone didn't want to use the fallback producer.

exporters/autoexport/metrics.go
```diff
@@ -170,3 +221,31 @@ func getenv(key, fallback string) string {
...
+	expType := os.Getenv(pr.envKey)
+	if expType == "" {
+		if pr.fallbackProducer != nil {
+			return pr.fallbackProducer(ctx)
+		}
+
+		return nil, nil
+	}
+
+	return pr.registry.load(ctx, expType)
```
dashpole, line 250: This currently loads a single producer from the registry, but the env var is plural: `OTEL_METRICS_PRODUCERS`.  Should we support multiple comma-separated values here?

exporters/autoexport/metrics.go
gouthamve: Good point, done!

exporters/autoexport/metrics.go
```diff
@@ -71,10 +76,37 @@ func RegisterMetricReader(name string, factory func(context.Context) (metric.Rea
...
+		producers, err := metricsProducers.create(ctx)
+		if err != nil {
+			return nil, err
+		}
+		readerOpts := []metric.PeriodicReaderOption{}
+		if len(producers) > 0 {
+			for _, producer := range producers {
+				readerOpts = append(readerOpts, metric.WithProducer(producer))
+			}
+		}
```
dashpole, line 108:
```suggestion
		for _, producer := range producers {
			readerOpts = append(readerOpts, metric.WithProducer(producer))
		}
```
nit: If statement isn't neccessary

exporters/autoexport/metrics.go
```diff
@@ -86,33 +118,58 @@ func init() {
...
```
dashpole, line 142:
```suggestion
		for _, producer := range producers {
			readerOpts = append(readerOpts, metric.WithProducer(producer))
		}
```
Same as above

exporters/autoexport/metrics.go
```diff
@@ -86,33 +118,58 @@ func init() {
...
+
+		producers, err := metricsProducers.create(ctx)
+		if err != nil {
+			return nil, err
+		}
+		if len(producers) > 0 {
+			for _, producer := range producers {
+				exporterOpts = append(exporterOpts, promexporter.WithProducer(producer))
+			}
+		}
```
dashpole, line 170:
```suggestion
		for _, producer := range producers {
			exporterOpts = append(exporterOpts, promexporter.WithProducer(producer))
		}
```
same as above

exporters/autoexport/noop.go
```diff
@@ -46,3 +47,20 @@ func IsNoneMetricReader(e metric.Reader) bool {
...
+	return nil, nil
+}
+
+func newNoopMetricProducer() noopMetricProducer {
+	return noopMetricProducer{}
+}
+
+// IsNoneMetricReader returns true for the exporter returned by [NewMetricReader]
+// when OTEL_METRICS_PRODUCERS environment variable is set to "none".
+func IsNoneMetricProducer(e metric.Producer) bool {
```
dashpole, line 63: I'm not sure this function is needed.  The IsNoneMetricReader is useful for people who want to avoid initializing the SDK.  Since we don't ever return a metric producer to users, they wouldn't every be able to use this.

dashpole (approved):

CHANGELOG.md
```diff
@@ -17,6 +17,9 @@ This project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0.htm
 ### Added
 
 - `NewSDK` in `go.opentelemetry.io/contrib/config` now returns a configured SDK with a valid `MeterProvider`. (#4804)
+- Add an EXPERIMENTAL `OTEL_METRICS_PRODUCERS` environment variable to `go.opentelemetry.io/contrib/autoexport` to be set metrics producers. (#5281)
+  -  `prometheus` and `none` are supported values. You can specify multiple producers separated by a comma.
```
pellared, line 21:
```suggestion
  - `prometheus` and `none` are supported values. You can specify multiple producers separated by a comma.
```

CHANGELOG.md
```diff
@@ -17,6 +17,9 @@ This project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0.htm
 ### Added
 
 - `NewSDK` in `go.opentelemetry.io/contrib/config` now returns a configured SDK with a valid `MeterProvider`. (#4804)
+- Add an experimental `OTEL_METRICS_PRODUCERS` environment variable to `go.opentelemetry.io/contrib/autoexport` to be set metrics producers. (#5281)
+  - `prometheus` and `none` are supported values. You can specify multiple producers separated by a comma.
+  - `WithFallbackMetricProducer(func(context.Context) (metric.Producer, error))` allows adding a fallback option if the env var is not set.
```
pellared, line 22:
```suggestion
  - Add `WithFallbackMetricProducer` option allowing adding a fallback if the `OTEL_METRICS_PRODUCERS` is not set or empty.
```

pellared (approved):

PR #5092, opened 2024-02-14
Add semconv v1.24.0 to `otelhttp`

MadVikingGod:
This change adds semantic convention v1.24.0 to `otlehttp` and the [compatibility mode](https://github.com/open-telemetry/semantic-conventions/blob/v1.21.0/docs/http/README.md) that was required in the v1.21.0 semantic convention.

### Scope
This is the first in a series of patches to fully finish out this feature.  The goal is to show how the environment variable can be supported and how it can be used.

What was excluded to aid in reviewability, all can be added after:
- Templating - Changes need to be made to each instrumentation to opt into the new methods.
- Client Instrumentation - This will follow a similar pattern to the `HTTPServer` interface and supplied by new types.
- Metrics Instrumentation - The Methods are in the interface but aren't used.

### Benchmarks
A benchmark was added to aid in comparing.  I've added the raw data to this [gist](https://gist.github.com/MadVikingGod/7fd77d87c582358e23c305901cd1c4eb).

First I wanted to make sure that this didn't introduce performance regressions to users using the default/old attributes:
```
benchstat bench-orig.txt bench-old.txt 
goos: linux
goarch: amd64
pkg: go.opentelemetry.io/contrib/instrumentation/net/http/otelhttp/internal/semconv
cpu: AMD EPYC 7763 64-Core Processor                
                           │ bench-orig.txt │             bench-old.txt             │
                           │     sec/op     │    sec/op     vs base                 │
HTTPServerRequest-2            478.5n ± 19%   504.8n ± 22%       ~ (p=0.218 n=10)
HTTPServerRequestMetrics-2     279.9n ± 13%
geomean                        366.0n         504.8n        +5.51%                ¹
¹ benchmark set differs from baseline; geomeans may not be comparable

                           │ bench-orig.txt │            bench-old.txt            │
                           │      B/op      │    B/op     vs base                 │
HTTPServerRequest-2              644.0 ± 0%   644.0 ± 0%       ~ (p=1.000 n=10) ¹
HTTPServerRequestMetrics-2       388.0 ± 0%
geomean                          499.9        644.0       +0.00%                ²
¹ all samples are equal
² benchmark set differs from baseline; geomeans may not be comparable

                           │ bench-orig.txt │            bench-old.txt            │
                           │   allocs/op    │ allocs/op   vs base                 │
HTTPServerRequest-2              2.000 ± 0%   2.000 ± 0%       ~ (p=1.000 n=10) ¹
[...]

PR #5291, opened 2024-03-18
otelhttp: Introduce internal wrapper package

dmathieu:
I am looking into cleaning up the otelhttp package, with the end goal of having separate handlers/transports for metrics and traces. (with #2782 in mind).

This is the first PR towards that goal. It moves the `bodyWrapper` and `respWriterWrapper` into a `wrapper` package, with additionnal tests, so the utility structs can be used across a future `transport` and `handler` package.


dashpole:
I think we should consider waiting to do refactoring work until after we have updated to the stable semconv: https://github.com/open-telemetry/opentelemetry-go-contrib/pull/5092

dmathieu:
Closing for now. I'll bring this back once the semconv migration is done.

PR #5292, opened 2024-03-18
Add skeleton of slog bridge

MrAlias:
Part of https://github.com/open-telemetry/opentelemetry-go-contrib/pull/5195


bridges/otelslog/handler.go
```diff
@@ -0,0 +1,93 @@
...
+
+	// TODO: implement.
+}
+
+// Compile-time check *Handler implements slog.Handler.
+var _ slog.Handler = (*Handler)(nil)
+
+// New returns a new [Handler] to be used as an [slog.Handler].
+//
+// If lp is nil, [noop.LoggerProvider] will be used as the default.
```
dashpole, line 57: Some ideas to consider:

* Would it be better to fall back to the global LoggerProvider?
* If so, should we make the LoggerProvider an option and just accept options in New?

dashpole: It would allow users to do `slog.SetDefault(otelslog.New())`

bridges/otelslog/handler.go
```diff
@@ -0,0 +1,93 @@
...
+//
+// By default the returned Handler will use a [log.Logger] that is identified
+// with this bridge package information. [WithInstrumentationScope] should be
+// used to override this with details about the package or module the handler
+// will instrument.
+func New(lp log.LoggerProvider, options ...Option) *Handler {
```
dashpole, line 63: Should this be `NewHandler`?  The package doesn't include handler in the name.  `otelslog.NewHandler()` reads better to me than `otelslog.New()`, since slog contains other interfaces.

MrAlias: Sounds good to me. I think we can have both `New` and `NewHandler`. The `New` function can return an `slog.Logger` instance. Thoughts @dashpole?

dashpole: I'm OK with that.  WDYT about NewLogger vs New?  Standard usage would be:

* `slog.SetDefault(otelslog.New())` vs
* `slog.SetDefault(otelslog.NewLogger())`

MrAlias: Update. PTAL @dashpole

dashpole (approved):
Looks great

MrAlias:
@dashpole do you think exporting the `Handler` type is needed? Should we only export the `New` func that returns an `slog.Logger`?

dashpole:
I suspect it will still be needed, as wrapping handlers is how users commonly make global modifications to their logs (e.g. stick things in context, and extract using the logger).  If we aren't sure, i'm happy to keep it non-exported for now.

MrAlias:
> (e.g. stick things in context, and extract using the logger)

Like they may want to wrap our `Handler` and have `Handle` unpack things from a context then call our `Handler`?

dashpole:
> Like they may want to wrap our Handler and have Handle unpack things from a context then call our Handler?

Yes.

pellared (approved):

Issue #1341, opened 2021-10-11
Runtime instrumentation: Make runtime.uptime an Asynchronous Gauge


cwildman:
Currently the `runtime.uptime` metric is reported as an [Asynchronous Counter](https://github.com/open-telemetry/opentelemetry-go-contrib/blob/8e44f3f02293665b0547280bf151da43d9a8fb2a/instrumentation/runtime/runtime.go#L120-L121). Uptime as a concept is certainly monotonic which makes it seem like it should be a Counter, but it's also non-additive, it doesn't make sense to sum the uptime of different processes.

Additionally reporting uptime as a Counter presents a problem when trying to use delta temporality. The obvious question you want to answer with a metric like this is "How long has my process been running?". Using delta temporality with uptime would force you to track the running sum of all data points since the process started.

Because runtime.uptime is not additive and we don't actually care about rates or sum I think it is best served as an Asynchronous Gauge.

krak3n:
In the prometheus exporter in [open-telemetry/opentelemetry-collector-contrib](https://github.com/open-telemetry/opentelemetry-collector-contrib/tree/main/exporter/prometheusexporter) this metric fails to get exported due to this error:

```
{
   "level":"error",
   "ts":1659451541.6164072,
   "caller":"prometheusexporter@v0.53.0/log.go:34",
   "msg":"error gathering metrics: collected metric runtime_uptime counter:<value:8.415e+07 >  should be a Gauge\n",
   "kind":"exporter",
   "name":"prometheus",
   "stacktrace":"github.com/open-telemetry/opentelemetry-collector-contrib/exporter/prometheusexporter.(*promLogger).Println\n\tgithub.com/open-telemetry/opentelemetry-collector-contrib/exporter/prometheusexporter@v0.53.0/log.go:34\ngithub.com/prometheus/client_golang/prometheus/promhttp.HandlerFor.func1\n\tgithub.com/prometheus/client_golang@v1.12.2/prometheus/promhttp/http.go:129\nnet/http.HandlerFunc.ServeHTTP\n\tnet/http/server.go:2084\nnet/http.(*ServeMux).ServeHTTP\n\tnet/http/server.go:2462\nnet/http.serverHandler.ServeHTTP\n\tnet/http/server.go:2916\nnet/http.(*conn).serve\n\tnet/http/server.go:1966"
}
```

starzhanganz:
[...]

PR #5293, opened 2024-03-18
Rename  "runtime.uptime" metric from instrumentation/runtime package

AkhigbeEromo:
Renamed `runtime.uptime` in `/instrumentation/runtime/runtime.go` to `process.runtime.uptime` (#2625)

PR #5294, opened 2024-03-19
Make runtime.uptime an asynchronous gauge

AkhigbeEromo:
Changed `runtime.uptime` in `instrumentation/runtime/runtime.go` from an asynchronous counter to an asynchronous gauge (#1341)


dmathieu:
Both this PR and #5293 introduce breaking changes, bu renaming the metric (which means users have to change their dashboards) and changing its type (which can cause display weirdness).
So how about, instead of the current approach to rename and change the type, introduce a new metric with the new type/name, and mark the old metric as deprecated.
Then, after the next release, remove the old metric.

AkhigbeEromo:
Understood, will work on that now @dmathieu

PR #5314, opened 2024-03-19
Implement the slog Handler functionality

MrAlias:


bridges/otelslog/handler_test.go
```diff
@@ -30,6 +31,9 @@ type recorder struct {
 	embedded.LoggerProvider
 	embeddedLogger // nolint:unused  // Used to embed embedded.Logger.
 
+	// Records are the records emited.
```
MrAlias, line 34:
```suggestion
	// Records are the records emitted.
```

bridges/otelslog/handler.go
```diff
@@ -124,10 +129,46 @@ func NewHandler(options ...Option) *Handler {
...
 // Handle handles the passed record.
 func (h *Handler) Handle(ctx context.Context, record slog.Record) error {
-	// TODO: implement.
+	h.logger.Emit(ctx, h.convertRecord(record))
 	return nil
 }
 
+func (h *Handler) convertRecord(r slog.Record) log.Record {
+	var record log.Record
+	record.SetTimestamp(r.Time)
```
XSAM, line 138: From https://opentelemetry.io/docs/specs/otel/logs/data-model/#field-observedtimestamp

> This field SHOULD be set once the event is observed by OpenTelemetry.

Could we set the `ObservedTimestamp` to `time.Now`?

bridges/otelslog/handler.go
```diff
@@ -140,13 +181,258 @@ func (h *Handler) Enabled(ctx context.Context, l slog.Level) bool {
...
+
+// AddAttrs adds attrs to b.
+func (b *kvBuffer) AddAttrs(attrs []slog.Attr) {
+	b.data = slices.Grow(b.data, len(attrs))
+	for _, a := range attrs {
+		_ = b.AddAttr(a)
+	}
+}
+
+// AddAttrs adds attr to b and returns true.
```
XSAM, line 370:
```suggestion
// AddAttr adds attr to b and returns true.
```

bridges/otelslog/handler.go
MrAlias: Yeah, it's not clear to me what constitutes "OpenTelemetry". It's it the bridge, bridge API, or SDK?

I'm happy to add it here if there's no objections.

XSAM: This is a good question.

> For events that originate in OpenTelemetry (e.g. using OpenTelemetry Logging SDK) this timestamp is typically set at the generation time and is equal to Timestamp. For events originating externally and collected by OpenTelemetry (e.g. using Collector) this is the time when OpenTelemetry’s code observed the event measured by the clock of the OpenTelemetry code.

I feel like OpenTelemetry here refers to the API and SDK (whatever is in the https://github.com/open-telemetry/opentelemetry-go)

https://opentelemetry.io/docs/specs/otel/logs/#direct-to-collector

This PR looks like a part of the `logging library` to me.

---

I think we can keep the current implementation.

XSAM (approved):

bridges/otelslog/handler.go
```diff
@@ -140,13 +181,258 @@ func (h *Handler) Enabled(ctx context.Context, l slog.Level) bool {
...
+		return newKVBuffer(5)
+	},
+}
+
+func getKVBuffer() (buf *kvBuffer, free func()) {
+	buf = kvBufferPool.Get().(*kvBuffer)
+	return buf, func() {
+		// TODO: limit returned size so the pool doesn't hold on to very large
+		// buffers. Idea is based on
+		// https://cs.opensource.google/go/x/exp/+/814bf88c:slog/internal/buffer/buffer.go;l=27-34
```
pellared, line 322: Should we create an issue for it for tracking?
Alternatively, adding a TODO item in https://github.com/open-telemetry/opentelemetry-go-contrib/pull/5195 would be also good enough.

bridges/otelslog/handler.go
```diff
@@ -140,13 +181,258 @@ func (h *Handler) Enabled(ctx context.Context, l slog.Level) bool {
...
+		return convertValue(v.Resolve())
+	default:
+		// Try to handle this as gracefully as possible.
+		//
+		// Don't panic here. The goal here is to have developers find this
+		// first if a new slog.Kind is added. A test on the new kind will find
+		// this malformed attribute as well as a panic. However, it is
+		// preferable to have user's open issue asking why their attributes
+		// have a "unhandled: " prefix than say that their code is panicking.
+		return log.StringValue(fmt.Sprintf("unhandled: %+v", v.Any()))
```
pellared, line 436: How about also "logging" `v.Kind()`?

bridges/otelslog/handler.go
```diff
@@ -140,13 +181,258 @@ func (h *Handler) Enabled(ctx context.Context, l slog.Level) bool {
...
+		}
+	}
+	b.data = append(b.data, log.KeyValue{
+		Key:   attr.Key,
+		Value: convertValue(attr.Value),
+	})
+	return true
+}
+
+func convertValue(v slog.Value) log.Value {
```
pellared, line 403: Should we document (in Go doc so it is clear for the users) how the bridge slog types are converted to OTel log records?
Especially how `slog.KindAny`, `slog.KindDuration`, `slog.KindTime` are converted.

This can be addressed in a separate PR.

pellared (approved):

bridges/otelslog/handler.go
MrAlias: There is already an item in the "Outstanding tasks" of #5195

bridges/otelslog/handler.go
```diff
@@ -4,6 +4,43 @@
...
+// Severity types. For example:
+//
+//   - [slog.LevelDebug] is transformed to [log.SeverityDebug]
+//   - [slog.LevelInfo] is transformed to [log.SeverityInfo]
+//   - [slog.LevelWarn] is transformed to [log.SeverityWarn]
+//   - [slog.LevelError] is transformed to [log.SeverityError]
+//
+// Attribute values are transformed based on their [slog.Kind]:
+//
+//   - [slog.KindAny] are transformed to [log.StringValue]. The value will be
```
pellared, line 29:
```suggestion
//   - [slog.KindAny] are transformed to [log.StringValue]. The value is
```

pellared (approved):

bridges/otelslog/handler.go
```diff
@@ -140,13 +218,258 @@ func (h *Handler) Enabled(ctx context.Context, l slog.Level) bool {
...
+		g.attrs = newKVBuffer(len(attrs))
+	}
+	g.attrs.AddAttrs(attrs)
+}
+
+var kvBufferPool = sync.Pool{
+	New: func() any {
+		// Based on slog research (https://go.dev/blog/slog#performance), 95%
+		// of use-cases will use 5 or less attributes.
+		return newKVBuffer(5)
```
jba, line 350: I used 5 for the array inside a `Record` because there was a tradeoff: using too much space would hurt performance. There's no similar tradeoff here, so you could make this a bit larger. Maybe 10?

bridges/otelslog/handler.go
```diff
@@ -140,13 +218,258 @@ func (h *Handler) Enabled(ctx context.Context, l slog.Level) bool {
...
+	}
+	return len(b.data)
+}
+
+// Clone returns a copy of b.
+func (b *kvBuffer) Clone() *kvBuffer {
+	if b == nil {
+		return nil
+	}
+	return &kvBuffer{data: slices.Clone(b.data)}
```
jba, line 388: You lose any extra capacity when you clone b.data, so the subsequent AddAttrs will always allocate. (I think.)

In general though, it's probably best not to worry too much about these details. You can always tweak the performance later.

PR #5317, opened 2024-03-19
Create a new metric `process.runtime.uptime` and deprecate `runtime.uptime` metric in `instrumentation/runtime/runtime.go` file

AkhigbeEromo:
- Added a new metric called `process.runtime.uptime` and made it an asynchronous gauge function.
- Deprecated the `runtime.uptime` metric
- Link (#5293, #5294)


instrumentation/runtime/runtime.go
```diff
@@ -108,6 +108,7 @@ func Start(opts ...Option) error {
 
 func (r *runtime) register() error {
 	startTime := time.Now()
+	// DEPRECATED: This metric is deprecated in favor of processUptime
```
dmathieu, line 111: I don't think this is enough. AFAIK, it won't show in godoc, as `// DEPRECATED` is for methods and structs.
I'm not finding any precedent where a metric was deprecated.
But how about specifying it in the metric description?

AkhigbeEromo: Just edited the metric description to indicate the metric is deprecated. Any other corrections @dmathieu

instrumentation/runtime/runtime.go
```diff
@@ -111,6 +111,15 @@ func (r *runtime) register() error {
 	uptime, err := r.meter.Int64ObservableCounter(
 		"runtime.uptime",
 		metric.WithUnit("ms"),
+		metric.WithDescription("This metric is deprecated in favor of processUptime"),
```
dmathieu, line 114: Even though this is deprecated, the description should still mention what the metric does.

```suggestion
		metric.WithDescription("Milliseconds since application was initialized. Deprecated metric. Use `process.runtime.runtime` instead."),
```

instrumentation/runtime/runtime.go
```diff
@@ -111,6 +111,15 @@ func (r *runtime) register() error {
 	uptime, err := r.meter.Int64ObservableCounter(
 		"runtime.uptime",
 		metric.WithUnit("ms"),
+		metric.WithDescription("Milliseconds since application was initialized. Deprecated metric. Use `process.runtime.runtime` instead."),
```
dmathieu, line 114: @open-telemetry/go-approvers what do you all think about this way of deprecating/renaming a metric?
It does mean for one release we emit two metrics.

AkhigbeEromo: Is there anything you would like me to change  @dmathieu

AkhigbeEromo:
Hello @dmathieu, any update on this PR?

Or is there anything you would like me to change?

instrumentation/runtime/runtime.go
dashpole: There is a typo in the replacement metric name here (`process.runtime.runtime` instead of `process.runtime.uptime`), though note that #5780 already deprecated `runtime.uptime` in v1.29.0 and closed #1341 and #2625 after `open-telemetry/semantic-conventions#981` removed `runtime.uptime`, so we can probably close this PR.

Issue #5331, opened 2024-03-28
Make `otelhttp` produce old, new and duplicate Server Trace Attributes


MadVikingGod:
Make `instrumentation/net/http/otelhttp` produce Server Traces following the migration plan #5132.

A user should be able to produce Traces from the `handler` in current semconv (default), the latest semconv (1.24.0), and a duplicate form based on the Environment variable `OTEL_HTTP_CLIENT_COMPATIBILITY_MODE`

- [x] [Produce Old Server Attributes](https://github.com/open-telemetry/opentelemetry-go-contrib/pull/5333)
- [x] ~[Produce Duplicate Server Attributes](https://github.com/open-telemetry/opentelemetry-go-contrib/pull/5400)~ Will not do to simplify rollout
- [x] [Produce New Server Attributes](https://github.com/open-telemetry/opentelemetry-go-contrib/pull/5401)
- [x] Create a changelog for Server Attributes availability.

MrAlias:
For changelog entry: New environment variable that will produce both old and new needs to be added to the changelog.

MrAlias:
Done

PR #5333, opened 2024-03-29
HTTP Semconv migration Part1 Server - v1.20.0 support

MadVikingGod:
This change adds an absctraction layer to the semconv of otlehttp.  This is to enable [compatibility mode](https://github.com/open-telemetry/semantic-conventions/blob/v1.21.0/docs/http/README.md) that was required in the v1.21.0 semantic convention.

The full PR is #5092 
Part of #5331


CHANGELOG.md
```diff
@@ -11,6 +11,7 @@ This project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0.htm
 ### Added
 
 - Add the new `go.opentelemetry.io/contrib/instrgen` package to provide auto-generated source code instrumentation. (#3068, #3108)
+- Add support additional semantic conventions `go.opentelemetry.io/contrib/instrumentation/net/http/otelhttp`. (#????)
```
MrAlias, line 14: Is a changelog entry needed? It seems like this is just moving code around. From the user perspective nothing changed, right?

instrumentation/net/http/otelhttp/doc.go
```diff
@@ -4,4 +4,9 @@
 // Package otelhttp provides an http.Handler and functions that are intended
 // to be used to add tracing by wrapping existing handlers (with Handler) and
 // routes WithRouteTag.
+//
+// Warning: migration of semantic conventions to v1.24.0 is in progress. Because
+// this will break most existing dashboards we have developed a migration plan
+// detailed [here](). Use the environment variable `OTEL_HTTP_CLIENT_COMPATIBILITY_MODE`
+// to opt into the new conventions. This will be removed in a future release.
```
MrAlias, line 11:
```suggestion
```
The migration is not happening in this PR.

instrumentation/net/http/otelhttp/internal/semconv/v1.20.0.go
```diff
@@ -0,0 +1,86 @@
...
+// (http://nginx.org/en/docs/http/ngx_http_core_module.html#server_name) for an
+// nginx server. More generically, the primary server name would be the host
+// header value that matches the default virtual host of an HTTP server. It
+// should include the host identifier and if a port is used to route to the
+// server that port identifier should be included as an appropriate port
+// suffix.
+//
+// If the primary server name is not known, server should be an empty string.
+// The req Host will be used to determine the server instead.
+func (o oldHTTPServer) TraceRequest(server string, req *http.Request) []attribute.KeyValue {
```
MrAlias, line 46: `TraceRequest` implies that this is going to trace the `req`.

```suggestion
func (o oldHTTPServer) RequestAttrs(server string, req *http.Request) []attribute.KeyValue {
```

Or,

```suggestion
func (o oldHTTPServer) RequestTraceAttrs(server string, req *http.Request) []attribute.KeyValue {
```

Or, given all these just return attributes so that is implied:

```suggestion
func (o oldHTTPServer) Request(server string, req *http.Request) []attribute.KeyValue {
```

instrumentation/net/http/otelhttp/internal/semconv/v1.20.0.go
```diff
@@ -0,0 +1,86 @@
...
+	return semconvutil.HTTPServerRequest(server, req)
+}
+
+// TraceRequest returns trace attributes for telemetry from an HTTP response.
+//
+// If any of the fields in the ResponseTelemetry are not set the attribute will be omitted.
+func (o oldHTTPServer) TraceResponse(resp ResponseTelemetry) []attribute.KeyValue {
```
MrAlias, line 53:
```suggestion
func (o oldHTTPServer) ResponseAttrs(resp ResponseTelemetry) []attribute.KeyValue {
```

instrumentation/net/http/otelhttp/internal/semconv/v1.20.0_test.go
```diff
@@ -0,0 +1,56 @@
...
+package semconv
+
+import (
+	"testing"
+
+	"go.opentelemetry.io/otel/attribute"
+)
+
+func TestV120TraceRequest(t *testing.T) {
+	// Anything but "http" or "http/dup" works
```
MrAlias, line 24:
```suggestion
	// Anything but "http" or "http/dup" works.
```

instrumentation/net/http/otelhttp/internal/semconv/env.go
```diff
@@ -0,0 +1,97 @@
...
+	"go.opentelemetry.io/otel"
+	"go.opentelemetry.io/otel/attribute"
+	"go.opentelemetry.io/otel/codes"
+)
+
+type ResponseTelemetry struct {
+	StatusCode int
+	ReadBytes  int
+	ReadError  error
+	WriteBytes int
```
MrAlias, line 37:
```suggestion
	ReadBytes  int64
	ReadError  error
	WriteBytes int64
```

The passed values are always converted. Instead, accept `int64` here given it will be convert in the attribute anyway.

instrumentation/net/http/otelhttp/internal/semconvutil/httpconv_benchmark_test.go
MrAlias, line 1: Why are benchmarks for another package being added in this PR?

Can this be removed?

instrumentation/net/http/otelhttp/internal/semconv/v1.20.0_test.go
```diff
@@ -0,0 +1,56 @@
...
+func TestV120TraceResponse(t *testing.T) {
+	// Anything but "http" or "http/dup" works
+	t.Setenv("OTEL_HTTP_CLIENT_COMPATIBILITY_MODE", "old")
+	serv := NewHTTPServer()
+	want := []attribute.KeyValue{
+		attribute.Int("http.request_content_length", 701),
+		attribute.String("http.read_error", "read error"),
+		attribute.Int("http.response_content_length", 802),
+		attribute.String("http.write_error", "write error"),
+		attribute.Int("http.status_code", 200),
```
MrAlias, line 53: These all seem like magic values. Why not structure this similar to the request test where they are passed from the testing harness?

instrumentation/net/http/otelhttp/internal/semconv/util.go
```diff
@@ -0,0 +1,86 @@
...
+		xForwardedFor = xForwardedFor[:idx]
+	}
+	return xForwardedFor
+}
+
+func netProtocol(proto string) (name string, version string) {
+	name, version, _ = strings.Cut(proto, "/")
+	name = strings.ToLower(name)
+	return name, version
+}
```
MrAlias, line 86: These functions are unused.

```suggestion
```

MrAlias:
Please address the lint failures.

instrumentation/net/http/otelhttp/internal/semconv/v1.20.0.go
MadVikingGod: I have gone with `RequestTraceAttrs` Because we will need to be able to differentiate Trace Attributes from Metric Attributes.

instrumentation/net/http/otelhttp/internal/semconv/v1.20.0.go
MadVikingGod: I have gone with ResponseTraceAttrs Because we will need to be able to differentiate Trace Attributes from Metric Attributes.

instrumentation/net/http/otelhttp/internal/semconvutil/httpconv_benchmark_test.go
MadVikingGod: This is to ensure the interface approach does not add any extra overhead.  This is directly comparable to the benchmark that is in the semconv package

instrumentation/net/http/otelhttp/internal/semconv/v1.20.0_test.go
MadVikingGod: Because only the status code is contained within a response, the other fields are created from the package that use `semconv`.  This is to test that this package produces the Correct attributes from that data.

So all the data is made up by the test harness.

CHANGELOG.md
MadVikingGod: Removed.

instrumentation/net/http/otelhttp/doc.go
MadVikingGod: Removed.

instrumentation/net/http/otelhttp/internal/semconvutil/httpconv_benchmark_test.go
MrAlias: Okay. I'm still not following why they are added here.

- This is adding code that says it is generate, it is not generated. This is going to cause confusion or worse conflicts going forward
- Nothing added here couldn't be added in its own PR and reference from this PRs description or comments.
- All of the code being added to semconv is being copied from semconvutil or otelhttp directly. It doesn't seem like benchmarks are needed given the direct copy without a change to implementation.

Do we need to add these at all? If so, can we do it in a separate PR?

instrumentation/net/http/otelhttp/internal/semconv/v1.20.0_test.go
MrAlias: > So all the data is made up by the test harness.

Right, but by magically referencing that data from this test it means any changes there will break here.

The `testTraceRequest` function takes a `func(req testServerReq) []attribute.KeyValue`. Why not have `testTraceResponse` accept a `func(req testServerResp) []attribute.KeyValue` argument and have things like the request/response length and status passed to it?

instrumentation/net/http/otelhttp/internal/semconv/common_test.go
```diff
@@ -0,0 +1,93 @@
...
+	peerPort   int
+	clientIP   string
+}
+
+func testTraceRequest(t *testing.T, serv HTTPServer, want func(testServerReq) []attribute.KeyValue) {
+	t.Helper()
+
+	got := make(chan *http.Request, 1)
+	handler := func(w http.ResponseWriter, r *http.Request) {
+		got <- r
```
MrAlias, line 44: This only expects one request right? Closing will ensure this.

```suggestion
		got <- r
		close(got)
```

instrumentation/net/http/otelhttp/internal/semconv/v1.20.0_test.go
MrAlias, line 24:
```suggestion
	// Anything but "http" or "http/dup" works.
```

instrumentation/net/http/otelhttp/internal/semconv/util.go
```diff
@@ -0,0 +1,60 @@
...
+)
+
+// splitHostPort splits a network address hostport of the form "host",
+// "host%zone", "[host]", "[host%zone], "host:port", "host%zone:port",
+// "[host]:port", "[host%zone]:port", or ":port" into host or host%zone and
+// port.
+//
+// An empty host is returned if it is not provided or unparsable. A negative
+// port is returned if it is not provided or unparsable.
+func splitHostPort(hostport string) (host string, port int) {
```
dashpole, line 30: Is this only used for unit tests?  If so, it could be in a _test.go file.

instrumentation/net/http/otelhttp/internal/semconv/env.go
```diff
@@ -0,0 +1,92 @@
...
+	ResponseTraceAttrs(ResponseTelemetry) []attribute.KeyValue
+
+	// Route returns the attribute for the route.
+	Route(string) attribute.KeyValue
+}
+
+// var warnOnce = sync.Once{}
+
+func NewHTTPServer() HTTPServer {
+	// env := strings.ToLower(os.Getenv("OTEL_HTTP_CLIENT_COMPATIBILITY_MODE"))
```
dashpole, line 67: I would prefer removing commented out code.

instrumentation/net/http/otelhttp/internal/semconv/httpconv_benchmark_test.go
```diff
@@ -0,0 +1,58 @@
+// Code created by gotmpl. DO NOT MODIFY.
+// source: internal/shared/semconvutil/httpconv_test.go.tmpl
```
dashpole, line 2: This isn't templatised yet, right?

instrumentation/net/http/otelhttp/internal/semconv/util.go
MadVikingGod: This is used by the [duplicate semconv](https://github.com/MadVikingGod/opentelemetry-go-contrib/blob/30adb7e838d73484398ef40dfc910b6ef690f5fa/instrumentation/net/http/otelhttp/internal/semconv/dup.go#L57) and [1.24.0](https://github.com/MadVikingGod/opentelemetry-go-contrib/blob/30adb7e838d73484398ef40dfc910b6ef690f5fa/instrumentation/net/http/otelhttp/internal/semconv/v1.24.0.go#L52).  I did remove the test accidentally, I will add that back.

instrumentation/net/http/otelhttp/internal/semconvutil/httpconv_benchmark_test.go
pellared: The benchmark has been removed.

instrumentation/net/http/otelhttp/internal/semconv/httpconv_benchmark_test.go
```diff
@@ -0,0 +1,45 @@
...
+	"testing"
+
+	"go.opentelemetry.io/otel/attribute"
+)
+
+var benchHTTPServerRequestResults []attribute.KeyValue
+
+// BenchmarkHTTPServerRequest allows comparison between different version of the HTTP server.
+// To use an alternative start this test with OTEL_HTTP_CLIENT_COMPATIBILITY_MODE set to the
+// version under test.
```
pellared, line 18: Cannot we simply have a table-driven benchmark for different values of `OTEL_HTTP_CLIENT_COMPATIBILITY_MODE`?

instrumentation/net/http/otelhttp/internal/semconv/httpconv_benchmark_test.go
pellared, line 1: nit: Can we rename to `bench_test.go`.

pellared (commented):
_I haven't reviewed the code. I just focused on resolving some of the addressed comments._

dashpole (approved):

instrumentation/net/http/otelhttp/internal/semconv/httpconv_benchmark_test.go
MadVikingGod: We can.  The reason not to is when you want to compare one version to a different version the benchmark needs to have the same name for `benchstat` to compare them.  

This is why I put the same benchmark in the `semconvutil`.

instrumentation/net/http/otelhttp/internal/semconvutil/httpconv_benchmark_test.go
MadVikingGod: This one was removed.  The one in internal/semconv remains. 

I will just have to add it back if asked for any data to show that passing the semconvutil through an interface doesn't add any extra allocations or significant processing

MrAlias (approved):

dmathieu (approved):

pellared:
Side note, please do not wait on my review.

instrumentation/net/http/otelhttp/internal/semconv/env.go
```diff
@@ -0,0 +1,69 @@
...
+func NewHTTPServer() HTTPServer {
+	// TODO (#5331): Detect version based on environment variable OTEL_HTTP_CLIENT_COMPATIBILITY_MODE.
+	// TODO (#5331): Add warning of use of a deprecated version of Semantic Versions.
+	return oldHTTPServer{}
+}
+
+// ServerStatus returns a span status code and message for an HTTP status code
+// value returned by a server. Status codes in the 400-499 range are not
+// returned as errors.
+func ServerStatus(code int) (codes.Code, string) {
```
krim, line 61: @MadVikingGod Could you explain why we cannot use codes.Ok for statuses in the range of 200..400?

Issue #5186, opened 2024-02-26
Document slog log bridge


pellared:
Document how to use the slog bridge it in opentelemetry.io.

Blocked by:
- https://github.com/open-telemetry/opentelemetry-go-contrib/issues/5138

PR #5355, opened 2024-04-03
otelslog: Add example

pellared:
Towards https://github.com/open-telemetry/opentelemetry-go-contrib/issues/5186

### DIFF
--- a/bridges/otelslog/exampe_test.go
+++ b/bridges/otelslog/exampe_test.go
@@ -1,0 +1,19 @@
+// Copyright The OpenTelemetry Authors
+// SPDX-License-Identifier: Apache-2.0
+
+package otelslog_test
+
+import (
+	"log/slog"
+
+	"go.opentelemetry.io/contrib/bridges/otelslog"
+	"go.opentelemetry.io/otel/log/noop"
+)
+
+func Example() {
+	// Use a working LoggerProvider implementation instead e.g. using go.opentelemetry.io/otel/sdk/log.
+	provider := noop.NewLoggerProvider()
+
+	// Create an *slog.Logger with *otelslog.Handler and use it in your application.
+	slog.New(otelslog.NewHandler(otelslog.WithLoggerProvider(provider)))
+}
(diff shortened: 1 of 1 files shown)


dmathieu (approved):

bridges/otelslog/exampe_test.go
```diff
@@ -0,0 +1,19 @@
...
+	"go.opentelemetry.io/contrib/bridges/otelslog"
+	"go.opentelemetry.io/otel/log/noop"
+)
+
+func Example() {
+	// Use a working LoggerProvider implementation instead e.g. using go.opentelemetry.io/otel/sdk/log.
+	provider := noop.NewLoggerProvider()
+
+	// Create an *slog.Logger with *otelslog.Handler and use it in your application.
+	slog.New(otelslog.NewHandler(otelslog.WithLoggerProvider(provider)))
```
dashpole, line 18: Should we use otelslog.NewLogger instead to simplify it?
```suggestion
	otelslog.NewLogger(otelslog.WithLoggerProvider(provider))
```

pellared: I propose `otelslog.NewHandler` as most (all?) packages I am aware have only `Handler` factories. The reason is that one may want to use it together with some handler decorator e.g. https://github.com/samber/slog-multi. The `slog` package has `NewJSONHandler` and `NewTextHandler`.  See also: https://go.dev/wiki/Resources-for-slog

I even propose removing `otelslog.NewLogger` for sake of consistency and in order to remove redundant functionality.

@MrAlias WDYT?

dashpole: See https://github.com/open-telemetry/opentelemetry-go-contrib/pull/5292 for the previous discussion.  I agree with the need for NewHandler for decorators.  I like the simplicity of being able to do `slog.SetDefault(otelslog.NewLogger())`, but I don't think NewLogger is strictly necessary.

MrAlias: I think @dashpole is correct, the simplicity of `slog.SetDefault(otelslog.NewLogger())` is likely what most users are going to want.

pellared: Personally, I lean towards smaller API surface. But it is just a preference.

pellared: Applying suggestion and resolving

dashpole:
Should the example show how to call SetDefault?

pellared:
> Should the example show how to call slog.SetDefault?

I did not want to do that as many people find global logger as an anti-pattern.

XSAM (approved):

MrAlias (approved):

Issue #5383, opened 2024-04-14
prometheusbridge: Creates invalid exemplars for Prometheus counters


gouthamve:
### Description

I was testing https://github.com/open-telemetry/opentelemetry-go-contrib/pull/5281 and I found errors in Mimir which looked like the following:

```
ts=2024-04-14T14:27:50.976027765Z caller=push.go:171 level=error user=12690 msg="push error" err="received an exemplar with no valid labels, timestamp: 0 series: blackbox_module_unknown_total{job=\"unknown_service:blackbox_exporter\"} labels: {} (err-mimir-exemplar-labels-missing)"
 ts=2024-04-14T14:27:30.96869299Z caller=push.go:171 level=error user=12690 msg="push error" err="received an exemplar with no valid labels, timestamp: 0 series: go_memstats_mallocs_total{job=\"unknown_service:blackbox_exporter\"} labels: {} (err-mimir-exemplar-labels-missing)"
```

Digging into the errors showed me that the Prometheus bridge is actually adding `0` exemplars even when none existed before. This is only happening with counters and not histograms from what I could see.

### Environment

- OS: MacOS
- Architecture: arm64
- Go Version: go version go1.22.2 darwin/arm64
- `go.opentelemetry.io/contrib` version: go.opentelemetry.io/contrib/bridges/prometheus v0.50.0

### Steps To Reproduce

A minimal reproduction is below:

```go
package main

import (
	"time"

	"github.com/prometheus/client_golang/prometheus"
	prometheusbridge "go.opentelemetry.io/contrib/bridges/prometheus"
	"go.opentelemetry.io/otel/exporters/stdout/stdoutmetric"
	"go.opentelemetry.io/otel/sdk/metric"
)

func main() {
	reg := prometheus.NewRegistry()
	someDummyMetric := prometheus.NewCounter(prometheus.CounterOpts{
		Name: "dummy_metric",
		Help: "dummy metric",
	})
	reg.MustRegister(someDummyMetric)

	someDummyMetric.Inc()
	someDummyMetric.Add(2.0)
	someDummyMetric.Add(3.0)

	producer := prometheusbridge.NewMetricProducer(prometheusbridge.WithGatherer(reg))
	exporter, err := stdoutmetric.New()
	checkErr(err)

	reader := metric.NewPeriodicReader(exporter, metric.WithProducer(producer), metric.WithInterval(1*time.Second))
[...]

PR #5395, opened 2024-04-15
Fix bug where an empty exemplar was added to counters

dashpole:
Fixes https://github.com/open-telemetry/opentelemetry-go-contrib/issues/5383

I also found out that the +Inf bucket is only _sometimes_ added to the proto format: https://github.com/prometheus/client_golang/blob/d038ab96c0c7b9cd217a39072febd610bcdf1fd8/prometheus/metric.go#L189

That meant we were dropping the highest histogram bucket if users were not using exemplars.

@gouthamve

### DIFF
--- a/bridges/prometheus/producer_test.go
+++ b/bridges/prometheus/producer_test.go
@@ -218,7 +218,7 @@ func TestProduce(t *testing.T) {
 								{
 									Count:        1,
 									Sum:          578.3,
-									Bounds:       prometheus.DefBuckets,
+									Bounds:       []float64{0.005, 0.01, 0.025, 0.05, 0.1, 0.25, 0.5, 1, 2.5, 5, 10},
 									BucketCounts: []uint64{0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 1},
 									Attributes:   attribute.NewSet(attribute.String("foo", "bar")),
 									Exemplars:    []metricdata.Exemplar[float64]{},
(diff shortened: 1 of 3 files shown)


bridges/prometheus/producer_test.go
```diff
@@ -181,15 +254,15 @@ func TestProduce(t *testing.T) {
...
+						Name:        "test_histogram_metric_with_exemplar",
 						Description: "A histogram metric for testing",
 						Data: metricdata.Histogram[float64]{
 							Temporality: metricdata.CumulativeTemporality,
 							DataPoints: []metricdata.HistogramDataPoint[float64]{
 								{
 									Count:        1,
 									Sum:          578.3,
-									Bounds:       []float64{0.005, 0.01, 0.025, 0.05, 0.1, 0.25, 0.5, 1, 2.5, 5, 10},
+									Bounds:       prometheus.DefBuckets,
```
MrAlias, line 265: Doesn't `prometheus.DefBuckets` equal `[]float64{0.005, 0.01, 0.025, 0.05, 0.1, 0.25, 0.5, 1, 2.5, 5, 10}`? Changing this seems to obfuscate the expected value.

dashpole: reverted.

MrAlias (approved):

dmathieu (approved):

gouthamve (approved):
LGTM

pellared (approved):

PR #5399, opened 2024-04-16
Add a host resource detector

### LINKED ISSUE #5398: Add a resource detector for physical hosts
### Problem Statement

Currently, there is no possibility to obtain useful host-related attributes (`host.id`, or `host.name`) in non-cloud scenarios.

### Proposed Solution

A resource detector for physical hosts should be provided which sets values according to [semantic conventions for host resources](https://github.com/open-telemetry/semantic-conventions/blob/main/docs/resource/host.md).

#### Alternatives

Setting relevant attributes manually without resource detectors.

pyohannes:
Fixes #5398 

This PR adds a resource detector for physical hosts which sets values according to [semantic conventions for host resources](https://github.com/open-telemetry/semantic-conventions/blob/main/docs/resource/host.md):

* `host.arch`
* `host.id`
* `host.name`
* `host.ip`
* `host.mac`


MrAlias:
This seems duplicative of [`WithHost`](https://pkg.go.dev/go.opentelemetry.io/otel/sdk@v1.25.0/resource#WithHost). Why not update that option?

pyohannes:
> This seems duplicative of [WithHost](https://pkg.go.dev/go.opentelemetry.io/otel/sdk@v1.25.0/resource#WithHost). Why not update that option?

Indeed it is duplicative.

@MrAlias Do you think it a feasible option to populate `host.arch` via `WithHost`, and add additional `WithHostIP` and `WithHostMAC` to opt into `host.ip` and `host.mac` (all implemented in the SDK)?

MrAlias:
Taking a look at the host semantic conventions, it looks like they are all experimental (even `host.name` :grimacing:).

I think in the long term we want to make the `WithHost` option include all the recommended/required stable semantic conventions for a host. But in the interim, having a detector here is probably the correct approach.

I think this PR is the correct approach. I'll plan to review.

.github/dependabot.yml
MrAlias, line 1: Please also update the `versions.yaml` file. This needs to be released as an experimental package. It will likely need its own modset if there isn't already one for experimental detectors.

detectors/host/host.go
```diff
@@ -0,0 +1,137 @@
...
+func newConfig(options ...Option) *config {
+	c := &config{}
+	for _, option := range options {
+		option.apply(c)
+	}
+
+	return c
+}
+
+// Option applies an Azure VM detector configuration option.
```
MrAlias, line 31:
```suggestion
// Option applies a host detector configuration option.
```

detectors/host/host.go
```diff
@@ -0,0 +1,137 @@
...
+}
+
+// WithIPAddresses adds the optional attribute `host.ip`.
+func WithIPAddresses() Option {
+	return optionFunc(func(c *config) {
+		c.optInIPAddresses = true
+	})
+}
+
+// WithMACAddresses adds the optional attribute `host.ip`.
```
MrAlias, line 49:
```suggestion
// WithMACAddresses adds the optional attribute "host.mac".
```

detectors/host/host.go
```diff
@@ -0,0 +1,137 @@
...
+	return optionFunc(func(c *config) {
+		c.optInMACAddresses = true
+	})
+}
+
+type resourceDetector struct {
+	config *config
+}
+
+// NewResourceDetector returns a resource detector that will detect host resources.
```
MrAlias, line 60:
```suggestion
// NewResourceDetector returns a [resource.Detector] that will detect host resources.
```

detectors/host/README.md
```diff
@@ -0,0 +1,35 @@
...
+hostResourceDetector := host.NewResourceDetector(
+	WithIPAddresses(),
+	WithMACAddresses(),
+)
+resource, err := hostResourceDetector.Detect(context.Background())
+```
+
+## Supported attributes
+
+According to [semantic conventions for host resources](https://github.com/open-telemetry/semantic-conventions/blob/main/docs/resource/host.md), each of the following attributes is added if it is available:
```
MrAlias, line 26:
```suggestion
According to [semantic conventions for host resources](https://opentelemetry.io/docs/specs/semconv/resource/host/), each of the following attributes is added if it is available:
```

detectors/host/host.go
```diff
@@ -0,0 +1,137 @@
...
+	for _, option := range options {
+		option.apply(c)
+	}
+
+	return c
+}
+
+// Option applies an Azure VM detector configuration option.
+type Option interface {
+	apply(*config)
```
MrAlias, line 33: An allocation can be avoided here:

```suggestion
	apply(config) config
```

https://github.com/open-telemetry/opentelemetry-go/blob/main/CONTRIBUTING.md#config

detectors/host/host_test.go
```diff
@@ -0,0 +1,73 @@
...
+	"testing"
+
+	"github.com/stretchr/testify/assert"
+
+	"go.opentelemetry.io/otel/attribute"
+	"go.opentelemetry.io/otel/sdk/resource"
+	semconv "go.opentelemetry.io/otel/semconv/v1.24.0"
+)
+
+func Test_Detect(t *testing.T) {
```
MrAlias, line 19:
```suggestion
func TestDetect(t *testing.T) {
```

detectors/host/host_test.go
```diff
@@ -0,0 +1,73 @@
...
+	if err == nil {
+		attributes = append(attributes, semconv.HostID(machineId))
+	}
+
+	expectedResource := resource.NewWithAttributes(semconv.SchemaURL, attributes...)
+
+	assert.Equal(t, expectedResource, hostResource)
+}
+
+func Test_Detect_WithOptIns(t *testing.T) {
```
MrAlias, line 44:
```suggestion
func TestDetectWithOptIns(t *testing.T) {
```

detectors/host/host_test.go
```diff
@@ -0,0 +1,73 @@
...
+	"go.opentelemetry.io/otel/sdk/resource"
+	semconv "go.opentelemetry.io/otel/semconv/v1.24.0"
+)
+
+func Test_Detect(t *testing.T) {
+	detector := NewResourceDetector()
+
+	hostResource, err := detector.Detect(context.Background())
+
+	assert.True(t, err == nil)
```
MrAlias, line 24:
```suggestion
	assert.NoError(t, err)
```

detectors/host/host_test.go
```diff
@@ -0,0 +1,73 @@
...
+
+	hostName, _ := os.Hostname()
```
MrAlias, line 26: This error should not be ignored. The test should be failed.

detectors/host/host_id_windows.go
```diff
@@ -0,0 +1,24 @@
...
+// +build windows
+
+package host // import "go.opentelemetry.io/contrib/detectors/host"
+
+import (
+	"golang.org/x/sys/windows/registry"
+)
+
+func getHostId() (string, error) {
+	key, err := registry.OpenKey(registry.LOCAL_MACHINE, `SOFTWARE\Microsoft\Cryptography`, registry.QUERY_VALUE)
```
MrAlias, line 14:
```suggestion
	key, err := registry.OpenKey(registry.LOCAL_MACHINE, `SOFTWARE\Microsoft\Cryptography`, registry.QUERY_VALUE|registry.WOW64_64KEY)
```

https://github.com/open-telemetry/opentelemetry-go/blob/main/sdk/resource/host_id_windows.go

MrAlias (commented):
- Missing support for BSD: https://github.com/open-telemetry/opentelemetry-go/blob/main/sdk/resource/host_id_bsd.go
- Missing fallback when the host system is not supported for the host ID: https://github.com/open-telemetry/opentelemetry-go/blob/main/sdk/resource/host_id_unsupported.go

.github/dependabot.yml
pyohannes: Done.

detectors/host/host_test.go
pyohannes: I adapted the tests so that the `host.name` is only added to the list of expected attributes when `os.Hostname()` doesn't return an error.

This covers cases where `os.Hostname()` returns an error, but a `host.id` might still be obtained.

I'm happy to both make the tests and detector logic fail when `os.Hostname()` fails, if folks think that's better.

pyohannes:
> Missing support for BSD: https://github.com/open-telemetry/opentelemetry-go/blob/main/sdk/resource/host_id_bsd.go
> Missing fallback when the host system is not supported for the host ID: https://github.com/open-telemetry/opentelemetry-go/blob/main/sdk/resource/host_id_unsupported.go

Thanks @MrAlias, this should be ready for another review.

MrAlias:
> > But in the interim, having a detector here is probably the correct approach.
> 
> @MrAlias Should we use feature environment variables instead of a new detector module?

That's a good idea.

MrAlias:
> > > But in the interim, having a detector here is probably the correct approach.
> > 
> > 
> > @MrAlias Should we use feature environment variables instead of a new detector module?
> 
> That's a good idea.

I've added this to the SIG meeting agenda to discuss.

MrAlias:
@pyohannes can this be closed? We plan to move this into [this](https://github.com/open-telemetry/opentelemetry-go/blob/6d45f283c7abf2ee5284c769f9ae7109096d3564/sdk/resource/host_id.go) (behind the experimental envar), right?

PR #5420, opened 2024-04-21
Fix datadog incorrect resource names added semconv attribute http.route

MrwanBaghdad:
Added `http.route` key in the span attribute to fix Datadog of resource name as it depends on `http.method` + `http.route`

### DIFF
--- a/instrumentation/github.com/gorilla/mux/otelmux/internal/semconvutil/httpconv.go
+++ b/instrumentation/github.com/gorilla/mux/otelmux/internal/semconvutil/httpconv.go
@@ -318,13 +318,13 @@ func (c *httpConv) ServerRequest(server string, req *http.Request) []attribute.K
 	net.protocol.name       string Note: not set if the value is "http".
 	net.protocol.version    string
 	http.target             string Note: doesn't include the query parameter.
+	http.route              string Note: doesn't include the query parameter.
 	*/
 
 	/* The following semantic conventions are not returned:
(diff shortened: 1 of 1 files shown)


instrumentation/github.com/gorilla/mux/otelmux/internal/semconvutil/httpconv.go
```diff
@@ -407,6 +407,7 @@ func (c *httpConv) ServerRequest(server string, req *http.Request) []attribute.K
 
 	if target != "" {
 		attrs = append(attrs, c.HTTPTargetKey.String(target))
+		attrs = append(attrs, c.HTTPTargetKey.String(target))
```
dmathieu, line 410: This is using `HTTPTargetKey`, so it's setting the `http.target` attribute twice, not `http.route`.

dmathieu (changes requested):
Here, you would be setting `http.route` as the target, not the route.
The target is `/posts/42`, while the route is `/posts/:id`.

If we cannot get the route, it should not be set.
See https://github.com/open-telemetry/semantic-conventions/blob/b77c881ad68dca7bb7af2c97899a0ef64357c47c/docs/http/http-spans.md#http-server-semantic-conventions

MrAlias:
This looks blocked and stale, closing. @MrwanBaghdad if this is in error, please reopen and update/respond to the feedback provided by @dmathieu.

Issue #5423, opened 2024-04-23
Add a resource detector that sets a default value for `service.instance.id` according to semantic conventions


pyohannes:
### Problem Statement

Currently, there is no resource detector provided that populates `service.instance.id` as specified in semantic conventions.

### Proposed Solution

A resource detector for service attributes should be provided which sets values according to [experimental semantic conventions for services](https://github.com/open-telemetry/semantic-conventions/tree/main/docs/resource#service-experimental).

#### Alternatives

Setting `service.instance.id` manually and ensuring that it is unique for each service instance.

dashpole:
If this eventually is moved to stable, it will become part of the default resource detector.  Could we instead make this a feature of the default detector, similar to the [defaultServiceNameDetector](https://github.com/open-telemetry/opentelemetry-go/blob/fe8e3a1b42997767073598aaf6c5923b664b1254/sdk/resource/resource.go#L223)?

We could use environment variables to gate the feature, similar to other experimental parts of the SDK.

@MrAlias do you think we could move https://github.com/open-telemetry/opentelemetry-go/tree/main/sdk/metric/internal/x to sdk/internal/x so it can be used for non-metric SDK features?

MrAlias:
> @MrAlias do you think we could move https://github.com/open-telemetry/opentelemetry-go/tree/main/sdk/metric/internal/x to sdk/internal/x so it can be used for non-metric SDK features?

Yeah, we could add a similar package there. We wouldn't want to move it there as that would cross module boundaries. But we can implement the same thing there for other experimental features of that module.

ayushrakesh:
@pyohannes  Can you help me how to start work on this issue?

PR #5424, opened 2024-04-23
Service resource detector

pyohannes:
Fixes https://github.com/open-telemetry/opentelemetry-go-contrib/issues/5423

This PR adds a resource detector for detecting `service.instance.id` which sets its values according to to [experimental semantic conventions for service attributes](https://github.com/open-telemetry/semantic-conventions/tree/main/docs/resource#service-experimental):

- `service.instance.id`


detectors/service/service_test.go
```diff
@@ -0,0 +1,50 @@
...
+
+	var uuid string
+
+	for _, kv := range serviceResource.Attributes() {
+		if kv.Key == "service.instance.id" {
+			uuid = kv.Value.AsString()
+		}
+	}
+
+	matched, err := regexp.MatchString("^[A-Fa-f0-9]{8}-[A-Fa-f0-9]{4}-[A-Fa-f0-9]{4}-[A-Fa-f0-9]{4}-[A-Fa-f0-9]{12}$", uuid)
```
maryliag, line 28: tip for the future: you can add `/i` at the end of the regex to make it case insensitive, this way you don't need to add `A-F` and `a-f`, which have the risk of forgetting one of them

maryliag (approved):

pyohannes:
Closing in favor of https://github.com/open-telemetry/opentelemetry-go/pull/5520.

PR #5427, opened 2024-04-24
config: add support for logger provider configuration

codeboten:


config/log.go
```diff
@@ -0,0 +1,144 @@
...
+			errs = append(errs, err)
+		}
+	}
+
+	if len(errs) > 0 {
+		return noop.NewLoggerProvider(), noopShutdown, errors.Join(errs...)
+	}
+
+	mp := sdklog.NewLoggerProvider(opts...)
+	return mp, mp.Shutdown, nil
```
MrAlias, line 42: nit

```suggestion
	lp := sdklog.NewLoggerProvider(opts...)
	return lp, lp.Shutdown, nil
```

config/log_test.go
```diff
@@ -0,0 +1,343 @@
...
+								"test": "test1",
+							},
+						},
+					},
+				},
+			},
+			wantErr: errors.New("unsupported compression \"invalid\""),
+		},
+		{
+			name: "simple/no-exporter",
```
MrAlias, line 313: Missing test for simple with a valid exporter.

config/log.go
```diff
@@ -0,0 +1,144 @@
...
+		return batchLogProcessor(processor.Batch, exp)
+	}
+	if processor.Simple != nil {
+		exp, err := logExporter(ctx, processor.Simple.Exporter)
+		if err != nil {
+			return nil, err
+		}
+		return sdklog.NewSimpleProcessor(exp), nil
+	}
+	return nil, fmt.Errorf("unsupported log processor type %v", processor)
```
MrAlias, line 63:
```suggestion
	return nil, fmt.Errorf("unsupported log processor type %[1]#v (%[1]T)", processor)
```

config/log.go
```diff
@@ -0,0 +1,144 @@
...
+}
+
+func logExporter(ctx context.Context, exporter LogRecordExporter) (sdklog.Exporter, error) {
+	if exporter.OTLP != nil {
+		switch exporter.OTLP.Protocol {
+		case protocolProtobufHTTP:
+			return otlpHTTPLogExporter(ctx, exporter.OTLP)
+		default:
+			return nil, fmt.Errorf("unsupported protocol %q", exporter.OTLP.Protocol)
```
MrAlias, line 72: Missing test case.

config/log.go
codeboten: did you mean to leave both v and T in here? with both it ends up logging something like:

```
"unsupported log processor type {<nil> <nil> <nil>} (config.LogRecordProcessor)"
```

config/log_test.go
codeboten: Added

config/log.go
codeboten: test added

config/log.go
MrAlias: Yeah, I was mostly thinking we need to include the type to support the rest of the message. If including the instance value as well isn't going to be helpful I'm fine removing that.

codeboten: i updated the error message to specify the type of processors that are expected instead. the type of processor isn't support helpful and neither are the null values

MrAlias (approved):

codeboten:
@MrAlias is there anything else on this PR that's needed to move it forward?

MrAlias:
It needs another review from a non-Splunker.

cc @open-telemetry/go-approvers

dashpole (approved):

codeboten:
Thanks @MrAlias @dashpole!

PR #5433, opened 2024-04-26
config: add yaml/json struct tags

codeboten:
This will make parsing the configuration easier

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -12,6 +12,7 @@ This project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0.htm
 
 - The `Severitier` and `SeverityVar` types are added to `go.opentelemetry.io/contrib/processors/minsev` allowing dynamic configuration of the severity used by the `LogProcessor`. (#6116)
 - Move examples from `go.opentelemetry.io/otel` to this repository under `examples` directory. (#6158)
+- Support yaml/json struct tags for generated code in `go.opentelemetry.io/contrib/config`. (#5433)
 
 ### Changed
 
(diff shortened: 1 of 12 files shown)


MadVikingGod:
This needs a changelog.

I see this adds additional tags, but I'm not 100% sure how this is used. It doesn't necessarily need to be part of the PR, but could you share an example of how this was used before and how it works now?

pellared:
Can you add tests that verify that the YAML/JSON files are parsed (thanks to the tags) as expected?

codeboten:
@MadVikingGod @pellared happy to add some functionality to test it, i didn't want to overstep on this PR https://github.com/open-telemetry/opentelemetry-go-contrib/pull/4826

@carrbs if you're ok with it I can incorporate your work here, or you can pull this change in your PR, it's all the same to me :)

carrbs:
> @MadVikingGod @pellared happy to add some functionality to test it, i didn't want to overstep on this PR #4826
> 
> @carrbs if you're ok with it I can incorporate your work here, or you can pull this change in your PR, it's all the same to me :)

I'm good with incorporating here, thanks @codeboten!

MadVikingGod:
So, what I was hoping for to understand better why this change is necessary was something like:
Before:
```go
input := map[string]interface{}
err := json.Unmarshal(rawJson, &input)
config := config{}
err = mapstructure.Decode(input, &config)
```

But afterwards we have
```
Some happy path code
```

While this looks to add direct ways to decode directly into our Config structures, it also means we have to test three different decoders (Mapstructure, Json, and Yaml) and find all the edge cases of each.

bogdandrutu (approved):

MrAlias (approved):

pellared (approved):

MrAlias:
@MadVikingGod PTAL

codeboten:
Added test to validate json/yaml parsing is working as expected, also pulled in part of https://github.com/open-telemetry/opentelemetry-go-contrib/pull/4826 for adding a ParseYAML api

pellared (approved):

PR #5559, opened 2024-05-10
detectors/gcp: Add support for Cloud Run jobs

nvx:
Adds support for Cloud Run jobs, previously on Cloud Run services were supported.

I've left the cloud platform as `semconv.CloudPlatformGCPCloudRun` as it's a part of Cloud Run ecosystem, but have no objections to making a "gcp_cloud_run_jobs" if needed although that would require an update to the `semconv` package first and I'm not quite sure I understand how the folder versioning works for that so would be looking for guidance (do I just add it to the latest version folder? do I make a new version copying the latest version folder?)


MrAlias:
cc @dashpole

MrAlias:
@nvx it looks like you pushed a commit with signature for that did not sign the CLA. Please be sure to sign for this alternate or re-sign the commit with the appropriate CLA. We are not able to accept this without doing so.

nvx:
> @nvx it looks like you pushed a commit with signature for that did not sign the CLA. Please be sure to sign for this alternate or re-sign the commit with the appropriate CLA. We are not able to accept this without doing so.

Looks like the changelog updater thing made a commit with a different email, and then even rebasing and squashing back to the original didn't make it happy so had to resign it again. Looks like it's happy now though!

detectors/gcp/detector_test.go
```diff
@@ -107,6 +107,28 @@ func TestDetect(t *testing.T) {
...
+		{
+			desc: "Cloud Run Job",
+			detector: &detector{detector: &fakeGCPDetector{
+				projectID:            "my-project",
+				cloudPlatform:        gcp.CloudRunJob,
+				faaSID:               "1472385723456792345",
+				faaSCloudRegion:      "us-central1",
+				faaSName:             "my-service",
+				cloudRunJobExecution: "my-service-ekdih",
+				cloudRunJobTaskIndex: "0",
```
dashpole, line 119: can you add a test for a non-integer response, which expects an error?

nvx:
I added that test @dashpole, was there anything else needed before this can be merged?

dashpole (approved):
Thanks!

PR #5563, opened 2024-05-12
otelhttptrace: add TLS info to the "http.tls" span

dotwaffle:
I noticed that the `http.tls` span was not including any useful information for otelhttptrace, and considering httptrace is supposed to show excess debug information about an HTTP connection, I figured it would be good to add some simple attributes on that span.

When committing, I noticed that the golangci-lint checks were failing because of some deprecated semconv functions, namely `semconv.NetHostName()`, so I fixed them to what those values they set are to eliminate the linter failures. It looks like `net.host.name` has been replaced with `server.address` in semconv, but I didn't want to break existing code -- happy to change them to the new semconv version if requested!

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -24,7 +24,7 @@ This project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0.htm
 - The `WithSchemaURL` option function in `go.opentelemetry.io/contrib/bridges/otelslog`.
   This option function is used as a replacement of `WithInstrumentationScope` to specify the semantic convention schema URL for the logged records. (#5588)
 - Add support for Cloud Run jobs in `go.opentelemetry.io/contrib/detectors/gcp`. (#5559)
-- Add TLS information to otelhttptrace http.tls attributes, providing information on the cipher and protocol version used, whether the session was resumed, the SHA256 hash of the leaf certificate, "not before"/"not after" dates, and verified certificate chains.
+- Add TLS information from semantic conventions v1.24.0 to `go.opentelemetry.io/contrib/instrumentation/net/http/httptrace`. (#5563)
 
 ### Changed
 
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -24,6 +24,7 @@ This project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0.htm
 - The `WithSchemaURL` option function in `go.opentelemetry.io/contrib/bridges/otelslog`.
   This option function is used as a replacement of `WithInstrumentationScope` to specify the semantic convention schema URL for the logged records. (#5588)
 - Add support for Cloud Run jobs in `go.opentelemetry.io/contrib/detectors/gcp`. (#5559)
+- Add TLS information to otelhttptrace http.tls attributes, providing information on the cipher and protocol version used, whether the session was resumed, the SHA256 hash of the leaf certificate, "not before"/"not after" dates, and verified certificate chains.
 
 ### Changed
 
--- a/instrumentation/net/http/httptrace/otelhttptrace/clienttrace.go
+++ b/instrumentation/net/http/httptrace/otelhttptrace/clienttrace.go
@@ -18,7 +18,6 @@ import (
 	"go.opentelemetry.io/otel"
 	"go.opentelemetry.io/otel/attribute"
 	"go.opentelemetry.io/otel/codes"
-	semconv "go.opentelemetry.io/otel/semconv/v1.24.0"
 	"go.opentelemetry.io/otel/trace"
 )
 
@@ -38,6 +37,19 @@ var (
 	HTTPConnectionDoneNetwork  = attribute.Key("http.conn.done.network")
 	HTTPConnectionDoneAddr     = attribute.Key("http.conn.done.addr")
 	HTTPDNSAddrs               = attribute.Key("http.dns.addrs")
+)
+
+// TLS attributes.
+//
+// From https://opentelemetry.io/docs/specs/semconv/attributes-registry/tls/ but only present from semconv v1.24.
+var (
+	TLSCipher                 = attribute.Key("tls.cipher")
+	TLSProtocolVersion        = attribute.Key("tls.protocol.version")
+	TLSResumed                = attribute.Key("tls.resumed")
+	TLSServerCertificateChain = attribute.Key("tls.server.certificate_chain")
+	TLSServerHashSha256       = attribute.Key("tls.server.hash.sha256")
+	TLSServerNotAfter         = attribute.Key("tls.server.not_after")
+	TLSServerNotBefore        = attribute.Key("tls.server.not_before")
 )
 
 var hookMap = map[string]string{
@@ -323,9 +335,9 @@ func (ct *clientTracer) tlsHandshakeStart() {
 func (ct *clientTracer) tlsHandshakeDone(state tls.ConnectionState, err error) {
 	attrs := make([]attribute.KeyValue, 0, 7)
 	attrs = append(attrs,
-		semconv.TLSCipher(tls.CipherSuiteName(state.CipherSuite)),
-		semconv.TLSProtocolVersion(tls.VersionName(state.Version)),
-		semconv.TLSResumed(state.DidResume),
+		TLSCipher.String(tls.CipherSuiteName(state.CipherSuite)),
+		TLSProtocolVersion.String(tls.VersionName(state.Version)),
+		TLSResumed.Bool(state.DidResume),
 	)
 
 	if len(state.PeerCertificates) > 0 {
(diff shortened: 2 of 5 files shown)


dmathieu:
Please don't change the semconv attributes, despite the deprecations. This work is happening in other PRs, as part of https://github.com/open-telemetry/opentelemetry-go-contrib/pull/5400.

dotwaffle:
Ok, noted. How would you like me to proceed for the time being? Revert to 1.20 and manually enter the attribute keys? Is there any merit in the idea proposed?

dmathieu:
I think we could add the new attributes as string for now.

dotwaffle:
Super, I'll make the changes and come back to you. Thanks!

dotwaffle:
Sorry for the delay, hopefully that's more acceptable now! Let me know if you want anything else changing!

instrumentation/net/http/httptrace/otelhttptrace/clienttrace.go
```diff
@@ -262,7 +278,7 @@ func (ct *clientTracer) span(hook string) trace.Span {
 }
 
 func (ct *clientTracer) getConn(host string) {
-	ct.start("http.getconn", "http.getconn", semconv.NetHostName(host))
+	ct.start("http.getconn", "http.getconn", attribute.Key("net.host.name").String(host))
```
dmathieu, line 281: Don't change this please.

instrumentation/net/http/httptrace/otelhttptrace/clienttrace.go
```diff
@@ -287,7 +303,7 @@ func (ct *clientTracer) gotFirstResponseByte() {
 }
 
 func (ct *clientTracer) dnsStart(info httptrace.DNSStartInfo) {
-	ct.start("http.dns", "http.dns", semconv.NetHostName(info.Host))
+	ct.start("http.dns", "http.dns", attribute.Key("net.host.name").String(info.Host))
```
dmathieu, line 306: Same here.

instrumentation/net/http/httptrace/otelhttptrace/clienttrace.go
dotwaffle: Whoops, sorry about that, that was changed as it was being called out as a deprecated attribute in semconv v1.24.0 -- I've restored the original call.

instrumentation/net/http/httptrace/otelhttptrace/clienttrace.go
dotwaffle: Fixed the same as the previous instance.

CHANGELOG.md
```diff
@@ -24,6 +24,7 @@ This project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0.htm
 - The `WithSchemaURL` option function in `go.opentelemetry.io/contrib/bridges/otelslog`.
   This option function is used as a replacement of `WithInstrumentationScope` to specify the semantic convention schema URL for the logged records. (#5588)
 - Add support for Cloud Run jobs in `go.opentelemetry.io/contrib/detectors/gcp`. (#5559)
+- Add TLS information to otelhttptrace http.tls attributes, providing information on the cipher and protocol version used, whether the session was resumed, the SHA256 hash of the leaf certificate, "not before"/"not after" dates, and verified certificate chains.
```
dmathieu, line 27: I would simply specify that we add the semantic convention attributes for TLS.

```suggestion
- Add TLS attributes from semantic conventions 1.24.0 to `go.opentelemetry.io/contrib/instrumentation/net/http/httptrace`. (#5563)
```

instrumentation/net/http/httptrace/otelhttptrace/clienttrace.go
```diff
@@ -36,6 +40,19 @@ var (
...
+//
+// From https://opentelemetry.io/docs/specs/semconv/attributes-registry/tls/ but only present from semconv v1.24.
+var (
+	TLSCipher                 = attribute.Key("tls.cipher")
+	TLSProtocolVersion        = attribute.Key("tls.protocol.version")
+	TLSResumed                = attribute.Key("tls.resumed")
+	TLSServerCertificateChain = attribute.Key("tls.server.certificate_chain")
+	TLSServerHashSha256       = attribute.Key("tls.server.hash.sha256")
+	TLSServerNotAfter         = attribute.Key("tls.server.not_after")
+	TLSServerNotBefore        = attribute.Key("tls.server.not_before")
```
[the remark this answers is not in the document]
dotwaffle: I just followed the convention of the HTTP attributes above (which do not have the "Key" name in them) -- would you like me to change those to match too?

instrumentation/net/http/httptrace/otelhttptrace/httptrace_test.go
```diff
@@ -30,7 +31,7 @@ func TestRoundtrip(t *testing.T) {
 	props := otelhttptrace.WithPropagators(propagation.NewCompositeTextMapPropagator(propagation.TraceContext{}, propagation.Baggage{}))
 
 	// Mock http server
-	ts := httptest.NewServer(
+	ts := httptest.NewTLSServer(
```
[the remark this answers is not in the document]
dotwaffle: Have reverted this change, as it wasn't testing `http.tls` anyway, that is tested within `clienttrace_test.go` instead.

CHANGELOG.md
dotwaffle: Implemented as requested.

instrumentation/net/http/httptrace/otelhttptrace/test/clienttrace_test.go
```diff
@@ -48,7 +52,7 @@ func TestHTTPRequestWithClientTrace(t *testing.T) {
 	tr := tp.Tracer("httptrace/client")
 
 	// Mock http server
-	ts := httptest.NewServer(
+	ts := httptest.NewTLSServer(
```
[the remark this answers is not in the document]
dotwaffle: Implemented as requested. Thank you humbly for your patience!

instrumentation/net/http/httptrace/otelhttptrace/clienttrace.go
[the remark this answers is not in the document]
dotwaffle: Have changed these as requested, but just the TLS attributes, have left the HTTP attributes alone.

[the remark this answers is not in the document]
dmathieu: Yes, let's not change the other ones. We can do that in another PR to make things more consistent.

instrumentation/net/http/httptrace/otelhttptrace/test/clienttrace_test.go
```diff
@@ -42,107 +47,153 @@ func getSpansFromRecorder(sr *tracetest.SpanRecorder, name string) []trace.ReadO
...
-			parent, ok := getSpanFromRecorder(sr, tl.parent)
-			if assert.True(t, ok) {
-				assert.Equal(t, span.Parent().SpanID(), parent.SpanContext().SpanID())
-			}
+		u, err := url.Parse(ts.URL)
+		if err != nil {
+			panic("unexpected error in parsing httptest server URL: " + err.Error())
+		}
+		// http.tls only exists on HTTPS connections.
+		if u.Scheme == "https" {
```
dmathieu, line 136: There shouldn't be conditions like this in a subtest loop. These should be part of the attributes defined in the loop definition.

instrumentation/net/http/httptrace/otelhttptrace/clienttrace.go
```diff
@@ -316,8 +333,29 @@ func (ct *clientTracer) tlsHandshakeStart() {
 	ct.start("http.tls", "http.tls")
 }
 
-func (ct *clientTracer) tlsHandshakeDone(_ tls.ConnectionState, err error) {
-	ct.end("http.tls", err)
+func (ct *clientTracer) tlsHandshakeDone(state tls.ConnectionState, err error) {
+	attrs := make([]attribute.KeyValue, 0, 7)
```
dashpole: When the TLS handshake fails (`err != nil`), `net/http` invokes `TLSHandshakeDone` with a zero `tls.ConnectionState{}`, which causes `tls.CipherSuiteName` and `tls.VersionName` to record `"0x0000"`. We should skip extracting `state` attributes when `err != nil`:
```suggestion
	if err != nil {
		ct.end("http.tls", err)
		return
	}
	attrs := make([]attribute.KeyValue, 0, 7)
```

instrumentation/net/http/httptrace/otelhttptrace/clienttrace.go
```diff
@@ -316,8 +333,29 @@ func (ct *clientTracer) tlsHandshakeStart() {
...
+	attrs = append(attrs,
+		TLSCipherKey.String(tls.CipherSuiteName(state.CipherSuite)),
+		TLSProtocolVersionKey.String(tls.VersionName(state.Version)),
```
dashpole: `tls.VersionName(state.Version)` returns values like `"TLS 1.3"`, whereas the `tls.protocol.version` semantic convention expects only the numeric version string (e.g. `"1.3"` or `"1.2"`, with `"tls"` in `tls.protocol.name`). Maybe strip the prefix or map the `tls.VersionTLS12` / `tls.VersionTLS13` constants to their numeric strings?

PR #5564, opened 2024-05-13
Fix otelhttptrace example to avoid duplicating the generated data

### LINKED ISSUE #5562: why otelhttptrace duplicate?
in example:
https://github.com/open-telemetry/opentelemetry-go-contrib/blob/main/instrumentation/net/http/httptrace/otelhttptrace/example/client/client.go

```
client := http.Client{
		Transport: otelhttp.NewTransport(
			http.DefaultTransport,
			otelhttp.WithClientTrace(func(ctx context.Context) *httptrace.ClientTrace {
				return otelhttptrace.NewClientTrace(ctx)
			}),
		),
	}

	bag, _ := baggage.Parse("username=donuts")
	ctx := baggage.ContextWithBaggage(context.Background(), bag)

	var body []byte

	tr := otel.Tracer("example/client")
	err = func(ctx context.Context) error {
		ctx, span := tr.Start(ctx, "say hello", trace.WithAttributes(semconv.PeerService("ExampleService")))
		defer span.End()

		ctx = httptrace.WithClientTrace(ctx, otelhttptrace.NewClientTrace(ctx))
		req, _ := http.NewRequestWithContext(ctx, "GET", *url, nil)

		fmt.Printf("Sending request...\n")
		res, err := client.Do(req)
		if err != nil {
			panic(err)
		}
		body, err = io.ReadAll(res.Body)
		_ = res.Body.Close()

		return err
	}(ctx)
```
why func  otelhttptrace.NewClientTrace have called twice?

which is correct?

dmathieu:
Closes #5562

This example shows both ways to setup otelhttptrace, but [WithClientTrace](https://pkg.go.dev/net/http/httptrace#WithClientTrace) appends tracers, it does not replace them. So someone running the example gets duplicated data, which is confusing.

I have chosen to remove the example from the client, as that's our own option. Passing it to the context is "the httptrace way".


dashpole (approved):

MrAlias (approved):

zdyj3170101136:
> Closes #5562
> 
> This example shows both ways to setup otelhttptrace, but [WithClientTrace](https://pkg.go.dev/net/http/httptrace#WithClientTrace) appends tracers, it does not replace them. So someone running the example gets duplicated data, which is confusing.
> 
> I have chosen to remove the example from the client, as that's our own option. Passing it to the context is "the httptrace way".

i recommend you reserve the first way.

first way:
```
Transport: otelhttp.NewTransport(
			http.DefaultTransport,
			otelhttp.WithClientTrace(func(ctx context.Context) *httptrace.ClientTrace {
				return otelhttptrace.NewClientTrace(ctx)
			}),
		),
```
截屏2024-05-14 下午2 38 28

second way:
```
req, _ := http.NewRequestWithContext(ctx, "GET", *url, nil)
```

截屏2024-05-14 下午2 38 42



the first way's jaeger ui is easier for user to understand.

and is much easier for user to use(it could just replace global http.Transport)

dmathieu:
You're right, thank you.
When we add it to the context before the HTTP span is started, the used context doesn't have the proper span parent.

instrumentation/net/http/httptrace/otelhttptrace/example/client/client.go
```diff
@@ -60,6 +60,9 @@ func main() {
 	client := http.Client{
 		Transport: otelhttp.NewTransport(
 			http.DefaultTransport,
+			// By setting the otelhttptrace client in this transport, it can be
+			// injected into the context after the span is started, which makes the
+			// httptrace spans childs of the transport one.
```
MrAlias, line 65:
```suggestion
			// httptrace spans children of the transport one.
```

MadVikingGod (approved):

PR #5566, opened 2024-05-13
otelzap: Skeleton for zap encoder

khushijain21:

Why we might need a custom encoder - 
zap code forces you to implement your own object encoder to intercept field types. We cannot use their default console and json encoder -  as they write to the provided `io.Writer` in JSON format.
 
[MemoryEncoder](https://github.com/uber-go/zap/blob/v1.27.0/zapcore/memory_encoder.go) from zap would have been a suitable solution for our use case - but it is not recommended to be used in production. 

A custom encoder will also ensure support for all compatible types


MrAlias (approved):

pellared (changes requested):
From https://pkg.go.dev/go.uber.org/zap#hdr-Extending_Zap:

>  requires implementing the zapcore.Encoder, zapcore.WriteSyncer, **or zapcore.Core** interfaces

Can we have proof we need to implement the encoder? AFAIK we just need to implement `zapcore.Core`. Zap encoder implementation would be needed only if we would like the users to create the `zapcore.Core` using https://pkg.go.dev/go.uber.org/zap/zapcore#NewCore. What are the benefits of supporting such use-case?

PS. I already mentioned it as the first bullet in https://github.com/open-telemetry/opentelemetry-go-contrib/issues/5191#issuecomment-2098462295

khushijain21:
https://github.com/open-telemetry/opentelemetry-go-contrib/blob/ba244b3fa89280388f869c42e75e36765103c30b/bridges/zap/zap_logger.go#L103

I did use switch statements at the beginning and this is the problem I ran into - 

When a user chooses to implement MarshalLogObject/MarshalLogArray - we need an encoder to unmarshal its contents since `MarshalLogObject` uses methods defined on the encoder interface - in this case `enc.AddString` and `enc.AddInt`. A switch statement cannot recursively unmarshall our data as we see in other bridges. 

```
// Person is a custom type that holds information about a person.
type Person struct {
    Name string
    Age  int
}

// MarshalLogObject implements the zapcore.ObjectMarshaler interface.
func (p Person) MarshalLogObject(enc zapcore.ObjectEncoder) error {
    enc.AddString("name", p.Name)
    enc.AddInt("age", p.Age)
    return nil
}

func main() {
    // Create a zap logger.
    logger, _ := zap.NewProduction()
    defer logger.Sync()

    // Create an instance of Person.
    person := Person{
        Name: "Jane Doe",
        Age:  25,
    }

    // Log the Person struct using the Object field.
    logger.Info("Logging custom object", zap.Object("person", person))
}
```

The only other way to do this w/o an enoder would be to read from the destination that zap writes to and convert the data to OTel values (cannot vouch if this may be a good solution tho)

pellared:
> When a user chooses to implement MarshalLogObject/MarshalLogArray

Now, I get it. Thanks for the explanation 🙏

bridges/otelzap/encoder.go
```diff
@@ -0,0 +1,120 @@
...
+
+import (
+	"time"
+
+	"go.uber.org/zap/zapcore"
+
+	"go.opentelemetry.io/otel/log"
+)
+
+var _ zapcore.ObjectEncoder = (*objectEncoder)(nil)
```
pellared, line 14: What about `zapcore.ArrayEncoder`? Separate PR?

khushijain21: yes that was the idea, should I add it in this PR itself

khushijain21:
>To sum up, this is needed for handling ArrayMarshalerType and ObjectMarshalerType, right?
yes correct

bridges/otelzap/encoder.go
pellared: No. This is fine.

bridges/otelzap/encoder.go
```diff
@@ -0,0 +1,119 @@
...
+	keyval := make([]log.KeyValue, 0, len)
+
+	return &objectEncoder{
+		kv: keyval,
+	}
+}
+
+// TODO
+// Converts Array to log.Slice using ArrayEncoder.
+func (m *objectEncoder) AddArray(key string, v zapcore.ArrayMarshaler) error {
```
pellared, line 32: From https://go.dev/wiki/CodeReviewComments#comment-sentences:

> Comments should begin with the name of the thing being described 

The same comment applies to documentation of all methods.

pellared (approved):

pellared:
```
golangci-lint ./bridges/otelzap
Error: encoder.go:19:2: field `kv` is unused (unused)
	kv []log.KeyValue
	^
Error: encoder.go:22:6: func `newObjectEncoder` is unused (unused)
func newObjectEncoder(len int) *objectEncoder {
     ^
```

You can force the linter to ignore the error by adding a comment looking like `//nolint:unused TODO.`

bridges/otelzap/encoder.go
khushijain21: added documentation comments for primitive types that we are sure. Will add for the rest when implementing them.

XSAM (approved):

PR #5594, opened 2024-05-16
otelzap: Implement Enabled method

khushijain21:

This PR implements `Enabled` method on Core

### DIFF
--- a/bridges/otelzap/core.go
+++ b/bridges/otelzap/core.go
@@ -107,10 +107,12 @@ func NewCore(opts ...Option) *Core {
 	}
 }
 
-// TODO
 // LevelEnabler decides whether a given logging level is enabled when logging a message.
 func (o *Core) Enabled(level zapcore.Level) bool {
-	return true
+	r := log.Record{}
+	r.SetSeverity(convertLevel(level))
+	// TODO How to pass context to Enabled method
+	return o.logger.Enabled(context.Background(), r)
 }
 
 // TODO
--- a/bridges/otelzap/core.go
+++ b/bridges/otelzap/core.go
@@ -111,7 +111,6 @@ func NewCore(opts ...Option) *Core {
 func (o *Core) Enabled(level zapcore.Level) bool {
 	r := log.Record{}
 	r.SetSeverity(convertLevel(level))
-	// TODO How to pass context to Enabled method
 	return o.logger.Enabled(context.Background(), r)
 }
 
--- a/bridges/otelzap/core_test.go
+++ b/bridges/otelzap/core_test.go
@@ -46,11 +46,12 @@ func TestCoreEnabled(t *testing.T) {
 	}
 
 	assert.Empty(t, r.Result()[1].Records)
+	logger.Debug(testMessage)
+	assert.Empty(t, r.Result()[1].Records)
 
 	if ce := logger.Check(zap.InfoLevel, testMessage); ce != nil {
 		ce.Write()
 	}
-
 	require.Len(t, r.Result()[1].Records, 1)
 	got := r.Result()[1].Records[0]
 	assert.Equal(t, testMessage, got.Body().AsString())
--- a/bridges/otelzap/core_test.go
+++ b/bridges/otelzap/core_test.go
@@ -4,6 +4,7 @@
 package otelzap
 
 import (
+	"context"
 	"testing"
 
 	"github.com/stretchr/testify/assert"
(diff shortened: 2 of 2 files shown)


bridges/otelzap/core.go
```diff
@@ -107,10 +107,12 @@ func NewCore(opts ...Option) *Core {
 	}
 }
 
-// TODO
 // LevelEnabler decides whether a given logging level is enabled when logging a message.
 func (o *Core) Enabled(level zapcore.Level) bool {
-	return true
+	r := log.Record{}
+	r.SetSeverity(convertLevel(level))
+	// TODO How to pass context to Enabled method
```
pellared, line 114: You cannot 😉 

```suggestion
```

bridges/otelzap/core_test.go
```diff
@@ -31,6 +32,18 @@ func TestCore(t *testing.T) {
...
+	enabledFunc := func(c context.Context, r log.Record) bool {
+		return r.Severity() >= log.SeverityInfo
+	}
+
+	r := logtest.NewRecorder(logtest.WithEnabledFunc(enabledFunc))
+	zc := NewCore(WithLoggerProvider(r))
+
+	assert.False(t, zc.Enabled(zap.DebugLevel), "level conversion: permissive")
+	assert.True(t, zc.Enabled(zap.InfoLevel), "level conversion: restrictive")
+}
```
pellared, line 45: Can you please test the functionality in a producation-like way (like in `TestCore`) by using https://pkg.go.dev/go.uber.org/zap#Logger.Check?

bridges/otelzap/core_test.go
```diff
@@ -31,6 +33,29 @@ func TestCore(t *testing.T) {
...
+	}
+
+	r := logtest.NewRecorder(logtest.WithEnabledFunc(enabledFunc))
+	logger := zap.New(NewCore(WithLoggerProvider(r)))
+
+	if ce := logger.Check(zap.DebugLevel, testMessage); ce != nil {
+		ce.Write()
+	}
+
+	assert.Nil(t, r.Result()[1].Records)
```
pellared, line 48:
```suggestion
	assert.Empty(t, r.Result()[1].Records)
```

bridges/otelzap/core_test.go
```diff
@@ -31,6 +33,29 @@ func TestCore(t *testing.T) {
...
+
+	if ce := logger.Check(zap.InfoLevel, testMessage); ce != nil {
+		ce.Write()
+	}
+
+	got := r.Result()[1].Records[0]
```
pellared, line 54: let's make sure that this test will not panic if the implementation does not work properly

```suggestion
	require.Len(t, r.Result()[1].Records, 1)
	got := r.Result()[1].Records[0]
```

bridges/otelzap/core_test.go
```diff
@@ -31,6 +33,29 @@ func TestCore(t *testing.T) {
 	assert.Equal(t, log.SeverityInfo, got.Severity())
 }
 
+func TestCoreEnabled(t *testing.T) {
+	enabledFunc := func(c context.Context, r log.Record) bool {
+		return r.Severity() >= log.SeverityInfo
+	}
+
+	r := logtest.NewRecorder(logtest.WithEnabledFunc(enabledFunc))
```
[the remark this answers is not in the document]
khushijain21: yes, but I was thinking from the user perspective. There is no WithEnabledFunc method available on the SDK implementation of `LoggerProvider` right?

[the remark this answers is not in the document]
pellared: Right now the user would have to create a custom [`log.Processor`](https://pkg.go.dev/go.opentelemetry.io/otel/sdk/log#Processor). We have an issue to add a "minimum-level-processor":  https://github.com/open-telemetry/opentelemetry-go/issues/5078. However, I think it is better to postpone it before we have clarification how to do it in the specification: https://github.com/open-telemetry/opentelemetry-specification/issues/3917

bridges/otelzap/core_test.go
```diff
@@ -31,6 +33,30 @@ func TestCore(t *testing.T) {
 	assert.Equal(t, log.SeverityInfo, got.Severity())
 }
 
+func TestCoreEnabled(t *testing.T) {
```
pellared, line 36: What about `logger.Debug(testMessage)`. Will it be recorded or not? I think it deserves a test case.

pellared (approved):

XSAM (approved):

Issue #5551, opened 2024-05-09
Deprecate and remove go.opentelemetry.io/contrib/instrumentation/go.mongodb.org/mongo-driver/mongo/otelmongo


MrAlias:
This module has been identified to not have an [owner](https://github.com/open-telemetry/opentelemetry-go-contrib/blob/5d8c6d20b1852e47e15b989abe813c20cb935e9b/CODEOWNERS). Based on the [project's ownership policy](https://github.com/open-telemetry/opentelemetry-go-contrib/issues/2834#issuecomment-1282698918), this module will be deprecated and then removed.

### How to keep this module

For this module to continue in this repository, it needs a sponsor.

If you would like to sponsor this module and become an owner, please comment in this issue about your desire. As an owner you will assume the following responsibilities:

- You will need to [be a member of the OpenTelemetry organization](https://github.com/open-telemetry/community/blob/main/community-membership.md#member) and maintain that membership.
  - If you are not an existing member, we can add you to the community. This is not a disqualification, but something you will need in the process of becoming a sponsor.
- You will be responsible for keeping up with the instrumented library. Any "upstream" changes that impact this module need to be proactively handle by you.
- You will be expected to review any Pull Requests or Issues created that relate to this module.
- You will be responsible for the stability and versioning compliance of the module.

You will need to have a good working knowledge of the code this module is instrumenting and, ideally, familiarity with the existing module code.

### How this module will be removed

This module is in the process of being deprecated. After that deprecation notice has been published, we will wait 3 months or 2 two releases (whichever is the longer time period). After that time period, this module will be removed from this repository and no more versions of the module will be published.

#### Resurrection

If a sponsor is found after the module has been deprecated or removed, these operations can be reversed (i.e. coded added back, deprecation notice removed).

pellared:
Duplicates https://github.com/open-telemetry/opentelemetry-go-contrib/issues/4678

There is an informal agreement that the MongoDB (Go driver) team will take over the instrumentation library.
Reference: https://jira.mongodb.org/browse/GODRIVER-2938

However, there is no progress so I propose to follow the plan to deprecate it in 4 weeks if no code owner is found.

dmathieu:
I have posted about this removal in the mongo community forum.
https://www.mongodb.com/community/forums/t/otelmongo-at-risk-of-removal/282989

pellared:
I also added a comment in https://jira.mongodb.org/browse/GODRIVER-2938

[...]

PR #5598, opened 2024-05-16
Add deprecation notice to otelmongo

MrAlias:
Towards https://github.com/open-telemetry/opentelemetry-go-contrib/issues/5551

First deprecation notice to add for the [Go: Remove unowned modules project](https://github.com/orgs/open-telemetry/projects/92/views/1). If merged, this design will be applied to all the other modules we plan to deprecate.


instrumentation/go.mongodb.org/mongo-driver/mongo/otelmongo/doc.go
```diff
@@ -12,4 +12,10 @@
 // This code was originally based on the following:
 // - https://github.com/DataDog/dd-trace-go/tree/02f0449efa3cb382d499fadc873957385dcb2192/contrib/go.mongodb.org/mongo-driver/mongo
 // - https://github.com/DataDog/dd-trace-go/tree/v1.23.3/ddtrace/ext
+//
+// Deprecated: otelmongo is deprecated. This module will be frozen and no
+// longer supported in two minor releases unless an owner can be found. See
+// [this issue] if you would like to become the owner of this module.
```
pellared, line 18: 1. `Deprecated: otelmongo is deprecated.` - this is repetitive. We should say why it is deprecated. Reference: https://go.dev/wiki/Deprecated
2. I think using a date will be more informative for the user than saying (in two minor releases). Otherwise, do we plan to change this text after next release? Also if we remove it later it is nothing bad happens and people can prepare better if they know what is the date when they could expect "end of support".
3. I suggest using the term "Code Owner" like here: https://github.com/open-telemetry/opentelemetry-go-contrib/blob/main/CONTRIBUTING.md#code-owners



```suggestion
// Deprecated: otelmongo has no Code Owner.
// After August 21, 2024, it may no longer be supported and may stop
// receiving new releases unless a new Code Owner is found. See
// [this issue] if you would like to become the Code Owner of this module.
```

instrumentation/go.mongodb.org/mongo-driver/mongo/otelmongo/go.mod
```diff
@@ -1,3 +1,4 @@
+// Deprecated: otelmongo is deprecated.
```
pellared, line 1:
```suggestion
// Deprecated: otelmongo has no Code Owner.
```

CHANGELOG.md
```diff
@@ -29,6 +29,10 @@ This project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0.htm
 ### Deprecated
 
 - The `InterceptorFilter` type in `go.opentelemetry.io/contrib/instrumentation/google.golang.org/grpc/otelgrpc` is deprecated. (#5196)
+- The `go.opentelemetry.io/contrib/instrumentation/go.mongodb.org/mongo-driver/mongo/otelmongo` package is deprecated.
+  If you would like to become an owner of this module and prevent it from being frozen, see [#5551]. (#5598)
```
pellared, line 33: 1. I suggest using the term "Code Owner" like here: https://github.com/open-telemetry/opentelemetry-go-contrib/blob/main/CONTRIBUTING.md#code-owners
2. I think "frozen" can be understood as "does not receive features but is still released".  Especially we already have such definition in OTel Spec: https://opentelemetry.io/docs/specs/otel/document-status/#feature-freeze. I would rather use "removed" (or "dropped", "abandoned").

```suggestion
  If you would like to become an Code Owner of this module and prevent it from being removed, see [#5551]. (#5598)
```

pellared (changes requested):

pellared:
Do we want to add PRs to https://github.com/orgs/open-telemetry/projects/92?

CHANGELOG.md
MrAlias:
```suggestion
  If you would like to become a Code Owner of this module and prevent it from being removed, see [#5551]. (#5598)
```

dashpole (approved):

pellared (approved):

MadVikingGod (approved):

dmathieu (approved):

PR #5602, opened 2024-05-17
otelzap: Implement methods on `objectEncoder`

khushijain21:

This PR implements some more methods on `objectEncoder`


bridges/otelzap/encoder.go
```diff
@@ -89,16 +87,17 @@ func (m *objectEncoder) OpenNamespace(k string) {
...
-func (m *objectEncoder) AddInt32(k string, v int32)     { m.AddInt64(k, int64(v)) }
-func (m *objectEncoder) AddInt16(k string, v int16)     { m.AddInt64(k, int64(v)) }
-func (m *objectEncoder) AddInt8(k string, v int8)       { m.AddInt64(k, int64(v)) }
+func (m *objectEncoder) AddComplex64(k string, v complex64)    { m.AddComplex128(k, complex128(v)) }
+func (m *objectEncoder) AddDuration(k string, v time.Duration) { m.AddInt64(k, v.Nanoseconds()) }
+func (m *objectEncoder) AddTime(k string, v time.Time)         { m.AddInt64(k, v.UnixNano()) }
+func (m *objectEncoder) AddFloat32(k string, v float32)        { m.AddFloat64(k, float64(v)) }
+func (m *objectEncoder) AddInt32(k string, v int32)            { m.AddInt64(k, int64(v)) }
+func (m *objectEncoder) AddInt16(k string, v int16)            { m.AddInt64(k, int64(v)) }
+func (m *objectEncoder) AddInt8(k string, v int8)              { m.AddInt64(k, int64(v)) }
```
pellared, line 96: Can you please format the methods in "regular" way. E.g.

```go
func (m *objectEncoder) AddComplex64(k string, v complex64) {
   m.AddComplex128(k, complex128(v)) 
}

func (m *objectEncoder) AddDuration(k string, v time.Duration) {
...
}
```

bridges/otelzap/encoder.go
```diff
@@ -50,12 +51,9 @@ func (m *objectEncoder) AddBool(k string, v bool) {
...
 }
 
-func (m *objectEncoder) AddDuration(k string, v time.Duration) {
-	// TODO
-}
-
 func (m *objectEncoder) AddComplex128(k string, v complex128) {
-	// TODO.
+	stringValue := strconv.FormatComplex(v, 'f', -1, 64)
+	m.kv = append(m.kv, log.String(k, stringValue))
```
pellared, line 56: What do you think about representing it in a structured way as a "map"?

```suggestion
	r := log.Float64("r", real(v))
	i := log.Float64("i", imag(v))
	m.kv = append(m.kv, log.Map(k, r, i))
```

khushijain21:
```
func main() {
    c := complex(1, 2)
    fmt.Println(c) // Output: (1+2i)
}
```

Here, the output would be 1+2i, log.String would match this more closely than a map right. wdyt

pellared: I just think it is better to have more accurate structure than string formatting. By formatting we can also lose more precision.

pellared (approved):

XSAM (approved):

PR #5611, opened 2024-05-20
otelzap: Add skeleton for array encoder

khushijain21:

This PR adds skeleton for arrayEncoder


bridges/otelzap/encoder.go
```diff
@@ -121,3 +124,45 @@ func (m *objectEncoder) AddUint32(k string, v uint32)   {}
 func (m *objectEncoder) AddUint16(k string, v uint16)   {}
 func (m *objectEncoder) AddUint8(k string, v uint8)     {}
 func (m *objectEncoder) AddUintptr(k string, v uintptr) {}
+
+// arrayEncoder implements [zapcore.ArrayEncoder].
+type arrayEncoder struct {
```
pellared, line 129: Do we need to have seperate structs (`objectEncoder` and `arrayEncoder`)? Is there any reason why `objectEncoder` would not implement `zapcore.ArrayEncoder`?  Can we have a single one instead? Then we could rename `objectEncoder` to `encoder`.

khushijain21: 1. If we have `objectEncoder` implement `arrayEncoder` then in `AddArray` method we will be creating an entire `objectEncoder` - which has many methods not required for unmarshalling the array. If we have nested arrays then this may bloat our function right

```
func (m *objectEncoder) AddArray(key string, v zapcore.ArrayMarshaler) error {
	arr, free := getArrayEncoder()
	defer free()
	err := v.MarshalLogArray(arr)
	m.cur.kv = append(m.cur.kv, log.Slice(key, arr.elems...))
	return err
}
``` 

2. and the objectEncoder will always have to be initialized like this 
```
struct objectEncoder{
   kv []log.KeyValue
   elems []log.Value
}
```
This will create more heap memory that could potentially be unused.

wdyt

pellared: You are right. Better to keep them separated 👍

MrAlias (approved):

dmathieu (approved):

pellared:
@khushijain21, can you sync the branch please.

bridges/otelzap/core.go
```diff
@@ -101,7 +101,7 @@ func NewCore(name string, opts ...Option) *Core {
 	}
 }
 
-// LevelEnabler decides whether a given logging level is enabled when logging a message.
+// Enabled method decides whether a given logging level is enabled when logging a message.
```
khushijain21, line 104: Corrected this comment

PR #5619, opened 2024-05-21
baggagetrace: Add baggage key predicate

### LINKED ISSUE #5437: [Baggage span processor] Key predicate
This issue is to track adding a method of selecting what baggage key entries should be copied.

Feedback in the JS contrib PR was to allow a user-provided predicate function. This puts the responsibility on the user to ensure sensitive baggage keys are not copied while also not prescribing how that is determined.
- https://github.com/open-telemetry/opentelemetry-js-contrib/issues/2166

We had a similar feedback in the .NET contrib project but thought it was more complicated than just using a set of prefixes so created an issue to continue the discussion. The plain processor that copies all baggage entries (like using `*` in your example) is likely to be accepted first.
- https://github.com/open-telemetry/opentelemetry-dotnet-contrib/issues/1695

MikeGoldsmith:
## Which problem is this PR solving?
- Closes #5437 

## Short description of the changes
- Add `BaggageKeyPredicate` type and `AllowAllBaggageKeys` implementation that allows all keys
- Add baggage key predicate as constructor parameter to processor
- Add unit tests to verify behaviour for prefix and regex
- Update README with examples of setting a filter


MikeGoldsmith:
Ping @open-telemetry/go-approvers @open-telemetry/go-maintainers - please can you take a look?

processors/baggage/baggagetrace/README.md
```diff
@@ -0,0 +1,47 @@
+# Baggage Span Processor
```
pellared, line 1: Please follow https://github.com/open-telemetry/opentelemetry-go/blob/main/CONTRIBUTING.md#documentation for documentation and exmaples

processors/baggage/baggagetrace/processor.go
```diff
@@ -11,23 +11,36 @@ import (
...
 var _ trace.SpanProcessor = (*SpanProcessor)(nil)
 
 // New returns a new SpanProcessor.
 //
 // The Baggage span processor duplicates onto a span the attributes found
 // in Baggage in the parent context at the moment the span is started.
-func New() trace.SpanProcessor {
-	return &SpanProcessor{}
+// The predicate function is used to filter which baggage keys are added to the span.
+func New(baggageKeyPredicate BaggageKeyPredicate) trace.SpanProcessor {
```
pellared, line 32: Please use the options pattern so that the new configuration option is optional https://github.com/open-telemetry/opentelemetry-go/blob/main/CONTRIBUTING.md#configuration

processors/baggage/baggagetrace/processor_test.go
```diff
@@ -45,7 +47,72 @@ func TestSpanProcessorAppendsBaggageAttributes(t *testing.T) {
...
+		trace.WithSpanProcessor(trace.NewSimpleSpanProcessor(exporter)),
+	)
+
+	// create tracer and start/end span
+	tracer := tp.Tracer("test")
+	_, span := tracer.Start(ctx, "test")
+	span.End()
+
+	assert.Len(t, exporter.spans, 1)
+	assert.Len(t, exporter.spans[0].Attributes(), 1)
```
pellared, line 92: this will panic if the previous assertion fails

processors/baggage/baggagetrace/processor_test.go
```diff
@@ -45,7 +47,72 @@ func TestSpanProcessorAppendsBaggageAttributes(t *testing.T) {
...
+
+	want := []attribute.KeyValue{attribute.String("baggage.test", "baggage value")}
+	assert.Equal(t, want, exporter.spans[0].Attributes())
```
pellared, line 95: this will panic if the previous assertion fails

processors/baggage/baggagetrace/processor_test.go
```diff
@@ -45,7 +47,72 @@ func TestSpanProcessorAppendsBaggageAttributes(t *testing.T) {
...
+		trace.WithSpanProcessor(trace.NewSimpleSpanProcessor(exporter)),
+	)
+
+	// create tracer and start/end span
+	tracer := tp.Tracer("test")
+	_, span := tracer.Start(ctx, "test")
+	span.End()
+
+	assert.Len(t, exporter.spans, 1)
+	assert.Len(t, exporter.spans[0].Attributes(), 1)
```
pellared, line 60: this will panic if the previous assertion fails

processors/baggage/baggagetrace/processor_test.go
```diff
@@ -45,7 +47,72 @@ func TestSpanProcessorAppendsBaggageAttributes(t *testing.T) {
...
+
+	want := []attribute.KeyValue{attribute.String("baggage.test", "baggage value")}
+	assert.Equal(t, want, exporter.spans[0].Attributes())
```
pellared, line 63: this will panic if the previous assertion fails

processors/baggage/baggagetrace/processor.go
MikeGoldsmith: I've intentionally used a constructor parameter to make the predicate function required. If it's an optional parameter it's easy for a user to miss and not set an appropriate filter. This helps it be safe by default and offer a connivence allow all filter.

MikeGoldsmith:
Thanks for your feedback @pellared - I've addressed your feedback. Please re-review and let me know if there are more things you'd like to me to consider.

processors/baggage/baggagetrace/processor.go
pellared: How would you add a new configuration without introducing a breaking change?

I do not think it is convenient to require a filter. 

> offer a connivence allow all filter.

Users may miss it as well.

MikeGoldsmith: Agree, this is a breaking change but I think it's worth it.

Requiring a filter isn't convenient but is safer than presuming all baggage keys should be copied, which is the current behaviour.

It's possible users miss the convenience function `baggagetrace.AllowAllBaggageKeys`, but they won't be able to add the processor without providing a filter function so they at least have to consider what should be allowed.

Hopefully the documentation is extensive enough they discover it and use it.

processors/baggage/baggagetrace/processor.go
```diff
@@ -11,23 +11,36 @@ import (
...
 
 var _ trace.SpanProcessor = (*SpanProcessor)(nil)
 
 // New returns a new SpanProcessor.
 //
 // The Baggage span processor duplicates onto a span the attributes found
 // in Baggage in the parent context at the moment the span is started.
-func New() trace.SpanProcessor {
-	return &SpanProcessor{}
+// The predicate function is used to filter which baggage keys are added to the span.
```
codeboten, line 31: Might be useful to document the `AllowAllBaggageKeys` filter for discoverability

codeboten (approved):
Thanks @MikeGoldsmith, I agree that making this a required parameter is less convenient but also more secure. Additionally, this will be consistent with the other language implementations.

processors/baggage/baggagetrace/doc.go
```diff
@@ -2,4 +2,56 @@
...
+//		return strings.HasPrefix(baggageKey, "my-key")
+//	})
+//
+// For example, to only copy baggage entries that match the regex '^key.+':
+//
+//	expr := regexp.MustCompile(`^key.+`)
+//
+//	baggagetrace.New(func(baggageKey string) bool {
+//		return expr.MatchString(baggageKey)
+//	})
```
pellared, line 53: From 

> Prefer using [Examples](https://pkg.go.dev/testing#hdr-Examples) instead of putting code snippets in Go doc comments. In some cases, you can even create [Testable Examples](https://go.dev/blog/examples).

processors/baggage/baggagetrace/doc.go
```diff
@@ -2,4 +2,56 @@
...
+//
+// Keys and values added to Baggage will appear on all subsequent child spans for
+// a trace within this service *and* will be propagated to external services via
+// propagation headers.
+// If the external services also have a Baggage span processor, the keys and
+// values will appear in those child spans as well.
+//
+// ⚠️ Waning ⚠️
+// To repeat: a consequence of adding data to Baggage is that the keys and values
+// will appear in all outgoing HTTP baggage headers from the application.
```
pellared, line 18: typo + this is redundant

processors/baggage/baggagetrace/processor.go
pellared: Resolving also per https://github.com/open-telemetry/opentelemetry-go-contrib/pull/5619#pullrequestreview-2094879032

processors/baggage/baggagetrace/doc.go
MikeGoldsmith: I don't think this is redundant - it's a valid security concern that should be called out to ensure users understand the implication of using baggage. I can remove it if you feel strongly that it shouldn't be there though.

processors/baggage/baggagetrace/processor.go
MikeGoldsmith: It did have a docs entry, but because it also re-declared the type in the var, it thought it was another type definition. Fixed in [a8fbc83](https://github.com/open-telemetry/opentelemetry-go-contrib/pull/5619/commits/a8fbc837f2d3432755e90c7f2fa4c2bff3fcba6c)

processors/baggage/baggagetrace/doc.go
MikeGoldsmith: Added examples in [111bf75](https://github.com/open-telemetry/opentelemetry-go-contrib/pull/5619/commits/111bf75ee014958330702c54cee0758418fe54b9)

processors/baggage/baggagetrace/doc.go
pellared: > *and* will be propagated to external services via propagation headers.

This is already mentioned here. Maybe it could be reworded to make it more concerning?

>  keys and values will appear in all outgoing HTTP baggage headers from the application.

This is not correct. It would be only correct if e.g. `otelhttp` is used everywhere to make HTTP calls. Also baggage can appear in other outgoing traffic (e.g. message queues) in other form than HTTP headers.

MikeGoldsmith: I think you're right - the warning may be a little heavy handed and does repeat itself a little. 

I've updated to remove the warning and repeated statement about HTTP propagation headers but left in the "don't put sensitive data in baggage".

What do you think of this? [b759d50](https://github.com/open-telemetry/opentelemetry-go-contrib/pull/5619/commits/b759d50d85940684aacd88eebefda7e643083a8e)

MadVikingGod (approved):
Looks good.

I would suggest a test with more than one attribute added and a test showing that it will skip baggage that doesn't match.

MikeGoldsmith:
Added extra test and updated all tests to be a littel clearer.

Thanks for the feedback @MadVikingGod @pellared

processors/baggage/baggagetrace/doc.go
```diff
@@ -1,5 +1,27 @@
 // Copyright The OpenTelemetry Authors
 // SPDX-License-Identifier: Apache-2.0
 
-// Package baggagetrace implements a baggage span processor.
+// Package baggagetrace is an OpenTelemetry [Span Processor] that reads key/values
+// stored in [Baggage] in the starting span's parent context and adds them as
+// attributes to the span.
+//
+// Keys and values added to Baggage will appear on all subsequent child spans for
+// a trace within this service *and* will be propagated to external services via
```
pellared, line 9: nit: check with `pkgsite` how it renders.

I think it does NOT support *italic*

more: https://github.com/open-telemetry/opentelemetry-go/blob/main/CONTRIBUTING.md#documentation

```suggestion
// a trace within this service and will be propagated to external services via
```

processors/baggage/baggagetrace/processor_test.go
```diff
@@ -54,9 +51,101 @@ func TestSpanProcessorAppendsBaggageAttributes(t *testing.T) {
...
+	require.Len(t, exporter.spans, 1)
+	require.Len(t, exporter.spans[0].Attributes(), 1)
+
+	want := attribute.String("baggage.test", "baggage value")
+	require.Equal(t, want, exporter.spans[0].Attributes()[0])
+}
+
+func addEntryToBaggage(baggage otelbaggage.Baggage, key, value string) otelbaggage.Baggage {
+	member, _ := otelbaggage.NewMemberRaw(key, value)
+	baggage, _ = baggage.SetMember(member)
```
pellared, line 149: Can you ensure that there are no errors? E.g. pass `t` as first parameter, and use `require.NoError` to fail the test if there is a problem with creating+adding an entry.

MikeGoldsmith: Fixed [bd60f43](https://github.com/open-telemetry/opentelemetry-go-contrib/pull/5619/commits/bd60f43ad6a414331314a5f94875cad80a8b57cd)

processors/baggage/baggagetrace/doc.go
MikeGoldsmith: Fixed in [5a137c9](https://github.com/open-telemetry/opentelemetry-go-contrib/pull/5619/commits/5a137c96fd17b6045364b0466aa5752cef97c83c)

pellared (approved):

MikeGoldsmith:
Thanks for all your feedback @pellared - much appreciated.

PR #5626, opened 2024-05-21
fix Incorrect conversion between integer types

Nanmozhi22:


dmathieu:
Could you add a test (and sign the CLA) ?

samplers/probability/consistent/tracestate.go
```diff
@@ -187,18 +187,20 @@ func parseOTelTraceState(ts string, isSampled bool) (otelTraceState, error) { //
 	return otts, nil
 }
 
-func parseNumber(key string, input string, maximum uint8) (uint8, error) {
+// Here`strconv.ParseInt` is used to parse the input string into an integer. The bit size is specified as 8, which is the bit size of `int8`. The parsed value is then checked against the maximum value, and if it's greater, an error is returned. If the parsed value is within the allowed range, it's returned as an `int8` value.
```
dmathieu, line 190: This kind of comment is too verbose imho. There's no need to describe what the method does, reading the code is the best doc.

brianwarner:
Hi @dmathieu, we've signed the CLA and are configured in EasyCLA to accept anyone who is a member of our public-facing org, but for some reason it's not picking up @Nanmozhi22. I've filed a ticket with support asking for help.

samplers/probability/consistent/tracestate.go
```diff
@@ -187,18 +187,20 @@ func parseOTelTraceState(ts string, isSampled bool) (otelTraceState, error) { //
...
 	if input == "" {
 		return maximum + 1, nil
 	}
-	value, err := strconv.ParseUint(input, 10, 64)
+	value, err := strconv.ParseInt(input, 10, 8)
 	if err != nil {
 		return maximum + 1, parseError(key, err)
 	}
-	if value > uint64(maximum) {
+	if value > int64(maximum) {
```
merbhavik, line 200: Comparing int8(value) with int64(maximum) can I know why?

samplers/probability/consistent/tracestate.go
```diff
@@ -187,18 +187,20 @@ func parseOTelTraceState(ts string, isSampled bool) (otelTraceState, error) { //
 	return otts, nil
 }
 
-func parseNumber(key string, input string, maximum uint8) (uint8, error) {
+// Here`strconv.ParseInt` is used to parse the input string into an integer. The bit size is specified as 8, which is the bit size of `int8`. The parsed value is then checked against the maximum value, and if it's greater, an error is returned. If the parsed value is within the allowed range, it's returned as an `int8` value.
+
+func parseNumber(key string, input string, maximum int8) (int8, error) {
```
MadVikingGod, line 192: This change doesn't make sense. The output of this is going to be used as a uint8 anyway, so we are now throwing away half the usable space. This change will break builds that haven't been run.

samplers/probability/consistent/tracestate.go
```diff
@@ -187,18 +187,20 @@ func parseOTelTraceState(ts string, isSampled bool) (otelTraceState, error) { //
...
+	value, err := strconv.ParseInt(input, 10, 8)
 	if err != nil {
 		return maximum + 1, parseError(key, err)
 	}
-	if value > uint64(maximum) {
+	if value > int64(maximum) {
 		return maximum + 1, parseError(key, strconv.ErrRange)
 	}
-	return uint8(value), nil
+	return int8(value), nil
```
MadVikingGod, line 203: This comparison will prevent the uint8 conversion from ever overflowing.  Furthermore, this is always called with maximum = 62, so we aren't near overflow anyways.

MadVikingGod (changes requested):
These changes don't effectively make this function any safer and break builds. Therefore, this should not be applied.

samplers/probability/consistent/tracestate.go
Nanmozhi22: Thanks for the response! Do you have any suggestions on how to make it , so that I can try to make those changes

PR #5634, opened 2024-05-22
Introduce respWriter.Flush so we don't write the status twice

### LINKED ISSUE #5438: otelhttp response writer implementation does not track `WriteHeader` when `Flush` is called
### Description

When using the otelhttp wrapper to instrument an http endpoint, the wrapper is doing its own tracking to determine if `WriteHeader` should be called during a `Write`.

However the wrapper does not implement `http.Flusher`, as such  when `Flush()` is called on the http.ResponseWriter the undrlying flusher from net/http is called, however otelhttp is not aware that the headers have already been written.
When a subsequent call to `Write` is made, otelhttp calls `WriteHeader` first, which has already been called on the underlying net/http implementation, and we wend up with a log printed from net/http:

> http: superfluous response.WriteHeader call from go.opentelemetry.io/contrib/instrumentation/net/http/otelhttp.(*respWriterWrapper).WriteHeader (wrap.go:98)

### Environment

- OS: N/A
- Architecture: N/A
- Go Version: N/A
- `otelhttp` version: v0.46.1

### Steps To Reproduce

Basic example:

```go
func someHandler(w http.ResponseWriter, _ *http.Request) {
  w.(http.Flusher).Flush()
  w.Write([]byte("hello world"))
}
```

### Expected behavior

Since otelhttp is doing its own tracking in its `Write` implementation for when to call `WriteHeader`, it needs to be aware of when `Flush` is called.

dmathieu:
Closes #5438.

### DIFF
--- a/instrumentation/net/http/otelhttp/handler.go
+++ b/instrumentation/net/http/otelhttp/handler.go
@@ -204,6 +204,9 @@ func (h *middleware) serveHTTP(w http.ResponseWriter, r *http.Request, next http
 		WriteHeader: func(httpsnoop.WriteHeaderFunc) httpsnoop.WriteHeaderFunc {
 			return rww.WriteHeader
 		},
+		Flush: func(httpsnoop.FlushFunc) httpsnoop.FlushFunc {
+			return rww.Flush
+		},
 	})
 
 	labeler := &Labeler{}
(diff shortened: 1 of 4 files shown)


pellared (approved):

cpuguy83 (approved):

dmathieu:
@open-telemetry/go-approvers this needs a review from a non-splunk employee.

dashpole (approved):

PR #5639, opened 2024-05-22
Deprecate otelmongo/test

MrAlias:
Part of #5551

Follow up to a missing change not included in #5598. We want to deprecate `otelmongo` and `otelmongo/test` at the same time.


pellared (approved):

dashpole (approved):
Non-blocking: WDYT of not including a changelog entry for this PR, and changing the doc.go to say that otelmongo (instead of test) has no codeowner.  I don't think it really makes sense to separately call for a CODEOWNER of the test directory.

MrAlias:
> Non-blocking: WDYT of not including a changelog entry for this PR, and changing the doc.go to say that otelmongo (instead of test) has no codeowner. I don't think it really makes sense to separately call for a CODEOWNER of the test directory.

I'm onboard for this if others are. @pellared?

pellared:
I support  @dashpole proposal

MrAlias:
Updated, PTAL

dashpole (approved):

pellared (approved):

dmathieu (approved):

PR #5664, opened 2024-05-28
Use more efficient `WithAttributeSet()`

ash2k:
https://github.com/open-telemetry/opentelemetry-go/blob/5661ff0ded32cf1b83f1147dae96ca403c198504/metric/instrument.go#L349-L364

Since all the usage places of `WithAttributes()` operate on a locally allocated slice, there is no need to make a defensive copy.

PR #5681, opened 2024-05-29
Add otelhttp Handler.ServeHTTP and Transport.RoundTrip benchmarks

dmathieu:
This adds high-level benchmarks for `otelhttp.Handler.ServeHTTP` and `otelhttp.Transport.RoundTrip`, as well as a benchmark comparison without those wrappers.

Related: #5664.

Current results on my machine:

```
goos: darwin
goarch: arm64
pkg: go.opentelemetry.io/contrib/instrumentation/net/http/otelhttp/test
BenchmarkHandlerServeHTTP/without_the_otelhttp_handler-10               37971928                31.35 ns/op           28 B/op          0 allocs/op
BenchmarkHandlerServeHTTP/with_the_otelhttp_handler-10                    435142              2723 ns/op            5585 B/op         35 allocs/op
BenchmarkTransportRoundTrip/without_the_otelhttp_transport-10              11430            401735 ns/op           24436 B/op        117 allocs/op
BenchmarkTransportRoundTrip/with_the_otelhttp_transport-10                   278           4367998 ns/op            5416 B/op         85 allocs/op
PASS
ok      go.opentelemetry.io/contrib/instrumentation/net/http/otelhttp/test      10.056s
```


instrumentation/net/http/otelhttp/test/handler_test.go
```diff
@@ -528,3 +528,43 @@ func TestWithRouteTag(t *testing.T) {
 		}
 	}
 }
+
+func BenchmarkHandlerServeHTTP(b *testing.B) {
+	tp := sdktrace.NewTracerProvider()
+	mp := sdkmetric.NewMeterProvider()
+
+	r, err := http.NewRequest(http.MethodGet, "http://localhost/", nil)
```
dashpole, line 536: Is it possible to structure this so the "extra" attributes are also set?
"net.host.port", "net.sock.peer.addr", "net.sock.peer.port", "user_agent.original", "http.client_ip".

dashpole (approved):

instrumentation/net/http/otelhttp/test/handler_test.go
dmathieu: I can, but there's already a benchmark in the semconv package, which I believe would be better to test specific behavior.
https://github.com/open-telemetry/opentelemetry-go-contrib/blob/fa9fc0cf0b33211ff66d836e6636aa98b9e6f293/instrumentation/net/http/otelhttp/internal/semconv/bench_test.go#L19

dashpole: Thats fine.  I'm hoping this benchmark can be somewhat of a canonical benchmark we can use to evaluate the impact of changes to the SDKs as well.  So as long as this is reasonably representative of instrumentation, its fine

dmathieu: Yeah, but they only give a high-level overview, and a performance change could be hidden by something else.
They shouldn't replace lower-level benchmarks

instrumentation/net/http/otelhttp/test/handler_test.go
```diff
@@ -528,3 +528,43 @@ func TestWithRouteTag(t *testing.T) {
...
+		},
+		{
+			name: "with the otelhttp handler",
+			handler: otelhttp.NewHandler(http.HandlerFunc(func(w http.ResponseWriter, r *http.Request) {
+				fmt.Fprint(w, "Hello World")
+			}),
+				"test_handler",
+				otelhttp.WithTracerProvider(tp),
+				otelhttp.WithMeterProvider(mp),
+			),
```
MrAlias, line 557:
```suggestion
			handler: otelhttp.NewHandler(
				http.HandlerFunc(func(w http.ResponseWriter, r *http.Request) {
					fmt.Fprint(w, "Hello World")
				}),
				"test_handler",
				otelhttp.WithTracerProvider(tp),
				otelhttp.WithMeterProvider(mp),
			),
```

or

```suggestion
			handler: otelhttp.NewHandler(http.HandlerFunc(func(w http.ResponseWriter, r *http.Request) {
				fmt.Fprint(w, "Hello World")
			}), "test_handler", otelhttp.WithTracerProvider(tp), otelhttp.WithMeterProvider(mp)),
```

MrAlias (approved):

PR #5700, opened 2024-06-01
Fix and detect race condition in clienttrace.go

gregorynisbet-google:
ct.root can be updated without holding the mutex on ct. Also, addEvent can be called on a nil ct.root.

Fix both of these issues by moving the mutex acquisition logic so that ct.root is only touched while the mutex is held.

### DIFF
--- a/instrumentation/net/http/httptrace/otelhttptrace/clienttrace.go
+++ b/instrumentation/net/http/httptrace/otelhttptrace/clienttrace.go
@@ -186,6 +186,9 @@ func NewClientTrace(ctx context.Context, opts ...ClientTraceOption) *httptrace.C
 }
 
 func (ct *clientTracer) start(hook, spanName string, attrs ...attribute.KeyValue) {
+	ct.mtx.Lock()
+	defer ct.mtx.Unlock()
+
 	if !ct.useSpans {
 		if ct.root == nil {
 			ct.root = trace.SpanFromContext(ct.Context)
@@ -193,9 +196,6 @@ func (ct *clientTracer) start(hook, spanName string, attrs ...attribute.KeyValue
 		ct.root.AddEvent(hook+".start", trace.WithAttributes(attrs...))
 		return
 	}
-
-	ct.mtx.Lock()
-	defer ct.mtx.Unlock()
 
 	if hookCtx, found := ct.activeHooks[hook]; !found {
 		var sp trace.Span
@@ -214,6 +214,13 @@ func (ct *clientTracer) start(hook, spanName string, attrs ...attribute.KeyValue
 }
 
 func (ct *clientTracer) end(hook string, err error, attrs ...attribute.KeyValue) {
+	ct.mtx.Lock()
+	defer ct.mtx.Unlock()
+
+	if ct.root == nil {
+		return
+	}
+
 	if !ct.useSpans {
 		if err != nil {
 			attrs = append(attrs, attribute.String(hook+".error", err.Error()))
@@ -222,8 +229,6 @@ func (ct *clientTracer) end(hook string, err error, attrs ...attribute.KeyValue)
 		return
 	}
 
-	ct.mtx.Lock()
-	defer ct.mtx.Unlock()
 	if ctx, ok := ct.activeHooks[hook]; ok {
 		span := trace.SpanFromContext(ctx)
 		if err != nil {
@@ -321,6 +326,9 @@ func (ct *clientTracer) tlsHandshakeDone(_ tls.ConnectionState, err error) {
 }
 
 func (ct *clientTracer) wroteHeaderField(k string, v []string) {
+	if ct.root == nil {
+		return
+	}
 	if ct.useSpans && ct.span("http.headers") == nil {
 		ct.start("http.headers", "http.headers")
 	}
--- a/instrumentation/net/http/httptrace/otelhttptrace/clienttrace_test.go
+++ b/instrumentation/net/http/httptrace/otelhttptrace/clienttrace_test.go
@@ -8,6 +8,8 @@ import (
 	"fmt"
 	"net/http"
 	"net/http/httptrace"
+	"sync"
+	"testing"
 
 	"go.opentelemetry.io/contrib/instrumentation/net/http/otelhttp"
 )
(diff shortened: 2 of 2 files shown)


instrumentation/net/http/httptrace/otelhttptrace/constants_test.go
```diff
@@ -0,0 +1,9 @@
+// Copyright The OpenTelemetry Authors
+// SPDX-License-Identifier: Apache-2.0
+
+// +build !race
```
dmathieu, line 4: We will never run the tests with this tag. We should remove it.

instrumentation/net/http/httptrace/otelhttptrace/clienttrace_test.go
```diff
@@ -32,3 +34,38 @@ func ExampleNewClientTrace() {
...
+	for i := 1; i < workers; i++ {
+		wg.Add(1)
+		go func() {
+			resp, err := client.Get("https://example.com")
+			if err != nil {
+				fmt.Println(err)
+				return
+			}
+			resp.Body.Close()
+			fmt.Println(resp.Status)
```
dmathieu, line 65: This should either be an `assert`, or not be there (tests souldn't use `Println`).

instrumentation/net/http/httptrace/otelhttptrace/clienttrace.go
```diff
@@ -214,16 +214,20 @@ func (ct *clientTracer) start(hook, spanName string, attrs ...attribute.KeyValue
...
 
 func (ct *clientTracer) end(hook string, err error, attrs ...attribute.KeyValue) {
+	ct.mtx.Lock()
+	defer ct.mtx.Unlock()
+
 	if !ct.useSpans {
 		if err != nil {
 			attrs = append(attrs, attribute.String(hook+".error", err.Error()))
 		}
+		if ct.root == nil {
```
dmathieu, line 224: Can the conditions of this be tested?

instrumentation/net/http/httptrace/otelhttptrace/clienttrace.go
```diff
@@ -332,6 +336,9 @@ func (ct *clientTracer) wroteHeaderField(k string, v []string) {
 	if _, ok := ct.redactedHeaders[k]; ok {
 		value = "****"
 	}
+	if ct.root == nil {
```
dmathieu, line 339: Could the conditions of this be tested?

instrumentation/net/http/httptrace/otelhttptrace/clienttrace_test.go
```diff
@@ -32,3 +34,38 @@ func ExampleNewClientTrace() {
...
+			otelhttp.WithClientTrace(makeClientTrace),
+		),
+	}
+
+	var wg sync.WaitGroup
+
+	for i := 1; i < workers; i++ {
+		wg.Add(1)
+		go func() {
+			resp, err := client.Get("https://example.com")
```
dmathieu, line 59: This shouldn use otelhttptest, not make an external network call.

gregorynisbet-google: Done

instrumentation/net/http/httptrace/otelhttptrace/clienttrace.go
gregorynisbet-google: I'm not very familiar with this codebase, I just tried to fix a race condition that I could repro locally.

What should I test instead of checking whether `ct.root` is nil?

In my latest revision, I moved the test for `ct.root` being nil to the top of the function and am successfully exiting early if it is nil.

I'm not sure that's right. I'm operating under the assumption that most functions that call events on `ct.root` cannot do useful work until `ct.root` is set and should just move on.

instrumentation/net/http/httptrace/otelhttptrace/clienttrace_test.go
gregorynisbet-google: Done.

instrumentation/net/http/httptrace/otelhttptrace/constants_test.go
gregorynisbet-google: Done.

gregorynisbet-google:
Friendly ping.

instrumentation/net/http/httptrace/otelhttptrace/clienttrace_test.go
dmathieu: This still won't work without an active network / if example.com is unavailable.

instrumentation/net/http/httptrace/otelhttptrace/clienttrace_test.go
```diff
@@ -32,3 +34,45 @@ func ExampleNewClientTrace() {
...
+
+func (zeroTripper) RoundTrip(_ *http.Request) (*http.Response, error) {
+	return &http.Response{StatusCode: 200}, nil
+}
+
+var _ http.RoundTripper = zeroTripper{}
+
+// TestNewClientParallelismWithoutSubspans tests running many Gets on a client simultaneously,
+// which would trigger a race condition if root were not protected by a mutex.
+func TestNewClientParallelismWithoutSubspans(t *testing.T) {
```
MrAlias, line 48: This test passes when run against main:

```
$ go test -race -count=20 .
ok  	go.opentelemetry.io/contrib/instrumentation/net/http/httptrace/otelhttptrace	5.443s
```

It does not seem to be testing the fix being applied here.

instrumentation/net/http/httptrace/otelhttptrace/clienttrace_test.go
```diff
@@ -32,3 +34,45 @@ func ExampleNewClientTrace() {
...
+	client := http.Client{
+		Transport: otelhttp.NewTransport(
+			zeroTripper{},
+			otelhttp.WithClientTrace(makeClientTrace),
+		),
+	}
+
+	var wg sync.WaitGroup
+
+	for i := 1; i < 10000; i++ {
```
MrAlias, line 64: This is a lot of goroutines being spawned here. What's the reasoning behind this? It is making this test quite resource intensive.

instrumentation/net/http/httptrace/otelhttptrace/clienttrace.go
```diff
@@ -214,6 +214,13 @@ func (ct *clientTracer) start(hook, spanName string, attrs ...attribute.KeyValue
 }
 
 func (ct *clientTracer) end(hook string, err error, attrs ...attribute.KeyValue) {
+	ct.mtx.Lock()
+	defer ct.mtx.Unlock()
+
+	if ct.root == nil {
+		return
+	}
```
MrAlias, line 222: It looks like this is what is breaking https://github.com/open-telemetry/opentelemetry-go-contrib/blob/b0dce52c2d8fe6b347a4e4827889450635083902/instrumentation/net/http/httptrace/otelhttptrace/test/clienttrace_test.go#L239-L251

https://github.com/open-telemetry/opentelemetry-go-contrib/actions/runs/9620444944/job/26539940748?pr=5700

PR #5703, opened 2024-06-01
otelzap: Implement Reflect method

khushijain21:

### DIFF
--- a/bridges/otelzap/encoder.go
+++ b/bridges/otelzap/encoder.go
@@ -4,6 +4,7 @@
 package otelzap // import "go.opentelemetry.io/contrib/bridges/otelzap"
 
 import (
+	"encoding/json"
 	"time"
 
 	"go.uber.org/zap/zapcore"
@@ -19,7 +20,8 @@ var (
 // objectEncoder implements zapcore.ObjectEncoder.
 // It encodes given fields to OTel key-values.
 type objectEncoder struct {
-	kv []log.KeyValue
+	kv         []log.KeyValue
+	reflectval log.Value
 }
 
 // nolint:unused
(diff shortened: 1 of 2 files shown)


bridges/otelzap/encoder.go
```diff
@@ -93,11 +95,24 @@ func (m *objectEncoder) AddUint64(k string, v uint64) {
...
 
-// TODO.
+// AddReflected converts all non-primitive type to JSON string.
 func (m *objectEncoder) AddReflected(k string, v interface{}) error {
+	// TODO use json Encoder from a pool.
+	enc := json.NewEncoder(m)
+	if err := enc.Encode(v); err != nil {
+		return err
+	}
+	m.AddString(k, m.reflectval.AsString())
```
pellared, line 105: Shouldn't we do something like here: https://github.com/open-telemetry/opentelemetry-go-contrib/blob/b9072109ea81db7e384917c9dd4528623e44e09b/bridges/otellogrus/hook.go#L186-L237?

khushijain21: actually `zap` does this conversion internally. It falls back to reflect only if necessary. Doing this mapping again would be overkill right? 

https://github.com/uber-go/zap/blob/273f9ea8c8870d5f468f33f5026cd2e7ed4bd061/field.go#L473

pellared: Nice.

Still, wouldn't this part would be still helpful?

https://github.com/open-telemetry/opentelemetry-go-contrib/blob/b9072109ea81db7e384917c9dd4528623e44e09b/bridges/otellogrus/hook.go#L202-L231

bridges/otelzap/encoder_test.go
```diff
@@ -65,6 +65,13 @@ func TestObjectEncoder(t *testing.T) {
 			},
 			expected: []interface{}{wantTurducken, wantTurducken},
 		},
+		{
+			desc: "AddReflected",
+			f: func(e zapcore.ObjectEncoder) {
+				assert.NoError(t, e.AddReflected("k", map[string]interface{}{"foo": 5}), "Expected AddReflected to succeed.")
+			},
+			expected: "{\"foo\":5}\n",
```
pellared, line 73: It would be great if this would pass:

```suggestion
			expected: map[string]interface{}{"foo": 5},
```

(similar comment bellow)

pellared (approved):
I am fine approving even without full code coverage.

PR #5704, opened 2024-06-01
Set explicit boundaries for internal otlp grpc/http histogram metrics

jangaraj:
I found that implicit boundaries of internal grpc metrics are not very useful. For example I was debugging issues with huge RPC requests (>4MB which is default limit on the collector), but histogram max bucket size is only `9.77kiB - +Inf`, so it is not visible how many huge requests I received:
image
It will be nice to have better explicit boundaries, so I set them. It is opinionated configuration, so someone should review boundaries. I think we need to provide support for huge requests (e.g. max 16MB) and long timeouts (e.g. max 100sec).

### DIFF
--- a/instrumentation/google.golang.org/grpc/otelgrpc/config.go
+++ b/instrumentation/google.golang.org/grpc/otelgrpc/config.go
@@ -86,7 +86,8 @@ func newConfig(opts []Option, role string) *config {
 	var err error
 	c.rpcDuration, err = c.meter.Float64Histogram("rpc."+role+".duration",
 		metric.WithDescription("Measures the duration of inbound RPC."),
-		metric.WithUnit("ms"))
+		metric.WithUnit("ms"),
+		metric.WithExplicitBoundaries([]float64{0, 5, 25, 50, 75, 100, 250, 500, 750, 1000, 2500, 5000, 7500, 10000, 25000, 50000, 75000, 100000}...))
 	if err != nil {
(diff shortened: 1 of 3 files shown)


dmathieu (changes requested):
The specification sets specific explicit boundaries that are defaults:
https://github.com/open-telemetry/opentelemetry-specification/blob/31e3af555156f0b0d6317e6211c4e630ec584066/specification/metrics/sdk.md#explicit-bucket-histogram-aggregation

You should override this with a view. But I fear we can't change the default values without a specification change (and that could also be seen as a breaking change).

dashpole:
I believe the [RPC conventions](https://github.com/open-telemetry/semantic-conventions/blob/6a0a75e73e0caed8c5f9bbb6cff34a752dbda310/docs/rpc/rpc-metrics.md) could choose to add different explicit bucket boundaries, similar to the [HTTP conventions](https://github.com/open-telemetry/semantic-conventions/blob/6a0a75e73e0caed8c5f9bbb6cff34a752dbda310/docs/http/http-metrics.md#metric-httpserverrequestduration).  That would be a breaking change only for the RPC conventions, which are experimental.  Regardless, you should open an issue in the semantic conventions repo: https://github.com/open-telemetry/semantic-conventions if you want to change the default bucket boundaries

dmathieu:
Closing as invalid.

PR #5707, opened 2024-06-03
otelzap: Allow context injection via fields

khushijain21:


bridges/otelzap/core.go
```diff
@@ -155,17 +158,21 @@ func (o *Core) Write(ent zapcore.Entry, fields []zapcore.Field) error {
...
+		r.AddAttributes(convertField(&o.ctx, fields)...)
 	}
 
-	o.logger.Emit(context.Background(), r)
+	o.logger.Emit(o.ctx, r)
 	return nil
 }
 
-func convertField(fields []zapcore.Field) []log.KeyValue {
+func convertField(ctx *context.Context, fields []zapcore.Field) []log.KeyValue {
```
pellared, line 168: Can we return the context instead?

```suggestion
func convertField(fields []zapcore.Field) (context.Context, []log.KeyValue) {
```

pellared (commented):
We should have at least 2 tests where the context is passed using `.With` or directly as a field.
These could be added as test cases to `TestCore` and with names like `WriteContext` and `WithContext`.

@dmathieu, I see that it will be problematic as the Recorder in https://pkg.go.dev/go.opentelemetry.io/otel/log/logtest does not capture the context. Are you able to modify https://pkg.go.dev/go.opentelemetry.io/otel/log/logtest#ScopeRecords to allow it?

bridges/otelzap/core.go
```diff
@@ -152,31 +154,32 @@ func (o *Core) Write(ent zapcore.Entry, fields []zapcore.Field) error {
...
+		o.ctx, attrbuf = convertField(o.ctx, fields)
+		r.AddAttributes(attrbuf...)
 	}
 
 	o.logger.Emit(o.ctx, r)
 	return nil
 }
 
-func convertField(ctx *context.Context, fields []zapcore.Field) []log.KeyValue {
+func convertField(ctx context.Context, fields []zapcore.Field) (context.Context, []log.KeyValue) {
```
pellared, line 171: Do we need to pass `ctx context.Context`?

bridges/otelzap/core.go
```diff
@@ -152,31 +154,32 @@ func (o *Core) Write(ent zapcore.Entry, fields []zapcore.Field) error {
...
 	// TODO: Handle attributes passed via With (exceptions: context.Context and zap.Namespace).
-	// TODO: Handle context.Context containing trace context.
 	// TODO: Handle zap.Namespace.
 	// TODO: Handle ent.LoggerName.
 
 	r.AddAttributes(o.attr...)
 	if len(fields) > 0 {
-		r.AddAttributes(convertField(&o.ctx, fields)...)
+		var attrbuf []log.KeyValue
+		o.ctx, attrbuf = convertField(o.ctx, fields)
```
pellared, line 163: should it not be something like the following?

```suggestion
		ctx, attrbuf := convertField(fields)
		if ctx != nil { 
			o.ctx = ctx
		}
```

bridges/otelzap/core.go
```diff
@@ -116,7 +116,9 @@ func (o *Core) Enabled(level zapcore.Level) bool {
 func (o *Core) With(fields []zapcore.Field) zapcore.Core {
 	cloned := o.clone()
 	if len(fields) > 0 {
-		cloned.attr = append(cloned.attr, convertField(&cloned.ctx, fields)...)
+		var attrbuf []log.KeyValue
+		cloned.ctx, attrbuf = convertField(cloned.ctx, fields)
+		cloned.attr = append(cloned.attr, attrbuf...)
```
pellared, line 121: should it not be something like the following?


```suggestion
		ctx, attrbuf := convertField(fields)
		if ctx != nil { 
			cloned.ctx = ctx
		}
		cloned.attr = append(cloned.attr, attrbuf...)
```

bridges/otelzap/core_test.go
```diff
@@ -44,6 +44,9 @@ func TestCore(t *testing.T) {
 
 	rec.Reset()
 
+	// TODO WriteContext
+	// TOOD WithContext
```
pellared, line 48:
```suggestion
	// TODO: Add WriteContext test case
	// TOOD: Add WithContext test case
```

bridges/otelzap/core.go
khushijain21: oh yes you are right, we don't need to pass context

pellared (approved):

pellared (approved):

dmathieu (approved):

PR #5721, opened 2024-06-04
otelzap: Implement namespace method

khushijain21:

The approach is to use linked list to open a namespace and keep track of it.
`calculate` method iterates to the most nested namespace and appends data in reverse order.


khushijain21:
The approach is to use linked list to open a namespace and keep track of it.

`getObjValue` method iterates to the most nested namespace and appends data in reverse order.

pellared:
`build_and_test / lint (pull_request)` is failing. Can you fix it before I make a review?

bridges/otelzap/encoder.go
```diff
@@ -18,47 +18,69 @@ var (
 	_ zapcore.ArrayEncoder  = (*arrayEncoder)(nil)
 )
 
+type newNameSpace struct {
+	ns   string
+	kv   []log.KeyValue
```
pellared, line 23: Can we rename it to "attrs" (for attributes)?

```suggestion
	attrs []log.KeyValue
```

bridges/otelzap/encoder.go
```diff
@@ -18,47 +18,69 @@ var (
...
 	}
+	return &objectEncoder{
+		root: m,
+		cur:  m,
+	}
+}
+
+// It iterates to the end of the linked list and appends namespace data.
+// Run this function before accessing complete result.
+func (m *objectEncoder) getEncodedResult(o *newNameSpace) {
```
pellared, line 49: Can we rename it to something like `calculate` or `aggregate`? The current name suggests that it returns something and that it is a pure function (accessor)

pellared (approved):

dmathieu (approved):

PR #5732, opened 2024-06-05
otelgin: Using `c.FullPath()` to set `http.route` attribute

NeoCN:
As `spanName` can be custumized with `SpanNameFormatter`, so the `spanName` may not be the same with `http.route`, e.g. the `spanName` can be `GET /users/:id`, but the `http.route` is `/users/:id`, so, using  [`c.FullPath()`](https://github.com/gin-gonic/gin/blob/v1.10.0/context.go#L165-L173) to set `http.route` attribute in `otelgin`

### DIFF
--- a/instrumentation/github.com/gin-gonic/gin/otelgin/option.go
+++ b/instrumentation/github.com/gin-gonic/gin/otelgin/option.go
@@ -23,8 +23,8 @@ type config struct {
 // be traced. A Filter must return true if the request should be traced.
 type Filter func(*http.Request) bool
 
-// SpanNameFormatter is used to set span name by http.request.
-type SpanNameFormatter func(r *http.Request) string
+// SpanNameFormatter is used to set span name by http.Request.
+type SpanNameFormatter func(routeName string, r *http.Request) string
 
 // Option specifies instrumentation configuration options.
 type Option interface {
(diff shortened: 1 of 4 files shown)


dmathieu:
Could you add a test, as well as a changelog entry?

NeoCN:
@dmathieu Done.

instrumentation/github.com/gin-gonic/gin/otelgin/option.go
```diff
@@ -23,8 +23,8 @@ type config struct {
 // be traced. A Filter must return true if the request should be traced.
 type Filter func(*http.Request) bool
 
-// SpanNameFormatter is used to set span name by http.request.
-type SpanNameFormatter func(r *http.Request) string
+// SpanNameFormatter is used to set span name by http.Request.
+type SpanNameFormatter func(routeName string, r *http.Request) string
```
dmathieu, line 27: This changes the public API, and isn't what the PR describes. It should be its own PR.

NeoCN: OK, I will create two PR for those changes.

NeoCN: https://github.com/open-telemetry/opentelemetry-go-contrib/pull/5734 created for `http.route` change. @dmathieu

Issue #5655, opened 2024-05-23
Update runtime instrumentation to new semantic conventions


MrAlias:
https://github.com/open-telemetry/semantic-conventions/pull/981

dashpole:
Do we want a migration plan?  Or do we want to just want to switch to the new conventions?

MrAlias:
+1 to just switching.

MrAlias:
Migration plan:

- Add the new semantic conventions, these will be used by default
- Add option to turn on old metrics (in addition to the new which are default)

The old metrics will be supported for 2 releases. We do not plan to make any changes to the old metrics. If there is any CVEs this they will be removed prior to the 2 release timeline.

dashpole:
I just need to implement the histogram metric now.

dashpole:
Some notes from my rough attempt to implement the histogram metric:

* It has a lot of buckets (160?).  See the source: https://github.com/golang/go/blob/355711821eea51c6456a31ab61d0dc2e9db034f7/src/runtime/histogram.go#L71
* It seems like the buckets may map to our exponential histogram.  We might want to consider using that instead of a fixed-bucket histogram.
* It doesn't have a Sum.  I'm not sure if we can reasonably GA it without a sum.  We may need to ask the Go team to add the Sum.

MrAlias:
> * It seems like the buckets may map to our exponential histogram.  We might want to consider using that instead of a fixed-bucket histogram.

["buckets are power-of-2 sized"](https://github.com/golang/go/blob/355711821eea51c6456a31ab61d0dc2e9db034f7/src/runtime/histogram.go#L16) seems to motivate this :+1:

dashpole:
> ["buckets are power-of-2 sized"](https://github.com/golang/go/blob/355711821eea51c6456a31ab61d0dc2e9db034f7/src/runtime/histogram.go#L16) seems to motivate this 👍

It is powers-of-2 sized, but with linear "sub-buckets" between the power-of-two buckets.  I asked, and the buckets are also subject to change over time, so using an exponential histogram seems like not the best idea.  Slightly disappointing, but maybe an opportunity to improve in the future.

dashpole:
I'm going to track adding the histogram metrics separately: https://github.com/open-telemetry/opentelemetry-go-contrib/issues/5974.

MrAlias:
[...]

PR #5747, opened 2024-06-09
Prepare for migration to new runtime metrics

dashpole:
Part of #5655

This is a refactoring to prepare for the implementation of the new runtime metrics.  It:

* Adds support for `OTEL_GO_X_DEPRECATED_RUNTIME_METRICS`, which can be set to `true` or `false` to enable/disable the existing runtime metrics. It initially defaults to `true` while the new metrics are under development.
* Moves the existing runtime metrics implementation to its own internal package, deprecatedruntime, to clearly separate it from the new metrics being added, and to make the eventual removal easier.

This does not change any of the metrics generated, or the public API surface of the runtime metrics package.


instrumentation/runtime/internal/x/x.go
```diff
@@ -0,0 +1,53 @@
...
+// conventions, it is enabled by default.
+//
+// To disable this feature set the OTEL_GO_X_DEPRECATED_RUNTIME_METRICS environment variable
+// to the case-insensitive string value of "false" (i.e. "False" and "FALSE"
+// will also enable this).
+var DeprecatedRuntimeMetrics = newFeature("DEPRECATED_RUNTIME_METRICS", true)
+
+// BoolFeature is an experimental feature control flag. It provides a uniform way
+// to interact with these feature flags and parse their values.
+type BoolFeature struct {
```
dashpole, line 26: This differs from the "Feature" pattern used elsewhere because I wanted to be able to have a default value of "true".  Rather than make the existing pattern more complex to support this, I simplified it to fit this packages' needs.

instrumentation/runtime/internal/x/README.md
```diff
@@ -0,0 +1,38 @@
...
+export OTEL_GO_X_DEPRECATED_RUNTIME_METRICS=true
+```
+
+After two additional releases, the deprecated runtime metrics will be removed,
+and setting the environment variable will no longer have any effect.
+
+The value set must be the case-insensitive string of `"true"` to enable the
+feature, and `"false"` to disable the feature. All other values are ignored.
+
+[previous runtime metrics conventions]: go.opentelemetry.io/contrib/instrumentation/runtime/internal/deprecatedruntime
```
dashpole, line 37: This link will not work until after the next release.

MadVikingGod (approved):
This looks good to me.

Are we going to keep the same scope for the deprecated metrics, and will it be the same as the scope of the new metrics?

dashpole:
I was going to keep the same scope for the deprecated metrics.  It doesn't make sense to me to modify the scope if they are going away anyways.

dmathieu (approved):

MrAlias:
Looks like this introduced a broken link. I'll look into submitting a fix.

PR #5780, opened 2024-06-18
Implement new runtime metrics

dashpole:
Part of https://github.com/open-telemetry/opentelemetry-go-contrib/issues/5655

Changes:

* Move the configuration options to `options.go` without modification.
* Implements the metrics defined here: https://github.com/open-telemetry/semantic-conventions/blob/main/docs/runtime/go-metrics.md.  These are still disabled by default while it is under development.
* Add unit tests

Notes:

It doesn't implement `go.schedule.duration`, as the histogram will need some additional work to figure out.

Based on https://github.com/prometheus/client_golang/pull/955#issue-1090830520, using go's runtime metrics should is more efficient than reading memstats.


CHANGELOG.md
```diff
@@ -15,6 +15,7 @@ This project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0.htm
 - Add support to configure views when creating MeterProvider using the config package. (#5654)
 - Add log support for the autoexport package. (#5733)
 - Add support for disabling the old runtime metrics using the `OTEL_GO_X_DEPRECATED_RUNTIME_METRICS=false` environment variable. (#5747)
+- Add new runtime metrics to `go.opentelemetry.io/contrib/instrumentation/runtime`, which are still disabled by default. (#5870)
```
```suggestion
- Add new runtime metrics to `go.opentelemetry.io/contrib/instrumentation/runtime`, which are still disabled by default. (#5780)
```

dmathieu (approved):

MrAlias (approved):

MrAlias:
Waiting to merge so @XSAM can review.

XSAM (changes requested):
Should we add a test for `newCollector` or `refresh` method to ensure the minimum refresh interval can actually work via the collector created by `newCollector`?

instrumentation/runtime/runtime.go
```diff
@@ -97,6 +50,166 @@ func Start(opts ...Option) error {
...
+}
+
+func newCollector(minimumInterval time.Duration) *goCollector {
+	g := &goCollector{
+		sampleBuffer: make([]metrics.Sample, 0, len(runtimeMetrics)),
+		sampleMap:    make(map[string]*metrics.Sample, len(runtimeMetrics)),
+	}
+	for _, runtimeMetric := range runtimeMetrics {
+		g.sampleBuffer = append(g.sampleBuffer, metrics.Sample{Name: runtimeMetric})
+		g.sampleMap[runtimeMetric] = &g.sampleBuffer[len(g.sampleBuffer)-1]
```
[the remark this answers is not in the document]
dashpole: This broke the implementation because we need to reference the position in the slice, not the underlying Sample.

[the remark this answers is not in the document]
dashpole: I've reverted it

dashpole:
Working on tests.  metrics.Read is being strange during unit testing, and seems to always return the same result.

instrumentation/runtime/runtime.go
```diff
@@ -97,6 +50,167 @@ func Start(opts ...Option) error {
...
+
+func newCollector(minimumInterval time.Duration) *goCollector {
+	g := &goCollector{
+		sampleBuffer:    make([]metrics.Sample, 0, len(runtimeMetrics)),
+		sampleMap:       make(map[string]*metrics.Sample, len(runtimeMetrics)),
+		minimumInterval: minimumInterval,
+	}
+	for _, runtimeMetric := range runtimeMetrics {
+		g.sampleBuffer = append(g.sampleBuffer, metrics.Sample{Name: runtimeMetric})
+		g.sampleMap[runtimeMetric] = &g.sampleBuffer[len(g.sampleBuffer)-1]
```
XSAM, line 195:
```suggestion
		// We are referencing the position of the slice. 
		// Any growth to this `sampleBuffer` slice would break the implementation.
		g.sampleBuffer = append(g.sampleBuffer, metrics.Sample{Name: runtimeMetric})
		g.sampleMap[runtimeMetric] = &g.sampleBuffer[len(g.sampleBuffer)-1]
```

Should we add a comment here to declare that we are referencing the position of the slice and that any growth to this `sampleBuffer` slice would break the implementation?

dashpole: added a similar disclaimer.

dashpole:
I had to add code that allocates memory to ensure the metric actually changes for the refresh test.  I ran it locally with `-count 10000`, and it passes, so hopefully it won't be flaky.

dashpole:
`/gc/heap/allocs:bytes` and `/gc/heap/allocs:objects` don't seem to be populated on windows.  I'm investigating

dashpole:
I think I got the test passing

instrumentation/runtime/test/runtime_test.go
```diff
@@ -0,0 +1,161 @@
...
+			t.Fatalf("unexpected data type %v", a)
+		}
+	}
+}
+
+func allocateMemory(buffer [][]byte) {
+	newBuffer := make([]byte, 1000000)
+	for i := range newBuffer {
+		newBuffer[i] = 0
+	}
```
MrAlias, line 159: Go does this by default for all slices created with `make`.

```suggestion
```

instrumentation/runtime/test/runtime_test.go
```diff
@@ -0,0 +1,161 @@
...
+	buffer = append(buffer, newBuffer)
+}
```
MrAlias, line 161: The passed slice is never used again. The GC will collect the allocated space right after this returns, right?

Maybe pass a pointer, return the slice, or just replace this call with `buffer append(buffer, make([]byte, 1000000))`

instrumentation/runtime/test/runtime_test.go
dashpole: removed.

instrumentation/runtime/test/runtime_test.go
dashpole: returned the slice.

MrAlias (approved):

MrAlias:
Looks like `go mod tidy` needs to be run.

MrAlias:
@XSAM PTAL.

MrAlias:
@XSAM planning to merge this in an hour or two unless I hear from you otherwise.

XSAM (approved):

PR #5789, opened 2024-06-20
Remove pdata dependency and use proto-go instead

dmathieu:
It's a bit weird that this autoexport package has a dependency on the collector (well, pdata).
I believe it makes more sense to use [proto-go](https://github.com/open-telemetry/opentelemetry-proto-go).

pdata also has a rather high release cadence, even though it isn't modified very frequently. So this change should slightly reduce the number of renovabot PRs.


dmathieu:
Possibly, we could also move those tests into a `test` submodule, so there is no indirect dependency on grpc for folks using this.

MrAlias (approved):

dashpole (approved):

PR #5718, opened 2024-06-03
xray client: return an error if the HTTP request failed

jaedle:

PR #5795, opened 2024-06-23
Fix: Silently failing of sampling rules

### LINKED ISSUE #5717: Obtaining sampling rules may silently fail
### Description

The X-Ray remote sampler may silently fail obtaining the sampling rules from the [awsproxy](https://aws-otel.github.io/docs/getting-started/remote-sampling).

We are running the OpenTelemetry Collector behind a reverse proxy as we have about a dozen services ingesting traces which are forwarded to AWS X-Ray. As this adds forward headers (i.e. x-forwarded-*), the request signing of the awsproxy was not working correctly. 

Even though the response of AWS results in a client error, there are no signs of an error. The JSON response body is unmarshaled and treated as empty ruleset.

The missing check on status code of the [response](https://github.com/open-telemetry/opentelemetry-go-contrib/blob/b9072109ea81db7e384917c9dd4528623e44e09b/samplers/aws/xray/internal/client.go#L136) should be the problem in this case. 

The only response code for a successful request should be [200/OK](https://docs.aws.amazon.com/xray/latest/api/API_GetSamplingRules.html) according the documentation.

### Environment

Should not be relevant for the error.

### Steps To Reproduce

The simplest way to reproduce may be a test case for [`client_test.go`](https://github.com/open-telemetry/opentelemetry-go-contrib/blob/b9072109ea81db7e384917c9dd4528623e44e09b/samplers/aws/xray/internal/client_test.go#L1):

```go
func TestRespondsWithErrorStatusCode(t *testing.T) {
	testServer := httptest.NewServer(http.HandlerFunc(func(res http.ResponseWriter, _ *http.Request) {
		res.WriteHeader(http.StatusForbidden)
		_, err := res.Write([]byte(`{}`))
		require.NoError(t, err)
	}))
	t.Cleanup(testServer.Close)

	u, err := url.Parse(testServer.URL)
	require.NoError(t, err)

	client, err := newClient(*u)
	require.NoError(t, err)

	samplingRules, err := client.getSamplingRules(context.Background())
	require.Error(t, err)
	require.Nil(t, samplingRules)
}
```

### Current behavior

```gherkin
GIVEN sampling rules should be obtained from the xray proxy
WHEN the response code incidcates an error 4xx/5xx
THEN an empty manifest is created
AND no error is indicated
```

### Expected behavior

```gherkin
GIVEN sampling rules should be obtained from the xray proxy
WHEN the response code incidcates an error 4xx/5xx
THEN an error should occur on loading the rulset
AND no manifest should be present 
```

I would love to provide a PR to fix that issue.

ayushrakesh:
## Description 
 `getSamplingRules` method in the AWS X-Ray sampler did not check the HTTP response status code, potentially leading to incorrect handling of non-200 responses.

- Added a check for the HTTP response status code in the getSamplingRules method.
- If the status code is not 200, the method now returns an error instead of proceeding to unmarshal the response body.
- Included test for this status code check.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -22,7 +22,7 @@ This project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0.htm
 - Custom attributes targeting metrics recorded by the `go.opentelemetry.io/contrib/instrumentation/net/http/otelhttp` are not ignored anymore. (#5129)
 - Use `c.FullPath()` method to set `http.route` attribute in `go.opentelemetry.io/contrib/instrumentation/github.com/gin-gonic/gin/otelgin`. (#5734)
 - The double setup in `go.opentelemetry.io/contrib/instrumentation/net/http/httptrace/otelhttptrace/example` that caused duplicate traces. (#5564)
-- Added status code check in `getSamplingRules` method to ensure only successful responses with status 200 are processed in `go.opentelemetry.io/contrib/samplers/aws/xray/internal/client.go`.
+- Added status code check in `getSamplingRules` method to ensure only successful responses with status 200 are processed in `go.opentelemetry.io/contrib/samplers/aws/xray/internal/client.go`. (#5795)
 
 
 ### Deprecated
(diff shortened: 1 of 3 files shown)


dmathieu:
Duplicate of https://github.com/open-telemetry/opentelemetry-go-contrib/pull/5718

PR #5804, opened 2024-06-26
Add errorlint linter

dmathieu:
This is the contrib counterpart of https://github.com/open-telemetry/opentelemetry-go/pull/5535, adding the errorlint linter.


instrumentation/google.golang.org/grpc/otelgrpc/example/server/main.go
```diff
@@ -88,7 +89,7 @@ func (s *server) SayHelloBidiStream(stream api.HelloService_SayHelloBidiStreamSe
 	for {
 		in, err := stream.Recv()
 
-		if err == io.EOF {
+		if errors.Is(err, io.EOF) {
```
pellared, line 92: Have you checked the performance overhead? While I have nothing against using `errors.Is` and `errors.As` in business logic I am a uncertain if we to use it in instrumentation libraires. As far as I remember it uses some kind of reflection (`internal/reflectlight`).

dmathieu: Related: https://github.com/open-telemetry/opentelemetry-go/issues/5551

pellared: Does not look like an issue.

pellared (approved):

MrAlias (approved):

PR #5817, opened 2024-06-27
Add the minsev package

MrAlias:
Resolve https://github.com/open-telemetry/opentelemetry-go/issues/5078

Add the experimental `minsev` module. This module provides a log processor that thresholds all records recorded to being above a minimum severity.

### Benchmarks

```
goos: linux
goarch: amd64
pkg: go.opentelemetry.io/contrib/processors/minsev
cpu: Intel(R) Core(TM) i7-8550U CPU @ 1.80GHz
                        │   new.txt   │
                        │   sec/op    │
LogProcessor/Base-8       20.53n ± 2%
LogProcessor/Enabled-8    41.27n ± 5%
LogProcessor/Disabled-8   22.34n ± 8%
geomean                   26.65n

                        │   new.txt    │
                        │     B/op     │
LogProcessor/Base-8       0.000 ± 0%
LogProcessor/Enabled-8    0.000 ± 0%
LogProcessor/Disabled-8   0.000 ± 0%
geomean                              ¹
¹ summaries must be >0 to compute geomean

                        │   new.txt    │
                        │  allocs/op   │
LogProcessor/Base-8       0.000 ± 0%
LogProcessor/Enabled-8    0.000 ± 0%
LogProcessor/Disabled-8   0.000 ± 0%
geomean                              ¹
¹ summaries must be >0 to compute geomean
```


processors/minsev/minsev.go
```diff
@@ -0,0 +1,68 @@
...
+}
+
+var defaultProcessor = noopProcessor{}
+
+type noopProcessor struct{}
+
+func (p noopProcessor) OnEmit(context.Context, log.Record) error { return nil }
+func (p noopProcessor) Enabled(context.Context, log.Record) bool { return false }
+func (p noopProcessor) Shutdown(context.Context) error           { return nil }
+func (p noopProcessor) ForceFlush(context.Context) error         { return nil }
```
pellared, line 68: Maybe we could provide a No-Op processor as part of SDK or Contrib?

processors/minsev/minsev_test.go
```diff
@@ -0,0 +1,191 @@
...
+	p.ForceFlushCalls = p.ForceFlushCalls[:0]
+}
+
+func TestLogProcessorOnEmit(t *testing.T) {
+	t.Run("Passthrough", func(t *testing.T) {
+		wrapped := &processor{ReturnErr: assert.AnError}
+
+		p := NewLogProcessor(wrapped, api.SeverityTrace1)
+		ctx := context.Background()
+		r := &log.Record{}
```
pellared, line 72: If we would move this inside the `for` loop here and in other table-driven-tests then we would not need to use (and define) `wrapped.Reset()`, right?

pellared (approved):

processors/minsev/minsev_test.go
MrAlias: No, the `OnEmitCalls` slice would still be appended to and need to be reset.

processors/minsev/minsev.go
MrAlias: I think the only use for a no-op processor is for testing. If we are going to add it somewhere I expect [`logtest`](https://pkg.go.dev/go.opentelemetry.io/otel/sdk/log@v0.3.0/logtest) would be where we want to add it. However, I am reluctant to add something there until there is another example use.

dmathieu (approved):

PR #5818, opened 2024-06-27
HTTP Semconv migration Part1 Server Metrics - v1.20.0 support

MadVikingGod:
This change moves the metric creation and use into the semconv package.  This is because metrics names are defined by the semantic convention, so to be able to change them seamlessly they should be within the scope of the semconv package.


instrumentation/net/http/otelhttp/internal/semconv/env.go
```diff
@@ -80,3 +82,40 @@ func ServerStatus(code int) (codes.Code, string) {
...
+	StatusCode           int
+	AdditionalAttributes []attribute.KeyValue
+
+	RequestSize  int64
+	ResponseSize int64
+	ElapsedTime  float64
+}
+
+func (s HTTPServer) RecordMetrics(ctx context.Context, md MetricData) {
+	if s.requestBytesCounter == nil {
```
MrAlias, line 98:
```suggestion
	if s.requestBytesCounter == nil || s.responseBytesCounter == nil || s.serverLatencyMeasure == nil {
```

instrumentation/net/http/otelhttp/internal/semconv/env.go
```diff
@@ -80,3 +82,40 @@ func ServerStatus(code int) (codes.Code, string) {
...
+		// This will happen if an HTTPServer{} is used insted of NewHTTPServer.
+		return
+	}
+
+	attributes := oldHTTPServer{}.MetricAttributes(md.ServerName, md.Req, md.StatusCode, md.AdditionalAttributes)
+	o := metric.WithAttributeSet(attribute.NewSet(attributes...))
+	addOpts := []metric.AddOption{o} // Allocate vararg slice once.
```
MrAlias, line 105: Can we track that these should be pooled?

MrAlias (approved):

dashpole (approved):

instrumentation/net/http/otelhttp/internal/semconv/env.go
MadVikingGod: I made #5968 to track this

VinozzZ (approved):

PR #5842, opened 2024-07-02
Remove redundant otelhttp allocation

gaiaz-iusipov:
There is no need to wrap `MeasurementOption` to a slice


dmathieu (approved):

ash2k:
@gaiaz-iusipov @MrAlias @dmathieu May I ask why the change, approved in https://github.com/open-telemetry/opentelemetry-go-contrib/pull/5664, was reverted? Vararg slice was allocated once and used 2/multiple times. Now it is allocated on each `Add()` call. There is/was even a comment, explaining the "why".

See https://go.dev/ref/spec#Passing_arguments_to_..._parameters

See https://fs.blog/chestertons-fence/

```go
var sink []string

func abc(vals ...string) {
	sink = vals
}

func BenchmarkFuncCall(b *testing.B) {
	b.ReportAllocs()
	b.ResetTimer()
	arg := []string{"a", "b", "c"}
	for i := 0; i < b.N; i++ {
		abc(arg...)
	}
}
```
gives me
```
BenchmarkFuncCall-10    	1000000000	         0.9481 ns/op	       0 B/op	       0 allocs/op
```

```go
var sink []string

func abc(vals ...string) {
	sink = vals
}

func BenchmarkFuncCall(b *testing.B) {
	b.ReportAllocs()
	b.ResetTimer()
	for i := 0; i < b.N; i++ {
		abc("a", "b", "c")
	}
}
```
gives me

```
BenchmarkFuncCall-10    	42442630	        26.69 ns/op	      48 B/op	       1 allocs/op
```

Put the fence back, please 😁

PR #5861, opened 2024-07-05
[WIP] Add isolating log processor

pellared:


processors/isolate/processor.go
```diff
@@ -0,0 +1,58 @@
...
+	return p.Processor.OnEmit(ctx, record)
+}
+
+// Enabled clones the record and calls the wrapped downstream processor.
+func (p *LogProcessor) Enabled(ctx context.Context, record log.Record) bool {
+	record = record.Clone()
+	return p.Processor.Enabled(ctx, record)
+}
+
+var defaultProcessor = noopProcessor{}
```
pellared, line 51: Here is another usage of an no-op processor in a processor implementation.

I think we can think about adding it the the SDK.

See also: https://github.com/open-telemetry/opentelemetry-go-contrib/pull/5817#discussion_r1658924196

dmathieu (approved):
There should be a codeowners entry.

MrAlias (commented):
It's not clear to me this is needed currently. We still do not comply with specification regarding our log record type begin passed to the log processor: https://github.com/open-telemetry/opentelemetry-go/issues/5219. The resolution of that may make this type of log processor redundant and unneeded.

I would prefer we resolve prior to adding this kind of extension.

pellared:
> It's not clear to me this is needed currently. We still do not comply with specification regarding our log record type begin passed to the log processor: [open-telemetry/opentelemetry-go#5219](https://github.com/open-telemetry/opentelemetry-go/issues/5219). The resolution of that may make this type of log processor redundant and unneeded.
> 
> I would prefer we resolve prior to adding this kind of extension.

Even if the processor would to accept a fully mutable record - this processor would still be helpful. However, we can postpone this if this is your preference, then please move it to "blocked".

MrAlias:
> Even if the processor would to accept a fully mutable record - this processor would still be helpful. However, we can postpone this if this is your preference, then please move it to "blocked".

With our changes and [this](https://github.com/open-telemetry/opentelemetry-specification/pull/4067) still pending, I would prefer to not add this as it has a good potential to be cruft.

PR #5876, opened 2024-07-09
 added option on http transport for having default attributes to be i…

luca-filipponi:
…ncluded in request metric


dmathieu:
You shouldn't need this option. The same can be achieved with a view.
https://opentelemetry.io/docs/languages/go/instrumentation/#registering-views

luca-filipponi:
> You shouldn't need this option. The same can be achieved with a view.
Ohh nice, so i guess this is not useful! Thanks, will close the other issue

luca-filipponi:
@dmathieu i've checked out the documentation of the view, but I don't see an option to add attributes with that

you can only filter them?
https://github.com/open-telemetry/opentelemetry-go/blob/82184aa44bb892ad95d98e09196a7c331f17797e/sdk/metric/instrument.go#L146

dmathieu:
Urgh yes sorry, my bad.

luca-filipponi:
Created an issue: https://github.com/open-telemetry/opentelemetry-go-contrib/issues/5877, not sure if this is the best way to add this functionality (maybe there can be an option to add attributes in a the view? but it seems more invasive)

CHANGELOG.md
```diff
@@ -13,6 +13,7 @@ This project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0.htm
 - Add the new `go.opentelemetry.io/contrib/instrgen` package to provide auto-generated source code instrumentation. (#3068, #3108)
 - Support for stdoutlog exporter in `go.opentelemetry.io/contrib/config`. (#5850)
 - Add macOS ARM64 platform to the compatibility testing suite. (#5868)
+- Added option for adding default attributes to otel http transport in `go.opentelemetry.io/contrib/instrumentation/net/http/config`. (#5876)
```
```suggestion
- Added option for adding default attributes to otel http transport in `go.opentelemetry.io/contrib/instrumentation/net/http/otelhttp`. (#5876)
```

instrumentation/net/http/otelhttp/config.go
```diff
@@ -194,3 +197,11 @@ func WithServerName(server string) Option {
 		c.ServerName = server
 	})
 }
+
+// WithDefaultAttributes returns an option that sets the default attributes to be
+// included in metrics.
+func WithDefaultAttributes(attributes []attribute.KeyValue) Option {
```
dmathieu, line 203: Those attributes will only be used in metrics. Using `DefaultAttributes` as the name could be misleading, as otelhttp also does spans.
How about `WithMetricAttributes`?

dmathieu (commented):
Typically, with this kind of attributes in otelhttp and other instrumentations, the follow request we have seen is to be able to set data dynamically, depending on the request data.

I wonder if it would be worth going directly with a callback that accepts the request as a parameter?

instrumentation/net/http/otelhttp/config.go
luca-filipponi: Sure, will change it

luca-filipponi:
> I wonder if it would be worth going directly with a callback that accepts the request as a parameter?

yea, pretty sure I can do that 😄  will send changes in a bit

instrumentation/net/http/otelhttp/config.go
```diff
@@ -194,3 +198,11 @@ func WithServerName(server string) Option {
 		c.ServerName = server
 	})
 }
+
+// WithReqAttributesMapper returns an Option to set a function that maps an HTTP request to a slice of attribute.KeyValue.
+// These attributes will be included in metrics for every request.
+func WithReqAttributesMapper(reqAttributesMapper func(r *http.Request) []attribute.KeyValue) Option {
```
dmathieu, line 204: This method name is still not metric-specific, when the data is. How about `WithMetricAttributesFn`?

luca-filipponi: Yea, is more appropriate, done ✅

instrumentation/net/http/otelhttp/transport.go
```diff
@@ -27,13 +27,14 @@ import (
...
-	spanNameFormatter func(string, *http.Request) string
-	clientTrace       func(context.Context) *httptrace.ClientTrace
+	tracer              trace.Tracer
+	meter               metric.Meter
+	propagators         propagation.TextMapPropagator
+	spanStartOptions    []trace.SpanStartOption
+	filters             []Filter
+	spanNameFormatter   func(string, *http.Request) string
+	clientTrace         func(context.Context) *httptrace.ClientTrace
+	reqAttributesMapper func(*http.Request) []attribute.KeyValue
```
dmathieu, line 37:
```suggestion
	metricAttributesFn func(*http.Request) []attribute.KeyValue
```

instrumentation/net/http/otelhttp/transport.go
```diff
@@ -170,7 +172,7 @@ func (t *Transport) RoundTrip(r *http.Request) (*http.Response, error) {
 	}
 
 	// metrics
-	metricAttrs := append(labeler.Get(), semconvutil.HTTPClientRequestMetrics(r)...)
+	metricAttrs := append(append(labeler.Get(), semconvutil.HTTPClientRequestMetrics(r)...), t.attributesFromRequest(r)...)
```
dmathieu, line 175:
```suggestion
	metricAttrs := append(append(labeler.Get(), semconvutil.HTTPClientRequestMetrics(r)...), t.metricAttributesFromRequest(r)...)
```

instrumentation/net/http/otelhttp/transport.go
```diff
@@ -196,6 +198,14 @@ func (t *Transport) RoundTrip(r *http.Request) (*http.Response, error) {
 	return res, err
 }
 
+func (t *Transport) attributesFromRequest(r *http.Request) []attribute.KeyValue {
```
dmathieu, line 201:
```suggestion
func (t *Transport) metricAttributesFromRequest(r *http.Request) []attribute.KeyValue {
```

dmathieu (approved):

luca-filipponi:
@dmathieu what is the process for merging? do you want me to squash commits? is there anything else that i should do?

dmathieu:
A second approval from a member of @open-telemetry/go-approvers is required before we can merge.

dashpole:
I think this has been discussed, but I don't remember the outcome.  Can you update the description to include it?

Why can't the Labeler be used to accomplish this?

dmathieu:
It's mentioned in https://github.com/open-telemetry/opentelemetry-go-contrib/issues/5877.

dashpole:
Thanks.  Does it need to be a function, then, if the goal is to add constant labels not based on the request?  This seems to just duplicate the Labeler's functionality right now.

dmathieu:
I asked for it to be a function [here](https://github.com/open-telemetry/opentelemetry-go-contrib/pull/5876#pullrequestreview-2168250634), as every other option we've set like this before was always followed with a request to be able to dynamically add data rather than statically.
I don't mind reverting that if you feel against it though.

dashpole:
Just trying to confirm my understanding of the problem we are solving:

We _can_ add labels to the http metrics with the Labeler, but we are trying to make it simpler?  Should this replace the Labeler then?

luca-filipponi:
This is the use case:

* I have 2 http clients, one calling `service-a`, other calling `service-b`. I want to have an attribute `"http-client-type":"service-a" which is included in every request.

The only way to achieve that right now, is to create a wrapper for the transport:

```
type CustomTransport struct {
	defaultAttr []attributes
	base   http.RoundTripper
}

var _ http.RoundTripper = &Transport{}

func (t *Transport) RoundTrip(req *http.Request) (*http.Response, error) {
	labeler := &otelhttp.Labeler{}

        // add the attributes from the wrapper
	labeler.Add(attributes)
	req = req.WithContext(otelhttp.ContextWithLabeler(req.Context(), labeler))
	return t.base.RoundTrip(req)
}

func NewCustomTransport(base http.RoundTripper, attributes []attributes) *Transport {
	base = otelhttp.NewTransport(base,
		otelhttp.WithTracerProvider(cfg.tracerProvider),
		otelhttp.WithMeterProvider(cfg.metricProvider),
	)

	return &Transport{
		attributes: attributes,
		base:   base,
	}
}

```

So with this option, you can add default attributes without the need of a wrapper. This could potentially replace the labeler (since attributes are based on the request)

hanyuancheung (approved):

luca-filipponi:
@hanyuancheung @dmathieu i've squashed all the commits and rebased with main. Let me know if you need me to do something else.

CHANGELOG.md
```diff
@@ -17,6 +17,7 @@ This project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0.htm
 - The `go.opentelemetry.io/contrib/config` package supports configuring `with_resource_constant_labels` for the prometheus exporter. (#5890)
 - Add new runtime metrics to `go.opentelemetry.io/contrib/instrumentation/runtime`, which are still disabled by default. (#5870)
 - Support for the `OTEL_HTTP_CLIENT_COMPATIBILITY_MODE=http/dup` environment variable in `go.opentelemetry.io/contrib/instrumentation/net/http/otelhttp` to emit attributes for both the v1.20.0 and v1.26.0 semantic conventions. (#5401)
+- Added option for extracting attributes from the http request in http transport in `go.opentelemetry.io/contrib/instrumentation/net/http/otelhttp`. (#5876)
```
dmathieu, line 20: Sorry for not pointing this out later. If the changelog entry mentions the new method, that's way easier for folks reading it to notice it.

```suggestion
- Add the `WithMetricsAttirbutesFn` option to allow setting dynamic, per-request metric attributes in `go.opentelemetry.io/contrib/instrumentation/net/http/otelhttp`. (#5876)
```

luca-filipponi: np, i had to fix some conflicts anyway

CHANGELOG.md
```diff
@@ -20,6 +20,7 @@ This project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0.htm
 - The `go.opentelemetry.io/contrib/bridges/otelzerolog` module.
   This module provides an OpenTelemetry logging bridge for `github.com/rs/zerolog`. (#5405)
 - Add `WithGinFilter` filter parameter in `go.opentelemetry.io/contrib/instrumentation/github.com/gin-gonic/gin/otelgin` to allow filtering requests with `*gin.Context`. (#5743)
+- Add the `WithMetricsAttirbutesFn` option to allow setting dynamic, per-request metric attributes in `go.opentelemetry.io/contrib/instrumentation/net/http/otelhttp`. (#5876)
```
dmathieu, line 23:
```suggestion
- Add the `WithMetricsAttributesFn` option to allow setting dynamic, per-request metric attributes in `go.opentelemetry.io/contrib/instrumentation/net/http/otelhttp`. (#5876)
```

luca-filipponi: argh right didn't see the typo. Thanks!

instrumentation/net/http/otelhttp/test/transport_test.go
```diff
@@ -534,30 +538,85 @@ func TestCustomAttributesHandling(t *testing.T) {
...
+
+	ts := httptest.NewServer(http.HandlerFunc(func(w http.ResponseWriter, _ *http.Request) {
+		w.WriteHeader(http.StatusOK)
+	}))
+	defer ts.Close()
+
+	r, err := http.NewRequest(http.MethodGet, ts.URL, nil)
+	require.NoError(t, err)
+
+	_, err = client.Do(r)
```
dmathieu, line 587:
```suggestion
	resp, err := client.Do(r)
	resp.Body.Close()
```

PR #5895, opened 2024-07-11
otelzap: remove TODO comments for using pooled objects

khushijain21:
Part of #5191

Pre-work #5279


bridges/otelzap/encoder.go
```diff
@@ -196,15 +194,13 @@ type arrayEncoder struct {
 }
 
 func (a *arrayEncoder) AppendArray(v zapcore.ArrayMarshaler) error {
-	// TODO: Use arrayEncoder from a pool.
 	arr := &arrayEncoder{}
```
pellared, line 197: Wouldn't pre-allocating the slice in `arrayEncoder` reduce the amount of heap allocations (similarly to `newObjectEncoder(2)` for object encoder)?

This can be handled in a separate PR.

bridges/otelzap/encoder.go
```diff
@@ -55,15 +55,13 @@ func (m *objectEncoder) calculate(o *namespace) {
 }
 
 func (m *objectEncoder) AddArray(key string, v zapcore.ArrayMarshaler) error {
-	// TODO: Use arrayEncoder from a pool.
 	arr := &arrayEncoder{}
```
pellared, line 58: Wouldn't pre-allocating the slice in `arrayEncoder` reduce the amount of heap allocations (similarly to `newObjectEncoder(2)` for object encoder)?

This can be handled in a separate PR.

bridges/otelzap/encoder.go
```diff
@@ -55,15 +55,13 @@ func (m *objectEncoder) calculate(o *namespace) {
...
 	err := v.MarshalLogArray(arr)
 	m.cur.attrs = append(m.cur.attrs, log.Slice(key, arr.elems...))
 	return err
 }
 
 func (m *objectEncoder) AddObject(k string, v zapcore.ObjectMarshaler) error {
-	// TODO: Use objectEncoder from a pool.
 	newobj := newObjectEncoder(2)
```
pellared, line 65: It is always `2`. Maybe we can move the magic value inside `newObjectEncoder` (and just have `newobj := newObjectEncoder()` and also explain why it is 2 and not any different value.

This can be handled in a separate PR.

bridges/otelzap/encoder.go
```diff
@@ -196,15 +194,13 @@ type arrayEncoder struct {
...
-	// TODO: Use arrayEncoder from a pool.
 	arr := &arrayEncoder{}
 	err := v.MarshalLogArray(arr)
 	a.elems = append(a.elems, log.SliceValue(arr.elems...))
 	return err
 }
 
 func (a *arrayEncoder) AppendObject(v zapcore.ObjectMarshaler) error {
-	// TODO: Use objectEncoder from a pool.
 	m := newObjectEncoder(2)
```
pellared, line 204: It is always `2`. Maybe we can move the magic value inside `newObjectEncoder` (and just have `newobj := newObjectEncoder()` and also explain why it is 2 and not any different value.

This can be handled in a separate PR.

pellared (approved):

bridges/otelzap/encoder.go
khushijain21: we don't do that because here we initialize it based on length of the field

```
func convertField(fields []zapcore.Field) (context.Context, []log.KeyValue) {
	var ctx context.Context
	enc := newObjectEncoder(len(fields))
	.... 
}
```

pellared (approved):

dmathieu (approved):

Issue #5902, opened 2024-07-11
Upgrade v1.24.0 to v1.26.0 semconv for otelhttp


MrAlias:
An audit of the new semconv for any additions missing will be needed.

VinozzZ:
Hi @MrAlias , I would like to work on this issue.

FYI, based on this message in [CNCF's slack ](https://cloud-native.slack.com/archives/C01NJ7V1KRC/p1716316290206639), it's called out by the Semconv Maintainers that adoption for HTTP should wait until final migration to stabilized semconv is ready.

MrAlias:
> Hi @MrAlias , I would like to work on this issue.
> 
> FYI, based on this message in [CNCF's slack ](https://cloud-native.slack.com/archives/C01NJ7V1KRC/p1716316290206639), it's called out by the Semconv Maintainers that adoption for HTTP should wait until final migration to stabilized semconv is ready.

The note you linked to is referring to database semantic conventions:

> Quite a few DB related breaking changes are in this release.  As with HTTP, please hold on adopting these until final migration to stabilized semconv is ready.

The HTTP semantic conventions are already stable: https://github.com/open-telemetry/semantic-conventions/blob/main/docs/http/http-spans.md

Though an audit of any additions between v1.24.0 and v1.26.0 needs to be done. These are allowed stable changes.

VinozzZ:
Thank you for the clairification!

It looks like we have a few new attributes that currently are all under `experimental` status
[Add url.template to HTTP client attributes](https://github.com/open-telemetry/semantic-conventions/commit/eb859d55d66a63bd6b2cde3d57c8758d5b023caf)
[Add experimental HTTP attributes to HTTP spans](https://github.com/open-telemetry/semantic-conventions/commit/30402c468f38c8d0250fa37b1dda590bfb9f81f2). The list of attributes are

For both client and server side:
- `http.request.size`
- `http.response.size`
- `http.request.body.size`
- `http.response.body.size`

For only client side:
- `url.template`

MrAlias:
Gotcha :+1: . We'll want to not add anything experimental at this point.

Sounds like the upgrade without those additions should just be a version bump then.

VinozzZ:
I assume we will do this after we finish adding support for v1.24.0. Is that correct?

MrAlias:
> I assume we will do this after we finish adding support for v1.24.0. Is that correct?

[...]

PR #5923, opened 2024-07-17
instrumentation/net/http/otelhttp: upgrade semconv to v1.26.0

VinozzZ:
Bases on the discussion in #5902 , we have came to the conclusion that there's no new stable attributes introduced in semconv v1.26.0 for otelhttp. We can directly bump from v1.24.0 to v1.26.0

Changes in this PR:
- bump semconv from v1.24.0 to v1.26.0 for otelhttp
- rename the `instrumentation/net/http/otelhttp/internal/semconv/v1.24.0.go` to `instrumentation/net/http/otelhttp/internal/semconv/httpconv.go` to match with client side naming convention


MrAlias:

I think [this](https://github.com/open-telemetry/opentelemetry-go-contrib/blob/dedcf91a55a36a5a8589c56f2e43c188eb42f4f2/CHANGELOG.md?plain=1#L19) entry could be updated to address this.

MrAlias (commented):
The `util.go` import of v1.24.0 needs to also be upgraded: https://github.com/open-telemetry/opentelemetry-go-contrib/blob/6fa80373e8eb0fea5f55afa514179ded020ddc00/instrumentation/net/http/otelhttp/internal/semconv/util.go#L13

The `v1.24.0_test.go` file needs to be renamed to match: `httpconv_test.go`

MrAlias (approved):

dmathieu (approved):

PR #5932, opened 2024-07-20
otelgokit: Logging bridge for go-kit/log

khushijain21:
This is a draft PR for [go-kit/log](https://pkg.go.dev/github.com/go-kit/log). Since it is widely used in OSS projects like Prometheus, and Thanos, the end-users may find this bridge useful.

Can we also move code common to all bridges to bridges/common folder?


pellared:
go-kit tends to have integrations as subpackages. See:
- https://github.com/go-kit/kit/tree/master/metrics
- https://github.com/go-kit/kit/tree/master/tracing

Have you considered to donate your implementation into https://github.com/go-kit/log? I think it would be a better place for the go-kit community. 

It is also no default preference to add more bridges: https://github.com/open-telemetry/opentelemetry-go-contrib/tree/main/instrumentation#new-instrumentation

Issue #5332, opened 2024-03-28
Make `otelhttp` produce old, new and duplicate Client Trace Attributes


MadVikingGod:
Make `instrumentation/net/http/otelhttp` produce Client Traces following the migration plan #5132.

A user should be able to produce Traces from the `Transport` in the current semconv (default), the latest semconv (1.26.0), and a duplicate form based on the Environment variable `OTEL_HTTP_CLIENT_COMPATIBILITY_MODE`

- [x] Produce Old Client Attributes
- ~Produce Duplicate Client Attributes~
- [x] Produce New Client Attributes
- [x] Add a changelog entry

VinozzZ:
I would like to work on the remaining task for this issue

VinozzZ:
Implementation Plan:
- Create a `newHTTPClient` that returns attributes from [semconv v1.26.0](https://opentelemetry.io/docs/specs/semconv/http/http-spans/#http-client)
- the `newHTTPClient` should implement `RequestTraceAttrs`, `ResponseTraceAttrs`, and `Status` methods
- Add tests to check `newHTTPClient` is returning correct attributes 
- Add logic to check for environment variable `OTEL_HTTP_CLIENT_COMPATIBILITY_MODE` in `HTTPClient` to also return attributes from `newHTTPClient` if the env is set to `http/dup`. Otherwise, only return attributes from the `oldHTTPClient`

MrAlias:
Done.

PR #5945, opened 2024-07-25
instrumentation/net/http/otelhttp: support duplicate both old and new attributes for HTTPClient

VinozzZ:
This PR implements the HTTPClient that can duplicate attributes for old semconv v1.20.0 and new semconv v1.26.0.
It's part of the migration plan #5332


MadVikingGod (approved):
Thank you for this.

The only attribute we might want to consider adding is `url.scheme`.

VinozzZ:
> Thank you for this.
> 
> The only attribute we might want to consider adding is `url.scheme`.

Thank you for the review!
I also had a discussion at the last sig meeting about the new attribute `error.type`. 
The consensus is to implement a new method to set `error.type`. The new method will replace the `RecordError()` in `RoundTrip()` in `otelhttp/transport.go`. I will make this change soon

instrumentation/net/http/otelhttp/internal/semconv/httpconv.go
```diff
@@ -195,3 +195,143 @@ func (n newHTTPServer) ResponseTraceAttrs(resp ResponseTelemetry) []attribute.Ke
 func (n newHTTPServer) Route(route string) attribute.KeyValue {
 	return semconvNew.HTTPRoute(route)
 }
+
+type newHTTPClient struct{}
+
+// RequestTraceAttrs returns trace attributes for an HTTP request made by a client.
+func (n newHTTPClient) RequestTraceAttrs(req *http.Request) []attribute.KeyValue {
```
dmathieu, line 202: Should those methods be tested?

VinozzZ: This is tested in the httpconv_test.go. Is there anywhere else I should add tests for this method?

instrumentation/net/http/otelhttp/internal/semconv/httpconv.go
```diff
@@ -195,3 +198,151 @@ func (n newHTTPServer) ResponseTraceAttrs(resp ResponseTelemetry) []attribute.Ke
...
+	}
+
+	if isErrorStatusCode(resp.StatusCode) {
+		errorType := strconv.Itoa(resp.StatusCode)
+		attrs = append(attrs, semconvNew.ErrorTypeKey.String(errorType))
+	}
+	return attrs
+}
+
+func (n newHTTPClient) ErrorType(err error) attribute.KeyValue {
```
VinozzZ, line 314: I'm not sure if this is the correct implementation that comply with the spec.
In the [spec](https://opentelemetry.io/docs/specs/semconv/attributes-registry/error/), it says `Instrumentations SHOULD document the list of errors they report.`. What error types do we want to return here?

dashpole (approved):

dmathieu (approved):

PR #5965, opened 2024-07-31
otelhttptrace: handle missing getconn hook without panic

krantideep95:
Fork of https://github.com/open-telemetry/opentelemetry-go-contrib/pull/5187 updated with main branch and tests, this PR adds nil dereference check for clientTracer.root in `end()` when span events are used instead of sub spans


krantideep95:
Hey @dmathieu, sorry for the tag. Would you mind reviewing this Pull Request? Thanks!

instrumentation/net/http/httptrace/otelhttptrace/clienttrace_test.go
```diff
@@ -5,9 +5,20 @@ package otelhttptrace
...
 	"net/http"
 	"net/http/httptrace"
+	"testing"
+
+	"github.com/stretchr/testify/assert"
+	"github.com/stretchr/testify/require"
+
+	"go.opentelemetry.io/otel"
+	"go.opentelemetry.io/otel/attribute"
+	sdktrace "go.opentelemetry.io/otel/sdk/trace"
```
dmathieu, line 19: This means the sdk is being added as a dependency, which instrumentations shouldn't have (someone using this instrumentation may want to use their own custom SDK).

Could you move those tests into the `test` subfolder?

krantideep95: I have removed these, and added `TestEndBeforeStartWithoutSubSpansDoesNotPanic` instead in `test` subfolder.

dmathieu (approved):

Issue #3368, opened 2023-02-07
otelaws AppendMiddlewares causes S3 pre-signed URLs to stop working


lgcmotta:
# Description

When adding the `otelaws.AppendMiddlewares(&cfg.APIOptions)`, all pre-signed URLs for downloading S3 files start to break with code `SignatureDoesNotMatch` and the message:

> "The request signature we calculated does not match the signature you provided. Check your key and signing method."

![image](https://user-images.githubusercontent.com/33238105/217125791-14a45626-7937-4fae-a1a3-f6f4997bc7f7.png)

When removing `otelaws.AppendMiddlewares(&cfg.APIOptions)` they immediately return to work.

I'm attaching an image from a diff window with the generated URLs. The only difference that is not related to the AWS account nor the link expiration time is the `X-Amz-SignedHeaders`. The URL generated from code with **otelaws middlewares** has two values `host` and `traceparent`, while the one without **otelaws middlewares** has only `host`.

![image](https://user-images.githubusercontent.com/33238105/217124120-f12c5afa-1625-49af-bf42-03bc38be119e.png)

# Packages installed

- github.com/aws/aws-sdk-go-v2 v1.17.4
- github.com/aws/aws-sdk-go-v2/config v1.18.12
- github.com/aws/aws-sdk-go-v2/service/s3 v1.30.2
- github.com/gin-gonic/gin v1.8.2
- github.com/swaggo/files v1.0.0
- github.com/swaggo/gin-swagger v1.5.3
- github.com/swaggo/swag v1.8.10
- go.opentelemetry.io/contrib/instrumentation/github.com/aws/aws-sdk-go-v2/otelaws v0.38.0
- go.opentelemetry.io/contrib/instrumentation/github.com/gin-gonic/gin/otelgin v0.38.0
- go.opentelemetry.io/otel v1.12.0
- go.opentelemetry.io/otel/exporters/otlp/otlptrace/otlptracegrpc v1.12.0
- go.opentelemetry.io/otel/sdk v1.12.0
- google.golang.org/grpc v1.52.3

## Code

I created a simple project to demonstrate this behavior. 

👉 **Code is available [here](https://github.com/lgcmotta/s3-otel-poc).** 👈 

This sample exports traces to _Jaeger_ using _OTLP_ with gRPC and  `opentelemetry-collector-contrib` at version `0.68.0`.

This is the code with AWS SDK v2 configuration and `otelaws` middleware:

```go
// ...
[...]

PR #5975, opened 2024-08-02
otelaws: Add finalize middleware after instead of before

jacksehr:
Fixes https://github.com/open-telemetry/opentelemetry-go-contrib/issues/3368.

## Summary
In any presigned URL flow, the AWS SDK skips all `Finalize` middlewares after the [PresignHTTPMiddleware](https://github.com/aws/aws-sdk-go-v2/blob/a2747951bd66c0b687a0a483195843b73f6448bb/internal/v4a/presign_middleware.go#L36-L38). So, if we move context propagation after the `PresignHTTPMiddleware`, we will not include context propagation headers in presigned URL requests, which should prevent the reported issue with signatures. 

## Context
The existing middleware in this package adds propagation headers (e.g. `traceparent`) to outgoing AWS requests, presumably to be used in things like AWS X-Ray. However, when making requests for presigned URLs, the AWS SDK adds a bunch of middleware of its own to requests. The most important of these is the [PresignHTTPMiddleware](https://github.com/aws/aws-sdk-go-v2/blob/a2747951bd66c0b687a0a483195843b73f6448bb/internal/v4a/presign_middleware.go#L36-L38). This does a lot of things, but for the purposes of this PR/issue, there are two particularly relevant things it does:

### Incorporating request headers into the presigned URL signature
When it is built, the middleware receives a presigner. Then, when it is actually run, it receives a `*smithyHTTP.Request`, and  it's entirely up to the presigner if the headers on the `*smithyHTTP.Request` are incorporated in the calculation of the presigned URL's signature. If we look in the [internals of the default v4 signer](https://github.com/aws/aws-sdk-go-v2/blob/main/internal/v4a/v4a.go#L305), it appears to just incorporate all existing headers and use them to sign the URL. This leads to the problem described in the above issue:

> The URL generated from code with **otelaws middlewares** has two values `host` and `traceparent`, while the one without **otelaws middlewares** has only `host`.

As observed by the comments in that issue, disabling the context propagation solves this issue, which makes sense as that removes the `traceparent` header from the signature calculation.

This leads us to the second relevant behaviour of the `PresignHTTPMiddleware`.

### Short circuiting remaining middleware
From the comments of the `PresignHTTPMiddleware`:
```golang
// PresignHTTPRequestMiddleware provides the Finalize middleware for creating a
// presigned URL for an HTTP request.
//
// Will short circuit the middleware stack and not forward onto the next
// Finalize handler.
```
In addition to this, the Deserialize middlewares get cleared, as seen [here](https://github.com/aws/aws-sdk-go-v2/blob/a2747951bd66c0b687a0a483195843b73f6448bb/service/s3/api_client.go#L904).

So, in the presigned URL flow, the last middleware that will get called will be the `PresignHTTPRequestMiddleware`. This is actually useful for us now, as if we just ensure our context propagation middleware comes last in the `Finalize` step, then we can be sure it will never run in the presign flow. Simply shifting from `middleware.Before` to `middleware.After` does just that.

I've also renamed the function and moved it so that its order reflects the order of middleware execution (i.e. [finalize before deserialize](https://aws.github.io/aws-sdk-go-v2/docs/middleware/))


pellared:
@akats7 PTAL as a code owner.

akats7:
Hey @jacksehr, will take a look at this later today, thanks

akats7:
Hey @jacksehr, can you please resolve the issues in the unit tests to use your updated method, also would you be able to add/update a test case for this change.

akats7:
Hey @jacksehr, any update on this?

jacksehr:
> Hey @jacksehr, any update on this?

Hey @akats7, sorry for the delay, circling back to this -- I've updated with a test and resolved the compilation issue.

pellared:
Please add a changelog entry

jacksehr:
@pellared @akats7 I can update with `main` again if needed, otherwise let me know if there's anything else this PR needs to get merged 🙂

dmathieu (approved):

akats7 (approved):
LGTM thanks!

jacksehr:
@akats7 I've just updated the branch, but I can't merge it even when the checks are all green -- can I get one of you to merge it then the checks run?

akats7:
> @akats7 I've just updated the branch, but I can't merge it even when the checks are all green -- can I get one of you to merge it then the checks run?

It can only be merged by one of the maintainers

cc @dmathieu @pellared

pellared (approved):

Issue #5974, opened 2024-08-02
Support go.schedule.duration runtime metric


dashpole:
### Problem Statement

We would like to support the go schedule duration:
https://github.com/open-telemetry/semantic-conventions/blob/main/docs/runtime/go-metrics.md#metric-goscheduleduration

From https://github.com/golang/go/issues/67120:

> Rationale: This metric is a measure of scheduling latency that is useful as a fine-grained proxy for overall system load. For example, diffing the distribution over short time windows can provide visibility into the latency impact of uneven load. This metric is battle-tested and has been found to be useful in a variety of scenarios.

However, this has a few challenges:

* The go runtime histogram type doesn't have a sum: https://pkg.go.dev/runtime/metrics#Float64Histogram.  It only has buckets and counts.  The sum is a critical part of the histogram, as it allows computing averages.
* The go runtime histogram is pre-computed.  This means we can't add this histogram using the OTel api.  Instead, we will need to model this as a metric.Producer.

dashpole:
For now, we could use the same approach as https://github.com/prometheus/client_golang/blob/46f77a97fa1e83e7285634745bd5c92c11bf20c7/prometheus/go_collector_latest.go#L541, and estimate the sum.

PR #5991, opened 2024-08-07
Implement go.schedule.duration

dashpole:
Fixes https://github.com/open-telemetry/opentelemetry-go-contrib/issues/5974.

This uses the same method for estimating the sum as the prometheus client-go: https://github.com/prometheus/client_golang/blob/46f77a97fa1e83e7285634745bd5c92c11bf20c7/prometheus/go_collector_latest.go#L543, which is to use the lower-bound of each bucket as the value of each observation.

It implements this using a metric producer.  It is designed to be used together with existing runtime instrumentation.


instrumentation/runtime/example/main.go
```diff
@@ -27,13 +27,17 @@ var res = resource.NewWithAttributes(
...
 		log.Fatal(err)
 	}
 
 	// Register the exporter with an SDK via a periodic reader.
-	read := metric.NewPeriodicReader(exp, metric.WithInterval(1*time.Second))
+	read := metric.NewPeriodicReader(
+		exp,
+		metric.WithInterval(1*time.Second),
+		metric.WithProducer(runtime.NewProducer()),
+	)
```
pellared, line 40: Given we add a dependency to the SDK, I think we can refactor this to a "testable example" in `go.opentelemetry.io/contrib/instrumentation/runtime` to make its docs better.

dashpole: Do you think it should use the stdoutmetric exporter?

pellared: We could do something like https://github.com/open-telemetry/opentelemetry-go/blob/4d29b35a1b6f78a85a91d8a8657ae521f57fb6b0/sdk/metric/example_test.go#L39

dashpole: done

instrumentation/runtime/producer.go
```diff
@@ -0,0 +1,115 @@
...
+func (p *Producer) Produce(context.Context) ([]metricdata.ScopeMetrics, error) {
+	p.collector.refresh()
+	// Use the last collection time (which may or may not be now) for the timestamp.
+	histDp := convertRuntimeHistogram(p.collector.getHistogram(goSchedLatencies), p.collector.lastCollect)
+	if len(histDp) == 0 {
+		return nil, fmt.Errorf("unable to obtain go.schedule.duration metric from the runtime")
+	}
+	return []metricdata.ScopeMetrics{
+		{
+			Scope: instrumentation.Scope{
```
dashpole, line 52: note that this is a separate instrumentation scope from the one that will be produced by runtime instrumentation.  The OTel SDK does not merge those today, but I'm not sure how problematic that will be.  I propose we merge this as-is, and then see what impact it has, if any.

instrumentation/runtime/example_test.go
```diff
@@ -0,0 +1,38 @@
...
+	"go.opentelemetry.io/otel/sdk/metric"
+)
+
+func Example() {
+	// This reader is used as a stand-in for a reader that will actually export
+	// data. See https://pkg.go.dev/go.opentelemetry.io/otel/exporters for
+	// exporters that can be used as or with readers.
+	reader := metric.NewManualReader(
+		// Add the runtime producer to get histograms from the Go runtime.
+		metric.WithProducer(runtime.NewProducer()),
```
pellared, line 22: Double-checking (I am not sure). Will it work correctly if the user would like to have multiple metrics  exporters? For instance: OTLP and Prometheus?

dashpole: They would have concurrent access issues, I think.  I can synchronize the goCollector to fix that.

instrumentation/runtime/go.mod
```diff
@@ -6,13 +6,17 @@ require (
 	github.com/stretchr/testify v1.9.0
 	go.opentelemetry.io/otel v1.28.0
 	go.opentelemetry.io/otel/metric v1.28.0
+	go.opentelemetry.io/otel/sdk v1.28.0
+	go.opentelemetry.io/otel/sdk/metric v1.28.0
```
MrAlias, line 10: Instrumentation should not depend on the the SDK. This looks to be coming in from the inclusion of the example. I would recommend keeping it as a separate module.

dashpole: This is from the use of metric.Producer.  LMK if you would like the producer in a separate module.

MrAlias: I see ...

This is all because we don't have an observable Histogram, right?

Does it make more sense to have all of the runtime metrics come from this `Producer`? That way they are all under the same scope.

dashpole: I considered that.  However, it would mean all other metrics would be pre-aggregated as well. Things like views or temporality configuration would not work unless we added + implemented those features within the producer

dashpole: I could probably use the metrics SDK itself within the producer if we wanted to support those.

MrAlias: Hmm, yeah. I'm not really sure which way would be preferable to end-users.

I think in the best case scenario, as an end-user, you would get instrumentation that included all of these metrics (including the `go.schedule.duration`) without having to depend on the SDK. Is this something we need to bring up to the specification? Is adding an observable "pre-binned" histogram instrument the right solution here?

dashpole: i'm honestly not sure what the right solution is.  Having an observable pre-binned histogram instrument (a "bridge" API) would eliminate the need to depend on the SDK here, but it doesn't solve things like views (e.g. you can't change the aggregation to ExponentialHistogram).  We also put producers on the Reader, rather than meter provider, intentionally since producers don't allow exporters to influence aggregations or temporality.

MrAlias: > i'm honestly not sure what the right solution is.

Same.

If we were to move everything into the producer, how would you use the SDK to accommodate views? Is the idea to setup some kind of pull exporter and when `Produce` is called pull the SDK data and merge it into our own `metricsdata.ScopeMetrics`?

instrumentation/runtime/options.go
```diff
@@ -36,7 +43,7 @@ const DefaultMinimumReadMemStatsInterval time.Duration = 15 * time.Second
 // WithMinimumReadMemStatsInterval sets a minimum interval between calls to
 // runtime.ReadMemStats(), which is a relatively expensive call to make
 // frequently.  This setting is ignored when `d` is negative.
-func WithMinimumReadMemStatsInterval(d time.Duration) Option {
+func WithMinimumReadMemStatsInterval(d time.Duration) ProducerOption {
```
XSAM, line 46:
```suggestion
func WithMinimumReadMemStatsInterval(d time.Duration) Option {
```

There is no need to change `Option` to `ProducerOption`.

instrumentation/runtime/example_test.go
dashpole: fixed

instrumentation/runtime/go.mod
dashpole: I don't have cycles to take this on this week.  If you are OK with the current implementation, i'd like to start with this.  Having the producer separate means at least our two options (API vs Producer) are at least non-overlapping and give users control

XSAM (approved):
👍

dmathieu (approved):

instrumentation/runtime/go.mod
MrAlias: SIG notes:

- Go ahead with this approach
- Look into merging scopes in the SDK
- Long term:
  - Metrics bridge API
  - Producer location in the SDK (maybe make it its own package?)

dashpole: I would like to go ahead with this as-is.  I thought about adding an experimental disclaimer, but I didn't want to imply that other parts of the package are non-experimental.  When we stabilize other portions of this package, we can decide if this needs to be moved.

Issue #5132, opened 2024-02-19
HTTP Semconv Migration Plan


MadVikingGod:
# Reasoning
The HTTP Semantic Conventions made breaking changes from v1.20.0 to v1.21.0; this means that if we did a hard cutover in our instrumentation, any users with dashboards or alarms that used these attributes would break.  To ease that, we propose the following plan to provide a path for users to migrate any existing dashboards. 

# Migration Plan
- Current release - v0.49.0
    - Support only for v1.20.0 semantic conventions
- v0.50.0
    - Add support for v1.20.0, v1.24.0, and both semantic conventions
    - Default to v1.20.0
    - Warn users when using v1.20.0
- v0.51.0
    - Default to v1.24.0
    - Warn users when using v1.20.0 or both semantic conventions
- v0.52.0
    - Remove support for v1.20.0
    - Remove support for both semantic conventions
    - All users will be using v1.24.0

# User Migration Story
The goal is to allow users to upgrade versions of the `otelhttp` package and enable incremental migrations of dashboards and alerts.

When the user first upgrades to v0.49, they will receive a warning that they are using v1.20.0 semantic conventions. This indicates that action needs to be taken, or their dashboards and alerts will be affected.

They can then set the environment variable `OTEL_HTTP_CLIENT_COMPATIBILITY_MODE` to `http/dup` and receive both the original v1.20.0 attributes and the new v1.24.0 attributes. This will allow them to update their dashboards and alerts to use the new attributes and validate they are not missing any data.

MrAlias:
This doesn't include any of the details discussed during yesterdays SIG meeting regarding rolling out the "opt-in" phase over multiple releases for all the instrumentation libraries.

MadVikingGod:
Alternative Plan:

- Current Version:
    - Always export the current version of the semantic convention.
    - Use an environment variable to export the current Version and an incomplete new version. This allows us to release a portion at a time. In this mode, it is acceptable to have two sets of allocations, one for the old code path and one for the new.
    - This is complete when:
      - All supported HTTP instrumentation have this environment variable
      - Both Server instrumentation and Client instrumentation produce duplicates
      - Metrics have support for both
- Current Version + 1: 
    - Always export the newest version of the semantic convention.
    - Use an environment variable to export the newest version and the old version.    In this mode, it is acceptable to have two sets of allocations, one for the old code path and one for the new.
- Current Version + 2:
[...]

PR #6031, opened 2024-08-22
otelgrpc: update semconv to v1.26.0

zchee:
Update semconv to v1.26.0 instead of too old of v1.17.0.

### DIFF
--- a/instrumentation/google.golang.org/grpc/otelgrpc/interceptor.go
+++ b/instrumentation/google.golang.org/grpc/otelgrpc/interceptor.go
@@ -24,7 +24,7 @@ import (
 	"go.opentelemetry.io/otel/attribute"
 	"go.opentelemetry.io/otel/codes"
 	"go.opentelemetry.io/otel/metric"
-	semconv "go.opentelemetry.io/otel/semconv/v1.17.0"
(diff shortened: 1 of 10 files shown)


instrumentation/google.golang.org/grpc/otelgrpc/interceptor.go
```diff
@@ -479,13 +479,13 @@ func peerAttr(addr string) []attribute.KeyValue {
...
-			semconv.NetSockPeerPort(port),
+			semconv.NetworkPeerAddress(host),
+			semconv.NetworkPeerPort(port),
 		}
 	} else {
 		attr = []attribute.KeyValue{
-			semconv.NetPeerName(host),
-			semconv.NetPeerPort(port),
+			semconv.NetworkPeerAddress(host),
+			semconv.NetworkPeerPort(port),
```
dmathieu, line 488: These are breaking changes. We should either mark it explicitly in the changelog, or do as we do in otelhttp and provide a smooth migration path.
See https://github.com/open-telemetry/opentelemetry-go-contrib/issues/5132

dmathieu (commented):

dmathieu:
Closing. This PR is quite stale, and there's been many new semconv releases since.

PR #5916, opened 2024-07-16
Move otelhttp wrappers into internal package

dmathieu:
This moves the otelhttp wrappers into the internal package, so we cannot rely on private values, and are forced to think about race conditions when using that data.

Closes #5914.

PR #6054, opened 2024-08-27
Don't write headers twice

### LINKED ISSUE #6053: Error logs coming for `http: superfluous response.WriteHeader call` when using otelhttp handler
### Description

In the latest release the call to request.Write() is setting the response headers to `http.StatusOK` without checking if the WriteHeader is already called, this is causing issues when trying to set Status Codes and then trying to Write to the response.

### Environment

- OS: darwin
- Architecture: arm64
- Go Version: go 1.22
- `otelhttp` version: v0.54.0

### Steps To Reproduce

1. Using this code ...
```
package main

import (
	"fmt"

	"gofr.dev/pkg/gofr"
)

func main() {
	a := gofr.New()

	a.GET("/hello", HelloHandler)

	a.Run()
}

func HelloHandler(c *gofr.Context) (interface{}, error) {
	name := c.Param("name")
	if name == "" {
		c.Log("Name came empty")
		name = "World"
	}

	return fmt.Sprintf("Hello %s!", name), nil
}

```
2. Run ...
```
go run main.go

curl --location 'localhost:8000/hello'
```
3. See error ...
image

### Expected behavior

The Write() call should check if the wroteHeader is true before calling WriteHeader()

dmathieu:
The change in https://github.com/open-telemetry/opentelemetry-go-contrib/pull/5916 introduced a regression, as we check for `wroteHeader`, but then always write anyway.
We need to only write if the header wasn't written yet.

Fixes #6053.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -12,6 +12,10 @@ This project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0.htm
 
 - Drop support for [Go 1.21]. (#6046, #6047)
 
+### Fixes
+
+- Superfluous `response.WriteHeader` call in `go.opentelemetry.io/contrib/instrumentation/net/http/otelhttp`. (#6054)
+
 <!-- Released section -->
 <!-- Don't change this section unless doing release -->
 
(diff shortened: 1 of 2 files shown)


dmathieu:
Closing, as this breaks https://github.com/open-telemetry/opentelemetry-go-contrib/pull/3580

PR #6091, opened 2024-09-05
config: Add `WithPrometheusRegisterer` config option

codeboten:
This is to allow the OTel Collector to configure its own handler for prometheus metrics, allowing it to manage error reporting and shutdown.

I'm opening this PR to get feedback from contributors on whether this approach makes sense or not, will add tests once I get some feedback.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -11,7 +11,7 @@ This project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0.htm
 ### Added
 
 - Add `NewProducer` to `go.opentelemetry.io/contrib/instrumentation/runtime`, which allows collecting the `go.schedule.duration` histogram metric from the Go runtime. (#5991)
-- Add `WithRegistry` to `go.opentelemetry.io/contrib/config` to allow users to configure their own prometheus registry and server.
+- Add `WithRegistry` to `go.opentelemetry.io/contrib/config` to allow users to configure their own prometheus registry and server. (#6091)
 
 ### Removed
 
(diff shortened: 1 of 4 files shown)


CHANGELOG.md
```diff
@@ -11,6 +11,7 @@ This project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0.htm
 ### Added
 
 - Add `NewProducer` to `go.opentelemetry.io/contrib/instrumentation/runtime`, which allows collecting the `go.schedule.duration` histogram metric from the Go runtime. (#5991)
+- Add `WithRegistry` to `go.opentelemetry.io/contrib/config` to allow users to configure their own prometheus registry and server.
```
codeboten, line 14:
```suggestion
- Add `WithRegistry` to `go.opentelemetry.io/contrib/config` to allow users to configure their own prometheus registry and server. (#6091)
```

config/config.go
```diff
@@ -127,6 +129,17 @@ func WithOpenTelemetryConfiguration(cfg OpenTelemetryConfiguration) Configuratio
 	})
 }
 
+// WithPrometheusRegisterer sets a Prometheus registerer to be used by
+// any Prometheus exporters configured in this SDK. Note: if this option
+// is set, the caller is expected to initialize their own handler to
```
dashpole, line 134: Can you help explain this a bit more to me?  Doesn't this mean most of the prometheus configuration provided via config (e.g. port, path) is ignored?  Why does the collector need this behavior?

codeboten: So the bulk of the prometheus handler code in the collector exists here: https://github.com/open-telemetry/opentelemetry-collector/blob/e99074da2f6525cd177f26b46b8d03fef2fe4f35/service/internal/proctelemetry/config.go#L106-L116

I'm totally open to suggestions as to a better way to implement this, the problem I was trying to solve was handling errors without passing a channel around.

codeboten: So the error channel used terminates the collector on error:

```
    // AsyncErrorChannel is the channel that is used to report fatal errors.
    AsyncErrorChannel chan error
```

I suppose we could register an error handler and set it to return the error to the channel instead?

codeboten: I'll try to implement that and see how far i can get. will put this in draft until I get that working

codeboten:
Closing, this option is not required as using the otel handler can provide the functionality the collector needed

Issue #6112, opened 2024-09-12
superfluous response.WriteHeader call on errors


stmcginnis:
### Description

My basic web app has been seeing "superflous response.WriteHeader call" errors in the logs when someone hits on invalid endpoint. This same behavior has been observed here: https://github.com/prometheus/prometheus/issues/13678

I think I may see where this is happening. I could probably write up a small repro, but I hope this description is enough to go on based on what I've seen.

Many [tutorials](https://edgedelta.com/company/blog/how-to-instrument-a-go-application-using-opentelemtry-complete-guide-best-practices) out there show instrumenting the individual route:

```go
handleFunc := func(pattern string, handlerFunc func(http.ResponseWriter, *http.Request)) {
	handler := otelhttp.WithRouteTag(pattern, http.HandlerFunc(handlerFunc))
	mux.Handle(pattern, handler)
}

handleFunc("/spinwheel", spinwheel)
```

as well as installing a server-wide handler:

```go
// Add HTTP instrumentation for the whole server
handler := otelhttp.NewHandler(mux, "/")
```

This works fine under normal operations. But if there is a problem, like possibly the side effects from high memory usage mentioned in the issue linked above, I see this error message. It appears when `http.Error()` is used, the first `Error()` does is write the status code in the header. But `http.Error` also takes a string to return. So in the `Error()` code, [it writes the header, then writes to the ResponseWriter](https://github.com/golang/go/blob/eb975601a07ad25a47fdcb0cc6166ec695973d3e/src/net/http/server.go#L2251-L2252).

In the `RespWriterWrapper`, I'm seeing the `writeHeader` call happen twice for this error. The first time I believe is from the direct header write from `http.Error()`, then the second one from the write of the string content.

I think there is a bug in `RespWriterWrapper`'s logic for `writeHeader()`. In the [code](https://github.com/open-telemetry/opentelemetry-go-contrib/blob/d312469f21ca6606e872292bdca206b2680608d3/instrumentation/net/http/otelhttp/internal/request/resp_writer_wrapper.go#L76) it checks `if !w.wroteHeader`. If it hadn't written the header it will set the flag to true. But the problem here is that regardless of whether the flag is `true` or `false` it will still fall through and attempt to call `WriteHeader` at the end.

[...]

PR #6113, opened 2024-09-12
Protect against multiple WriteHeader calls

stmcginnis:
In the RespWriterWrapper there was a logic issue. The instance tracks whether WriteHeader has been called or not, but was still attempting to call it again even when it had already been called.

This updates the internal implementation to recognize if it's already been called and make any additional attempts a no-op. This avoids errors bubbling up for "superfluous response.WriteHeader calls".

Closes: #6112


dmathieu:
This is not the right fix, as you can see from the unit tests.

Are you using the latest release? 2 superfluous header issues were fixed in that latest release.

* https://github.com/open-telemetry/opentelemetry-go-contrib/pull/6055
* https://github.com/open-telemetry/opentelemetry-go-contrib/pull/6074

PR #6116, opened 2024-09-16
Update `minsev` to allow dynamic severities

MrAlias:
Similar to [`slog`](https://pkg.go.dev/log/slog#hdr-Levels), allow the `minsev.LogProcessor` to be configured with dynamic severity resolution.


processors/minsev/example_test.go
```diff
@@ -0,0 +1,51 @@
...
+	"fmt"
+	"os"
+	"strings"
+	"sync"
+
+	api "go.opentelemetry.io/otel/log"
+	"go.opentelemetry.io/otel/sdk/log"
+)
+
+const key = "OTEL_LOG_LEVEL"
```
pellared, line 17: `OTEL_LOG_LEVEL` is a level of the SDK's internal logger.
It is not well defined in the specification, but this is how (AFAIK) everyone is using it.

pellared (changes requested):
I do not think we should use `OTEL_LOG_LEVEL` here. AFAIK it is meant to be used for internal logging.

pellared (dismissed):

processors/minsev/severity.go
```diff
@@ -0,0 +1,124 @@
...
+	"sync/atomic"
+
+	"go.opentelemetry.io/otel/log"
+)
+
+// Severity represents a log record severity (also known as log level). Smaller
+// numerical values correspond to less severe log records (such as debug
+// events), larger numerical values correspond to more severe log records (such
+// as errors and critical events).
+type Severity int
```
dmathieu, line 16: What prevents us from using the severities defined in the API here? (and should we document it?)

processors/minsev/example_test.go
```diff
@@ -0,0 +1,51 @@
...
+
+var getSeverity = sync.OnceValue(func() api.Severity {
+	conv := map[string]api.Severity{
+		"":      api.SeverityInfo, // Default to SeverityInfo for unset.
+		"debug": api.SeverityDebug,
+		"info":  api.SeverityInfo,
+		"warn":  api.SeverityWarn,
+		"error": api.SeverityError,
+	}
+	// log.SeverityUnknown for unknown values.
```
dmathieu, line 27: There's no such thing as `log.SeverityUnknown`, so I suppose you mean `log.SeverityUndefined`?

processors/minsev/minsev.go
```diff
@@ -8,20 +8,28 @@ package minsev // import "go.opentelemetry.io/contrib/processors/minsev"
...
 // If downstream is nil a default No-Op [log.Processor] is used. The returned
 // processor will not be enabled for nor emit any records.
-func NewLogProcessor(downstream log.Processor, minimum api.Severity) *LogProcessor {
+func NewLogProcessor(downstream log.Processor, severity Severitier) *LogProcessor {
 	if downstream == nil {
 		downstream = defaultProcessor
 	}
-	p := &LogProcessor{Processor: downstream, Minimum: minimum}
+	if severity == nil {
+		severity = SeverityInfo
```
dmathieu, line 30: This what would be used in the default mentioned in the example. In which case, shouldn't this be `SeverityUndefined`?

processors/minsev/severity.go
MrAlias: The API doesn't implement `Severitier`.

pellared: I guess @dmathieu had in mind something like

```suggestion
type Severity log.Severity
```

processors/minsev/severity.go
```diff
@@ -0,0 +1,124 @@
...
+	val atomic.Int64
+}
+
+// Severity returns v's severity.
+func (v *SeverityVar) Severity() log.Severity {
+	return Severity(int(v.val.Load())).Severity()
+}
+
+// Set sets v's Severity to l.
+func (v *SeverityVar) Set(l Severity) {
```
pellared, line 117: Should it not accept `log.Severity`?
```suggestion
func (v *SeverityVar) Set(l log.Severity) {
```

processors/minsev/severity.go
MrAlias: What difference does that make?

processors/minsev/severity.go
MrAlias: No, you configure the package with the package severity. Similar to how you pass `Severity` to `NewLogProcessor`.

processors/minsev/severity.go
pellared: I think it is a little clearer. But maybe @dmathieu had something else in mind.

processors/minsev/severity.go
pellared: I do not understand what is the advantage of defining `type Severity` in this package and why we cannot simply rely on `log.Severity`. The `Severity` in this package looks tightly coupled to `log.Severity`.

processors/minsev/minsev.go
MrAlias: I don't follow. The default in the example is `log.SeverityInfo`. This is documented to default to `log.SeverityInfo`. We want to default to a usable value when we can, and that is what it is doing here. I'm not sure why you wan this to be `SeverityUndefined`.

processors/minsev/severity.go
pellared: Maybe then we could get rid of `translations` an users could simple create severities e.g. like this?

```
s := minsev.Severity(log.SeverityInfo)
```

MrAlias: > Maybe then we could get rid of `translations` an users could simple create severities like this?
> 
> ```
> s := minsev.Severity(log.SeverityInfo)
> ```

Is the idea that all the severity const are removed then? If so, this seems like a vague API and a poor design.

Instead of having users call:

```go
minsev.NewLogProcessor(p, minsev.SeverityInfo)
```

They would call:

```go
minsev.NewLogProcessor(p, minsev.Severity(log.SeverityInfo))
```

They have not syntactic hint about how to construct a `minsev.Severity`.

This would also mean we cannot default to `SeverityInfo` for zero values or things like:

```go
Severity(log.SeverityInfo).Severity().String()
```

will output `ERROR2`.

pellared (approved):

processors/minsev/severity.go
dmathieu: Yeah, I meant it that it would be nicer to avoid duplicating every severity level. But you're right, doing so would make the API weird.

dmathieu (approved):

PR #6125, opened 2024-09-18
[chore]: enable len and empty rules from testifylint

mmorel-35:
#### Description

Testifylint is a linter that provides best practices with the use of testify.

This PR enables [empty](https://github.com/Antonboom/testifylint?tab=readme-ov-file#empty) and [len](https://github.com/Antonboom/testifylint?tab=readme-ov-file#len) rules from [testifylint](https://github.com/Antonboom/testifylint)

It's linter provided by golangci-lint.

Here all available rules are activated except those who require to be fixed. This PR only fixes empty and len so the quantity of changes stays reasonnable for reviewers.


Makefile
```diff
@@ -154,6 +157,14 @@ golangci-lint/%: $(GOLANGCI_LINT)
 		&& cd $(DIR) \
 		&& $(GOLANGCI_LINT) run --allow-serial-runners $(ARGS)
 
+.PHONY: testifylint
+testifylint: $(OTEL_GO_MOD_DIRS:%=testifylint/%)
```
dmathieu, line 161: Same as in the SDK repo, I don't think we need testifylint specified here.

dmathieu (approved):

pellared (approved):

MrAlias (approved):

processors/minsev/minsev_test.go
```diff
@@ -159,7 +159,7 @@ func TestLogProcessorEnabled(t *testing.T) {
 			param.SetSeverity(sev)
 			assert.False(t, p.Enabled(ctx, param), sev.String())
 
-			if !assert.Lenf(t, wrapped.EnabledCalls, 0, "Record with severity %s passed-through", sev) {
+			if !assert.Emptyf(t, wrapped.EnabledCalls, "Record with severity %s passed-through", sev) {
 				wrapped.Reset()
 			}
 		}
```
dmathieu, line 165: It looks like a new issue crept in with another PR, at line 185.

Issue #5318, opened 2024-03-24
otelhttp: allow setting the start timestamp using go Context


dashpole:
### Problem Statement

Follow-up to https://github.com/open-telemetry/opentelemetry-go/issues/4993

Context: https://github.com/kubernetes/kubernetes/issues/103186

I would like otelhttp to allow me to pass the start time using the go `context.Context`.  This lets me move my instrumentation after my auth filter in my server handler chain while preserving the start time of the request.

### Proposed Solution

Similar to the Labeler, allow adding a start timestamp to a context in otelhttp.  This timestamp will be used by both otelhttp metric and trace instrumentation as the start timestamp. 

#### Alternative: Context-based StartSpanOptions

In the trace API package, add a `SpanStartContext` which is modeled after the current `Labeler`, but stores a list of SpanStartOptions, and has `Add` and `Get` options for the StartSpanOptions.

In the trace SDK, in [Start](https://github.com/open-telemetry/opentelemetry-go/blob/a5172ab422f4b9e09e8cfeba0e986f5f90c26092/sdk/trace/tracer.go#L29), it would fetch options from the SpanStartContext, and append them to the explicitly passed options.

This would allow users to override the StartSpanOptions for any instrumentation library, including adding attributes or links, or setting the span start time without any modifications to the instrumentation library itself.

However, this would not work very well for metrics.  First, there is no "start timestamp" option for metrics (nor should there be).  The start timestamp is an artifact of the otelhttp instrumentation library, not of the metrics API.  Second, while tracing instrumentation generally only creates one span, metric instrumentation generally makes multiple Record calls.  The rule of "these options apply to the span created" doesn't work well for metrics, because we probably want them to apply to all metrics in the instrumentation library.

PR #6137, opened 2024-09-21
Allow setting start time using go context for otelhttp

dashpole:
Fixes https://github.com/open-telemetry/opentelemetry-go-contrib/issues/5318

This allows overriding the start time for http server metrics and traces.  See the linked issue for the motivation.


instrumentation/net/http/otelhttp/start_time_context.go
```diff
@@ -0,0 +1,31 @@
...
+
+// StartTimeFromContext retrieves a time.Time from the provided context if one
+// is available. If no start time was found in the provided context, a new,
+// zero start time is returned and the second return value is false.
+func StartTimeFromContext(ctx context.Context) time.Time {
+	t, _ := ctx.Value(startTimeContextKey).(time.Time)
+	return t
+}
+
+var zeroTime time.Time
```
dmathieu, line 31:
```suggestion
```

instrumentation/net/http/otelhttp/start_time_context.go
```diff
@@ -0,0 +1,31 @@
...
+// instrumentation. Only one labeller can be injected into the context.
+// Injecting it multiple times will override the previous calls.
+func ContextWithStartTime(parent context.Context, start time.Time) context.Context {
+	return context.WithValue(parent, startTimeContextKey, start)
+}
+
+// StartTimeFromContext retrieves a time.Time from the provided context if one
+// is available. If no start time was found in the provided context, a new,
+// zero start time is returned and the second return value is false.
+func StartTimeFromContext(ctx context.Context) time.Time {
```
dmathieu, line 26: Does `StartTimeWithContext` need to be public? When would we want to retrieve it from outside the otelhttp package?

dashpole: K8s today only uses the tracing portion of otelhttp.  Instead, we are using the Prometheus client library.  I need to be able to get the start time for metrics there as well.

dmathieu (approved):

dashpole:
@open-telemetry/opamp-go-approvers this needs one more review if anyone has time

pellared (approved):
minor comments only

PR #6145, opened 2024-09-23
Add linter rules to prevent uses of the SDK in bridges and instrumentations

dmathieu:
Bridges and instrumentations shouldn't rely on the SDK (except in tests located in a test submodule).
So this adds a linter rule which enforces it.

I have fixes the obvious failures. For anything that wasn't trivial, I've set exceptions, as I don't want this PR to become huge.


.golangci.yml
```diff
@@ -75,6 +75,16 @@ issues:
 linters-settings:
   depguard:
     rules:
+      no-sdk-instrumentation:
+        files:
+          - "**/*/{bridges,instrumentation}/**/*.go"
+          - "!**/*test/*.go"
+          - "!**/*test/**/*.go"
+          - "!**/*example/*.go"
+          - "!**/*example/**/*.go"
```
dashpole, line 84: Can we omit the instrumentation/runtime and bridges/prometheus here instead of adding lots of nolint comments?  Metric bridges are expected to rely on the SDK.

dashpole (approved):

dashpole (approved):

pellared (approved):
👍

PR #6161, opened 2024-09-30
Use lychee cache

dmathieu:
A few months ago, we enabled lychee cache in the SDK repo, to prevent rate limit issues, but forgot to enable it here.
https://github.com/open-telemetry/opentelemetry-go/pull/5160


dmathieu:
Note that the check needs to succeed once for the cache to be stored (lychee doesn't update the cache if there's a failure anyway).

pellared (approved):

pellared:
@dmathieu, this does not look like fixing the following issue:

> * [429] [https://github.com/open-telemetry/opentelemetry-go/commits/main](https://github.com/open-telemetry/opentelemetry-go/commits/main) | Failed: Network error: Too Many Requests

dmathieu:
The check needs to run successfully once (we can trigger it at a moment of low activity) to trigger the cache.

pellared:
> add an auth token to the GitHub requests so the rate limit is increased

I think this is the way. We just need to make sure to limit the permissions for `GITHUB_TOKEN` using https://docs.github.com/en/actions/writing-workflows/choosing-what-your-workflow-does/controlling-permissions-for-github_token#defining-access-for-the-github_token-permissions

Reference: https://github.com/lycheeverse/lychee/blob/master/docs/TROUBLESHOOTING.md

pellared (approved):

dashpole (approved):

PR #6172, opened 2024-10-01
otelmongo: support OTEL_SEMCONV_STABILITY_OPT_IN for v1.26.0 semconv

### LINKED ISSUE #6171: [otelmongo] bump semconv to 1.26.0
### Problem Statement

otelmongo is currently using [semconv/v1.17.0](https://github.com/open-telemetry/opentelemetry-go/pull/3599), suggest updating to [semconv/v1.26.0](https://github.com/open-telemetry/opentelemetry-go/pull/5476).

### Proposed Solution
key/value pair changes migrating from 1.17.0 to 1.26.0 include:

<table>
  <tr>
   <td><strong>Old Func</strong>
   </td>
   <td><strong>New Func</strong>
   </td>
   <td><strong>Old Key</strong>
   </td>
   <td><strong>New Key</strong>
   </td>
   <td><strong>Notes</strong>
   </td>
  </tr>
  <tr>
   <td><code>DBOperation</code>
   </td>
   <td><code>DBOperationName</code>
   </td>
   <td><code>db.operation</code>
   </td>
   <td><code>db.operation.name</code>
   </td>
   <td>
   </td>
  </tr>
  <tr>
   <td><code>DBName</code>
   </td>
   <td><code>DBNamespace</code>
   </td>
   <td><code>db.name</code>
   </td>
   <td><code>db.namespace</code>
   </td>
   <td>
   </td>
  </tr>
  <tr>
   <td><code>NetPeerName</code>
   </td>
   <td><code>NetworkPeerAddress</code>
   </td>
   <td><code>net.peer.name</code>
   </td>
   <td><code>network.peer.address</code>
   </td>
   <td>Currently the hostname is used for the peer name. Documentation suggest using the server address.
   </td>
  </tr>
  <tr>
   <td><code>NetPeerPort</code>
   </td>
   <td><code>NetworkPeerPort</code>
   </td>
   <td><code>net.peer.port</code>
   </td>
   <td><code>network.peer.port</code>
   </td>
   <td>
   </td>
  </tr>
  <tr>
   <td><code>NetTransportTCP</code>
   </td>
   <td><code>NetworkingTransportTCP</code>
   </td>
   <td><code>net.transport</code>
   </td>
   <td><code>network.transport</code>
   </td>
   <td>Network transport no longer includes the <code>ip_</code> portion defined on the deprecated <code>net.trasnport</code>. Instead, this value is either <code>tcp</code> or <code>udp</code>.
   </td>
  </tr>
  <tr>
   <td><code>DBStatement</code>
   </td>
   <td><code>DBQueryText</code>
   </td>
   <td><code>db.statement</code>
   </td>
   <td><code>db.query.text</code>
   </td>
   <td>There is no deprecated analogue to `db.statement` according to the documentation. <code>Db.query.text</code> seemed like a natural fit.
   </td>
  </tr>
</table>

#### Alternatives

Since these fields are deprecated, there is no alternative other than leaving the semconv version static.

#### Prior Art

NA

### Additional Context

It's worth noting that this is a breaking change if telemetry key/values are considered stable.

prestonvasquez:
Resolves #6171, #2165

The specifications for maintaining multiple semantic convention versions is documented here: https://github.com/open-telemetry/semantic-conventions/blob/main/docs/database/database-spans.md. 

To use a non-default version:
```
export OTEL_SEMCONV_STABILITY_OPT_IN=database
```

To include both v1.26.0 and v1.21.0:
```
export OTEL_SEMCONV_STABILITY_OPT_IN=database/dup
```

To include the default (v1.21.0) version:
```
unset OTEL_SEMCONV_STABILITY_OPT_IN
```


prestonvasquez:
@pellared @dmathieu following up on issue #5551

dmathieu:
Updating semconv is definitely not just a chore, especially since, as you mention, there are field name changes (which can cause issues with users, since they need to change their queries and dashboards).

We should either mark these breaking changes explicitely in a changelog entry, or follow a similar migration path as the otelhttp instrumentation is doing: https://github.com/open-telemetry/opentelemetry-go-contrib/issues/5132

CHANGELOG.md
```diff
@@ -16,6 +16,7 @@ This project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0.htm
 ### Changed
 
 - The function signature of `NewLogProcessor` in `go.opentelemetry.io/contrib/processors/minsev` has changed to accept the added `Severitier` interface instead of a `log.Severity`. (#6116)
+- Attributes for otelmongo will change to reflect [semconv/v1.26.0](https://github.com/open-telemetry/opentelemetry-go/pull/5476). `db.operation` will change to `db.operation.name`, `db.name` will change to `db.namespace`, `net.peer.port` will change to `network.peer.port`, `net.transport` will change to `network.transport`, and `db.statement` will change to `db.query.text`.
```
dmathieu, line 19: If we do keep a one-step direct migration, the changelog entry should mention that this is because of the semconv upgrade.

```suggestion
- Upgrade the `go.opentelemetry.io/otel/semconv/v1.17.0` dependency to `go.opentelemetry.io/otel/semconv/v1.26.0` in `go.opentelemetry.io/contrib/instrumentation/go.mongodb.org/mongo-driver/mongo/otelmongo`. (#6172)
- Rename attributes `db.operation` to db.operation.name`, db.name` to `db.namespace`, `net.peer.port` to `network.peer.port`, `net.transport` to `network.transport` and `db.statement` to `db.query.text` in `go.opentelemetry.io/contrib/instrumentation/go.mongodb.org/mongo-driver/mongo/otelmongo`. (#6172)
```

pellared:
> We should either mark these breaking changes explicitely in a changelog entry, or follow a similar migration path as the otelhttp instrumentation is doing: https://github.com/open-telemetry/opentelemetry-go-contrib/issues/5132

@open-telemetry/specs-semconv-maintainers,  @open-telemetry/semconv-db-approvers, do you have some recommendations or feedback regarding changes of database semantic conventions?

pellared (changes requested):
I would rather first have a PR which bumps semconv to `v1.24.0`

prestonvasquez:
> I would rather first have a PR which bumps semconv to `v1.24.0`

@pellared Could you clarify this?

trask:
> I would rather first have a PR which bumps semconv to `v1.24.0`

would this go against the semconv recommendation?

> SHOULD NOT change the version of the database conventions that they emit by default until the database semantic conventions are marked stable.

pellared:
> > I would rather first have a PR which bumps semconv to `v1.24.0`
> 
> would this go against the semconv recommendation?
> 
> > SHOULD NOT change the version of the database conventions that they emit by default until the database semantic conventions are marked stable.

I thought that breaking changes started since v1.25.0. But I need to double check it later

pellared:
> @pellared Do you have an action item for this PR?

Let's first just bump semconv to `v1.24.0`.

In next PRs we can add support for `v1.26.0` and `OTEL_SEMCONV_STABILITY_OPT_IN`.

Notice also that the specification only defines following values for `OTEL_SEMCONV_STABILITY_OPT_IN`:
- `database`
- `database/dup`
- (none)

prestonvasquez:
@pellared 

> Notice also that the OTel Semantic Conventions only defines following values for `OTEL_SEMCONV_STABILITY_OPT_IN`:
> 
> * `database`
> * `database/dup`
> * (none)

IIUC this pattern is misguided and doesn't explain how to deal with multiple experimental bumps within a major version. Or doesn't explain why that would never happen.

trask:
@prestonvasquez `OTEL_SEMCONV_STABILITY_OPT_IN` is a comma-separated list of values, see

https://github.com/open-telemetry/semantic-conventions/blob/main/docs/database/database-spans.md

> * SHOULD introduce an environment variable `OTEL_SEMCONV_STABILITY_OPT_IN`
>   in the existing major version which is a comma-separated list of values.
>   If the list of values includes:
>   * `database` - emit the new, stable database conventions,
>     and stop emitting the old experimental database conventions
>     that the instrumentation emitted previously.
>   * `database/dup` - emit both the old and the stable database conventions,
>     allowing for a seamless transition.
>   * The default behavior (in the absence of one of these values) is to continue
>     emitting whatever version of the old experimental database conventions
>     the instrumentation was emitting previously.
>   * Note: `database/dup` has higher precedence than `database` in case both values are present

prestonvasquez:
> @prestonvasquez `OTEL_SEMCONV_STABILITY_OPT_IN` is a comma-separated list of values, see

@trask Would you mind reading the PR description and letting me know if the proposal is in line with the spirit of this requirement?

pellared:
> This pattern does not account for > 2 semantic version update within the lifetime of a major version. 

You would need to address it in https://github.com/open-telemetry/semantic-conventions.

This does not prevent bumping semconv to `v1.24.0` beforehand.

trask:
> @trask Would you mind reading the PR description and letting me know if the proposal is in line with the spirit of this requirement?

my recommendation would be to stick to what's defined in the semantic conventions

if you're interested, there was a lot of discussion around the `OTEL_SEMCONV_STABILITY_OPT_IN` mechanism when it was introduced in https://github.com/open-telemetry/opentelemetry-specification/pull/3404 and https://github.com/open-telemetry/opentelemetry-specification/pull/3443 (and refined in some later PRs)

> > This pattern does not account for > 2 semantic version update within the lifetime of a major version.
> 
> You would need to address it in https://github.com/open-telemetry/semantic-conventions.

I'm happy to discuss if you open an issue in https://github.com/open-telemetry/semantic-conventions describing the limitations you are running into with the existing `OTEL_SEMCONV_STABILITY_OPT_IN` mechanism.

prestonvasquez:
@pellared 

> This does not prevent bumping semconv to `v1.24.0` beforehand.

What is the purpose of doing this? It seems like this will maintain the existing behavior.

prestonvasquez:
> > @trask Would you mind reading the PR description and letting me know if the proposal is in line with the spirit of this requirement?
> 
> my recommendation would be to stick to what's defined in the semantic conventions
> 
> if you're interested, there was a lot of discussion around the `OTEL_SEMCONV_STABILITY_OPT_IN` mechanism when it was introduced in [open-telemetry/opentelemetry-specification#3404](https://github.com/open-telemetry/opentelemetry-specification/pull/3404) and [open-telemetry/opentelemetry-specification#3443](https://github.com/open-telemetry/opentelemetry-specification/pull/3443) (and refined in some later PRs)
> 
> > > This pattern does not account for > 2 semantic version update within the lifetime of a major version.
> > 
> > 
> > You would need to address it in [open-telemetry/semantic-conventions](https://github.com/open-telemetry/semantic-conventions).
> 
> I'm happy to discuss if you open an issue in [open-telemetry/semantic-conventions](https://github.com/open-telemetry/semantic-conventions) describing the limitations you are running into with the existing `OTEL_SEMCONV_STABILITY_OPT_IN` mechanism.

Issue: https://github.com/open-telemetry/semantic-conventions/issues/1502

dmathieu:
> What is the purpose of doing this? It seems like this will maintain the existing behavior.

Yes, this would be a noop change. It would (a bit virtually) reduce the scope of the upgrade to 1.26 though, as a single version bump would then be required.

prestonvasquez:
> > What is the purpose of doing this? It seems like this will maintain the existing behavior.
> 
> Yes, this would be a noop change. It would (a bit virtually) reduce the scope of the upgrade to 1.26 though, as a single version bump would then be required.

@dmathieu In this case, should we just close this PR and leave the semconv dependency as-is for now?

dmathieu:
Sure, if you think there's still too much work for this PR. Maybe document the outcome from the discussion that happened here in #6171?

prestonvasquez:
@dmathieu I think we should maintain 1.17.0 since upgrading to 1.24.0 has no obvious benefit. However, I'm happy to open a separate PR for 1.24.0 if you want. I would suggest not using the pattern for a non-breaking upgrade, though.

prestonvasquez:
> Sure, if you think there's still too much work for this PR. Maybe document the outcome from the discussion that happened here in #6171?

@dmathieu TBH it's not clear to me why we don't want to upgrade to 1.26.0.

pellared:
> @pellared
> 
> > This does not prevent bumping semconv to `v1.24.0` beforehand.
> 
> What is the purpose of doing this? It seems like this will maintain the existing behavior.

Like bumping a dependency - we are closer to the latest version.
Even if nothing changed at least it was double-checked during the bump. Maybe some new attribute has been introduced?

dmathieu:
Also, folks using the package will have less surprises seeing a newer version in their dependency tree rather than something older that looks like it's not being cared for.

pellared:
@prestonvasquez, I changed it to draft as it looks like it is still WIP. Please change to ready for review once applicable.

prestonvasquez:
@pellared Can you explain why this looks like a WIP? This PR is ready for review.

instrumentation/go.mongodb.org/mongo-driver/mongo/otelmongo/mongo.go
```diff
@@ -32,25 +33,28 @@ type monitor struct {
...
-		semconv.NetTransportTCP,
-	}
+	attrs := []attribute.KeyValue{semconv.DBSystemMongoDB}
+
+	attrs = appendOpNameAttrs(attrs, evt.CommandName)
+	attrs = appendDBNamespace(attrs, evt.DatabaseName)
+	attrs = appendNetworkPort(attrs, port)
+	attrs = appendNetworkHost(attrs, hostname)
+	attrs = appendNetworkAddress(attrs, net.JoinHostPort(hostname, strconv.Itoa(port)))
+	attrs = appendNetworkTransport(attrs)
```
dmathieu, line 45: Rather than creating one method for each attribute, how about one method which returns all attributes, and can decide their key name.
Kind of like what the `otelhttp` library does, where we define two structs, one for the old and another for the new version.
https://github.com/open-telemetry/opentelemetry-go-contrib/tree/main/instrumentation/net/http/otelhttp/internal/semconv

I would also move the logic into an internal package, for a better separation of concerns.

prestonvasquez: That's an interesting idea. If I understand correctly, you're suggesting we define two separate command monitors specifically for the semantic conventions. Could you explain the benefits of this pattern over the more atomic append* solution proposed in the PR? The goal of the existing solution is to minimize code changes when advancing the library to a desired version, while also avoiding duplication in the monitoring logic.

dmathieu: Your solution works if some values are renamed. But if a new value is added, or one is removed then, it becomes trickier to handle since the root method has to know about everything.

With one struct for each semconv version, that struct knows exactly what to do for that version, and that's it. There aren't two versions mixed.
It would also avoid having to define one method for each attribute.

Finally, doing the separation this way also makes it easier to add a third version if we someday need it (I really hope we don't).

pellared: At last using the same patterns in multiple modules help in maintenance of the repository.

prestonvasquez: @dmathieu Here are my arguments for the factory pattern over struct duplication:

> if a new value is added, or one is removed then, it becomes trickier to handle since the root method has to know about everything.

In this case you would just add `nil` to the unused version, `networkHost` for example:

```go
func appendNetworkHost(attrs []attribute.KeyValue, h string) []attribute.KeyValue {
	return appendAttrs(attrs, semconv1210.NetPeerName, nil, h)
}
```

> It would also avoid having to define one method for each attribute.

This is pretty low complexity, exchanging 1 line of code for 3 to avoid refactoring the library.

> Finally, doing the separation this way also makes it easier to add a third version if we someday need it (I really hope we don't).

By design, there can never be a third version. The clarification of this point can be found here: https://github.com/open-telemetry/semantic-conventions/issues/1502. Additionally, this logic should be rolled back entirely once we release a stable version:

> we only considered using OTEL_SEMCONV_STABILITY_OPT_IN for the initial unstable to stable migration, and not subsequent major version bumps from stable -> (breaking) stable

pellared: @prestonvasquez, please first create a PR which refactors `otelhttp`. We need to try using similar patterns across the codebase to make it easier to maintain and more readable.

prestonvasquez: @pellared I'm concerned a bug could be introduced by refactoring so much code. I've updated otelmongo to use the otelhttp pattern.

instrumentation/go.mongodb.org/mongo-driver/mongo/otelmongo/internal/semconv/event_monitor.go
```diff
@@ -0,0 +1,166 @@
...
+
+	if parsedPort, err := strconv.Atoi(portStr); err == nil {
+		port = parsedPort
+	}
+
+	return host, port
+}
+
+// sanitizeCommand converts a BSON command to a sanitized JSON string.
+// TODO: Sanitize values where possible and limit maximum size.
```
dmathieu, line 103: Having this as two different TODOs made it clearer that it's two different things that will be tackled separately.

dmathieu (approved):

pellared (changes requested):
This PR does not follow the semantic conventions.
https://github.com/open-telemetry/semantic-conventions/issues/1502 has been kind of rejected.
Also https://github.com/open-telemetry/semantic-conventions/pull/1758

Sorry for the delay. I missed this PR because of winter holidays.

prestonvasquez:
@pellared Do you have an action item for this PR?

pellared:
> @pellared Do you have an action item for this PR?

mongo/dup -> database/dup
mongo -> database

prestonvasquez:
@pellared 

> Please remember to also update the PR description.

Could you clarify this?

pellared:
> @pellared
> 
> > Please remember to also update the PR description.
> 
> Could you clarify this?

It does not reflect current changes

pellared:
@prestonvasquez, I meant the PR description (not title) :wink:

CHANGELOG.md
```diff
@@ -72,6 +72,10 @@ migration](https://github.com/open-telemetry/semantic-conventions/blob/main/docs
 - `NewSDK` handles empty `OpenTelemetryConfiguration.Resource` properly in `go.opentelemetry.io/contrib/config/v0.3.0`. (#6606)
 - Fix a possible nil dereference panic in `NewSDK` of `go.opentelemetry.io/contrib/config/v0.3.0`. (#6606)
 
+### Changed 
+
+- Support both `go.opentelemetry.io/otel/semconv/v1.21.0` (default) and `go.opentelemetry.io/otel/semconv/v1.26.0` (opt-in) in `go.opentelemetry.io/contrib/instrumentation/go.mongodb.org/mongo-driver/mongo/otelmongo` per the [Database semantic convention stability migration guide](https://github.com/open-telemetry/semantic-conventions/blob/main/docs/non-normative/db-migration.md#database-semantic-convention-stability-migration-guide) (#6172)
```
pellared, line 77: permalink

```suggestion
- Support both `go.opentelemetry.io/otel/semconv/v1.21.0` (default) and `go.opentelemetry.io/otel/semconv/v1.26.0` (opt-in) in `go.opentelemetry.io/contrib/instrumentation/go.mongodb.org/mongo-driver/mongo/otelmongo` per the [Database semantic convention stability migration guide](https://github.com/open-telemetry/semantic-conventions/blob/cb11bb9bac24f4b0e95ad0f61ce01813d8ceada8/docs/non-normative/db-migration.md#database-semantic-convention-stability-migration-guide) (#6172)
```

prestonvasquez:
> @prestonvasquez, I meant the PR description (not title) :wink: 

Whoops I was doing this on my phone, guess I didn't hit save.

dmathieu (changes requested):

instrumentation/go.mongodb.org/mongo-driver/mongo/otelmongo/internal/semconv/event_monitor.go
dmathieu: Ping about this comment?

```suggestion
// TODO: Sanitize values where possible.
// TODO: Limit maximum size.
```

dmathieu (approved):

pellared (approved):

pellared:
@prestonvasquez, thank you for your work :+1:

PR #6185, opened 2024-10-03
config: return a noop SDK if disabled is set to true

codeboten:
This change ensures the config package returns a noop SDK if disabled has been set to true by the config.

Additionally, returning the noop SDK on errors to reduce the chances for end users to use an uninitialized sdk struct.


dashpole (approved):

MrAlias (approved):

pellared (approved):

CHANGELOG.md
```diff
@@ -18,6 +18,7 @@ This project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0.htm
 ### Changed
 
 - The function signature of `NewLogProcessor` in `go.opentelemetry.io/contrib/processors/minsev` has changed to accept the added `Severitier` interface instead of a `log.Severity`. (#6116)
+- `NewSDK` now returns a no-op SDK if `disabled` is set to `true` via configuration. (#6185)
```
dmathieu, line 21: We usually specify the exact package the change it made in, as there are so many in this repository.

```suggestion
- `NewSDK` in `go.opentelemetry.io/contrib/config` now returns a no-op SDK if `disabled` is set to `true`. (#6185)
```

PR #6205, opened 2024-10-04
Ensure all XRay Sampler functionality is under ParentBased logic

jj22ee:
This fixes the bug where:
- Currently, XRay Sampler in OTel Go will use `ShouldSample` logic for every span, but for child spans, they should always respect the parent span sampling decision in XRay Sampler (as is already done in OTel Java and OTel .NET)

The changes introduced in this PR are as follows:
- Wrap the returned `remote_sampler` of `NewRemoteSampler` under a single ParentBased Sampler

**Existing Issue(s):**
Related
- https://github.com/aws-observability/aws-otel-js-instrumentation/issues/79
- https://github.com/aws-observability/aws-otel-js-instrumentation/pull/80

**Testing:**

- Tested Sampler functionality with XRay Sampler test bed - https://github.com/aws-observability/aws-otel-community/tree/master/centralized-sampling-tests

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -24,6 +24,7 @@ This project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0.htm
 
 - Possible nil dereference panic in `go.opentelemetry.io/contrib/instrumentation/net/http/httptrace/otelhttptrace`. (#5965)
 - Non-200 HTTP status codes when retrieving sampling rules in `go.opentelemetry.io/contrib/samplers/aws/xray` now return an error. (#5718)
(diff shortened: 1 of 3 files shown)


wangzlei (approved):

dmathieu (changes requested):
👎 
Ensuring this is the responsibility of the developer using this sampler in their own application.
This change prevents folks from setting up their own parent-based logic and using that.

MrAlias:
Thank you for your interest in the project. We have recently removed this sampler (https://github.com/open-telemetry/opentelemetry-go-contrib/pull/6187) as it does not have an owner. I'm closing this as we are no longer accepting changes to this code.

Issue #5192, opened 2024-03-01
Add logr log bridge


pellared:
Add [logr.LogSink](https://pkg.go.dev/github.com/go-logr/logr#LogSink) log bridge.

Prototype implementation: https://github.com/pellared/opentelemetry-go/blob/147762ffb5c4394417b3f973274cc570723db79b/log/internal/logr.go

TODO:
- [x] Implementation: (draft: https://github.com/open-telemetry/opentelemetry-go-contrib/pull/5357)
- [x] Benchmarks
- [ ] Add to published modules
- [ ] Add to OpenTelemetry Registry. Example PR: https://github.com/open-telemetry/opentelemetry.io/pull/4857
- [ ] [Handle key-naming conventions](https://github.com/open-telemetry/opentelemetry-go-contrib/pull/5357#discussion_r1680750975) (separate issue?)

scorpionknifes:
Keen to pick this one up

pellared:
@scorpionknifes You are assigned. I suggest also looking at https://github.com/open-telemetry/opentelemetry-go-contrib/tree/main/bridges/otelslog as reference.

pellared:
@scorpionknifes, I think that as the point we can publish the `otellogr` module. I find it safer to track https://github.com/open-telemetry/opentelemetry-go-contrib/pull/5357#discussion_r1680750975 as a separate issue.

PR #6206, opened 2024-10-05
otellogr: Implement LevelSeverity, Enabled

scorpionknifes:
This PR implements a few methods for otellogr package
- Configurable LevelSeverity
- Enabled

Please add skip changelog thx 🙏

Part of https://github.com/open-telemetry/opentelemetry-go-contrib/issues/5192


bridges/otellogr/example_test.go
```diff
@@ -17,6 +18,17 @@ func Example() {
 	// Create an logr.Logger with *otellogr.LogSink and use it in your application.
 	logr.New(otellogr.NewLogSink(
 		"my/pkg/name",
-		otellogr.WithLoggerProvider(provider)),
-	)
+		otellogr.WithLoggerProvider(provider),
+		// Optionally, set the log level severity mapping.
+		otellogr.WithLevelSeverity(func(i int) log.Severity {
```
dmathieu, line 23: I wonder what the use case is for this. When would folks want to change the severity mapping?

scorpionknifes: The log level in logr is the following:
```
>= 0, higher meaning "less important"
```

dmathieu: Should we instead provide better default values?

bridges/otellogr/logsink.go
```diff
@@ -10,13 +10,26 @@
...
+// not provided it would default to a function that adds an offset to the logr
+// such that [logr.Info] is transformed to [log.SeverityInfo]. For example:
+//
+//   - [logr.Info] is transformed to [log.SeverityInfo].
+//   - [logr.V(0)] is transformed to [log.SeverityInfo].
+//   - [logr.V(1)] is transformed to [log.SeverityInfo2].
+//   - [logr.V(2)] is transformed to [log.SeverityInfo3].
+//   - ...
+//   - [logr.V(15)] is transformed to [log.SeverityFatal4].
+//   - [logr.Error] is transformed to [log.SeverityError].
```
pellared, line 31: This should be the other way.

> The higher the V-level of a log line, the less critical it is considered

See https://pkg.go.dev/github.com/go-logr/logr

pellared (changes requested):
The high V level - the less important (more verbose) logs are emitted

bridges/otellogr/example_test.go
pellared: The defaults are currently wrong but even if they would better we would need to provide such option.

From pkg documentation

> This gives developers a chance to indicate arbitrary grades of importance for info logs, without assigning names with semantic meaning such as "warning", "trace", and "debug." Superficially this may feel very similar, but the primary difference is the lack of semantics. Because verbosity is a numerical value, it's safe to assume that an app running with higher verbosity means more (and less important) logs will be generated.

I really feel we need to give the users the possibility to map the numerical value into the severity leval semantics of OTel.

Also, I think we should probably encourage people using other bridges whenever possible (like `otelslog`, `otelzap`, `otellogrus`) as these are less abstract and should provide better translations for people that are already using these libraries as "backend" for `logr`.

bridges/otellogr/logsink.go
```diff
@@ -10,13 +10,37 @@
...
+//
+//   - [logr.Info] is transformed to [log.SeverityInfo].
+//   - [logr.V(0)] is transformed to [log.SeverityInfo].
+//   - [logr.V(1)] is transformed to [log.SeverityDebug4].
+//   - [logr.V(2)] is transformed to [log.SeverityDebug3].
+//   - ...
+//   - [logr.V(8)] is transformed to [log.SeverityTrace].
+//   - [logr.V(9)] is transformed to [log.SeverityUndefined].
+//   - [logr.V(10)] is transformed to log.Severity(-1).
+//   - [logr.Error] is transformed to [log.SeverityError].
```
pellared, line 36: After some thoughts I have a lot simpler proposal inspired by https://git-scm.com/docs/git-branch#Documentation/git-branch.txt--v

```suggestion
// The V-level is transformed by using the [WithLevelSeverity] option. If this is
// not provided there are transformed in the following way:
//
//   - logr.Info] and logr.V(0) are transformed to [log.SeverityInfo].
//   - logr.V(1) is transformed to [log.SeverityDebug].
//   - logr.V(2) and higher are transformed to [log.SeverityTrace].
```

pellared (approved):

pellared:
@dmathieu, PTAL

bridges/otellogr/logsink.go
```diff
@@ -10,13 +10,27 @@
...
 //     is used.
 //
+// The V-level is transformed by using the [WithLevelSeverity] option. If option is
+// not provided then V-level is transformed in the following way:
+//
+//   - logr.Info and logr.V(0) are transformed to [log.SeverityInfo].
+//   - logr.V(1) is transformed to [log.SeverityDebug].
+//   - logr.V(2) and higher are transformed to [log.SeverityTrace].
+//
+// If possible, look at alternative log bridges that provide less abstract log
```
dmathieu, line 28: This comment is a bit weird. Is that intended at folks using the library, and telling them to look at how other bridges are doing it, or is that a todo?

pellared: Blame is on me.

I will add a context why I proposed it. Imagine you have an application that uses `logr` for emitting logs that is configured to have `zap` as its backend. It might be better if this application bridges the `zap` to use `otelzap` than `otellogr` to get more precise translations based on their current usage...

However, after thinking the second time I think it may indeed be a strange comment and probably it is better to remove it as it would probably just cause more confusion.

TL;DR;
Lets remove it.

pellared: https://github.com/open-telemetry/opentelemetry-go-contrib/pull/6206/commits/b171aad6ab794e21a60f27c355ea87fe38e7175d

dmathieu (approved):

Issue #6183, opened 2024-10-03
Lychee CI failure: 429 too many requests


MrAlias:
```
### Errors in ./RELEASING.md

* [429] [https://github.com/open-telemetry/opentelemetry-go/commits/main](https://github.com/open-telemetry/opentelemetry-go/commits/main) | Failed: Network error: Too Many Requests
```

Example: https://github.com/open-telemetry/opentelemetry-go-contrib/actions/runs/11151936706/job/30996462798?pr=6002

dmathieu:
The two fixes here could be to get https://github.com/lycheeverse/lychee/pull/1403 to move forward, so a links check that fails would still update the cache with URLs that succeeded, making retries only fetch the previously failing URLs, and fixing rate limit issues.

If we can't get that moving, we'll have to disable the links check temporarily, or have it consider 429s as success.

PR #6210, opened 2024-10-07
Temporarily consider 429s as success in links check

dmathieu:
This considers 429s as a success, until we can get lychee to cache success requests even if there's a failure.
This should only be a temporary measure. But ignoring one status code seems better than entirely disabling the check.

Related: #6183.


.github/workflows/links-fail-fast.yml
```diff
@@ -25,4 +25,4 @@ jobs:
           GITHUB_TOKEN: ${{ secrets.GITHUB_TOKEN }}
         with:
           fail: true
-          args: --max-concurrency 5 --cache --max-cache-age 1d .
+          args: --max-concurrency 5 --cache --max-cache-age 1d --accept 100..=103,200..=299,429 .
```
dashpole, line 28: can you add a TODO to remove the accept arg, since this is temporary?

dashpole: Just want to make sure we don't forget to revert this.

XSAM: And, add this issue link to the TODO comment: https://github.com/open-telemetry/opentelemetry-go-contrib/issues/6183

dashpole (approved):

XSAM (approved):

hanyuancheung (approved):

MrAlias (approved):

dmathieu:
The remaining failure is due to the recent lychee action upgrade, not the 429s.
Merging this, and I'll look into fixing the v2 issue in another PR.

PR #6215, opened 2024-10-07
config: add support for parsing env variables in configuration

codeboten:
This replaces the last bit of functionality that was opened in https://github.com/open-telemetry/opentelemetry-go-contrib/pull/4826 to support env variable replacement. Pulled the envprovider.go code from https://github.com/open-telemetry/opentelemetry-collector/blob/main/confmap/provider/envprovider/provider.go

Fixes https://github.com/open-telemetry/opentelemetry-go-contrib/issues/4373

### DIFF
--- a/config/config.go
+++ b/config/config.go
@@ -6,6 +6,7 @@ package config // import "go.opentelemetry.io/contrib/config"
 import (
 	"context"
 	"errors"
+	"regexp"
 
 	"gopkg.in/yaml.v3"
 
@@ -142,6 +143,20 @@ func WithOpenTelemetryConfiguration(cfg OpenTelemetryConfiguration) Configuratio
 
 // ParseYAML parses a YAML configuration file into an OpenTelemetryConfiguration.
 func ParseYAML(file []byte) (*OpenTelemetryConfiguration, error) {
+	re := regexp.MustCompile(`\$\{([a-zA-Z_][a-zA-Z0-9_]*[-]?.*)\}`)
+
+	replaceEnvVars := func(input []byte) []byte {
+		return re.ReplaceAllFunc(input, func(s []byte) []byte {
+			match := re.FindSubmatch(s)
+			if len(match) < 2 {
+				return s
+			}
+			return replaceEnvVar(string(match[1]))
+		})
+	}
+
+	file = replaceEnvVars(file)
+
 	var cfg OpenTelemetryConfiguration
 	err := yaml.Unmarshal(file, &cfg)
 	if err != nil {
--- a/config/envprovider.go
+++ b/config/envprovider.go
@@ -1,0 +1,38 @@
+// Copyright The OpenTelemetry Authors
+// SPDX-License-Identifier: Apache-2.0
+
+package config // import "go.opentelemetry.io/contrib/config"
+
+import (
+	"os"
+	"regexp"
+	"strings"
+)
+
+const ValidationPattern = `^[a-zA-Z_][a-zA-Z0-9_]*$`
+
+var validationRegexp = regexp.MustCompile(ValidationPattern)
+
+func replaceEnvVar(uri string) []byte {
+	envVarName, defaultValuePtr := parseEnvVarURI(uri)
+	if !validationRegexp.MatchString(envVarName) {
+		return nil
+	}
+
+	val, exists := os.LookupEnv(envVarName)
+	if !exists {
+		if defaultValuePtr != nil {
+			val = *defaultValuePtr
+		}
+	}
+	return []byte(val)
+}
+
+func parseEnvVarURI(uri string) (string, *string) {
+	const defaultSuffix = ":-"
+	if strings.Contains(uri, defaultSuffix) {
+		parts := strings.SplitN(uri, defaultSuffix, 2)
+		return parts[0], &parts[1]
+	}
+	return uri, nil
+}
--- a/config/config_test.go
+++ b/config/config_test.go
@@ -332,6 +332,40 @@ func TestParseYAML(t *testing.T) {
 	}
 }
 
+func TestParseYAMLWithEnvironmentVariables(t *testing.T) {
+	tests := []struct {
+		name     string
+		input    string
+		wantErr  error
+		wantType interface{}
+	}{
+		{
+			name:     "valid v0.2 config with env vars",
+			input:    "v0.2-env-var.yaml",
+			wantType: &v02OpenTelemetryConfig,
+		},
+	}
+
+	t.Setenv("OTEL_SDK_DISABLED", "false")
+	t.Setenv("OTEL_ATTRIBUTE_VALUE_LENGTH_LIMIT", "4096")
+	t.Setenv("OTEL_EXPORTER_OTLP_PROTOCOL", "http/protobuf")
+
+	for _, tt := range tests {
+		t.Run(tt.name, func(t *testing.T) {
+			b, err := os.ReadFile(filepath.Join("testdata", tt.input))
+			require.NoError(t, err)
+
+			got, err := ParseYAML(b)
+			if tt.wantErr != nil {
+				require.Equal(t, tt.wantErr.Error(), err.Error())
+			} else {
+				require.NoError(t, err)
+				assert.Equal(t, tt.wantType, got)
+			}
+		})
+	}
+}
+
 func TestSerializeJSON(t *testing.T) {
 	tests := []struct {
 		name     string
--- a/config/testdata/v0.2-env-var.yaml
+++ b/config/testdata/v0.2-env-var.yaml
@@ -1,0 +1,428 @@
+# kitchen-sink.yaml demonstrates all configurable surface area, including explanatory comments.
+#
+# It DOES NOT represent expected real world configuration, as it makes strange configuration
+# choices in an effort to exercise the full surface area.
+#
+# Configuration values are set to their defaults when default values are defined.
+
+# The file format version
+file_format: ${fileformat:-0.2}
+
+# Configure if the SDK is disabled or not. This is not required to be provided
+# to ensure the SDK isn't disabled, the default value when this is not provided
+# is for the SDK to be enabled.
+#
+# Environment variable: OTEL_SDK_DISABLED
+disabled: ${OTEL_SDK_DISABLED}
+
+# Configure general attribute limits. See also tracer_provider.limits, logger_provider.limits.
+attribute_limits:
+  # Configure max attribute value size.
+  #
+  # Environment variable: OTEL_ATTRIBUTE_VALUE_LENGTH_LIMIT
+  attribute_value_length_limit: ${OTEL_ATTRIBUTE_VALUE_LENGTH_LIMIT}
+  # Configure max attribute count.
+  #
+  # Environment variable: OTEL_ATTRIBUTE_COUNT_LIMIT
+  attribute_count_limit: 128
+
+# Configure logger provider.
+logger_provider:
+  # Configure log record processors.
+  processors:
+    # Configure a batch log record processor.
+    - batch:
+        # Configure delay interval (in milliseconds) between two consecutive exports.
+        #
+        # Environment variable: OTEL_BLRP_SCHEDULE_DELAY
+        schedule_delay: 5000
+        # Configure maximum allowed time (in milliseconds) to export data.
+        #
+        # Environment variable: OTEL_BLRP_EXPORT_TIMEOUT
+        export_timeout: 30000
+        # Configure maximum queue size.
+        #
+        # Environment variable: OTEL_BLRP_MAX_QUEUE_SIZE
+        max_queue_size: 2048
+        # Configure maximum batch size.
+        #
+        # Environment variable: OTEL_BLRP_MAX_EXPORT_BATCH_SIZE
+        max_export_batch_size: 512
+        # Configure exporter.
+        #
+        # Environment variable: OTEL_LOGS_EXPORTER
+        exporter:
+          # Configure exporter to be OTLP.
+          otlp:
+            # Configure protocol.
+            #
+            # Environment variable: OTEL_EXPORTER_OTLP_PROTOCOL, OTEL_EXPORTER_OTLP_LOGS_PROTOCOL
+            protocol: ${OTEL_EXPORTER_OTLP_PROTOCOL}
+            # Configure endpoint.
+            #
+            # Environment variable: OTEL_EXPORTER_OTLP_ENDPOINT, OTEL_EXPORTER_OTLP_LOGS_ENDPOINT
+            endpoint: http://localhost:4318
+            # Configure certificate.
+            #
+            # Environment variable: OTEL_EXPORTER_OTLP_CERTIFICATE, OTEL_EXPORTER_OTLP_LOGS_CERTIFICATE
+            certificate: /app/cert.pem
+            # Configure mTLS private client key.
+            #
+            # Environment variable: OTEL_EXPORTER_OTLP_CLIENT_KEY, OTEL_EXPORTER_OTLP_LOGS_CLIENT_KEY
+            client_key: /app/cert.pem
+            # Configure mTLS client certificate.
+            #
+            # Environment variable: OTEL_EXPORTER_OTLP_CLIENT_CERTIFICATE, OTEL_EXPORTER_OTLP_LOGS_CLIENT_CERTIFICATE
+            client_certificate: /app/cert.pem
+            # Configure headers.
+            #
+            # Environment variable: OTEL_EXPORTER_OTLP_HEADERS, OTEL_EXPORTER_OTLP_LOGS_HEADERS
+            headers:
+              api-key: "1234"
+            # Configure compression.
+            #
+            # Environment variable: OTEL_EXPORTER_OTLP_COMPRESSION, OTEL_EXPORTER_OTLP_LOGS_COMPRESSION
+            compression: gzip
+            # Configure max time (in milliseconds) to wait for each export.
+            #
+            # Environment variable: OTEL_EXPORTER_OTLP_TIMEOUT, OTEL_EXPORTER_OTLP_LOGS_TIMEOUT
+            timeout: 10000
+            # Configure client transport security for the exporter's connection.
+            #
+            # Environment variable: OTEL_EXPORTER_OTLP_INSECURE, OTEL_EXPORTER_OTLP_LOGS_INSECURE
+            insecure: false
+    # Configure a simple span processor.
+    - simple:
+        # Configure exporter.
+        exporter:
+          # Configure exporter to be console.
+          console: {}
+  # Configure log record limits. See also attribute_limits.
+  limits:
+    # Configure max log record attribute value size. Overrides attribute_limits.attribute_value_length_limit.
+    #
+    # Environment variable: OTEL_LOGRECORD_ATTRIBUTE_VALUE_LENGTH_LIMIT
+    attribute_value_length_limit: 4096
+    # Configure max log record attribute count. Overrides attribute_limits.attribute_count_limit.
+    #
+    # Environment variable: OTEL_LOGRECORD_ATTRIBUTE_COUNT_LIMIT
+    attribute_count_limit: 128
+
+# Configure meter provider.
+meter_provider:
+  # Configure metric readers.
+  readers:
+    # Configure a pull-based metric reader.
+    - pull:
+        # Configure exporter.
+        #
+        # Environment variable: OTEL_METRICS_EXPORTER
+        exporter:
+          # Configure exporter to be prometheus.
+          prometheus:
+            # Configure host.
+            #
+            # Environment variable: OTEL_EXPORTER_PROMETHEUS_HOST
+            host: localhost
+            # Configure port.
+            #
+            # Environment variable: OTEL_EXPORTER_PROMETHEUS_PORT
+            port: 9464
+            # Configure Prometheus Exporter to produce metrics without a unit suffix or UNIT metadata.
+            without_units: false
+            # Configure Prometheus Exporter to produce metrics without a type suffix.
+            without_type_suffix: false
+            # Configure Prometheus Exporter to produce metrics without a scope info metric.
+            without_scope_info: false
+            # Configure Prometheus Exporter to add resource attributes as metrics attributes.
+            with_resource_constant_labels:
+              # Configure resource attributes to be included, in this example attributes starting with service.
+              included:
+                - "service*"
+              # Configure resource attributes to be excluded, in this example attribute service.attr1.
+              excluded:
+                - "service.attr1"
+    # Configure a periodic metric reader.
+    - periodic:
+        # Configure delay interval (in milliseconds) between start of two consecutive exports.
+        #
+        # Environment variable: OTEL_METRIC_EXPORT_INTERVAL
+        interval: 5000
+        # Configure maximum allowed time (in milliseconds) to export data.
+        #
+        # Environment variable: OTEL_METRIC_EXPORT_TIMEOUT
+        timeout: 30000
+        # Configure exporter.
+        #
+        # Environment variable: OTEL_METRICS_EXPORTER
+        exporter:
+          # Configure exporter to be OTLP.
+          otlp:
+            # Configure protocol.
+            #
+            # Environment variable: OTEL_EXPORTER_OTLP_PROTOCOL, OTEL_EXPORTER_OTLP_METRICS_PROTOCOL
+            protocol: http/protobuf
+            # Configure endpoint.
+            #
+            # Environment variable: OTEL_EXPORTER_OTLP_ENDPOINT, OTEL_EXPORTER_OTLP_METRICS_ENDPOINT
+            endpoint: http://localhost:4318
+            # Configure certificate.
+            #
+            # Environment variable: OTEL_EXPORTER_OTLP_CERTIFICATE, OTEL_EXPORTER_OTLP_METRICS_CERTIFICATE
+            certificate: /app/cert.pem
+            # Configure mTLS private client key.
+            #
+            # Environment variable: OTEL_EXPORTER_OTLP_CLIENT_KEY, OTEL_EXPORTER_OTLP_METRICS_CLIENT_KEY
+            client_key: /app/cert.pem
+            # Configure mTLS client certificate.
+            #
+            # Environment variable: OTEL_EXPORTER_OTLP_CLIENT_CERTIFICATE, OTEL_EXPORTER_OTLP_METRICS_CLIENT_CERTIFICATE
+            client_certificate: /app/cert.pem
+            # Configure headers.
+            #
+            # Environment variable: OTEL_EXPORTER_OTLP_HEADERS, OTEL_EXPORTER_OTLP_METRICS_HEADERS
+            headers:
+              api-key: !!str 1234
+            # Configure compression.
+            #
+            # Environment variable: OTEL_EXPORTER_OTLP_COMPRESSION, OTEL_EXPORTER_OTLP_METRICS_COMPRESSION
+            compression: gzip
+            # Configure max time (in milliseconds) to wait for each export.
+            #
+            # Environment variable: OTEL_EXPORTER_OTLP_TIMEOUT, OTEL_EXPORTER_OTLP_METRICS_TIMEOUT
+            timeout: 10000
+            # Configure client transport security for the exporter's connection.
+            #
+            # Environment variable: OTEL_EXPORTER_OTLP_INSECURE, OTEL_EXPORTER_OTLP_METRICS_INSECURE
+            insecure: false
+            # Configure temporality preference.
+            #
+            # Environment variable: OTEL_EXPORTER_OTLP_METRICS_TEMPORALITY_PREFERENCE
+            temporality_preference: delta
+            # Configure default histogram aggregation.
+            #
+            # Environment variable: OTEL_EXPORTER_OTLP_METRICS_DEFAULT_HISTOGRAM_AGGREGATION
+            default_histogram_aggregation: base2_exponential_bucket_histogram
+    # Configure a periodic metric reader.
+    - periodic:
+        # Configure exporter.
+        exporter:
+          # Configure exporter to be console.
+          console: {}
+  # Configure views. Each view has a selector which determines the instrument(s) it applies to, and a configuration for the resulting stream(s).
+  views:
+    # Configure a view.
+    - selector:
+        # Configure instrument name selection criteria.
+        instrument_name: my-instrument
+        # Configure instrument type selection criteria.
+        instrument_type: histogram
+        # Configure the instrument unit selection criteria.
+        unit: ms
+        # Configure meter name selection criteria.
+        meter_name: my-meter
+        # Configure meter version selection criteria.
+        meter_version: 1.0.0
+        # Configure meter schema url selection criteria.
+        meter_schema_url: https://opentelemetry.io/schemas/1.16.0
+      # Configure stream.
+      stream:
+        # Configure metric name of the resulting stream(s).
+        name: new_instrument_name
+        # Configure metric description of the resulting stream(s).
+        description: new_description
+        # Configure aggregation of the resulting stream(s). Known values include: default, drop, explicit_bucket_histogram, base2_exponential_bucket_histogram, last_value, sum.
+        aggregation:
+          # Configure aggregation to be explicit_bucket_histogram.
+          explicit_bucket_histogram:
+            # Configure bucket boundaries.
+            boundaries: [ 0.0, 5.0, 10.0, 25.0, 50.0, 75.0, 100.0, 250.0, 500.0, 750.0, 1000.0, 2500.0, 5000.0, 7500.0, 10000.0 ]
+            # Configure record min and max.
+            record_min_max: true
+        # Configure attribute keys retained in the resulting stream(s).
+        attribute_keys:
+          - key1
+          - key2
+
+# Configure text map context propagators.
+#
+# Environment variable: OTEL_PROPAGATORS
+propagator:
+  composite: [tracecontext, baggage, b3, b3multi, jaeger, xray, ottrace]
+
+# Configure tracer provider.
+tracer_provider:
+  # Configure span processors.
+  processors:
+    # Configure a batch span processor.
+    - batch:
+        # Configure delay interval (in milliseconds) between two consecutive exports.
+        #
+        # Environment variable: OTEL_BSP_SCHEDULE_DELAY
+        schedule_delay: 5000
+        # Configure maximum allowed time (in milliseconds) to export data.
+        #
+        # Environment variable: OTEL_BSP_EXPORT_TIMEOUT
+        export_timeout: 30000
+        # Configure maximum queue size.
+        #
+        # Environment variable: OTEL_BSP_MAX_QUEUE_SIZE
+        max_queue_size: 2048
+        # Configure maximum batch size.
+        #
+        # Environment variable: OTEL_BSP_MAX_EXPORT_BATCH_SIZE
+        max_export_batch_size: 512
+        # Configure exporter.
+        #
+        # Environment variable: OTEL_TRACES_EXPORTER
+        exporter:
+          # Configure exporter to be OTLP.
+          otlp:
+            # Configure protocol.
+            #
+            # Environment variable: OTEL_EXPORTER_OTLP_PROTOCOL, OTEL_EXPORTER_OTLP_TRACES_PROTOCOL
+            protocol: http/protobuf
+            # Configure endpoint.
+            #
+            # Environment variable: OTEL_EXPORTER_OTLP_ENDPOINT, OTEL_EXPORTER_OTLP_TRACES_ENDPOINT
+            endpoint: http://localhost:4318
+            # Configure certificate.
+            #
+            # Environment variable: OTEL_EXPORTER_OTLP_CERTIFICATE, OTEL_EXPORTER_OTLP_TRACES_CERTIFICATE
+            certificate: /app/cert.pem
+            # Configure mTLS private client key.
+            #
+            # Environment variable: OTEL_EXPORTER_OTLP_CLIENT_KEY, OTEL_EXPORTER_OTLP_TRACES_CLIENT_KEY
+            client_key: /app/cert.pem
+            # Configure mTLS client certificate.
+            #
+            # Environment variable: OTEL_EXPORTER_OTLP_CLIENT_CERTIFICATE, OTEL_EXPORTER_OTLP_TRACES_CLIENT_CERTIFICATE
+            client_certificate: /app/cert.pem
+            # Configure headers.
+            #
+            # Environment variable: OTEL_EXPORTER_OTLP_HEADERS, OTEL_EXPORTER_OTLP_TRACES_HEADERS
+            headers:
+              api-key: !!str 1234
+            # Configure compression.
+            #
+            # Environment variable: OTEL_EXPORTER_OTLP_COMPRESSION, OTEL_EXPORTER_OTLP_TRACES_COMPRESSION
+            compression: gzip
+            # Configure max time (in milliseconds) to wait for each export.
+            #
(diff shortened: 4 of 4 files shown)


config/envprovider.go
```diff
@@ -0,0 +1,38 @@
...
+
+var validationRegexp = regexp.MustCompile(ValidationPattern)
+
+func replaceEnvVar(uri string) []byte {
+	envVarName, defaultValuePtr := parseEnvVarURI(uri)
+	if !validationRegexp.MatchString(envVarName) {
+		return nil
+	}
+
+	val, exists := os.LookupEnv(envVarName)
```
dmathieu, line 22: Should there be an allow list of env vars we can retrieve here, to avoid possible security problems?
For example, do we want to allow folks to inject `PATH` into the config?

codeboten: I suppose it would be up to the user if they set the configuration to include the whichever environment variable they choose to include.

It's true that someone could leak sensitive information through this mechanism, but it seems too restrictive to have an allow list. Either ways maybe this could be discussed further in the spec? Currently there are no such restrctions defined https://github.com/open-telemetry/opentelemetry-specification/blob/main/specification/configuration/data-model.md#environment-variable-substitution

dmathieu (dismissed):

config/envprovider.go
pellared, line 22: From https://github.com/open-telemetry/opentelemetry-specification/blob/main/specification/configuration/sdk-environment-variables.md#parsing-empty-value

Reasons are here: https://github.com/open-telemetry/opentelemetry-specification/pull/2045. Maybe I should add the reasons to spec as well, but it was I think my first contribution to the specification.

config/config.go
```diff
@@ -142,6 +143,20 @@ func WithOpenTelemetryConfiguration(cfg OpenTelemetryConfiguration) Configuratio
...
+		return re.ReplaceAllFunc(input, func(s []byte) []byte {
+			match := re.FindSubmatch(s)
+			if len(match) < 2 {
+				return s
+			}
+			return replaceEnvVar(string(match[1]))
+		})
+	}
+
+	file = replaceEnvVars(file)
```
pellared, line 158: The env var parsing is not compliant with the specification: https://github.com/open-telemetry/opentelemetry-specification/blob/main/specification/configuration/data-model.md#environment-variable-substitution

> Environment variable substitution MUST only apply to scalar values. Mapping keys are not candidates for substitution.

and

> It MUST NOT be possible to inject YAML structures by environment variables.

Are these handled anywhere? Can we add unit tests for unallowed substitutions?

pellared (changes requested):

config/config.go
pellared: I have a feeling that implementing it in a safe manner would require implementing https://pkg.go.dev/gopkg.in/yaml.v3#Unmarshaler for each scalar type (and using these scalar types instead of the regular ones). 

Such parsing would need to couple the package to some concrete YAML library (I am not against it).

pellared: The alternative could be to implement https://pkg.go.dev/encoding/json#Unmarshaler and use https://github.com/kubernetes-sigs/yaml. Maybe it would be simpler.

codeboten: Would using something like this work here? https://github.com/open-telemetry/opentelemetry-collector/blob/28d0d57c358d850166c7a0d14eaad3d162e2ce56/confmap/provider.go#L234

pellared: > Would using something like this work here? https://github.com/open-telemetry/opentelemetry-collector/blob/28d0d57c358d850166c7a0d14eaad3d162e2ce56/confmap/provider.go#L234

Maybe. You would have to elaborate as I do not know how would like to use it (or maybe simply try). However, notice that probably we would not want to support `[]any` and `map[string]any`.

MrAlias:

codeboten:
@MrAlias moved the code to the new package, this is no longer blocked by that. There's other comments to address that I'll get back to this week

config/config.go
codeboten: @pellared i've updated the code with what i think is a minimal solution that works. i've added unit tests to validate that the yaml parsing isn't supporting yaml structs, and that scalar values defined in the spec are supported.

config/envprovider.go
codeboten: an unset variable will now set the value to nil, which will be parsed by the yaml parser as an unset value

codeboten:
CLA faillures is being investigated https://github.com/communitybridge/easycla/issues/4605

CHANGELOG.md
```diff
@@ -50,6 +50,7 @@ migration](https://github.com/open-telemetry/semantic-conventions/blob/main/docs
 - Add support for configuring `HeadersList` field for OTLP exporters in `go.opentelemetry.io/contrib/config`. (#6657)
 - Add `go.opentelemetry.io/contrib/otelconf` module which is a replacement for `go.opentelemetry.io/contrib/config`. (#6796)
 - Added `WithFallbackLogExporter` to allow setting a fallback log exporter when `OTEL_LOGS_EXPORTER` is unset in `go.opentelemetry.io/contrib/exporters/autoexport`. (#6844)
+- Parsing via ParseYAML in `go.opentelemetry.io/contrib/config` now support environment variables in the format `${VAR_NAME}`. (#6215)
```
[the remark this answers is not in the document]
dmathieu:
```suggestion
- Parsing via ParseYAML in `go.opentelemetry.io/contrib/otelconf` now support environment variables in the format `${VAR_NAME}`. (#6215)
```

codeboten:
@pellared would love a review from you as you've requested changes which I made

pellared:
> @pellared would love a review from you as you've requested changes which I made

I will do my best to review this week. Thanks a lot.

otelconf/internal/provider/envprovider.go
```diff
@@ -0,0 +1,89 @@
...
+	if !validationRegexp.MatchString(envVarName) {
+		return nil, fmt.Errorf("invalid environment variable name: %s", envVarName)
+	}
+
+	val, exists := os.LookupEnv(envVarName)
+	if !exists {
+		if defaultValuePtr != nil {
+			val = *defaultValuePtr
+		}
+	}
```
pellared, line 50:
```suggestion
	val := os.Getenv(envVarName)
	if val == "" && defaultValuePtr != nil {
		val = *defaultValuePtr
	}
```

From https://github.com/open-telemetry/opentelemetry-specification/blob/main/specification/configuration/sdk-environment-variables.md#parsing-empty-value

Reasons are here: https://github.com/open-telemetry/opentelemetry-specification/pull/2045.

otelconf/v0.3.0/config_test.go
```diff
@@ -450,6 +450,40 @@ func TestParseYAML(t *testing.T) {
...
+	t.Setenv("OTEL_SDK_DISABLED", "false")
+	t.Setenv("OTEL_ATTRIBUTE_VALUE_LENGTH_LIMIT", "4096")
+	t.Setenv("OTEL_EXPORTER_OTLP_PROTOCOL", "http/protobuf")
+
+	for _, tt := range tests {
+		t.Run(tt.name, func(t *testing.T) {
+			b, err := os.ReadFile(filepath.Join("..", "testdata", tt.input))
+			require.NoError(t, err)
+
+			got, err := ParseYAML(b)
```
pellared, line 476: What about parsing JSON? Separate PR?

otelconf/internal/provider/envprovider.go
```diff
@@ -0,0 +1,89 @@
...
+
+func ReplaceEnvVars(input []byte) ([]byte, error) {
+	re := regexp.MustCompile(`\$\{([a-zA-Z_][a-zA-Z0-9_]*[-]?.*)\}`)
+
+	replaceEnvVars := func(input []byte) ([]byte, error) {
+		var err error
+		out := re.ReplaceAllFunc(input, func(s []byte) []byte {
+			match := re.FindSubmatch(s)
+			if len(match) < 2 {
+				return s
```
pellared, line 28: Can we add a test case to cover this line?

otelconf/internal/provider/envprovider.go
```diff
@@ -0,0 +1,89 @@
...
+
+func checkRawConfType(val []byte) error {
+	var rawConf any
+	err := yaml.Unmarshal(val, &rawConf)
+	if err != nil {
+		return err
+	}
+	if rawConf == nil {
+		return nil
+	}
```
pellared, line 80: Can we add a test case to cover this case?

CHANGELOG.md
```diff
@@ -17,6 +17,7 @@ This project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0.htm
 - Add `WithAttributes` option to set instrumentation scope attributes on the created `log.Logger` in `go.opentelemetry.io/contrib/bridges/otellogr`. (#6967)
 - Add the `WithGinMetricAttributes` option to allow setting dynamic, per-request metric attributes based on `*gin.Context` in `go.opentelemetry.io/contrib/instrumentation/github.com/gin-gonic/gin/otelgin`. (#6932)
 - Use Gin's own `ClientIP` method to detect the client's IP, which supports custom proxy headers in `go.opentelemetry.io/contrib/instrumentation/github.com/gin-gonic/gin/otelgin`. (#6095)
+- Parsing via ParseYAML in `go.opentelemetry.io/contrib/otelconf` now support environment variables in the format `${VAR_NAME}`. (#6215)
```
pellared, line 20:
```suggestion
- `ParseYAML` in `go.opentelemetry.io/contrib/otelconf` now support environment variables in the format `${VAR_NAME}`. (#6215)
```

otelconf/internal/provider/envprovider_test.go
```diff
@@ -0,0 +1,143 @@
...
+			if tt.wantErr != nil {
+				require.ErrorContains(t, err, tt.wantErr.Error())
+			} else {
+				require.NoError(t, err)
+			}
+		})
+	}
+}
+
+func TestReplaceEnvVars(t *testing.T) {
```
pellared, line 93: Can we add all test cases from https://github.com/open-telemetry/opentelemetry-specification/blob/main/specification/configuration/data-model.md#environment-variable-substitution (the ones that are in the table)?

otelconf/internal/provider/envprovider.go
codeboten: i couldn't think of a way to trigger this code path, since the same regexp is used both in the `ReplaceAllFunc` and in the `FindSubmatch`, the code should never get into a place where a `FindSubmatch` returns less than 2 elements in the slice. Removed the code

otelconf/internal/provider/envprovider.go
codeboten: added a test for empty data passed into `checkRawConfType` even though this should never be the case (the caller checks the length of the data passed in), validated the error reported would be useful.

CHANGELOG.md
```diff
@@ -17,6 +17,7 @@ This project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0.htm
 - Add `WithAttributes` option to set instrumentation scope attributes on the created `log.Logger` in `go.opentelemetry.io/contrib/bridges/otellogr`. (#6967)
 - Add the `WithGinMetricAttributes` option to allow setting dynamic, per-request metric attributes based on `*gin.Context` in `go.opentelemetry.io/contrib/instrumentation/github.com/gin-gonic/gin/otelgin`. (#6932)
 - Use Gin's own `ClientIP` method to detect the client's IP, which supports custom proxy headers in `go.opentelemetry.io/contrib/instrumentation/github.com/gin-gonic/gin/otelgin`. (#6095)
+- `ParseYAML` in `go.opentelemetry.io/contrib/otelconf` now support environment variables substitution in the format `${VAR_NAME}`. (#6215)
```
pellared, line 20:
```suggestion
- `ParseYAML` in `go.opentelemetry.io/contrib/otelconf` now support environment variables substitution in the format `${VAR_NAME}` and `${VAR_NAME:-default-value}. (#6215)
```

otelconf/v0.2.0/config.go
```diff
@@ -143,8 +144,13 @@ func WithOpenTelemetryConfiguration(cfg OpenTelemetryConfiguration) Configuratio
 
 // ParseYAML parses a YAML configuration file into an OpenTelemetryConfiguration.
```
pellared, line 145: I think we should document also the env var substitution somewhere in Go Doc. Here or/and in package documentation, maybe in example test... This should be handled in a separate PR as the package's documentation needs more care anyway: https://pkg.go.dev/go.opentelemetry.io/contrib/otelconf/v0.3.0.

pellared: I created https://github.com/open-telemetry/opentelemetry-go-contrib/issues/7037

Feel free to resolve

CHANGELOG.md
```diff
@@ -17,6 +17,7 @@ This project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0.htm
 - Add `WithAttributes` option to set instrumentation scope attributes on the created `log.Logger` in `go.opentelemetry.io/contrib/bridges/otellogr`. (#6967)
 - Add the `WithGinMetricAttributes` option to allow setting dynamic, per-request metric attributes based on `*gin.Context` in `go.opentelemetry.io/contrib/instrumentation/github.com/gin-gonic/gin/otelgin`. (#6932)
 - Use Gin's own `ClientIP` method to detect the client's IP, which supports custom proxy headers in `go.opentelemetry.io/contrib/instrumentation/github.com/gin-gonic/gin/otelgin`. (#6095)
+- `ParseYAML` in `go.opentelemetry.io/contrib/otelconf` now support environment variables substitution in the format `${VAR_NAME}` and `${VAR_NAME:-defaultvalue}. (#6215)
```
pellared, line 20: my bad...

```suggestion
- `ParseYAML` in `go.opentelemetry.io/contrib/otelconf` now support environment variables substitution in the format `${VAR_NAME}` and `${VAR_NAME:-defaultvalue}`. (#6215)
```

otelconf/v0.2.0/config.go
codeboten: Thanks!

pellared (commented):
Overall looks very good :+1: 

I would like only [this](https://github.com/open-telemetry/opentelemetry-go-contrib/pull/6215/files#r2017477935) to get resolved as e.g. I want to have tests that ensures that `$$` are escaped correctly :wink:

otelconf/internal/provider/envprovider_test.go
codeboten: Added a bunch more tests, still need to ensure all the scenarios are covered and passing, but it's underway

codeboten: There are 4 test cases missing, would it be ok to separate those somewhat edge cases into a followup PR? The env substitution support as is in this PR would be really useful

codeboten: found some time to address the remaining cases, i think all cases from the spec are covered now

otelconf/go.mod
```diff
@@ -26,6 +26,7 @@ require (
 	go.opentelemetry.io/proto/otlp v1.8.0
 	go.yaml.in/yaml/v3 v3.0.4
 	google.golang.org/grpc v1.75.1
+	gopkg.in/yaml.v3 v3.0.1
```
pellared: Use `go.yaml.in/yaml/v3` instead

CHANGELOG.md
```diff
@@ -8,6 +8,10 @@ This project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0.htm
 
 ## [Unreleased]
 
+### Added
+
+- `ParseYAML` in `go.opentelemetry.io/contrib/otelconf` now supports environment variables substitution in the format `${VAR_NAME}` and `${VAR_NAME:-defaultvalue}`. (#6215)
```
pellared: What about the `env:` prefix?

otelconf/testdata/v0.3-env-var.yaml
pellared, line 1: I quickly fuzzed the implementation

```go
func FuzzYAMLWithEnvVars(f *testing.F) {
	b, err := os.ReadFile(filepath.Join("..", "testdata", "v0.3-env-var.yaml"))
	require.NoError(f, err)

	// Add example values for fuzzing - YAML data and all env var values.
	f.Add(b, "false", "4096", "test_string", "true", "42", "3.14", "0xFF", "invalid", "dynamic_key", "replaced_value", "value\\nwith\\tescape")

	f.Fuzz(func(t *testing.T, data []byte, otelSDKDisabled, otelAttrValueLengthLimit, stringValue, boolValue, intValue, floatValue, hexValue, invalidMapValue, envVarInKey, replaceMe, valueWithEscape string) {
		t.Log("YAML with env vars:\n" + string(data))

		// Helper function to check if environment variable value is valid.
		isValidEnvValue := func(value string) bool {
			// Environment variable values cannot contain null bytes.
			for _, b := range []byte(value) {
				if b == 0 {
					return false
				}
			}
			return true
		}

		// Set environment variables used in the test YAML with fuzzed values.
		// Skip if any value contains invalid characters.
		envVars := map[string]string{
			"OTEL_SDK_DISABLED":                 otelSDKDisabled,
			"OTEL_ATTRIBUTE_VALUE_LENGTH_LIMIT": otelAttrValueLengthLimit,
			"STRING_VALUE":                      stringValue,
			"BOOL_VALUE":                        boolValue,
			"INT_VALUE":                         intValue,
			"FLOAT_VALUE":                       floatValue,
			"HEX_VALUE":                         hexValue,
			"INVALID_MAP_VALUE":                 invalidMapValue,
			"ENV_VAR_IN_KEY":                    envVarInKey,
			"REPLACE_ME":                        replaceMe,
			"VALUE_WITH_ESCAPE":                 valueWithEscape,
		}

		for key, value := range envVars {
			if !isValidEnvValue(value) {
				t.Skipf("Skipping test due to invalid env var value for %s", key)
			}
			t.Setenv(key, value)
		}

		cfg, err := ParseYAML(data)
		if err != nil {
			return
		}

		sdk, err := NewSDK(WithOpenTelemetryConfiguration(*cfg))
		if err != nil {
			return
		}

		ctx, cancel := context.WithTimeout(t.Context(), time.Millisecond)
		defer cancel()
		_ = sdk.Shutdown(ctx)
	})
}
```

```
$ go test -fuzz=FuzzYAMLWithEnvVars -fuzztime=1m
fuzz: elapsed: 1m1s, execs: 59962 (0/sec), new interesting: 7 (total: 21)
PASS
ok      go.opentelemetry.io/contrib/otelconf/v0.3.0     61.109s
```

I think I can add this fuzz test in a separate PR.

otelconf/internal/provider/envprovider.go
```diff
@@ -0,0 +1,96 @@
...
+const validationPattern = `^[a-zA-Z_][a-zA-Z0-9_]*$`
+
+var validationRegexp = regexp.MustCompile(validationPattern)
+
+func EscapeAndEncodeDollarSigns(input []byte) []byte {
+	return []byte(strings.ReplaceAll(string(input), "$$", "$|"))
+}
+
+func DecodeDollarSigns(input []byte) []byte {
+	return []byte(strings.ReplaceAll(string(input), "$|", "$"))
```
pellared: Why do you use `|` as a "marker character". This can be used in the YAML file.

`- x: asd$|2` is a valid YAML. Right now such YAML would be decoded into `- x: asd$2`

otelconf/v0.3.0/config.go
```diff
@@ -179,8 +181,15 @@ func WithTracerProviderOptions(opts ...sdktrace.TracerProviderOption) Configurat
 
 // ParseYAML parses a YAML configuration file into an OpenTelemetryConfiguration.
 func ParseYAML(file []byte) (*OpenTelemetryConfiguration, error) {
+	file, err := provider.ReplaceEnvVars(provider.EscapeAndEncodeDollarSigns(file))
+	if err != nil {
+		return nil, err
+	}
+
+	file = provider.DecodeDollarSigns(file)
```
pellared: I would strongly prefer if this would be a single function call. Also the unit tests for such function would be more natural and we could validate if [the env var substitution](https://github.com/open-telemetry/opentelemetry-specification/blob/main/specification/configuration/data-model.md#environment-variable-substitution) works as expected.

Right now for debugging purposed I extracted

```go
func preprocessYAML(file []byte) ([]byte, error) {
	file, err := provider.ReplaceEnvVars(provider.EscapeAndEncodeDollarSigns(file))
	if err != nil {
		return nil, err
	}

	file = provider.DecodeDollarSigns(file)
	return file, nil
}
```

so that I can create some unit tests

```go
func TestPreprocessYAML(t *testing.T) {
	tests := []struct {
		name     string
		input    string
		setupEnv func(t *testing.T)
		want     string
		wantErr  bool
	}{
		{
			name:  "no environment variables",
			input: "key: value\nother: data",
			want:  "key: value\nother: data",
		},
		{
			name:     "simple environment variable substitution",
			input:    "key: ${TEST_VAR}",
			setupEnv: func(t *testing.T) { t.Setenv("TEST_VAR", "test_value") },
			want:     "key: test_value",
		},
		{
			name:  "undefined environment variable",
			input: "key: ${UNDEFINED_VAR}",
			want:  "key: ",
		},
		{
			name:  "environment variable with default value",
			input: "key: ${UNDEFINED_VAR:-default_value}",
			want:  "key: default_value",
		},
		{
			name:     "environment variable with default value when var is set",
			input:    "key: ${DEFINED_VAR:-default_value}",
			setupEnv: func(t *testing.T) { t.Setenv("DEFINED_VAR", "actual_value") },
			want:     "key: actual_value",
		},
		{
			name:  "escaped dollar sign - single escape",
			input: "key: $${NOT_REPLACED}",
			want:  "key: ${NOT_REPLACED}",
		},
		{
			name:  "escaped dollar sign - double escape produces single dollar",
			input: "key: $$${NOT_REPLACED}",
			want:  "key: $",
		},
		{
			name:  "escaped dollar sign - triple escape",
			input: "key: $$$${NOT_REPLACED}",
			want:  "key: $${NOT_REPLACED}",
		},
		{
			name:     "mixed escaped and unescaped",
			input:    "key1: ${REPLACE_ME}\nkey2: $${NOT_REPLACED}",
			setupEnv: func(t *testing.T) { t.Setenv("REPLACE_ME", "replaced") },
			want:     "key1: replaced\nkey2: ${NOT_REPLACED}",
		},
		{
			name:     "environment variable in key position",
			input:    "${KEY_VAR}: value",
			setupEnv: func(t *testing.T) { t.Setenv("KEY_VAR", "dynamic_key") },
			want:     "dynamic_key: value",
		},
		{
			name:  "multiple environment variables in same line",
			input: "key: ${VAR1} and ${VAR2}",
			setupEnv: func(t *testing.T) {
				t.Setenv("VAR1", "first")
				t.Setenv("VAR2", "second")
			},
			want: "key: first and second",
		},
		{
			name:  "environment variable with spaces in default",
			input: "key: ${UNDEFINED:-default with spaces}",
			want:  "key: default with spaces",
		},
		{
			name:     "nested env vars in default are treated literally",
			input:    "key: ${UNDEFINED:-${FALLBACK_VAR}}",
			setupEnv: func(t *testing.T) { t.Setenv("FALLBACK_VAR", "fallback_value") },
			want:     "key: ${FALLBACK_VAR}",
		},
		{
			name:     "boolean environment variable",
			input:    "enabled: ${BOOL_VAR}",
			setupEnv: func(t *testing.T) { t.Setenv("BOOL_VAR", "true") },
			want:     "enabled: true",
		},
		{
			name:     "numeric environment variable",
			input:    "count: ${NUM_VAR}",
			setupEnv: func(t *testing.T) { t.Setenv("NUM_VAR", "42") },
			want:     "count: 42",
		},
		{
			name:     "hex environment variable",
			input:    "value: ${HEX_VAR}",
			setupEnv: func(t *testing.T) { t.Setenv("HEX_VAR", "0xFF") },
			want:     "value: 0xFF",
		},
		{
			name:     "alternative env syntax",
			input:    "key: ${env:TEST_VAR}",
			setupEnv: func(t *testing.T) { t.Setenv("TEST_VAR", "env_value") },
			want:     "key: env_value",
		},
		{
			name:     "quoted environment variable",
			input:    "key: \"${QUOTED_VAR}\"",
			setupEnv: func(t *testing.T) { t.Setenv("QUOTED_VAR", "quoted_value") },
			want:     "key: \"quoted_value\"",
		},
		{
			name:     "environment variable with special characters",
			input:    "key: ${SPECIAL_VAR}",
			setupEnv: func(t *testing.T) { t.Setenv("SPECIAL_VAR", "value\\nwith\\tescape") },
			want:     "key: value\\nwith\\tescape",
		},
		{
			name:  "escape sequence in regular text",
			input: "key: a $$ b",
			want:  "key: a $ b",
		},
		{
			name:  "no escape sequence with single dollar",
			input: "key: a $ b",
			want:  "key: a $ b",
		},
		{
			name: "complex YAML with multiple substitutions",
			input: `service:
  name: ${SERVICE_NAME:-default-service}
  version: ${SERVICE_VERSION}
config:
  endpoint: ${ENDPOINT}
  escaped: $${NOT_REPLACED}`,
			setupEnv: func(t *testing.T) {
				t.Setenv("SERVICE_VERSION", "1.0.0")
				t.Setenv("ENDPOINT", "http://localhost:8080")
			},
			want: `service:
  name: default-service
  version: 1.0.0
config:
  endpoint: http://localhost:8080
  escaped: ${NOT_REPLACED}`,
		},
		{
			name:     "YAML injection causes error",
			input:    "key: ${MALICIOUS_VAR}",
			setupEnv: func(t *testing.T) { t.Setenv("MALICIOUS_VAR", "value\\nkey2: injected") },
			wantErr:  true,
		},
		{
			name:     "error case - invalid YAML type",
			input:    "key: ${INVALID_TYPE_VAR}",
			setupEnv: func(t *testing.T) { t.Setenv("INVALID_TYPE_VAR", "!!int NaN") },
			wantErr:  true,
		},
		{
			name:     "pipe test",
			input:    "key: ${PIPE_VAR}",
			setupEnv: func(t *testing.T) { t.Setenv("PIPE_VAR", "value|with$|pipes") },
			want:     "key: value|with$|pipes",
		},
	}

	for _, tt := range tests {
			if tt.setupEnv != nil {
				tt.setupEnv(t)
			}

			got, err := preprocessYAML([]byte(tt.input))

			if tt.wantErr {
				require.Error(t, err)
			} else {
				require.NoError(t, err)
			}
			assert.Equal(t, tt.want, string(got))
		})
	}
}
```

Then it should be easier to have unit tests based on: https://github.com/open-telemetry/opentelemetry-specification/blob/0e1094949a0cf57dd82c48fdb2b8e9e80cf3220e/specification/configuration/data-model.md#L174-L203 

Notice that I found a bug:

```
--- FAIL: TestPreprocessYAML (0.00s)
    --- FAIL: TestPreprocessYAML/pipe_test (0.00s)
        config_test.go:688: 
                Error Trace:    /home/ropajak/repos/opentelemetry-go-contrib/otelconf/v0.3.0/config_test.go:688
                Error:          Not equal: 
                                expected: "key: value|with$|pipes"
                                actual  : "key: value|with$pipes"
                            
                                Diff:
                                --- Expected
                                +++ Actual
                                @@ -1 +1 @@
                                -key: value|with$|pipes
                                +key: value|with$pipes
                Test:           TestPreprocessYAML/pipe_test
```

pellared (changes requested):
Usage of `|` during replacements causes invalid behavior.

otelconf/go.mod
codeboten: updated

CHANGELOG.md
codeboten: added a note to changelog

otelconf/testdata/v0.3-env-var.yaml
codeboten: Thanks for the test, added it in

otelconf/internal/provider/envprovider.go
codeboten: Removed the need for a marker character by changing the logic in the now single func used to replace env variables. PTAL

otelconf/v0.3.0/config.go
codeboten: added these tests and moved the different functions into a single one

otelconf/internal/provider/envprovider.go
```diff
@@ -0,0 +1,117 @@
...
+	"go.yaml.in/yaml/v3"
+)
+
+const validationPattern = `^[a-zA-Z_][a-zA-Z0-9_]*$`
+
+var validationRegexp = regexp.MustCompile(validationPattern)
+
+func ReplaceEnvVars(input []byte) ([]byte, error) {
+	// start by replacing all $$ that are not followed by a {
+	doubleDollarSigns := regexp.MustCompile(`\$\$([^{$])`)
```
pellared: Why this is not as a global variable (or why we are not passing the already parsed regexp when doing recursion)? Parsing the regex each time is expensive.

Same comment applies to other `regexp.MustCompile` occurrences.

otelconf/internal/provider/envprovider.go
```diff
@@ -0,0 +1,117 @@
...
+
+	out := []byte(val)
+	if err := checkRawConfType(out); err != nil {
+		return nil, fmt.Errorf("invalid value type: %w", err)
+	}
+
+	return out, nil
+}
+
+func parseEnvVarURI(uri string) (string, *string) {
```
pellared: Why "URI"? It does not seem to be an URI. See: https://datatracker.ietf.org/doc/html/rfc2396

otelconf/v0.3.0/config.go
```diff
@@ -179,8 +181,13 @@ func WithTracerProviderOptions(opts ...sdktrace.TracerProviderOption) Configurat
 
 // ParseYAML parses a YAML configuration file into an OpenTelemetryConfiguration.
 func ParseYAML(file []byte) (*OpenTelemetryConfiguration, error) {
+	file, err := provider.ReplaceEnvVars(file)
+	if err != nil {
+		return nil, err
+	}
```
pellared, line 187: This should be part of `func (c *OpenTelemetryConfiguration) UnmarshalYAML` in `config_yaml.go` and `func (c *OpenTelemetryConfiguration) UnmarshalJSON(b []byte) error` in `config_json.go`.

CHANGELOG.md
```diff
@@ -8,6 +8,10 @@ This project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0.htm
 
 ## [Unreleased]
 
+### Added
+
+- `ParseYAML` in `go.opentelemetry.io/contrib/otelconf` now supports environment variables substitution in the format `${VAR_NAME}`, `${env:VAR_NAME}`, and `${VAR_NAME:-defaultvalue}`. (#6215)
```
pellared: maybe this?

```suggestion
- `ParseYAML` in `go.opentelemetry.io/contrib/otelconf` now supports environment variables substitution in the format `${[env:]VAR_NAME[:-defaultvalue]}`. (#6215)
```

otelconf/internal/provider/envprovider.go
codeboten: Updated

otelconf/internal/provider/envprovider.go
codeboten: moved to global variable

otelconf/v0.3.0/config.go
codeboten: @pellared Do you have any suggestions on how to go about doing this in UnmarshalYAML? As far as I can tell, using the [unmarshaler interface](https://pkg.go.dev/gopkg.in/yaml.v3#Unmarshaler) ends up requiring the implementer to go through each node to do the substitution which gets pretty gnarly. 

An alternative way may be to use a custom Decoder but at that point it would require end users wanting to parse directly without using `ParseYAML` to instantiate the same decoder.

pellared: TBH, I am not sure if the docs are accurate. Notice other methods in https://github.com/open-telemetry/opentelemetry-go-contrib/blob/main/otelconf/v0.3.0/config_yaml.go. Or maybe the library just needs `UnmarshalYAML`?

otelconf/internal/provider/envprovider_test.go
```diff
@@ -0,0 +1,231 @@
...
+		},
+		{
+			name:  "escaped dollar sign - single escape",
+			input: "key: $${NOT_REPLACED}",
+			want:  "key: ${NOT_REPLACED}",
+		},
+		{
+			name:  "escaped dollar sign - double escape produces single dollar",
+			input: "key: $$${NOT_REPLACED}",
+			want:  "key: $",
```
MrAlias: This value is replaced.

```suggestion
			input: "key: $$${TEST_VAR}",
			want:  "key: $test_value",
```

otelconf/internal/provider/envprovider_test.go
```diff
@@ -0,0 +1,231 @@
...
+			input:    "key: ${PIPE_VAR}",
+			setupEnv: func(t *testing.T) { t.Setenv("PIPE_VAR", "value|with$|pipes") },
+			want:     "key: value|with$|pipes",
+		},
+		{
+			name:     "pipe test",
+			input:    "key: ${PIPE_VAR}",
+			setupEnv: func(t *testing.T) { t.Setenv("PIPE_VAR", "value|with$|pipes") },
+			want:     "key: value|with$|pipes",
+		},
```
MrAlias: duplicate

```suggestion
```

otelconf/internal/provider/envprovider_test.go
```diff
@@ -0,0 +1,231 @@
...
+	err := checkRawConfType([]byte{})
+	require.Error(t, err)
+	require.ErrorContains(t, err, "unsupported type=<nil> for retrieved config")
+}
+
+func TestReplaceEnvVar(t *testing.T) {
+	tests := []struct {
+		name     string
+		input    string
+		setupEnv func(t *testing.T)
```
MrAlias: nit: all your setups are just setting environment variables. This can be replaced with a `map` of environment variable values that is iterated in the test run.

otelconf/internal/provider/envprovider.go
```diff
@@ -0,0 +1,121 @@
...
+
+func parseEnvVar(in string) (string, *string) {
+	in = strings.TrimPrefix(in, "env:")
+	const defaultSuffix = ":-"
+	if strings.Contains(in, defaultSuffix) {
+		parts := strings.SplitN(in, defaultSuffix, 2)
+		return parts[0], &parts[1]
+	}
+	return in, nil
+}
```
MrAlias: nit: Avoid unnecessary `strings.Contains` and `strings.SplitN` (each allocates).

```suggestion
func parseEnvVar(in string) (string, *string) {
	in = strings.TrimPrefix(in, "env:")
	const sep = ":-"
	if i := strings.Index(in, sep); i >= 0 {
		def := in[i+len(sep):]
		return in[:i], &def
	}
	return in, nil
}

```

otelconf/internal/provider/envprovider.go
```diff
@@ -0,0 +1,121 @@
...
+
+	out := []byte(val)
+	if err := checkRawConfType(out); err != nil {
+		return nil, fmt.Errorf("invalid value type: %w", err)
+	}
+
+	return out, nil
+}
+
+func parseEnvVar(in string) (string, *string) {
```
MrAlias: Returning a pointer here to distinguish unset and empty can be replaced with a type that won't add an allocation.  E.g.

```suggestion
type value type {
	data string
	valid bool
}

func parseEnvVar(in string) (string, value) {
```

otelconf/internal/provider/envprovider.go
```diff
@@ -0,0 +1,121 @@
...
+				return data
+			}
+			if len(dollarSigns) > 1 {
+				data = append(dollarSigns[0:(len(dollarSigns)/2)], data...)
+			}
+		} else {
+			// need to expand any default value env var to support the case $${STRING_VALUE:-${STRING_VALUE}}
+			_, defaultValuePtr := parseEnvVar(string(match[2]))
+			if defaultValuePtr == nil || !strings.Contains(*defaultValuePtr, "$") {
+				return append(dollarSigns[0:(len(dollarSigns)/2)], []byte(fmt.Sprintf("{%s}", match[2]))...)
```
MrAlias:
```suggestion
				return fmt.Appendf(dollarSigns[0:(len(dollarSigns)/2)], "{%s}", match[2])
```

otelconf/internal/provider/envprovider.go
```diff
@@ -0,0 +1,121 @@
...
+			}
+			// expand the default value
+			data, err = ReplaceEnvVars(append(match[2], byte('}')))
+			if err != nil {
+				return data
+			}
+			data = append(dollarSigns[0:(len(dollarSigns)/2)], []byte(fmt.Sprintf("{%s", data))...)
```
MrAlias:
```suggestion
			data = fmt.Appendf(dollarSigns[0:(len(dollarSigns)/2)], "{%s", data)
```

otelconf/internal/provider/envprovider.go
```diff
@@ -0,0 +1,121 @@
...
+		}
+		return data
+	})
+	if err != nil {
+		return []byte{}, err
```
MrAlias: Why not `nil`?

```suggestion
		return nil, err
```

otelconf/internal/provider/envprovider_test.go
codeboten: updated the test 👍🏻

otelconf/internal/provider/envprovider.go
codeboten: updated to a struct 👍🏻

MrAlias (approved):
LGTM :+1:

pellared:
https://github.com/open-telemetry/opentelemetry-go-contrib/pull/6215#discussion_r2403289205 is not addressed. Other than that, LGTM

otelconf/v0.3.0/config.go
codeboten: right in either the `obsoleteUnmarshaler` or the `Unmarshaler` interfaces, the bytes are already decoded by the time they get to the function call.

```
// The Unmarshaler interface may be implemented by types to customize their
// behavior when being unmarshaled from a YAML document.
type Unmarshaler interface {
	UnmarshalYAML(value *Node) error
}

type obsoleteUnmarshaler interface {
	UnmarshalYAML(unmarshal func(interface{}) error) error
}

```

This is why I was suggesting implementing env var decoding via a custom decoder, but even that wouldn't be enough because the yaml parser is what takes the `[]byte` and creates a `yaml.Node` object from it. So the env substitution needs to happen before the call to `parse`. Below is the func that `yaml.Unmarshal` calls which as far as I can tell doesn't provide any facility to override the parser:

```
func unmarshal(in []byte, out interface{}, strict bool) (err error) {
	defer handleErr(&err)
	d := newDecoder()
	p := newParser(in)
	defer p.destroy()
	node := p.parse()
	if node != nil {
		v := reflect.ValueOf(out)
		if v.Kind() == reflect.Ptr && !v.IsNil() {
			v = v.Elem()
		}
		d.unmarshal(node, v)
	}
	if len(d.terrors) > 0 {
		return &TypeError{d.terrors}
	}
	return nil
}

```

This is much easier to implement in parsing json where `UnmarshalJSON` has access to the `[]byte`.

codeboten:
> [#6215 (comment)](https://github.com/open-telemetry/opentelemetry-go-contrib/pull/6215#discussion_r2403289205) is not addressed. Other than that, LGTM

replied on the comment, i'm not sure how to move it forward

pellared (approved):

pellared:
@open-telemetry/go-approvers, would be nice to have more reviews. Otherwise, we can merge this on Thursday or Friday.

otelconf/internal/provider/envprovider_test.go
```diff
@@ -0,0 +1,226 @@
+// Copyright The OpenTelemetry Authors
+// SPDX-License-Identifier: Apache-2.0
+
+package provider // import "go.opentelemetry.io/contrib/internal/provider"
```
flc1125:
```suggestion
package provider // import "go.opentelemetry.io/contrib/otelconf/internal/provider"
```

pellared:
@codeboten, thank you!

PR #6229, opened 2024-10-10
Revert "chore(deps): update lycheeverse/lychee-action action to v2"

dmathieu:
Reverts open-telemetry/opentelemetry-go-contrib#6220

lychee-action has a regression in v2, where it deletes the config file, and we therefore try checking links for templates that shouldn't be checked.
See https://github.com/lycheeverse/lychee-action/pull/255


MrAlias (approved):

dmathieu:
Merging this with a single review as it's a revert, and our CI has been failing for way too long.

PR #6233, opened 2024-10-11
Update otel core to the latest release

dashpole:
I disabled exemplars in the gRPC and http tests to make tests pass without more substantial changes to the test.


pellared:
> without substantial refactoring

Just to double-check. Do you mean simply updating the expected test results or something different?

Do you want to address it as a separate PR?

dashpole:
Yeah, just updating the expected test.  I was also not sure if I would need to tell the test to ignore values or timestamps.

dmathieu (approved):

PR #6238, opened 2024-10-13
Skip SDK upgrades that are pinned to a specific commit

Victorsesan:


dmathieu:
This also changes some unrelated formatting of the file, making the diff harder to read. Could you remove those unrelated formats ?

Victorsesan:
@dmathieu I've maintained the fille structure and ensured formatting is consistent and easy to read I'm not sure if that fixed it. Also the renovate bot link; renovate-schema.js file is unable to load from my end not sure why but maybe it could be the problem. Please let me know if that changed anything, thanks

renovate.json
```diff
@@ -32,7 +28,16 @@
     },
     {
       "matchPackageNames": ["go.opentelemetry.io/otel/**"],
-      "allowedVersions": "/^v[0-9]+\\.[0-9]+\\.[0-9]+/"
+      "allowedVersions": "^v[0-9]+\\.[0-9]+\\.[0-9]+",
```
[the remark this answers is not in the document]
Victorsesan: I'm sure it will be crucial for maintaining consistency and preventing issues with the version mismatches or unexpected formats in a long run, but can change or remove if otherwise not needed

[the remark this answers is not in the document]
dmathieu: We control version names, and follow semver. So there shouldn't be such unexpected formats.
And if there are, we will get exceptional PRs and can handle them as well. So I'd be in favor of not keeping anything that is not necessary.

renovate.json
```diff
@@ -32,7 +28,22 @@
...
+      "ignore": {
+        "matchUpdateTypes": ["pin", "digest"]
+      }
+    },
+    {
+      "matchPackageNames": ["go.opentelemetry.io/otel/**"],
+      "sourceUrlPrefixes": ["https://github.com/open-telemetry/opentelemetry-go"],
+      "enabled": false,
+      
+      "_comment": {
```
dmathieu, line 42: This `_comment` section is a neat trick.
However, I think it would be easier to read (especially because of folks habits) if it were actual yaml comments rather than a string which may seem like part of the configuration to an untrained eye.

Victorsesan: I've tried to do something about the comments , since its my first time adding yaml comments in a json file I'm not sure if i did it right open for corrections. TIA

renovate.json
```diff
@@ -32,7 +28,16 @@
...
+      }
+    },
+    {
+      "matchPackageNames": ["go.opentelemetry.io/otel/**"],
+      "sourceUrlPrefixes": ["https://github.com/open-telemetry/opentelemetry-go"],
+      "enabled": false
     }
-  ]
+  ],
+  "yamlConfig": "# Configuration for package matching and source URL prefixes\n matchPackageNames:\n    - \"go.opentelemetry.io/otel/**\"  # Matches all packages under the OpenTelemetry Go library\n  # List of source URL prefixes that correspond to the matched packages.\n  sourceUrlPrefixes:\n    - \"https://github.com/open-telemetry/opentelemetry-go\"  # Base URL for the OpenTelemetry Go repository\n  # Flag to enable or disable this configuration.\n  enabled: false  # Currently set to false, meaning this configuration is not active"
```
dmathieu, line 42:
```suggestion
  ]
```

renovate.json
```diff
@@ -32,7 +28,16 @@
     },
     {
```
dmathieu, line 29: Urgh sorry, I confused json and yaml. Json doesn't take comments in its file content 😢 
Let's make the comment the very first key then.

```suggestion
    {
        "_description": "Ignore pin and digest upgrades. We only want published releases of the SDK to trigger upgrades",   
```

renovate.json
```diff
@@ -32,7 +28,16 @@
...
       "matchPackageNames": ["go.opentelemetry.io/otel/**"],
-      "allowedVersions": "/^v[0-9]+\\.[0-9]+\\.[0-9]+/"
+      "enabled": true,
+      "ignore": {
+        "matchUpdateTypes": ["pin", "digest"]
+      }
+    },
+    {
+      "matchPackageNames": ["go.opentelemetry.io/otel/**"],
```
dmathieu, line 37:
```suggestion
        "_description": "Any SDK upgrade that made it through here is a pin/digest. DIsable it",
      "matchPackageNames": ["go.opentelemetry.io/otel/**"],
```

renovate.json
Victorsesan: That's true i was so confused when you requested😅 i had to improve on the yaml config but i did at least learn something new with the yaml config for references in a json file👍. I guess everything is fine now? Or is there anything else we need to improve of fix?

renovate.json
```diff
@@ -31,8 +27,18 @@
...
+        "_description": "Ignore pin and digest upgrades. We only want published releases of the SDK to trigger upgrades",   
       "matchPackageNames": ["go.opentelemetry.io/otel/**"],
-      "allowedVersions": "/^v[0-9]+\\.[0-9]+\\.[0-9]+/"
+      "enabled": true,
+      "ignore": {
+        "matchUpdateTypes": ["pin", "digest"]
+      }
+    },
+    {
+        "_description": "Any SDK upgrade that made it through here is a pin/digest. DIsable it",
```
dmathieu, line 38:
```suggestion
        "_description": "Any SDK upgrade that made it through here is a pin/digest. Disable it",
```

renovate.json
```diff
@@ -31,8 +27,18 @@
       "groupName": "golang.org/x"
     },
     {
+        "_description": "Ignore pin and digest upgrades. We only want published releases of the SDK to trigger upgrades",
```
dmathieu, line 30:
```suggestion
      "_description": "Ignore pin and digest upgrades. We only want published releases of the SDK to trigger upgrades",   
```

dmathieu (approved):
Left two nits.

renovate.json
```diff
@@ -31,8 +27,18 @@
       "groupName": "golang.org/x"
     },
     {
+        "_description": "Ignore pin and digest upgrades. We only want published releases of the SDK to trigger upgrades",   
       "matchPackageNames": ["go.opentelemetry.io/otel/**"],
-      "allowedVersions": "/^v[0-9]+\\.[0-9]+\\.[0-9]+/"
+      "enabled": true,
+      "ignore": {
+        "matchUpdateTypes": ["pin", "digest"]
```
pellared, line 34: What is a pin?

dmathieu: It looks like it's a dependency set at an explicit version.
https://docs.renovatebot.com/dependency-pinning/

Pinning isn't something we can really do with Go. But just to be same, maybe we should actually remove that match.

Victorsesan: @pellared I would be happy to! sorry for messing it up , i'm  trying but my PR keeps getting tested here instead can you please give me some pointers on how to test directly on my fork repo . thanks

Victorsesan: @dmathieu Did we succeeded in fixing this issue i would love a feedback if there is anything else i need to do i will be happy to! Thanks

dmathieu: There are still open questions, in this very thread that need resolving.
Especially, the tricky thing here is that we can't really confirm we "succeeded", since there's no way to validate the behavior of the config until it's merged and runs. Hence @pellared's question to try this in a fork, or another repository.

pellared: @Victorsesan, you need to configure Renovate on your fork to check if it will work as expected.

Victorsesan: @pellared @dmathieu Are these the expected test we need to see? Thanks for the time
Screenshot 2024-10-24 004430
Screenshot 2024-10-24 004348
Screenshot 2024-10-24 004059

dmathieu: They would be, yes. But in your screenshot, they appear to say the configuration file is invalid.

Victorsesan: @dmathieu Finally succeeded in making renovate work as expected after a long try and errors, i had to make some minor changes for it to pass the test which I'm not sure if its good or bad but i mentioned all the errors i faced with the solution and modifications i did in my recent commit which i will be happy if you can check it out and let me know if we are in the right track. I would love to succeed in working on this . Thank you all for your time.
Screenshot 2024-10-25 010858
Screenshot 2024-10-25 010932

dmathieu: This shows the config is valid. But it doesn't seem to validate that commit upgrades would be skipped, while releases would be bumped?

Victorsesan: > hows the config is valid. But it doesn't seem to validate that commit upgrades would be skipped, while releases would be bumped

@dmathieu Sorry i missed that, is this last commit good i have added something to fix it. I was just worried with mend not accepting the package rules even when they are good sometimes which made me lost track validating the skip rule and rather focus on having renovate pass the test, but let me know if the few changes i made does it and it did also pass the test. thanks

dmathieu:
Sorry for the stale PR.
With #6436, I think this becomes unneeded.

Issue #6240, opened 2024-10-13
Allow GRPC Server stats handler defining code mapping for grpc error


Ferrany1:
### Problem Statement

Currently [serverStatus](https://github.com/open-telemetry/opentelemetry-go-contrib/blob/c3c853808383d1bce596cbb910546c305c9187e8/instrumentation/google.golang.org/grpc/otelgrpc/interceptor.go#L518C1-L530C2) functions defines which errors to treat as status.Error, however some codes in some logic (like InvalidArgument or Unauthenticated) may also be defined as errors

### Proposed Solution

Adding Option which may override default function (also defining this function in [serverHandler](https://github.com/open-telemetry/opentelemetry-go-contrib/blob/c3c853808383d1bce596cbb910546c305c9187e8/instrumentation/google.golang.org/grpc/otelgrpc/stats_handler.go#L33C1-L35C2)

#### Alternatives

Adding all grpc codes > 3 to be defined as errors

pellared:
The statuses of gRPC instrumentation are defined by OTel Semantic Convetions: https://github.com/open-telemetry/semantic-conventions/blob/main/docs/rpc/grpc.md#grpc-status.

The application can create it's own spans and set span statuses as desired.

The application can also use [`trace.SpanFromContext`](https://pkg.go.dev/go.opentelemetry.io/otel/trace#SpanFromContext) to set an error status for the "current" span.

Is adding another option really needed? 

Moreover, I am not sure if it is good to add options which can make the instrumentation library working in a way that would be not compliant with the semantic conventions.

PR #6241, opened 2024-10-13
Feature: GRPC status code server remap function addition

Ferrany1:
#6240


pellared (changes requested):
I do not think this option is needed. More: https://github.com/open-telemetry/opentelemetry-go-contrib/issues/6240#issuecomment-2409138278

PR #6250, opened 2024-10-15
Fix request/response metrics. In StatsHandler In means response for Client

bogdandrutu:


dashpole (approved):
Change LGTM.  Is there any more context you can share about how you found this?  E.g. linked issues in the collector?

bogdandrutu:
I was trying internally to measure the throughput, and we enabled grpc metrics to see how many bytes of logs we export… and found that always the request size was 2B, the I saw that response was way larger… from there only some small debugging

CHANGELOG.md
```diff
@@ -8,6 +8,10 @@ This project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0.htm
 
 ## [Unreleased]
 
+### Fixed
+
+- Fix request/response metrics. In StatsHandler In means response for Client. (#6250)
```
dmathieu, line 13: This should specify which instrumentation this happens in.

```suggestion
- Fix request/response metrics in `go.opentelemetry.io/contrib/instrumentation/google.golang.org/grpc/otelgrpc`, as in StatsHandler `in` means response for Client and request for Server. (#6250)
```

CHANGELOG.md
```diff
@@ -13,6 +13,7 @@ This project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0.htm
 - Transform nil attribute values to `log.Value` zero value instead of panicking in `go.opentelemetry.io/contrib/bridges/otellogrus`. (#6237)
 - Transform nil attribute values to `log.Value` zero value instead of panicking in `go.opentelemetry.io/contrib/bridges/otelzap`. (#6237)
 - Transform nil attribute values to `log.Value` zero value instead of `log.StringValue("<nil>")` in `go.opentelemetry.io/contrib/bridges/otelslog`. (#6246)
+- Fix request/response metrics in `go.opentelemetry.io/contrib/instrumentation/google.golang.org/grpc/otelgrpc`, as in StatsHandler `in` means response for Client and request for Server. (#6250)
```
pellared, line 16:
```suggestion
- Fix `rpc.client.request.*` metrics to measure requests instead of responses and `rpc.client.responses.*` metrics to measure responses instead of requests in `go.opentelemetry.io/contrib/instrumentation/google.golang.org/grpc/otelgrpc`. (#6250)
```

I think this would be a more user-facing changelog entry. Can anyone double-check if I have not messed anything?

pellared (approved):
Only the changelog should be improved. Thanks @bogdandrutu 🥇

bogdandrutu:
FYI: I feel that you are focusing too much on the nuance of the changelog message instead of speedup the process of accepting this bug fix (which is important and quite significant).

PR #6260, opened 2024-10-16
config: fix bug where WithResourceAsConstantLabels wasn't set

codeboten:
There was a bug in the config package where the WithResourceAsConstantLabels option wasn't appended to the list of options for the Prometheus reader. Fixed that issue and added a test to validate the correct number of options is set.

### DIFF
--- a/config/metric.go
+++ b/config/metric.go
@@ -297,28 +297,6 @@ func newIncludeExcludeFilter(lists *IncludeExclude) (attribute.Filter, error) {
 	}, nil
 }
 
-func prometheusReaderOpts(prometheusConfig *Prometheus) ([]otelprom.Option, error) {
-	var opts []otelprom.Option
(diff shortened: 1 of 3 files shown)


config/metric.go
```diff
@@ -319,7 +313,23 @@ func prometheusReader(ctx context.Context, prometheusConfig *Prometheus) (sdkmet
...
 			return nil, err
 		}
-		otelprom.WithResourceAsConstantLabels(f)
+		opts = append(opts, otelprom.WithResourceAsConstantLabels(f))
+	}
+
+	return opts, nil
+}
+
+func prometheusReader(ctx context.Context, prometheusConfig *Prometheus) (sdkmetric.Reader, error) {
```
pellared, line 322: nit: I would move the `prometheusReader` function before `prometheusReaderOpts`.

pellared (approved):

config/metric.go
codeboten: updated

dashpole (approved):

PR #6289, opened 2024-11-03
config: update resource to return only configured values

codeboten:
This changes the resource configured to match what an end user would choose to configure. Without this, Collector users would have no way to ensure that only the resource attributes they configure are added to the telemetry.


CHANGELOG.md
```diff
@@ -19,6 +19,7 @@ This project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0.htm
 ### Changed 
 
 - Upgrade `go.opentelemetry.io/otel/semconv/v1.17.0` to `go.opentelemetry.io/otel/semconv/v1.21.0` in `go.opentelemetry.io/contrib/instrumentation/go.mongodb.org/mongo-driver/mongo/otelmongo`. (#6272)
+- Resource doesn't merge with defaults if a valid resource is configured.
```
codeboten, line 22:
```suggestion
- Resource doesn't merge with defaults if a valid resource is configured. (#6289)
```

CHANGELOG.md
```diff
@@ -19,6 +19,7 @@ This project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0.htm
 ### Changed 
 
 - Upgrade `go.opentelemetry.io/otel/semconv/v1.17.0` to `go.opentelemetry.io/otel/semconv/v1.21.0` in `go.opentelemetry.io/contrib/instrumentation/go.mongodb.org/mongo-driver/mongo/otelmongo`. (#6272)
+- Resource doesn't merge with defaults if a valid resource is configured. (#6289)
```
dmathieu, line 22:
```suggestion
- Resource doesn't merge with defaults if a valid resource is configured in `go.opentelemetry.io/contrib/config`. (#6289)
```

dmathieu (approved):

PR #6299, opened 2024-11-06
config: don't log an error on close

codeboten:
This is an expected outcome of the server shutdown being called, no need to log an error about it.

Not sure if this is worth a changelog entry, reviewers please advise


pellared (approved):
I would add a changelog entry. Something saying that we are no longer calling otel.Handle for a successful shutdown of the Prometheus exporter.

dmathieu (approved):

dmathieu (approved):

MrAlias (approved):

PR #6323, opened 2024-11-12
Use lychee new cache-exclude-status option

dmathieu:
Lychee's `cache-exclude-status` option has shipped. So we can use it rather than ignore 429s.

### DIFF
--- a/.github/workflows/links-fail-fast.yml
+++ b/.github/workflows/links-fail-fast.yml
@@ -25,5 +25,4 @@ jobs:
           GITHUB_TOKEN: ${{ secrets.GITHUB_TOKEN }}
         with:
           fail: true
-          # TODO: Remove 429s exception once https://github.com/open-telemetry/opentelemetry-go-contrib/issues/6183 is resolved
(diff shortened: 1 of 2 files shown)


MrAlias (approved):

pellared (approved):

pellared:
@dmathieu 

can we somehow get rid of:

```
* [429] [https://github.com/open-telemetry/opentelemetry-go/commits/main](https://github.com/open-telemetry/opentelemetry-go/commits/main) | Cached: Error (cached)
```

dmathieu:
I think that means it works as expected. We currently cache 429s, and this PR uses the cache.
Once the cache is renewed, 429s won't be cached anymore, and they will be retried.

pellared:
> Once the cache is renewed, 429s won't be cached anymore, and they will be retried.

I am just not sure if the cache would be ever renewed given the cached error. Still, not a blocker. I think we can merge this PR.

dmathieu:
The cache has a finite duration. It is renewed every day.

PR #6333, opened 2024-11-15
[chore] Fix codecov action usage

pellared:
Same as https://github.com/open-telemetry/opentelemetry-go/pull/5979.

This change makes sure that an unsuccessful upload will fail the workflow as this does not happen at the moment. 

Additionally, it configures the codecov action to be tokenless.


dmathieu (approved):

pellared:
Merging as trvial

Issue #6190, opened 2024-10-04
Deprecate and remove go.opentelemetry.io/contrib/instrumentation/github.com/gin-gonic/gin/otelgin


pellared:
This module has been identified to not have an [owner](https://github.com/open-telemetry/opentelemetry-go-contrib/blob/main/CODEOWNERS). Based on the [project's ownership policy](https://github.com/open-telemetry/opentelemetry-go-contrib/issues/2834#issuecomment-1282698918), this module will be deprecated and then removed.

### How to keep this module

For this module to continue in this repository, it needs a sponsor.

If you would like to sponsor this module and become an owner, please comment in this issue about your desire. As an owner you will assume the following responsibilities:

- You will need to [be a member of the OpenTelemetry organization](https://github.com/open-telemetry/community/blob/main/community-membership.md#member) and maintain that membership.
  - If you are not an existing member, we can add you to the community. This is not a disqualification, but something you will need in the process of becoming a sponsor.
- You will be responsible for keeping up with the instrumented library. Any "upstream" changes that impact this module need to be proactively handle by you.
- You will be expected to review any Pull Requests or Issues created that relate to this module.
- You will be responsible for the stability and versioning compliance of the module.

You will need to have a good working knowledge of the code this module is instrumenting and, ideally, familiarity with the existing module code.

### How this module will be removed

This module is in the process of being deprecated. After that deprecation notice has been published, we will wait 3 months or 2 two releases (whichever is the longer time period). After that time period, this module will be removed from this repository and no more versions of the module will be published.

#### Resurrection

If a sponsor is found after the module has been deprecated or removed, these operations can be reversed (i.e. coded added back, deprecation notice removed).

abiabsurd:
I have a proposal to replace the existing implementation of `otelgin` with a relatively simple adapter package — which I've already written and am using in a private codebase — that enables usage of the [otelhttp](go.opentelemetry.io/contrib/instrumentation/net/http/otelhttp) package as a gin middleware. In addition to addressing the root of this github issue, the solution also provides metrics instrumentation, which addresses several other open issues ([example](https://github.com/open-telemetry/opentelemetry-go-contrib/issues/5096)).

How do you recommend I proceed @pellared?

dmathieu:
This kind of refactoring would provide breaking changes to the library.
So it would have to be a new library, not just a commit in the existing and deprecated one.

[...]

PR #6346, opened 2024-11-19
feat(otelgin): enhance gin error tracking with span recording

flc1125:


dmathieu:
Could you test this change?
By checking there's indeed an error here for example: https://github.com/open-telemetry/opentelemetry-go-contrib/blob/main/instrumentation/github.com/gin-gonic/gin/otelgin/test/gintrace_test.go#L123

Also, I wonder if we need to keep the `gin.errors` attribute. While removing it is a breaking change, keeping it is redundant.

dmathieu:
Side note that otelgin has no owner, so it may be removed soon.
https://github.com/open-telemetry/opentelemetry-go-contrib/issues/6190

flc1125:
> Side note that otelgin has no owner, so it may be removed soon. #6190

Okay, will this package still accept new PRs at present?

dmathieu:
It does.

flc1125:
https://github.com/open-telemetry/opentelemetry-go-contrib/issues/6190

I think I can give it a try.

CHANGELOG.md
```diff
@@ -13,6 +13,10 @@ This project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0.htm
 - Added support for providing `endpoint`, `pollingIntervalMs` and `initialSamplingRate` using environment variable `OTEL_TRACES_SAMPLER_ARG` in `go.opentelemetry.io/contrib/samples/jaegerremote`. (#6310)
 - Added support exporting logs via OTLP over gRPC in `go.opentelemetry.io/contrib/config`. (#6340)
 
+### Changed
+
+- feat(otelgin): enhance gin error tracking with span recording (#6346)
```
```suggestion
- Record an error instead of setting the `gin.errors` attribute in `go.opentelemetry.io/contrib/instrumentation/github.com/gin-gonic/gin/otelgin`. (#6346)
```

flc1125:
> Could you test this change? By checking there's indeed an error here for example: [`main`/instrumentation/github.com/gin-gonic/gin/otelgin/test/gintrace_test.go#L123](https://github.com/open-telemetry/opentelemetry-go-contrib/blob/main/instrumentation/github.com/gin-gonic/gin/otelgin/test/gintrace_test.go#L123)
> 
> Also, I wonder if we need to keep the `gin.errors` attribute. While removing it is a breaking change, keeping it is redundant.

It's done.

flc1125:
> Side note that otelgin has no owner, so it may be removed soon.

I spent some time reading through the requirements. I understand that I may still need to make some contributions before applying. If so, I need some time.

.golangci.yml
```diff
@@ -82,6 +82,7 @@ linters-settings:
           - "**/*/{bridges,instrumentation}/**/*.go"
           - "!**/*/bridges/prometheus/*.go" # prometheus bridge is meant to use the SDK
           - "!**/*/instrumentation/runtime/*.go" # runtime instrumentation is meant to use the SDK
+          - "!**/*/instrumentation/github.com/gin-gonic/gin/otelgin/*.go" # gin instrumentation is meant to use the SDK
```
dmathieu: No, gin instrumentation is not meant to use the SDK.
There is a `test` subfolder which is a separate module. You can put tests that need the SDK in there, they will be ignored by this linter automatically.

dmathieu (dismissed):

.golangci.yml
flc1125: This has been removed before, you can refresh to view the latest code.

instrumentation/github.com/gin-gonic/gin/otelgin/gintrace.go
```diff
@@ -95,7 +97,7 @@ func Middleware(service string, opts ...Option) gin.HandlerFunc {
 			span.SetAttributes(semconv.HTTPStatusCode(status))
 		}
 		if len(c.Errors) > 0 {
-			span.SetAttributes(attribute.String("gin.errors", c.Errors.String()))
+			span.RecordError(errors.New(c.Errors.String()))
```
dmathieu, line 100:
```suggestion
			span.RecordError(errors.Join(c.Errors))
```

instrumentation/github.com/gin-gonic/gin/otelgin/test/gintrace_test.go
```diff
@@ -120,7 +120,6 @@ func TestError(t *testing.T) {
 	attr := span.Attributes()
 	assert.Contains(t, attr, attribute.String("net.host.name", "foobar"))
 	assert.Contains(t, attr, attribute.Int("http.status_code", http.StatusInternalServerError))
-	assert.Contains(t, attr, attribute.String("gin.errors", "Error #01: oh no\n"))
```
dmathieu, line 123: This test already checks for errors. We can check it being recorded here, no need for a new test.
We can also add a check for empty `span.Events` in `TestTrace200`.

dmathieu (commented):
This is starting to look good 😸

instrumentation/github.com/gin-gonic/gin/otelgin/test/gintrace_test.go
```diff
@@ -321,3 +328,61 @@ func TestWithGinFilter(t *testing.T) {
 		assert.Len(t, sr.Ended(), 1)
 	})
 }
+
+func TestSpanRecordError(t *testing.T) {
```
dmathieu, line 332: Shouldn't this test be removed?

flc1125: Good idea, before this, I hadn't fully read the other tests. This is somewhat redundant.

dmathieu (approved):

instrumentation/github.com/gin-gonic/gin/otelgin/gintrace.go
pellared, line 100: Shouldn't we call `span.RecordError` for each error inside `c.Errors`?

flc1125: Can it be like this? I'll try.

flc1125: I tried 2 options to see which one would recommend? My decision is Option 2:

- one:

	```go
	if len(c.Errors) > 0 {
		var errs error
		for _, err := range c.Errors {
			errs = errors.Join(errs, err.Err)
		}
		span.RecordError(errs)
	}
	```

- two:

	```go
	if len(c.Errors) > 0 {
		for _, err := range c.Errors {
			span.RecordError(err.Err)
		}
	}
	```

dmathieu: One would end up being exactly the same thing as we had before. So two.

CHANGELOG.md
```diff
@@ -13,6 +13,10 @@ This project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0.htm
 - Added support for providing `endpoint`, `pollingIntervalMs` and `initialSamplingRate` using environment variable `OTEL_TRACES_SAMPLER_ARG` in `go.opentelemetry.io/contrib/samples/jaegerremote`. (#6310)
 - Added support exporting logs via OTLP over gRPC in `go.opentelemetry.io/contrib/config`. (#6340)
 
+### Changed
+
+- Record an error instead of setting the `gin.errors` attribute in `go.opentelemetry.io/contrib/instrumentation/github.com/gin-gonic/gin/otelgin`. (#6346)
```
```suggestion
- Record errors instead of setting the `gin.errors` attribute in `go.opentelemetry.io/contrib/instrumentation/github.com/gin-gonic/gin/otelgin`. (#6346)
```

instrumentation/github.com/gin-gonic/gin/otelgin/gintrace.go
```diff
@@ -10,13 +10,14 @@ import (
...
 
-	"go.opentelemetry.io/contrib/instrumentation/github.com/gin-gonic/gin/otelgin/internal/semconvutil"
 	"go.opentelemetry.io/otel"
 	"go.opentelemetry.io/otel/attribute"
 	"go.opentelemetry.io/otel/codes"
 	"go.opentelemetry.io/otel/propagation"
 	semconv "go.opentelemetry.io/otel/semconv/v1.20.0"
 	oteltrace "go.opentelemetry.io/otel/trace"
+
+	"go.opentelemetry.io/contrib/instrumentation/github.com/gin-gonic/gin/otelgin/internal/semconvutil"
```
pellared, line 20: Can you revert this import rearrange change? We are grouping `go.openteletry.io`. See https://github.com/open-telemetry/opentelemetry-go-contrib/blob/96b1ed1e7b0c1613290df1801ff6ba8474a33ce5/.golangci.yml#L108-L109

instrumentation/github.com/gin-gonic/gin/otelgin/test/gintrace_test.go
```diff
@@ -17,14 +17,14 @@ import (
...
 	"go.opentelemetry.io/otel"
+	"go.opentelemetry.io/otel/attribute"
 	"go.opentelemetry.io/otel/codes"
 	sdktrace "go.opentelemetry.io/otel/sdk/trace"
 	"go.opentelemetry.io/otel/sdk/trace/tracetest"
-
-	"go.opentelemetry.io/otel/attribute"
 	oteltrace "go.opentelemetry.io/otel/trace"
+
+	"go.opentelemetry.io/contrib/instrumentation/github.com/gin-gonic/gin/otelgin"
```
pellared, line 27: Can you revert this import rearrange change? We are grouping `go.openteletry.io`. See https://github.com/open-telemetry/opentelemetry-go-contrib/blob/96b1ed1e7b0c1613290df1801ff6ba8474a33ce5/.golangci.yml#L108-L109

instrumentation/github.com/gin-gonic/gin/otelgin/test/gintrace_test.go
```diff
@@ -120,7 +122,17 @@ func TestError(t *testing.T) {
 	attr := span.Attributes()
 	assert.Contains(t, attr, attribute.String("net.host.name", "foobar"))
 	assert.Contains(t, attr, attribute.Int("http.status_code", http.StatusInternalServerError))
-	assert.Contains(t, attr, attribute.String("gin.errors", "Error #01: oh no\n"))
+
+	// verify the error event
```
pellared, line 126:
```suggestion
	// verify the error events
```

pellared (approved):
Just nit comments. Can you please address them before we merge?

instrumentation/github.com/gin-gonic/gin/otelgin/test/gintrace_test.go
flc1125: I'll handle it together later.

dmathieu:
Could you resolve the conflicts?

flc1125:
> Could you resolve the conflicts?

It's done.

instrumentation/github.com/gin-gonic/gin/otelgin/gintrace.go
```diff
@@ -95,7 +95,9 @@ func Middleware(service string, opts ...Option) gin.HandlerFunc {
 			span.SetAttributes(semconv.HTTPStatusCode(status))
 		}
 		if len(c.Errors) > 0 {
-			span.SetAttributes(attribute.String("gin.errors", c.Errors.String()))
+			for _, err := range c.Errors {
+				span.RecordError(err.Err)
+			}
```
dmathieu, line 100: (Sorry, I know this has been approved already)
Should we also set the span status to error?

flc1125: I saw that the setting was here, so I didn't make any settings.

flc1125: Additionally to be supplemented: Even if approved, as long as there is a need for adjustment, it can be pointed out. Try to make this component perfect as much as possible.

flc1125: > However, there are some differences here. In the existing logic, the status is set based on the status code. But in fact, it can also be explained through Error. So there are actually some differences here.

I can try to find some information based on this, look at best practices, see how Status is determined, and then consider how to adjust.

flc1125: https://opentelemetry.io/docs/concepts/signals/traces/#span-status

> To reiterate: Unset represents a span that completed without an error. Ok represents when a developer explicitly marks a span as successful. In most cases, it is not necessary to explicitly mark a span as Ok.

I saw this sentence on the official website, and my understanding is:

As long as we are very sure that there is no problem with him, it can mean OK; as long as there is an error (which may include status codes or may include error), we can set it to ERROR.

Does my understanding make sense?

dmathieu: Could gin have a success HTTP status, but include errors?

dmathieu: Then, we should mark the span as error if the HTTP status is an error, or if otelgin provides errors.

flc1125: Another supplementary material: https://opentelemetry.io/docs/specs/otel/trace/api/#set-status

This is its priority: Ok > Error > Unset

However, I checked the logic of the status codes, and it will only return Error and Unset. Therefore, it will be much easier for me to handle.

dmathieu: Ok is a very specific status that should be set extremely rarely. We can keep Unset.

flc1125: Adjusted

instrumentation/github.com/gin-gonic/gin/otelgin/gintrace.go
```diff
@@ -95,7 +95,10 @@ func Middleware(service string, opts ...Option) gin.HandlerFunc {
 			span.SetAttributes(semconv.HTTPStatusCode(status))
 		}
 		if len(c.Errors) > 0 {
-			span.SetAttributes(attribute.String("gin.errors", c.Errors.String()))
+			span.SetStatus(codes.Error, "gin error")
```
dmathieu, line 98: `gin error` isn't very meaningful.

```suggestion
			span.SetStatus(codes.Error, c.Errors.String())
```

MrAlias: Wont this duplicate, in a less useful flattened string from, the information being added as events below?

instrumentation/github.com/gin-gonic/gin/otelgin/test/gintrace_test.go
```diff
@@ -193,7 +195,17 @@ func TestError(t *testing.T) {
 	attr := span.Attributes()
 	assert.Contains(t, attr, attribute.String("net.host.name", "foobar"))
 	assert.Contains(t, attr, attribute.Int("http.status_code", http.StatusInternalServerError))
-	assert.Contains(t, attr, attribute.String("gin.errors", "Error #01: oh no\n"))
+
+	// verify the error events
+	events := span.Events()
+	assert.Len(t, events, 2)
```
MrAlias, line 201:
```suggestion
	require.Len(t, events, 2)
```

Avoid the panic.

MrAlias:
> > @dmathieu I could pick up ownership of this module so that it isn't removed, should I just open a PR on the codeowners file?
> 
> Could you review this PR if you want to be an code owner? 😉

cc @akats7 (I think)

akats7 (approved):

flc1125:
> lgtm, my only real thought is I personally always find it convenient to have the error message at least partially included directly on the span instead of only having the span events. Since span events are sometimes (always?) stored in separate table it could becomes an additional step to view the message depending on the platform you use.
> 
> For example I know a lot of the java instrumentation just adds _error: true_ as an attribute and then creates a span event, and we've had teams complain that the error message has no contextual info

This is indeed a problem, and it usually depends on what backend tools we are using.

Fortunately, when the entire span contains some abnormal events, the tool I am currently using allows me to see the exception information directly in the span interface. Of course, for more detailed information, I still need to switch to view it in Events.

——

Back to this PR, due to the OTEL standard specifications, I suggest we still maintain storing exceptions in Events. After all, this project is for the majority of users, and we need to present ourselves as professionally as possible and in accordance with the standards. 😁

akats7:
We ofc need to record the exception as a span event, what I'm getting at is when it comes to:
```
	span.SetAttributes(attribute.String("gin.errors", c.Errors.String()))

or

	span.SetStatus(codes.Error, "gin error")

```

I do find including the error string to have value

instrumentation/github.com/gin-gonic/gin/otelgin/gintrace.go
flc1125: @akats7 We can discuss it here. It should be the same issue here.

flc1125: I recommend using this method to record: `span.SetStatus(codes.Error, c.Errors.String())`

I would like to hear everyone's suggestions.

pellared: From: https://github.com/open-telemetry/opentelemetry-go-contrib/pull/6346#issuecomment-2498987933

> I personally always find it convenient to have the error message at least partially included directly on the span instead of **only** having the span events. Since span events are sometimes (always?) stored in separate table it could becomes an additional step to view the message depending on the platform you use. 
>
> For example I know a lot of the java instrumentation just adds _error: true_ as an attribute and then creates a span event, and we've had teams complain that the error message has no contextual info

pellared: Personally, I do not mind duplication.

flc1125: Based on this reason, if I adjust it like this, is it appropriate?

```go
```

instrumentation/github.com/gin-gonic/gin/otelgin/test/gintrace_test.go
pellared: Reopened 😉

instrumentation/github.com/gin-gonic/gin/otelgin/gintrace.go
pellared: @MrAlias, most of us prefer `span.SetStatus(codes.Error, c.Errors.String())`.

Please create an issue or follow-up PR if you think that this is a wrong decision.

PR #6349, opened 2024-11-19
Add the tmp directory to .gitignore

evan-bradley:
Running `make genjsonschema` creates a `tmp` directory that is not present in the .gitignore file, this makes sure contributors don't need to worry about having it in their repo directory.

PR #6347, opened 2024-11-19
Update generated config types

evan-bradley:
This updates the generated config types to be created with go-jsonschema v0.17.0, which includes an update to add a struct tag the `AdditionalProperties` field that allows it to be marshaled.

Making this change fixes the following error that occurs when I marshal a basic `otelcol.Config` struct, then run it with the Collector:
```
'metrics.readers[0].pull.exporter' has invalid keys: additionalproperties
```

Related to https://github.com/open-telemetry/opentelemetry-collector/issues/11711.

cc @codeboten

PR #6353, opened 2024-11-20
Run genjsonschema in CI

dmathieu:
https://github.com/open-telemetry/opentelemetry-go-contrib/pull/6347 is updating the generated JSON schema, as it has changed with upgrades of `go-jsonschema`.
We update the binary semi automatically with dependabot. However, this means there could be unseen regressions or changes like this that we miss.

This runs the schema generation on CI, before checking for a clean repository, so we can keep the generated code up to date with the binary version.

This should fail until #6347 and #6349 are merged.


.github/workflows/ci.yml
```diff
@@ -45,7 +45,7 @@ jobs:
         touch -t ${touchtime} "${filename}"
         ls -la --time-style=full-iso "${filename}"
     - name: Generate
-      run: make generate
+      run: make generate genjsonschema
```
pellared, line 48: Why not adding it to `generate` target in Makefile?

dmathieu: I hesitated, and I can do it.
I went with not doing it because the command takes a bit of time to run, and needs to be executed much less often than generate.

pellared: Reasonable to me if it takes more than a few seconds.
Otherwise I would add it to `generate` target.

pellared (approved):

.github/workflows/ci.yml
dmathieu: The command runs the linter. If we remove it, running it on generate seems fine

pellared: Makes sense. I think not running linter will also be clenaer.

evan-bradley (approved):
Thanks! Wasn't sure where to add this when updating the types, but this is the obvious next step to make sure they don't go out-of-date.

pellared (approved):

pellared:
CC @codeboten

dmathieu:
@open-telemetry/go-approvers this needs a second review.

codeboten (approved):

Makefile
```diff
@@ -342,7 +342,6 @@ genjsonschema: genjsonschema-cleanup $(GOJSONSCHEMA)
 	@echo Modify jsonschema generated files.
 	sed -f ./config/jsonschema_patch.sed ${GENERATED_CONFIG} > ${GENERATED_CONFIG}.tmp
 	mv ${GENERATED_CONFIG}.tmp ${GENERATED_CONFIG}
-	$(MAKE) lint
```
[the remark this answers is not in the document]
MrAlias: I don't follow. We don't run `generate` on `lint`, but we do run `lint` at the end of `genjsonschema`. Why is this changing now that we are running `genjsonschema` in our CI?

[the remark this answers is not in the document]
dmathieu: Ah, I see the misunderstanding. I messed up my updates from the discussion above.

With my latest commit, `jsonschemagen` now runs alongside `generate`. We therefore don't want to run `lint`, because it's already being run in `precommit` and `ci`.

MrAlias (approved):

dmathieu:
This has two reviews from approvers from the same employer.
However, it's also reviewed by the code owner of the `config` package. I am therefore going to merge it.

Issue #6363, opened 2024-11-24
Support WithSpanNameFormatter to otelecho


gifff:
### Problem Statement

`otelecho` lacks support of custom span name formatter like one in `otelhttp`.

### Proposed Solution

Add `WithSpanNameFormatter` option to `otelecho`.

#### Prior Art

https://github.com/open-telemetry/opentelemetry-go-contrib/blob/73c41655248a583825e53729e0f0a5529901d7ea/instrumentation/net/http/otelhttp/config.go#L179-L183

pellared:
Why would we need to create a custom span name? I feel that the instrumentation library should follow the semantic conventions the semantic conventions: https://opentelemetry.io/docs/specs/semconv/http/http-spans/#name. The user can always create a child span with a custom name.

dmathieu:
I would add: is this issue based on an actual need (could you describe it?), or is it just for parity with otelhttp?

flc1125:
mark

gifff:
> Why would we need to create a custom span name? I feel that the instrumentation library should follow the semantic conventions the semantic conventions: https://opentelemetry.io/docs/specs/semconv/http/http-spans/#name. The user can always create a child span with a custom name. 

I’m new to opentelemetry and currently learning to integrate it with existing system. So my initial plan was to annotate the span name with some prefixes so the span name becomes `Echo/HTTP GET /some/actions` for handlers and `ServiceX/HTTP GET /some/other/actions` for outgoing HTTP calls. I decided to use span.kind  for filtering and add custom attribute to identify different http.Client(s).

Thanks for the pointer to the semantic conventions! I haven't gone through it.

flc1125:
> I decided to use span.kind for filtering and add custom attribute to identify different http.Client(s).

This is a very good choice. I am doing the same at the moment. And there are no problems at all.

flc1125:
@pellared For this reason, we need to re-evaluate whether we need to add this feature.

https://github.com/open-telemetry/opentelemetry-go-contrib/issues/6375#issuecomment-2511762722

pellared:
Closing per https://github.com/open-telemetry/opentelemetry-go-contrib/pull/9613

PR #6365, opened 2024-11-24
feat(otelecho): add SpanNameFormatter config and tests

flc1125:
closed https://github.com/open-telemetry/opentelemetry-go-contrib/issues/6363

### DIFF
--- a/instrumentation/github.com/labstack/echo/otelecho/config.go
+++ b/instrumentation/github.com/labstack/echo/otelecho/config.go
@@ -4,17 +4,27 @@
 package otelecho // import "go.opentelemetry.io/contrib/instrumentation/github.com/labstack/echo/otelecho"
 
 import (
+	"github.com/labstack/echo/v4"
 	"github.com/labstack/echo/v4/middleware"
 
 	"go.opentelemetry.io/otel/propagation"
 	oteltrace "go.opentelemetry.io/otel/trace"
 )
 
+// defaultSpanNameFormatter is the default function used for formatting span names.
+var defaultSpanNameFormatter = func(c echo.Context) string {
+	return c.Path()
+}
+
+// SpanNameFormatter is a function that takes an echo.Context and returns a span name.
+type SpanNameFormatter func(c echo.Context) string
+
 // config is used to configure the mux middleware.
 type config struct {
-	TracerProvider oteltrace.TracerProvider
-	Propagators    propagation.TextMapPropagator
-	Skipper        middleware.Skipper
+	TracerProvider    oteltrace.TracerProvider
+	Propagators       propagation.TextMapPropagator
+	Skipper           middleware.Skipper
+	spanNameFormatter SpanNameFormatter
 }
 
 // Option specifies instrumentation configuration options.
@@ -53,5 +63,14 @@ func WithTracerProvider(provider oteltrace.TracerProvider) Option {
 func WithSkipper(skipper middleware.Skipper) Option {
 	return optionFunc(func(cfg *config) {
 		cfg.Skipper = skipper
+	})
+}
+
+// WithSpanNameFormatter specifies a function to use for formatting span names.
+func WithSpanNameFormatter(formatter SpanNameFormatter) Option {
+	return optionFunc(func(cfg *config) {
+		if formatter != nil {
+			cfg.spanNameFormatter = formatter
+		}
 	})
 }
--- a/instrumentation/github.com/labstack/echo/otelecho/test/echo_test.go
+++ b/instrumentation/github.com/labstack/echo/otelecho/test/echo_test.go
@@ -16,11 +16,12 @@ import (
 	"github.com/stretchr/testify/assert"
 	"github.com/stretchr/testify/require"
 
-	"go.opentelemetry.io/contrib/instrumentation/github.com/labstack/echo/otelecho"
 	"go.opentelemetry.io/otel"
 	"go.opentelemetry.io/otel/codes"
 	"go.opentelemetry.io/otel/sdk/trace"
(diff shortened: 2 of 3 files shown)


pellared:
@scorpionknifes, PTAL as code owner.

instrumentation/github.com/labstack/echo/otelecho/config.go
```diff
@@ -4,17 +4,27 @@
...
 	oteltrace "go.opentelemetry.io/otel/trace"
 )
 
+// defaultSpanNameFormatter is the default function used for formatting span names.
+var defaultSpanNameFormatter = func(c echo.Context) string {
+	return c.Path()
+}
+
+// SpanNameFormatter is a function that takes an echo.Context and returns a span name.
+type SpanNameFormatter func(c echo.Context) string
```
scorpionknifes, line 20: I don't like how this takes in `echo.Context` and allows people to mutate it.

I would prefer `func(path string, r *http.Request) string` instead like in otelhttp https://github.com/open-telemetry/opentelemetry-go-contrib/blob/b508e33e7d8c2140cac4e9530ccd4e81b68be079/instrumentation/net/http/otelhttp/transport.go#L33

You would need to do `spanName := cfg.spanNameFormatter(c.Path(), c.Request())` instead. otherwise everything else looks great!

instrumentation/github.com/labstack/echo/otelecho/test/echo_test.go
```diff
@@ -202,3 +200,57 @@ func TestErrorNotSwallowedByMiddleware(t *testing.T) {
...
+				return "custom " + path
+			},
+			expected: "custom /user/:id",
+		},
+		{
+			name: "has_request",
+			formatter: func(path string, req *http.Request) string {
+				return req.Method + " " + path
+			},
+			expected: "GET /user/:id",
```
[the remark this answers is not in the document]
flc1125: Just to preserve the original logic. But I also recommend this as the default.
I'm going to adjust this logic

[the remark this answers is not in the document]
flc1125: Then it seems that we just need to focus more on implementing method and target processing.
I will make adjustments later. and try to resolve the mistakes you raise.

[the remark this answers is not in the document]
flc1125: > I do not see a reason to add `SpanNameFormatter` config for `otelecho`. I added a comment in #6363.
> 
> At the same time, I see that the current span name is not following [open-telemetry/opentelemetry-specification@`v1.20.0`/specification/trace/semantic_conventions/http.md#name](https://github.com/open-telemetry/opentelemetry-specification/blob/v1.20.0/specification/trace/semantic_conventions/http.md#name) which is a separate bug.
> 
> I think that most (all?) HTTP instrumentation libraries may have this bug. _It would be good to double-check it and create bugs for tracking._

I adjusted according to the standards of the document.

Additionally, this will serve as a reference for other components. The adjustments for other components will be supplemented after this PR is approved.

CHANGELOG.md
```diff
@@ -13,6 +13,10 @@ This project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0.htm
 - Added support for providing `endpoint`, `pollingIntervalMs` and `initialSamplingRate` using environment variable `OTEL_TRACES_SAMPLER_ARG` in `go.opentelemetry.io/contrib/samples/jaegerremote`. (#6310)
 - Added support exporting logs via OTLP over gRPC in `go.opentelemetry.io/contrib/config`. (#6340)
 
+### Changed
+
+- Change the logic of span to make it comply with OTEL specifications in `go.opentelemetry.io/contrib/instrumentation/github.com/labstack/echo/otelecho`. (#6365)
```
dmathieu, line 18: No need to specify it's the otel spec. We're within otel, that's assumed :)
Also, someone reading this may not know what the spec says. So mentioning the new value is good.

```suggestion
- Change the span name to be `GET /path` so it complies with the specifications in `go.opentelemetry.io/contrib/instrumentation/github.com/labstack/echo/otelecho`. (#6365)
```

pellared: _I edited @dmathieu suggestion so that it references semantic conventions instead of the specification._

instrumentation/github.com/labstack/echo/otelecho/test/echo_test.go
```diff
@@ -202,3 +200,59 @@ func TestErrorNotSwallowedByMiddleware(t *testing.T) {
...
+
+	tests := []struct {
+		name     string
+		method   string
+		path     string
+		url      string
+		expected string
+	}{
+		// Test for standard methods
+		{"", http.MethodGet, "/user/:id", "/user/123", "GET /user/:id"},
```
pellared, line 216: Just noticed. That you have forgotten to name the test cases 😉

flc1125:
> Side note: Similar fixes needs to be done for all other HTTP instrumentation libraries.
> 
> @flc1125, do you plan to work on it?

As stated here, I will handle it.

https://github.com/open-telemetry/opentelemetry-go-contrib/pull/6365#discussion_r1858622018

pellared (changes requested):

instrumentation/github.com/labstack/echo/otelecho/test/echo_test.go
flc1125: I still want to be a bit lazy; I'll add this part later. 😂

pellared (approved):

dmathieu (approved):

PR #6366, opened 2024-11-24
feat(otelgin): add panic recovery and test panic handling

flc1125:


instrumentation/github.com/gin-gonic/gin/otelgin/gintrace.go
```diff
@@ -10,13 +10,14 @@ import (
...
 
-	"go.opentelemetry.io/contrib/instrumentation/github.com/gin-gonic/gin/otelgin/internal/semconvutil"
 	"go.opentelemetry.io/otel"
 	"go.opentelemetry.io/otel/attribute"
 	"go.opentelemetry.io/otel/codes"
 	"go.opentelemetry.io/otel/propagation"
 	semconv "go.opentelemetry.io/otel/semconv/v1.20.0"
 	oteltrace "go.opentelemetry.io/otel/trace"
+
+	"go.opentelemetry.io/contrib/instrumentation/github.com/gin-gonic/gin/otelgin/internal/semconvutil"
```
pellared, line 20: Can you revert this import rearrange change? We are grouping `go.openteletry.io`. See https://github.com/open-telemetry/opentelemetry-go-contrib/blob/96b1ed1e7b0c1613290df1801ff6ba8474a33ce5/.golangci.yml#L108-L109

instrumentation/github.com/gin-gonic/gin/otelgin/gintrace.go
```diff
@@ -83,6 +84,15 @@ func Middleware(service string, opts ...Option) gin.HandlerFunc {
 		ctx, span := tracer.Start(ctx, spanName, opts...)
 		defer span.End()
 
+		// support recovery
```
pellared, line 87: None of our instrumentation libraries handles panics. _Exception: `otelgin.HTML`, which does not look right to me._
I am against adding such functionality by default as users may have their own panic recovery. 
This code would change the callstack of the panic received by the next panic recovery making it hard to find the main source of the error in their existing panic recovery solution.

pellared (changes requested):

instrumentation/github.com/gin-gonic/gin/otelgin/gintrace.go
flc1125: I was also thinking today whether to deprecate: `otelgin.HTML` in future versions.

flc1125: I am wondering if it will cause a `panic`, resulting in the span not end yet.

instrumentation/github.com/gin-gonic/gin/otelgin/gintrace.go
flc1125: No problem, this is the default behavior of the IDE I configured. Local packages and third-party packages are grouped. I can make some adjustments.

instrumentation/github.com/gin-gonic/gin/otelgin/gintrace.go
pellared: > I am wondering if it will cause a `panic`, resulting in the span not end yet.

It should end as there is a `defer span.End()`.

pellared: > I was also thinking today whether to deprecate: `otelgin.HTML` in future versions.

Works for me

dmathieu: We also recover in `span.End()`.
https://github.com/open-telemetry/opentelemetry-go/blob/8b510c36ed26fafe151df35a251f4589f5959802/sdk/trace/span.go#L420

So any panic happening should already be caught by the `defer span.End()`.

flc1125: OK, I understand. Then this PR should be closed. Thank you all very much.

PR #6367, opened 2024-11-24
[chore]: enable usestdlibvars linter

mmorel-35:
#### Description

[usestdlibvars](https://golangci-lint.run/usage/linters/#usestdlibvars) is a linter that detect the possibility to use variables/constants from the Go standard library.


.golangci.yml
```diff
@@ -2,7 +2,7 @@
 run:
   issues-exit-code: 1 #Default
   tests: true #Default
-  timeout: 2m
+  timeout: 5m
```
pellared, line 5: Why do we need to change from 2 minutes to 5?

dmathieu (approved):

pellared (approved):

Issue #6351, opened 2024-11-19
otlpHTTPLogExporter, otlpGRPCLogExporter, etc do not support fields like OTLP.Certificate


mattsains:
### Description

The OTLP configuration type has fields Certificate, ClientCertificate, ClientKey, HeadersList and Insecure:

https://github.com/open-telemetry/opentelemetry-go-contrib/blob/6e79f7ca22d58345e2ccfe3f9e8bb4ad71633ab3/config/generated_config.go#L258-L264
 

These fields are not read/supported by constructors like [otlpGRPCLogExporter](https://github.com/open-telemetry/opentelemetry-go-contrib/blob/6e79f7ca22d58345e2ccfe3f9e8bb4ad71633ab3/config/log.go#L163-L199), otlpHTTPLogExporter, otlpHTTPMetricExporter, otlpGRPCMetricExporter, otlpGRPCSpanExporter and otlpHTTPSpanExporter. 

### Environment

- OS: N/A
- Architecture: N/A
- Go Version: N/A
- `config` version: [6e79f7c](https://github.com/open-telemetry/opentelemetry-go-contrib/commit/6e79f7ca22d58345e2ccfe3f9e8bb4ad71633ab3), 1.32.0

### Steps To Reproduce
An example of how to trigger this is to call NewSDK specifying a custom certificate. This certificate will not be used when connecting to the trace endpoint specified.

### Expected behavior
These fields should be respected when creating log, trace and metric exporters.

mattsains:
I think that this code is used in the collector's internal telemetry exporting too: https://github.com/open-telemetry/opentelemetry-collector/blob/9d2685f404ef65b46152ae80c36ff84f303e3cc6/service/service.go#L133C5-L138C7

I believe the fix for this is as simple as something like this case for Certificate:

```
if otlpConfig.Certificate != nil {
	creds, err := credentials.NewClientTLSFromFile(*otlpConfig.Certificate, "")
	opts = append(opts, otlptracegrpc.WithTLSCredentials(creds))
}
```

If that sounds good to maintainers, I'd be glad to submit a PR for it, I'd just like to know I'm barking up the right tree first.

ms-jcorley:
I created a PR for a small part of this in the collector repo:
https://github.com/open-telemetry/opentelemetry-collector/pull/11633

dmathieu:
cc @pellared @codeboten as code owners.

codeboten:
[...]

PR #6376, opened 2024-11-27
config: add support for certificate configuration

codeboten:
Part of https://github.com/open-telemetry/opentelemetry-go-contrib/issues/6351


mattsains (commented):
I'm not sure if you consider HTTP (as opposed to GRPC) config to be in scope for this PR, but the `otlpHTTPLogExporter` has a similar problem - http in this context can also mean https

config/log.go
```diff
@@ -178,6 +180,13 @@ func otlpGRPCLogExporter(ctx context.Context, otlpConfig *OTLP) (sdklog.Exporter
 		if u.Scheme == "http" {
 			opts = append(opts, otlploggrpc.WithInsecure())
 		}
+		if otlpConfig.Certificate != nil {
```
mattsains, line 183: I'm curious about why this was put here and not the outer scope on line 190/191 as a sibling of `if otlp.Config.Compression != nil {` ? the other if statements in the inner block pertain to the result of `url.ParseRequestURI`, so if you think it's better to put otlpConfig checks in this block, it would be good to understand when to add them here and when to add them in the outer scope

codeboten:
> I'm not sure if you consider HTTP (as opposed to GRPC) config to be in scope for this PR, but the otlpHTTPLogExporter has a similar problem - http in this context can also mean https

I was planning on submitting a separate PR for the http exporter

mattsains:
> > I'm not sure if you consider HTTP (as opposed to GRPC) config to be in scope for this PR, but the otlpHTTPLogExporter has a similar problem - http in this context can also mean https
> 
> I was planning on submitting a separate PR for the http exporter

I'll do the rest of the changes if you're okay with that. Some of the other fields require overwriting this change anyway

mattsains:
I created https://github.com/open-telemetry/opentelemetry-go-contrib/pull/6378 which is based on this PR

config/log.go
codeboten: i ended up moving the code after thinking about this further. originally i thought it made sense to only have it in the case where an endpoint was configured, but after rethinking through it i moved it out at the same level as compression and the other parameters. figured if there was a problem with the cert config it would make sense to surface it regardless of what the endpoint configuration is

codeboten:
Ah I missed your comment, i ended up implementing the http certificate field in this same PR

.golangci.yml
```diff
@@ -70,6 +70,11 @@ issues:
         - gosec
         - perfsprint
         - usestdlibvars
+    # Ignoring gosec G402: TLS MinVersion too low
+    # as the https://pkg.go.dev/crypto/tls#Config handles MinVersion default well.
```
dmathieu, line 74: Should we move this to be only for the config package, or only for a specific line? Ignoring it globally seems prone to errors.

codeboten: done, is it worth opening an issue to do the same in the core repo?

dmathieu: I would think so, yes.

dmathieu (approved):

codeboten:
@dmathieu can this be merged? it will unblock https://github.com/open-telemetry/opentelemetry-go-contrib/pull/6378

dmathieu:
We need a second approval to merge.

MrAlias (approved):

PR #6381, opened 2024-11-28
otelgin: update default span name to align with semantic conventions

flc1125:
fixes https://github.com/open-telemetry/opentelemetry-go-contrib/issues/7088

> `SpanNameFormatter` has undergone a breaking change, which can be considered when reviewing whether the project necessitates it.  
>
> From my previous recollection, I believe this part was adjusted according to this logic. Moreover, otelgin is still in an unstable version, meaning it permits some breaking changes.


instrumentation/github.com/gin-gonic/gin/otelgin/option.go
```diff
@@ -86,8 +87,7 @@ func WithGinFilter(f ...GinFilter) Option {
 
 // WithSpanNameFormatter takes a function that will be called on every
 // request and the returned string will become the Span Name.
-func WithSpanNameFormatter(f func(r *http.Request) string) Option {
-	return optionFunc(func(c *config) {
-		c.SpanNameFormatter = f
-	})
+// Deprecated: since 0.58.0, remove in 0.59.0.
+func WithSpanNameFormatter(func(r *http.Request) string) Option {
+	return optionFunc(func(*config) {})
```
dmathieu, line 92: A deprecation should keep the original behavior. We shouldn't make this a noop.

instrumentation/github.com/gin-gonic/gin/otelgin/option.go
```diff
@@ -86,8 +87,7 @@ func WithGinFilter(f ...GinFilter) Option {
 
 // WithSpanNameFormatter takes a function that will be called on every
 // request and the returned string will become the Span Name.
-func WithSpanNameFormatter(f func(r *http.Request) string) Option {
-	return optionFunc(func(c *config) {
-		c.SpanNameFormatter = f
-	})
+// Deprecated: since 0.58.0, remove in 0.59.0.
```
dmathieu, line 90: It's helpful for deprecations to shortly explain their reason.

```suggestion
// Deprecated: [0.58.0] the span name now follows semantic conventions.
```

As for the version name, I've switched this to the deprecation strategy used by the collector.
Specify the first version where it's deprecated, since deprecations should always only last for one version.

dmathieu:
Based on the ongoing discussions with the need or not for `SpanNameFormatter`, could you update this PR to only include the default span name?

flc1125:
> Based on the ongoing discussions with the need or not for `SpanNameFormatter`, could you update this PR to only include the default span name?

OK, I will add another draft version of the PR (which may not have passed unit testing and CI).

flc1125:
@dmathieu A draft version: https://github.com/open-telemetry/opentelemetry-go-contrib/pull/6396

flc1125:
Temporarily closed for now; will be reprocessed once there are further developments.

akats7:
@flc1125 I missed this issue the first time around, but what was the blocker here?

flc1125:
> @flc1125 I missed this issue the first time around, but what was the blocker here?

It's been quite some time, and I recall it was because there were still some content or semantic agreements that needed confirmation. Moreover, discussions were still ongoing. So, this matter was temporarily put on hold.

Due to the prolonged lack of progress, I have decided to close it for now. Until there are new developments, I will consider reopening it.

If there are any new changes, I will continue to address this issue.

flc1125:
https://github.com/open-telemetry/opentelemetry-go-contrib/issues/7088 These two issues may have some correlation.

akats7:
Yep this seems to resolve that issue. I don’t see any blockers regarding updating only updating the default span names. Unless the goal was to have the logic sit in the semconv package and avoid duplication.

flc1125:
> Yep this seems to resolve that issue. I don’t see any blockers regarding updating only updating the default span names. Unless the goal was to have the logic sit in the semconv package and avoid duplication.

OK, if there are no problems hindered for now, I will continue to track this issue.

akats7:
> > Yep this seems to resolve that issue. I don’t see any blockers regarding updating only updating the default span names. Unless the goal was to have the logic sit in the semconv package and avoid duplication.
> 
> OK, if there are no problems hindered for now, I will continue to track this issue.

Do you mind either updating the PR title or opening a new PR to make it clear that we’re not deprecating SpanNameFormatter

flc1125:
> Do you mind either updating the PR title or opening a new PR to make it clear that we’re not deprecating SpanNameFormatter

It has been changed. If there is any mistake in the description, can you help me adjust it. (PS: My English is not very good)

flc1125:
Ready for review.

dmathieu (changes requested):
`FullPath()` is a high cardinality value. We should make this the HTTP pattern/route, not the path.

If the path is `/blog/1234`, and the route is defined as `/blog/{id}` then, the latter should be used.

instrumentation/github.com/gin-gonic/gin/otelgin/config.go
```diff
@@ -27,15 +29,35 @@ type config struct {
...
 // be traced. A Filter must return true if the request should be traced.
 type Filter func(*http.Request) bool
 
 // GinFilter filters an [net/http.Request] based on content of a [gin.Context].
 type GinFilter func(*gin.Context) bool
 
-// SpanNameFormatter is used to set span name by http.request.
-type SpanNameFormatter func(r *http.Request) string
+// SpanNameFormatter is used to set span name by gin's FullPath() and http.request.
+type SpanNameFormatter func(path string, r *http.Request) string
```
dmathieu, line 60: `http.Request` includes the full path. Why do we need a second attribute here?

flc1125: The `FullPath()` method in Gin might be somewhat misleading in its naming. In reality, it returns the route pattern, such as `/users/:id`, rather than specific paths like `/users/1` or `/users/2`.

The primary purpose of defining `path` here stems from the fact that `FullPath` defaults to retrieving the routing rules of the Gin framework, and these routing rules cannot be obtained from `*http.Request`.

> PS: Additionally, some designs or definitions from other components were referenced here. As a result, my understanding of `path` may lean more toward being the custom routing rules defined by the component package.

References:

- An example of other components: https://github.com/open-telemetry/opentelemetry-go-contrib/blob/eda91cfdbe1732aed416f457946eb092833bcbc0/instrumentation/github.com/gorilla/mux/otelmux/mux.go#L132C57-L132C65
- gin framework `FullPath` method: https://github.com/gin-gonic/gin/blob/49e9137c68e6dfaa529a2d0c9fe64d9e69a8554e/context.go#L170-L178

flc1125: I think this is acceptable. Additionally, there is one item to confirm:

Should we adjust the default `SpanName`? Currently, it is based on `c.FullPath()`. Do we need to change it to `req.Pattern`? (Of course, this would introduce a breaking change at the data level.)

Also, the details of this part (including whether it involves breaking data changes) might be related to https://github.com/open-telemetry/opentelemetry-go-contrib/pull/6982/files.

akats7: > role

I feel the same way, I’m not sure in which package this started but it feels like an anti pattern to me

flc1125: > If we do end up having a breaking change, how about using gin.Context then? That way, we reduce the risk of further breaking changes down the line.

I agree with this approach, and it offers excellent scalability.

instrumentation/github.com/gin-gonic/gin/otelgin/test/gin_test.go
```diff
@@ -268,24 +268,43 @@ func TestSpanStatus(t *testing.T) {
...
+	provider := sdktrace.NewTracerProvider(
+		sdktrace.WithSpanProcessor(sr),
+	)
+
 	testCases := []struct {
+		method            string
+		route             string
 		requestPath       string
 		spanNameFormatter otelgin.SpanNameFormatter
 		wantSpanName      string
```
akats7, line 281: Can we make this expectedSpanName please

akats7: Sorry just updated the comment, I'm asking if wantSpanName can be updated.

akats7 (approved):

CHANGELOG.md
```diff
@@ -43,6 +43,7 @@ This project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0.htm
 - Update the Jaeger remote sampler to use "github.com/jaegertracing/jaeger-idl/proto-gen/api_v2" in  `go.opentelemetry.io/contrib/samplers/jaegerremote`. (#7061)
 - Improve performance by reducing allocations in the gRPC stats handler in `go.opentelemetry.io/contrib/instrumentation/google.golang.org/grpc/otelgrpc`. (#7186)
 - Update `http.route` attribute to support `request.Pattern` in `go.opentelemetry.io/contrib/instrumentation/github.com/gorilla/mux/otelmux`. (#7108)
+- Change the default span name to be `GET /path` so it complies with the OTel HTTP semantic conventions in `go.opentelemetry.io/contrib/instrumentation/github.com/gin-gonic/gin/otelgin`. (#6381)
```
dmathieu, line 46:
```suggestion
- Change the default span name to be `GET /path` so it complies with the HTTP semantic conventions in `go.opentelemetry.io/contrib/instrumentation/github.com/gin-gonic/gin/otelgin`. (#6381)
```

instrumentation/github.com/gin-gonic/gin/otelgin/test/gin_test.go
```diff
@@ -268,27 +268,46 @@ func TestSpanStatus(t *testing.T) {
...
+		sdktrace.WithSpanProcessor(sr),
+	)
+
 	testCases := []struct {
+		method            string
+		route             string
 		requestPath       string
 		spanNameFormatter otelgin.SpanNameFormatter
-		wantSpanName      string
+		expectedSpanName  string
```
dmathieu, line 281: Actually, `want` is the proper convention we should be using in all tests. Why did you change that?

dmathieu (commented):
Left minor comments. It overall looks good.

instrumentation/github.com/gin-gonic/gin/otelgin/test/gin_test.go
flc1125: 😁 Later, I adjusted back.

instrumentation/github.com/gin-gonic/gin/otelgin/config.go
```diff
@@ -27,15 +29,35 @@ type config struct {
...
 // Filter is a predicate used to determine whether a given http.request should
 // be traced. A Filter must return true if the request should be traced.
 type Filter func(*http.Request) bool
 
 // GinFilter filters an [net/http.Request] based on content of a [gin.Context].
 type GinFilter func(*gin.Context) bool
 
-// SpanNameFormatter is used to set span name by http.request.
-type SpanNameFormatter func(r *http.Request) string
+// SpanNameFormatter is a function that takes a gin.Context and returns a string.
```
dmathieu, line 59: This comment is like showing a box of chocolate and saying "this is chocolate". That's not really useful.
The previous comment, explaining why we need that type is a better comment IMHO.

```suggestion
// SpanNameFormatter is used by `WithSpanNameFormatter` to customize the request's span name.
```

instrumentation/github.com/gin-gonic/gin/otelgin/test/gin_test.go
akats7: Sorry that’s my fault, I gave the wrong guidance

PR #6391, opened 2024-12-02
otelhttp: close wrappedBody span only once

dimitarvdimitrov:
The span was otherwise closed twice: once when the body is exhausted with an `io.EOF` and once when the body is explicitly `Close`d. As a result the span was shipped twice from the application and was leading to duplicate traces like so
Screenshot 2024-12-02 at 19 45 54

This is after this fix
Screenshot 2024-12-02 at 19 46 11

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -26,6 +26,7 @@ This project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0.htm
 - Fixed the value for configuring the OTLP exporter to use `grpc` instead of `grpc/protobuf` in `go.opentelemetry.io/contrib/config`. (#6338)
 - Allow marshaling types in `go.opentelemetry.io/contrib/config`. (#6347)
 - Removed the redundant handling of panic from the `HTML` function in `go.opentelemetry.io/contrib/instrumentation/github.com/gin-gonic/gin/otelgin`. (#6373)
+- Close spans only once in `go.opentelemetry.io/contrib/instrumentation/net/http/otelhttp. (#6391)
 
 <!-- Released section -->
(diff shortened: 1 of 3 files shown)


dmathieu:
Note that your need for this means your SDK is not compliant with the specifications.

https://github.com/open-telemetry/opentelemetry-specification/blob/50027a1036746dce293ee0a8592639f131fc1fb8/specification/trace/api.md#end

> Implementations SHOULD ignore all subsequent calls to End and any other Span methods, i.e. the Span becomes non-recording by being ended (there might be exceptions when Tracer is streaming events and has no mutable state associated with the Span).

(this shouldn't be one of those exceptions).

dmathieu:
Could you add a test to prevent regressions?

dmathieu:
I have doubts about this. As mentioned above, this should be handled by the SDK. Spans can be ended twice, the second time should be a noop.

The default SDK does this:
https://github.com/open-telemetry/opentelemetry-go/blob/9f82e51c894278a30980deae2639f005b8fedea2/sdk/trace/span.go#L459

So this would rather be something you'd have to fix in your SDK IMHO.

dimitarvdimitrov:
> So this would rather be something you'd have to fix in your SDK IMHO.

makes sense. I'll close this now. Thanks for taking the time

dmathieu:
> Do you think there's room for the spec verbiage on the godocs in [opentelemetry-go/trace](https://github.com/open-telemetry/opentelemetry-go/blob/trace/v1.32.0/trace/span.go#L28-L32)?

It could possibly be made clearer, but this is kind of what the doc already says with:

> Therefore, updates to the Span are not allowed after this method has been called.

Issue #5968, opened 2024-08-01
Pool metric options created in the otelhttp instrumentation


MadVikingGod:
Pooling the options created by otelhttp instrumentation should reduce allocations and improve performance.  

### Tasks

- [ ] Use a pool when creating metric options.
- [ ] Create benchmark showing this does reduce allocations.

_Originally posted by @MrAlias in https://github.com/open-telemetry/opentelemetry-go-contrib/pull/5818#discussion_r1669451847_

vordimous:
This looks to be a straightforward pool for the objects created in `RecordMetrics()`. What is the best way to benchmark the number of allocations? I would like to take on this issue.

dmathieu:
A benchmark with the `ReportAllocs` option will allow you to compare the number of allocations between two different implementations.

See this as an example:
https://github.com/open-telemetry/opentelemetry-go-contrib/blob/a59214f7c43e0cc0741a56a8c75c422817b49c3e/instrumentation/net/http/otelhttp/internal/semconv/bench_test.go#L19

vordimous:
I have implemnted two benchmark functions in `env_test.go` that run the same code as the `TestHTTPServerDoesNotPanic()`. I added a benchmark for the unpooled code and pooled code with both reporting `0 allocs/op`. I may have done something wrong or identified that pooling isn't needed?

```
BenchmarkRecordMetricsPooled-10    	1000000000	         0.0000000 ns/op	       0 B/op	       0 allocs/op
BenchmarkRecordMetrics-10          	1000000000	         0.0000001 ns/op	       0 B/op	       0 allocs/op
```

Full test results:

```
Running tool: /opt/homebrew/bin/go test -timeout 30s -run ^(TestHTTPServerDoesNotPanicPooled|TestHTTPServerDoesNotPanic|TestHTTPClientDoesNotPanic)$ go.opentelemetry.io/contrib/instrumentation/net/http/otelhttp/internal/semconv -v --bench . --benchmem

=== RUN   TestHTTPServerDoesNotPanicPooled
=== RUN   TestHTTPServerDoesNotPanicPooled/empty
=== RUN   TestHTTPServerDoesNotPanicPooled/nil_meter
=== RUN   TestHTTPServerDoesNotPanicPooled/with_Meter
--- PASS: TestHTTPServerDoesNotPanicPooled (0.00s)
    --- PASS: TestHTTPServerDoesNotPanicPooled/empty (0.00s)
[...]

PR #6394, opened 2024-12-02
Add method and benchmarks for pooling metric options

vordimous:
This PR improves the `RecordMetrics()` allocations using a pool for the `metric.AddOption` slice.

`benchstat main.txt pool.txt`:
```
goos: darwin
goarch: arm64
pkg: go.opentelemetry.io/contrib/instrumentation/net/http/otelhttp/internal/semconv
cpu: Apple M1 Max
                            │  main.txt   │              pool.txt              │
                            │   sec/op    │   sec/op     vs base               │
RecordMetrics/empty-10        4.513n ± 3%   4.620n ± 0%  +2.38% (p=0.037 n=10)
RecordMetrics/nil_meter-10    341.8n ± 1%   330.7n ± 4%  -3.26% (p=0.022 n=10)
RecordMetrics/with_Meter-10   341.1n ± 1%   330.0n ± 1%  -3.27% (p=0.000 n=10)
geomean                       80.74n        79.59n       -1.42%

                            │   main.txt   │              pool.txt               │
                            │     B/op     │    B/op     vs base                 │
RecordMetrics/empty-10        0.000 ± 0%     0.000 ± 0%       ~ (p=1.000 n=10) ¹
RecordMetrics/nil_meter-10    696.0 ± 0%     680.0 ± 0%  -2.30% (p=0.000 n=10)
RecordMetrics/with_Meter-10   696.0 ± 0%     680.0 ± 0%  -2.30% (p=0.000 n=10)
geomean                                  ²               -1.54%                ²
¹ all samples are equal
² summaries must be >0 to compute geomean

                            │   main.txt   │               pool.txt               │
                            │  allocs/op   │ allocs/op   vs base                  │
RecordMetrics/empty-10        0.000 ± 0%     0.000 ± 0%        ~ (p=1.000 n=10) ¹
RecordMetrics/nil_meter-10    6.000 ± 0%     5.000 ± 0%  -16.67% (p=0.000 n=10)
RecordMetrics/with_Meter-10   6.000 ± 0%     5.000 ± 0%  -16.67% (p=0.000 n=10)
geomean                                  ²               -11.45%                ²
¹ all samples are equal
² summaries must be >0 to compute geomean
```

Fixes https://github.com/open-telemetry/opentelemetry-go-contrib/issues/5968


instrumentation/net/http/otelhttp/internal/semconv/env_test.go
```diff
@@ -132,3 +169,73 @@ func NewTestHTTPClient() HTTPClient {
 		latencyMeasure:       &testInst{},
 	}
 }
+
+func BenchmarkRecordMetricsPooled(b *testing.B) {
```
dmathieu, line 173: These two benchmarks won't provide any meaningful value.
You need to split the test cases into sub benchmarks, so each of them provides its own value.

`ReportAllocs` and `ResetTimer` need to be called before the action we want to call, since they specify the data we want to record and when we want to start recording.

You may want to look at https://dave.cheney.net/2013/06/30/how-to-write-benchmarks-in-go as a guidance on how to write benchmarks.
This repository and the opentelemetry-go one also have several benchmarks you can look at as examples.

vordimous: I have fixed the benchmarks and am now getting results:

```
goos: darwin
goarch: arm64
pkg: go.opentelemetry.io/contrib/instrumentation/net/http/otelhttp/internal/semconv
cpu: Apple M1 Max
BenchmarkHTTPServerRequest
BenchmarkHTTPServerRequest-10      	 3890485	       287.3 ns/op	     712 B/op	       2 allocs/op
BenchmarkRecordMetricsPooled
BenchmarkRecordMetricsPooled-10    	  552594	      2192 ns/op	    4073 B/op	      28 allocs/op
BenchmarkRecordMetrics
BenchmarkRecordMetrics-10          	  537703	      2152 ns/op	    4104 B/op	      30 allocs/op
PASS
ok  	go.opentelemetry.io/contrib/instrumentation/net/http/otelhttp/internal/semconv	5.713s
```

The number of allocs/op grew linearly for both benchmarks with double the amount of calls resulting in 56 Pooled and 60 NotPooled allocs/op.

The Pooled implementation is slightly faster and has slightly fewer bytes and allocs per operation. Do you want me to keep this benchmark in this file or another file or just move forward with replacing the `RecordMetrics()` with the pooling code?

dmathieu: You only changed one benchmark. They need to all be changed.
Also, rather than setting up a brand new method (which kind of borks the test. We need to be able to benchmark the actual change that will be made), could you setup one benchmark, and run it against both the main branch, and this PR?

You can also use the [benchstat](https://pkg.go.dev/golang.org/x/perf/cmd/benchstat) tool to make the comparison between two runs of the same benchmark more readable.

Basically, this PR should include:

* The change you wish to make, as it will be in the end.
* One benchmark, that tests that change's performance.
* In the PR description, the benchstat comparison of that benchmark against the main branch, and your branch.

vordimous: Got it. I intended to remove the extra methods once there was proof that pooling was an improvement. I will update the PR to fit this structure and move it out of draft.

instrumentation/net/http/otelhttp/internal/semconv/env_test.go
```diff
@@ -132,3 +132,39 @@ func NewTestHTTPClient() HTTPClient {
...
+			_ = tt.server.RequestTraceAttrs("stuff", req)
+			_ = tt.server.ResponseTraceAttrs(ResponseTelemetry{StatusCode: 200})
+			tt.server.RecordMetrics(context.Background(), ServerMetricData{
+				ServerName: "stuff",
+				MetricAttributes: MetricAttributes{
+					Req: req,
+				},
+			})
+		}
+	}
```
dmathieu, line 169:
```suggestion
	for _, tt := range testCases {
		t.Run(tt.name, func(t *testing.T) {
			req, _ := http.NewRequest("GET", "http://example.com", nil)

			_ = tt.server.RequestTraceAttrs("stuff", req)
			_ = tt.server.ResponseTraceAttrs(ResponseTelemetry{StatusCode: 200})
			for i := 0; i < b.N; i++ {
				tt.server.RecordMetrics(context.Background(), ServerMetricData{
					ServerName: "stuff",
					MetricAttributes: MetricAttributes{
						Req: req,
					},
				})
			}
		})
	}
```

With subtests, you'll get each test case as its own benchmark.
Also, we don't want to run the `b.N` loop for the entire code. We don't care about the performance of the `http` package, or `RequestTraceAttrs`. So we should only call that loop for `RecordMetrics`.

instrumentation/net/http/otelhttp/internal/semconv/env.go
```diff
@@ -102,18 +103,27 @@ type MetricData struct {
...
+var metricAddOptionPool = &sync.Pool{
+	New: func() interface{} {
+		return &[]metric.AddOption{}
+	},
+}
+
 func (s HTTPServer) RecordMetrics(ctx context.Context, md ServerMetricData) {
 	if s.requestBytesCounter == nil || s.responseBytesCounter == nil || s.serverLatencyMeasure == nil {
-		// This will happen if an HTTPServer{} is used insted of NewHTTPServer.
+		// This will happen if an HTTPServer{} is used instead of NewHTTPServer.
```
dmathieu, line 114: Could you open another PR for those typo fixes?

instrumentation/net/http/otelhttp/internal/semconv/env_test.go
```diff
@@ -153,18 +153,20 @@ func BenchmarkRecordMetrics(b *testing.B) {
...
 			_ = tt.server.RequestTraceAttrs("stuff", req)
 			_ = tt.server.ResponseTraceAttrs(ResponseTelemetry{StatusCode: 200})
-			tt.server.RecordMetrics(context.Background(), ServerMetricData{
-				ServerName: "stuff",
-				MetricAttributes: MetricAttributes{
-					Req: req,
-				},
-			})
-		}
+			for i := 0; i < b.N; i++ {
```
dmathieu, line 162: You should also reset the timer right before this loop.

vordimous:
@dmathieu I implemented the feedback and followed the `BenchmarkCoreWrite()` to have each case benchmarked. 

I also created the spelling fix PR: https://github.com/open-telemetry/opentelemetry-go-contrib/pull/6407

instrumentation/net/http/otelhttp/internal/semconv/env_test.go
```diff
@@ -132,3 +132,42 @@ func NewTestHTTPClient() HTTPClient {
...
+	for _, bm := range benchmarks {
+		b.Run(bm.name, func(b *testing.B) {
+			req, _ := http.NewRequest("GET", "http://example.com", nil)
+			_ = bm.server.RequestTraceAttrs("stuff", req)
+			_ = bm.server.ResponseTraceAttrs(ResponseTelemetry{StatusCode: 200})
+
+			b.ReportAllocs()
+			b.ResetTimer()
+			for i := 0; i < b.N; i++ {
+				bm.server.RecordMetrics(context.Background(), ServerMetricData{
```
dmathieu, line 164: If you create the context outside the loop, you'll reduce the number of allocations within the benchmarks, making them more accurate.

instrumentation/net/http/otelhttp/internal/semconv/env.go
```diff
@@ -110,10 +117,13 @@ func (s HTTPServer) RecordMetrics(ctx context.Context, md ServerMetricData) {
 
 	attributes := oldHTTPServer{}.MetricAttributes(md.ServerName, md.Req, md.StatusCode, md.AdditionalAttributes)
 	o := metric.WithAttributeSet(attribute.NewSet(attributes...))
-	addOpts := []metric.AddOption{o}
-	s.requestBytesCounter.Add(ctx, md.RequestSize, addOpts...)
-	s.responseBytesCounter.Add(ctx, md.ResponseSize, addOpts...)
+	addOpts := metricAddOptionPool.Get().(*[]metric.AddOption)
```
dmathieu, line 120: Shouldn't we empty the slice here?

dmathieu (commented):
This is starting to look good :)
We will need a changelog entry though.

instrumentation/net/http/otelhttp/internal/semconv/env.go
vordimous: I empty the slice before I put it back in the pool. I didn't want to be redundant, but if it is better to empty it again after getting it from the pool, that sounds good to me.

vordimous: I updated the code in case it was better practice, but I can revert if needed. Let me know if it looks good.

dmathieu: Ah right, sorry, I missed it. Before or after doesn't change much IMHO (with your latest commit, you're doing it both before and after though).

CHANGELOG.md
```diff
@@ -14,6 +14,7 @@ This project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0.htm
 - Added support exporting logs via OTLP over gRPC in `go.opentelemetry.io/contrib/config`. (#6340)
 - The `go.opentelemetry.io/contrib/bridges/otellogr` module.
   This module provides an OpenTelemetry logging bridge for `github.com/go-logr/logr`. (#6386)
+- Updated the `go.opentelemetry.io/contrib/instrumentation/net/http/otelhttp/internal/semconv` `RecordMetrics()` to used a pool for the addOptions slice which reduced allocs/op. (#6394)
```
dmathieu, line 17:
```suggestion
- Use a pool for the semantic convention metrics options in `go.opentelemetry.io/contrib/instrumentation/net/http/otelhttp`. (#6394)
```

instrumentation/net/http/otelhttp/internal/semconv/env.go
vordimous: Ok, I will do it after since I am guessing it is less ideal to have objects with different states in the pool.

dmathieu (approved):

pellared (approved):
Can you please update the branch?

vordimous:
> Thanks for the PR. Can you please update the branch?

PR updated to resolve conflicts

instrumentation/net/http/otelhttp/internal/semconv/env.go
```diff
@@ -102,18 +103,27 @@ type MetricData struct {
...
+var metricAddOptionPool = &sync.Pool{
+	New: func() interface{} {
+		return &[]metric.AddOption{}
+	},
+}
+
 func (s HTTPServer) RecordMetrics(ctx context.Context, md ServerMetricData) {
 	if s.requestBytesCounter == nil || s.responseBytesCounter == nil || s.serverLatencyMeasure == nil {
-		// This will happen if an HTTPServer{} is used instead of NewHTTPServer.
+		// This will happen if an HTTPServer{} is used insted of NewHTTPServer.
```
MrAlias, line 114:
```suggestion
		// This will happen if an HTTPServer{} is used instead of NewHTTPServer.
```

instrumentation/net/http/otelhttp/internal/semconv/env.go
```diff
@@ -206,7 +216,7 @@ func (c HTTPClient) MetricOptions(ma MetricAttributes) MetricOpts {
 
 func (s HTTPClient) RecordMetrics(ctx context.Context, md MetricData, opts MetricOpts) {
 	if s.requestBytesCounter == nil || s.latencyMeasure == nil {
-		// This will happen if an HTTPClient{} is used instead of NewHTTPClient().
+		// This will happen if an HTTPClient{} is used insted of NewHTTPClient().
```
MrAlias, line 219:
```suggestion
		// This will happen if an HTTPClient{} is used instead of NewHTTPClient().
```

instrumentation/net/http/otelhttp/internal/semconv/env.go
```diff
@@ -218,7 +228,7 @@ func (s HTTPClient) RecordMetrics(ctx context.Context, md MetricData, opts Metri
 
 func (s HTTPClient) RecordResponseSize(ctx context.Context, responseData int64, opts metric.AddOption) {
 	if s.responseBytesCounter == nil {
-		// This will happen if an HTTPClient{} is used instead of NewHTTPClient().
+		// This will happen if an HTTPClient{} is used insted of NewHTTPClient().
```
MrAlias, line 231:
```suggestion
		// This will happen if an HTTPClient{} is used instead of NewHTTPClient().
```

MrAlias (approved):

CHANGELOG.md
```diff
@@ -15,6 +15,7 @@ This project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0.htm
 - The `go.opentelemetry.io/contrib/bridges/otellogr` module.
   This module provides an OpenTelemetry logging bridge for `github.com/go-logr/logr`. (#6386)
 - Added SNS instrumentation in `go.opentelemetry.io/contrib/instrumentation/github.com/aws/aws-sdk-go-v2/otelaws`. (#6388)
+- Use a pool for the semantic convention metrics options in `go.opentelemetry.io/contrib/instrumentation/net/http/otelhttp`. (#6394)
```
MrAlias, line 18:
```suggestion
- Use a pool for the semantic convention metric options in `go.opentelemetry.io/contrib/instrumentation/net/http/otelhttp`. (#6394)
```

PR #6400, opened 2024-12-04
Switch the otelhttp semconv tests to use metricdatatest

dmathieu:
This switches the otelhttp semconv tests to use the [metricdatatest](https://pkg.go.dev/go.opentelemetry.io/otel/sdk/metric/metricdata/metricdatatest) package rather than a custom struct with a partial metrics implementation.

This is meant to make it easier to start implementing the client and server metrics.


dmathieu:
This PR may look big based on its diff size, but it's mostly just moving files into a `test` subpackage (to avoid adding the SDK as a dependency of otelhttp), and making some private methods public (since we're in an internal package, that has no public impact) so they can be used by the subpackage.

instrumentation/net/http/otelhttp/internal/semconv/test/go.mod
```diff
@@ -0,0 +1,25 @@
+module go.opentelemetry.io/contrib/instrumentation/net/http/otelhttp/internal/semconv/test
+
+go 1.23.3
```
pellared, line 3:
```suggestion
go 1.22
```

dmathieu:
Ping @open-telemetry/go-approvers for review.

dashpole (approved):

MrAlias (approved):

Issue #6262, opened 2024-10-17
Produce New Server Metrics


MrAlias:

dennisme:
I'd like to do this one!

dmathieu:
This issue hasn't been updated in quite some time. Unassigning it so someone else can pick it up.

flc1125:
I can give it a try.

flc1125:
Thanks.

Give me some time, I need to understand it, including some information about Metric.

flc1125:
I've spent some time studying the code recently, and I found that the design might be better with some refactorings or improvements. 

Here are some of my ideas: (Of course, this part of the improvement has nothing to do with the current issue)__

**Question:**

- We use `oldHTTPServer` and `newHTTPServer` to distinguish the semantic differences between the corresponding 1.20 and 1.26 versions.
- Suppose there are still 1.27, 1.28... When the time comes, do we need to add many new methods to solve them?

**My thoughts:**

- Separate different versions of related methods, such as `RequestTraceAttrs`, etc., and abstract them into interfaces
- We implement this interface through multiple versions
- We inject these implementations through `semconv.HTTPServer`, and we can implement it at any time even if new versions change or add in the future.

**Other thoughts**, thoughts on the version section of `semconv`

- We can inject up to 3 versions through `semconv.HTTPServer`.
- One is the latest version, the other is the previous version that is maintained, and the other is the version that is planned to be scrapped and is not maintained and has an abandonment time

dmathieu:
We should keep things simple.
We currently have breaking changes because we used semantic conventions that were not stable yet.
The state of the HTTP semantic conventions is now that they are stable, with parts that are still experimental. Once something is stable, the only way it would have breaking changes would be with a major release.

This is why we only emit stable metrics, not the experimental ones so far. If we wish to start emiting experimental metrics, we should provide a way to mark that specific metric as experimental, to make upgrades smoother (with no migration path).

With that, I think we can assume there won't be three semantic convention versions with breaking changes at the same time.
So we shouldn't need to handle that case.

Also, the [migration plan](https://github.com/open-telemetry/opentelemetry-go-contrib/issues/5132#issuecomment-2137930048) keeps two versions running alongside each other for a fairly limited amount of time.
This is taking longer, because only tracing has been tackled so far. But once metrics are done, we should be able to remove the duplicate versions about a quarter later.

[...]

PR #6411, opened 2024-12-09
feat(otelhttp): generate New Server Metrics in `otelhttp`

flc1125:
closed https://github.com/open-telemetry/opentelemetry-go-contrib/issues/6262


flc1125:
Additionally, I would like to ask if there is a test package similar to tracetest for metric? I checked the code and did not find anything similar.

dmathieu:
There is metricdatatest. I am changing the test struct used in this PR to use that in https://github.com/open-telemetry/opentelemetry-go-contrib/pull/6400

flc1125:
> There is metricdatatest. I am changing the test struct used in this PR to use that in https://github.com/open-telemetry/opentelemetry-go-contrib/pull/6400

No problem. i can wait for it to merge.

MrAlias:
> > There is metricdatatest. I am changing the test struct used in this PR to use that in #6400
> 
> No problem. i can wait for it to merge.

Merged.

flc1125:
Recently quite busy, doing various year-end summaries and planning. The progress will be a bit slow, please understand.

flc1125:
Already prepared for review.

> If the unit test coverage requirement is very strict, I can supplement it later.

CHANGELOG.md
```diff
@@ -8,6 +8,10 @@ This project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0.htm
 
 ## [Unreleased]
 
+### Added
+
+- Set the environment variable `OTEL_SEMCONV_STABILITY_OPT_IN` to `http/dup` to generate new server metric information based on the `semconv` 1.26 version in `go.opentelemetry.io/contrib/instrumentation/net/http/otelhttp`. (#6411)
```
```suggestion
- Generate server metrics with semantic conventions v1.26 in `go.opentelemetry.io/contrib/instrumentation/net/http/otelhttp` when `OTEL_SEMCONV_STABILITY_OPT_IN` is set to `http/dup`. (#6411)
```

instrumentation/net/http/otelhttp/internal/semconv/v1.20.0.go
```diff
@@ -144,7 +143,7 @@ func (o OldHTTPServer) MetricAttributes(server string, req *http.Request, status
 
 	attributes := slices.Grow(additionalAttributes, n)
 	attributes = append(attributes,
-		standardizeHTTPMethodMetric(req.Method),
+		semconv.HTTPMethod(standardizeHTTPMethod(req.Method)),
```
dmathieu, line 146: `semconv.HTTPMethod` is just an alias for string. We don't need to force that type here.

```suggestion
		standardizeHTTPMethod(req.Method),
```

instrumentation/net/http/otelhttp/internal/semconv/v1.20.0.go
```diff
@@ -214,7 +213,7 @@ func (o OldHTTPClient) MetricAttributes(req *http.Request, statusCode int, addit
...
```
dmathieu, line 216:
```suggestion
		standardizeHTTPMethod(req.Method),
```

instrumentation/net/http/otelhttp/internal/semconv/httpconv.go
```diff
@@ -199,6 +202,86 @@ func (n CurrentHTTPServer) Route(route string) attribute.KeyValue {
...
+
+	return requestBodySizeHistogram, responseBodySizeHistogram, requestDurationHistogram
+}
+
+func (n CurrentHTTPServer) MetricAttributes(server string, req *http.Request, statusCode int, additionalAttributes []attribute.KeyValue) []attribute.KeyValue {
+	num := len(additionalAttributes) + 3
+	var host string
+	var p int
+	if server == "" {
+		host, p = SplitHostPort(req.Host)
```
dmathieu, line 239: Could we improve test coverage on this method?

instrumentation/net/http/otelhttp/internal/semconv/v1.20.0.go
flc1125: https://github.com/flc1125/opentelemetry-go-contrib/blob/fb25ee409b7c398366502223693f2a50909f7817/instrumentation/net/http/otelhttp/internal/semconv/util.go#L100-L108

Because the new and old logic are consistent. But the format of semconv is inconsistent. So I pulled it out.

dashpole (approved):

instrumentation/net/http/otelhttp/internal/semconv/v1.20.0.go
dmathieu: Yes, I'm not talking about the extraction of the method, but the use of `semconv.HTTPMethod`.
That's a nit though, so we can ship without this change.

PR #6412, opened 2024-12-09
config: separate schema version support into separate directories

codeboten:
This makes the migration from one version to another explicit in any downstream dependencies. This will make dealing with breaking changes much easier in the Collector as it will allow us to have end users toggle a feature gate to enable support for new schema versions.

Doing this today is not possible as we cannot import both a v0.2.0 and a v0.3.0 compatible config package. With the change in this PR, this will be possible.

v0.2.0 directory was a copy based on 355fbbdb83be04e56837c8f4d583dae8bcc137bb


MrAlias (commented):
Seems like a good approach to me. It matches our semconv packaging.

The test failures need to be addressed, but other than that looks good at a high level.

codeboten:
test fixed 👍🏻

dmathieu (approved):

MrAlias (approved):

PR #6425, opened 2024-12-11
config: fix panic on nil value in headers name/value pair

codeboten:
Found a bug during testing of v0.3.0 support with the Collector where the value passed in to the headers name/value pair list could be nil, causing a dereferencing error. This fixes that bug.


dmathieu (approved):

dashpole (approved):

config/v0.3.0/config_yaml.go
```diff
@@ -30,6 +31,27 @@ func (j *AttributeNameValueType) UnmarshalYAML(unmarshal func(interface{}) error
...
+// UnmarshalYAML implements json.Unmarshaler.
+func (j *NameStringValuePair) UnmarshalYAML(unmarshal func(interface{}) error) error {
+	var raw map[string]interface{}
+	if err := unmarshal(&raw); err != nil {
+		return err
+	}
+	if _, ok := raw["name"]; raw != nil && !ok {
+		return errors.New("yaml: cannot unmarshal field name in NameStringValuePair required")
+	}
+	if _, ok := raw["value"]; raw != nil && !ok {
```
MrAlias, line 43: Why check `raw != nil`? If `raw == nil` then `ok` will always be `false` and this will evaluate to `true` with only `!ok`.

config/v0.3.0/config_yaml.go
```diff
@@ -30,6 +31,27 @@ func (j *AttributeNameValueType) UnmarshalYAML(unmarshal func(interface{}) error
...
+		return errors.New("yaml: cannot unmarshal field value in NameStringValuePair required")
```
MrAlias, line 44: This looks like a duplicate. Am I missing something?

config/v0.3.0/config_yaml.go
```diff
@@ -30,6 +31,27 @@ func (j *AttributeNameValueType) UnmarshalYAML(unmarshal func(interface{}) error
...
+	}
+	type Plain NameStringValuePair
+	var plain Plain
+	if err := unmarshal(&plain); err != nil {
+		return err
+	}
```
MrAlias, line 50: Calling `unmarshal` seems quite inefficient. My not just use the already unmashaled `raw` to build `Plain`?

config/v0.3.0/config_yaml.go
codeboten: Fixed

config/v0.3.0/config_yaml.go
codeboten: it's a good question, this code is originally from the code generated with go-jsonschema, it could use some additional cleanups since we're not autogenerating the unmarshaling code anymore

codeboten: tried fixing this by using the raw fields and ensuring the raw types match what's expected. ptal

config/v0.3.0/config_json.go
```diff
@@ -112,16 +112,24 @@ func (j *NameStringValuePair) UnmarshalJSON(b []byte) error {
...
 	}
 	type Plain NameStringValuePair
-	var plain Plain
-	if err := json.Unmarshal(b, &plain); err != nil {
-		return err
+	plain := Plain{
+		Name:  name,
+		Value: &value,
 	}
 	*j = NameStringValuePair(plain)
```
MrAlias, line 134: nit: no need to the `Plain` type anymore:

```suggestion
	*j = NameStringValuePair{
		Name:  name,
		Value: &value,
	}

File: config/v0.3.0/config_yaml.go [hunk 5]
```diff
@@ -30,6 +31,35 @@ func (j *AttributeNameValueType) UnmarshalYAML(unmarshal func(interface{}) error
...
+	var name, value string
+	var ok bool
+	if name, ok = raw["name"].(string); !ok {
+		return errors.New("yaml: cannot unmarshal field name in NameStringValuePair must be string")
+	}
+	if value, ok = raw["value"].(string); !ok {
+		return errors.New("yaml: cannot unmarshal field value in NameStringValuePair must be string")
+	}
+	type Plain NameStringValuePair
```
MrAlias, line 54: Similar here.

MrAlias (approved):

config/v0.3.0/config_json.go
codeboten: Cleaned up

PR #6476, opened 2024-12-15
update(renovate): add Go constraint

flc1125:
This ensures that updates to versions that are not supported by 1.22 are avoided. Until you change this configuration item.

https://docs.renovatebot.com/language-constraints-and-upgrading/

> In addition, I think it can skip CHANGELOG.

### DIFF
--- a/renovate.json
+++ b/renovate.json
@@ -7,6 +7,9 @@
   "ignoreDeps": [ "go.opentelemetry.io/contrib/instrgen" ],
   "labels": ["Skip Changelog", "dependencies"],
   "separateMajorMinor": true,
+  "constraints": {
+    "go": "1.22"
+  },
   "postUpdateOptions" : [
     "gomodTidy"
   ],


pellared (commented):
Nice idea. Have you tested if it would also work for dependencies having Go version < 1.22 and 1.22.x in their `go.mod` files?

flc1125:
I have tested that after setting this, the packages that cause the version upgrade of this go version no longer have PRs created. Even if created, they will fail during CI.

pellared:
> I have tested that after setting this, the packages that cause the version upgrade of this go version no longer have PRs created. Even if created, they will fail during CI.

Sure but what about other packages. Are we not going to miss any other updates?

flc1125:
> Sure but what about other packages. Are we not going to miss any other updates?

I can test it. At that time, I will be providing test results.

flc1125:
> Nice idea. Have you tested if it would also work for dependencies having Go version < 1.22 and 1.22.x in their `go.mod` files?

I want to confirm what I want to test:

Test the **other dependent packages** that come from the current package. If go<1.22 or 1.22.x exists, does it support it?

dmathieu:
Nice idea!
Could we document upgrading this in the [release doc](https://github.com/open-telemetry/opentelemetry-go-contrib/blob/main/RELEASING.md), for releases that drop a Go version?

pellared:
> > Nice idea. Have you tested if it would also work for dependencies having Go version < 1.22 and 1.22.x in their `go.mod` files?
> 
> I want to confirm what I want to test:
> 
> Test the **other dependent packages** that come from the current package. If go<1.22 or 1.22.x exists, does it support it?

Examples of such dependencies are:
- 1.22.x : `github.com/grpc-ecosystem/grpc-gateway/v2`
- < 1.22: `github.com/cenkalti/backoff/v4`

You can downgrade their versions and check if renovate would bump both of them.

flc1125:
> Nice idea! Could we document upgrading this in the [release doc](https://github.com/open-telemetry/opentelemetry-go-contrib/blob/main/RELEASING.md), for releases that drop a Go version?

They might not be exactly the same. This configuration tends to rely on the current version relationship of go.mod, it restricts the upgrade of dependent packages in the current repository, and ensures that it must conform to the version configuration of the current go.mod. Otherwise, it will report an error in CI.

Here is an example I am currently using: https://github.com/go-kratos-ecosystem/components/pull/444

*I'm not sure if this meets the effect you want. @pellared*

--- 

Additionally, I found that many of the otel package versions are inconsistent, some are 1.22.0 and some are 1.22.7. Therefore, it caused my dependency upgrade to fail.

Is it allowed to exist like this?

dmathieu:
Yes, but when we drop support for Go 1.22, we need to upgrade this constraint to 1.23. That's what we should document.

flc1125:
> Yes, but when we drop support for Go 1.22, we need to upgrade this constraint to 1.23. That's what we should document.

Yes, that is very correct.

pellared:
> Additionally, I found that many of the otel package versions are inconsistent, some are 1.22.0 and some are 1.22.7. Therefore, it caused my dependency upgrade to fail.

> Is it allowed to exist like this?

Yes. In general, we are supporting the latest versions of the two newest minor versions of Go. Meaning latest the 1.22 and the latest 1.23. Related issue: 
- https://github.com/open-telemetry/opentelemetry-go/issues/5969

flc1125:
Under investigation: https://github.com/renovatebot/renovate/discussions/33222

flc1125:
This PR is about to be forgotten, and I will continue to study it recently.

flc1125:
A long time ago, I tried various methods to adapt the project, but found that it temporarily "does not support" (or perhaps I couldn't find a way) the coexistence of multiple different Go versions in a monorepo. It currently works well under a single, unified Go version.  

However, since we now have multiple different versions, I’ve temporarily disabled it for now. If there are any new developments later, I’ll make adjustments accordingly.

PR #6511, opened 2024-12-18
Fix: error logged on empty OTEL_TRACES_SAMPLER_ARG

matthewhughes934:
We were noticing errors like:

> time=2024-12-18T15:00:02.080Z level=ERROR
source=/go/pkg/mod/go.opentelemetry.io/contrib/samplers/jaegerremote@v0.27.0/sampler_remote_options.go:112 msg="env variable parsing failure" err="argument  is not of type '<key>=<value>'"

In applications using the Jaeger sampler but had not set the `OTEL_TRACES_SAMPLER_ARG` environment variable. The issue was:

* `os.Getenv` would return `""` when this var was not set, and
* `strings.Split("", ",")` would return `[]string{""}`

since `""` does not contain `=` an error is emitted.


samplers/jaegerremote/sampler_remote_options.go
```diff
@@ -46,7 +46,12 @@ func getEnvOptions() ([]Option, []error) {
 	// list of errors which will be logged once logger is set by the user
 	var errs []error
 
-	args := strings.Split(os.Getenv("OTEL_TRACES_SAMPLER_ARG"), ",")
+	rawEnvArgs, ok := os.LookupEnv("OTEL_TRACES_SAMPLER_ARG")
+	if !ok {
+		return nil, nil
+	}
```
MrAlias, line 52: https://github.com/open-telemetry/opentelemetry-specification/blob/e81d06bd5def98b9004e3959ec9d4563c707a818/specification/configuration/sdk-environment-variables.md#parsing-empty-value

```suggestion
	rawEnvArgs := os.Getenv("OTEL_TRACES_SAMPLER_ARG")
	if args == "" {
		return nil, nil
	}
```

MrAlias (commented):
Overall, this looks good. Thank you for adding tests to cover your issue. The last remaining thing is a changelog entry for the fix your adding needs to be included.

matthewhughes934:
> Overall, this looks good. Thank you for adding tests to cover your issue. The last remaining thing is a changelog entry for the fix your adding needs to be included.

Thanks! Changelog entry added with [4d72fb9](https://github.com/open-telemetry/opentelemetry-go-contrib/pull/6511/commits/4d72fb9a07ea523044b98515ad54e3f1946061ff)

dmathieu (approved):

Issue #6509, opened 2024-12-18
otellogr emits nil context causing panic in exporters


gregoryfranklin:
### Description
By default, otellogr emits logs with a nil context - this causes exporters that do not support nil contexts to panic.

This happens because it uses a context that is never initialised
```
func NewLogSink(name string, options ...Option) *LogSink {
	return &LogSink{
		// does not initialise ctx
	}
}

type LogSink struct {
	ctx           context.Context
}

func (l *LogSink) Info(level int, msg string, keysAndValues ...any) {
	ctx, attr := convertKVs(l.ctx, keysAndValues...)

	l.logger.Emit(ctx, record)  // emits nil context
}
```

### Environment
- `go.opentelemetry.io/contrib/bridges/otellogr v0.8.0`

### Steps To Reproduce

Run the following code:
```
package main

import (
	"github.com/go-logr/logr"
	"go.opentelemetry.io/contrib/bridges/otellogr"
	"go.opentelemetry.io/otel/exporters/stdout/stdoutlog"
	"go.opentelemetry.io/otel/sdk/log"
)

func main() {
	exporter, _ := stdoutlog.New()
	processor := log.NewSimpleProcessor(exporter)
	provider := log.NewLoggerProvider(log.WithProcessor(processor))
	sink := otellogr.NewLogSink("example", otellogr.WithLoggerProvider(provider))
	logger := logr.New(sink)
	logger.Info("hello")
}
```
Results in
```
panic: runtime error: invalid memory address or nil pointer dereference
[signal SIGSEGV: segmentation violation code=0x1 addr=0x28 pc=0x570653]

goroutine 1 [running]:
go.opentelemetry.io/otel/exporters/stdout/stdoutlog.(*Exporter).Export(0xc000016250, {0x0, 0x0}, {0xc0001284e0?, 0x0?, 0x0?})
	/tmp/.cache/go/pkg/mod/go.opentelemetry.io/otel/exporters/stdout/stdoutlog@v0.9.0/exporter.go:49 +0xf3
go.opentelemetry.io/otel/sdk/log.(*SimpleProcessor).OnEmit(0xc000070180, {0x0, 0x0}, 0xc000128340)
	/tmp/.cache/go/pkg/mod/go.opentelemetry.io/otel/sdk/log@v0.9.0/simple.go:58 +0x1ad
[...]

PR #6527, opened 2024-12-23
otellogr: Fix nil context panic

flc1125:
Fixes https://github.com/open-telemetry/opentelemetry-go-contrib/issues/6509


CHANGELOG.md
```diff
@@ -12,6 +12,10 @@ This project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0.htm
 
 - Generate server metrics with semantic conventions v1.26 in `go.opentelemetry.io/contrib/instrumentation/net/http/otelhttp` when `OTEL_SEMCONV_STABILITY_OPT_IN` is set to `http/dup`. (#6411)
 
+### Fixed
+
+- Fixed an issue where otellogr emits nil context which leads to panic in exporters `go.opentelemetry.io/contrib/bridges/otellogr`. (#6527)
```
dmathieu, line 17:
```suggestion
- Use `context.Background()` as default context instead of nil in `go.opentelemetry.io/contrib/bridges/otellogr`. (#6527)
```

bridges/otellogr/logsink.go
```diff
@@ -139,6 +143,18 @@ func WithLoggerProvider(provider log.LoggerProvider) Option {
 	})
 }
 
+// WithContext returns an [Option] that configures the [context.Context] used by
+// a [LogSink].
+//
+// By default if this Option is not provided, the LogSink will use the
+// background context.
+func WithContext(ctx context.Context) Option {
```
dmathieu, line 151: This is a different change. Please keep this PR constrained to fixing the nil panic, without the additional way to set the context.

flc1125: No problem~ I'll adjust it later.

bridges/otellogr/logsink.go
```diff
@@ -82,6 +82,10 @@ func newConfig(options []Option) config {
 		c.provider = global.GetLoggerProvider()
 	}
 
+	if c.ctx == nil {
```
dmathieu, line 85: This will always be nil since there's no way to pass the config. So we can remove the config attribute and pass `context.Background()` directly here.

bridges/otellogr/logsink.go
```diff
@@ -238,9 +242,20 @@ func (l *LogSink) Info(level int, msg string, keysAndValues ...any) {
 	l.logger.Emit(ctx, record)
 }
 
+// WithContext returns a new LogSink with the specified context.
+func (l *LogSink) WithContext(ctx context.Context) *LogSink {
```
dmathieu, line 246: Same as my previous comment, this adds a new feature, it's not needed to fix the panic issue. We should open it in another PR.

bridges/otellogr/logsink.go
```diff
@@ -201,9 +206,8 @@ var _ logr.LogSink = (*LogSink)(nil)
 // For example, commandline flags might be used to set the logging
 // verbosity and disable some info logs.
 func (l *LogSink) Enabled(level int) bool {
-	ctx := context.Background()
 	param := log.EnabledParameters{Severity: l.levelSeverity(level)}
-	return l.logger.Enabled(ctx, param)
+	return l.logger.Enabled(l.ctx, param)
```
dmathieu, line 210: This change should also be in its own PR, as it's not needed for the panic fix, and it adds a new feature (`Enabled` can now look into context values).

bridges/otellogr/logsink.go
```diff
@@ -65,10 +65,9 @@ import (
 )
 
 type config struct {
-	provider  log.LoggerProvider
-	version   string
-	schemaURL string
-
```
dmathieu, line 71: Why remove the new line? It makes sense.

bridges/otellogr/logsink.go
```diff
@@ -201,9 +201,8 @@ var _ logr.LogSink = (*LogSink)(nil)
 // For example, commandline flags might be used to set the logging
 // verbosity and disable some info logs.
 func (l *LogSink) Enabled(level int) bool {
-	ctx := context.Background()
 	param := log.EnabledParameters{Severity: l.levelSeverity(level)}
-	return l.logger.Enabled(ctx, param)
+	return l.logger.Enabled(context.Background(), param)
```
dmathieu, line 205: While not mandatory, the variable made sense too here, especially since it's rather a rare behavior to create empty contextes in here.

bridges/otellogr/logsink_test.go
```diff
@@ -30,44 +30,37 @@ func TestNewConfig(t *testing.T) {
 	customLoggerProvider := mockLoggerProvider{}
 
 	for _, tt := range []struct {
-		name    string
-		options []Option
-
-		wantConfig config
+		name     string
+		options  []Option
+		wantFunc func(config)
```
dmathieu, line 35: This change is making the test rather harder to grasp. Let's not make it until we do need it because of context.

flc1125: Because when writing test cases, it is necessary to add some default attributes at any time, otherwise it may lead to some exceptions.

However, the focus of unit testing is not on these default attributes. Therefore, when adjusting the context (ctx) at that time, this part of the method was rewritten.

dmathieu: Right, but we're not adjusting the context anymore. So there's no reason for this to be in the diff.

pellared:
> I'm not sure a consensus has been reached in #6509.

I think this is the proper fix and we probably do not need anything more than this.

flc1125:
Ready for approval.

dmathieu (approved):

bridges/otellogr/logsink_test.go
```diff
@@ -352,6 +352,40 @@ func TestLogSink(t *testing.T) {
 	}
 }
 
+// fix https://github.com/open-telemetry/opentelemetry-go-contrib/issues/6509
+func TestLogSinkCtxInInfo(t *testing.T) {
```
pellared, line 356: Couldn't we add test cases to `TestLogSink` instead?

scorpionknifes (approved):
Thanks, LGTM (not a go approver)

bridges/otellogr/logsink_test.go
flc1125: Upon checking, the structure of `TestLogSink` is currently not suitable for my testing scenario. If adjustments are needed, more changes might be required.

pellared: OK. Approving. I can look and try refactoring it tomorrow.

pellared (approved):

Issue #6586, opened 2025-01-09
config: Add fuzz tests


pellared:
Add [fuzz tests](https://go.dev/doc/security/fuzz) for loading the config from JSON and YAML files.
Update the GitHub workflow to run the fuzz tests for 3 minutes. Example: https://github.com/jidicula/go-fuzz-action/blob/main/action.yml

We can also separately look into using OSS-Fuzz. References:
- https://opentelemetry.io/blog/2024/fuzzing-audit-results/
- https://google.github.io/oss-fuzz/getting-started/new-project-guide/go-lang/

EDIT: https://github.com/open-telemetry/opentelemetry-go-contrib/issues/6605 created

PR #6604, opened 2025-01-13
config: Add fuzz tests

pellared:
Fixes https://github.com/open-telemetry/opentelemetry-go-contrib/issues/6586

A fix for a bug discovered thanks to these tests: 
- https://github.com/open-telemetry/opentelemetry-go-contrib/pull/6606

Issue for integration with OSS-Fuzz: https://github.com/open-telemetry/opentelemetry-go-contrib/issues/6605

Each fuzz test passed when running for more than 10 minutes. E.g.

```
11m0s, execs: 5055668 (10389/sec), new interesting: 840 (total: 859)
PASS
ok      go.opentelemetry.io/contrib/config/v0.3.0       660.154s
```


dashpole (approved):
cool!

Makefile
```diff
@@ -276,6 +277,18 @@ test-coverage/%:
...
+# Fuzzing
+
+.PHONY: fuzz-config-json
+fuzz-config-json:
+	cd config/v0.3.0 \
+	&& $(GO) test -run=^$ -fuzztime=$(FUZZTIME) -fuzz=FuzzJSON
+
+.PHONY: fuzz-config-yaml
+fuzz-config-yaml:
+	cd config/v0.3.0 \
```
MrAlias, line 289: How will this scale to multiple versions? Should this be finding related files and then iterating over them instead?

pellared: I was thinking about it as well but decided to keep it simple before we have scaling issues :wink:

codeboten (approved):
Thanks @pellared!

Makefile
pellared: Do we need it for only the latest version of config or for all possible fuzz tests?

pellared: Removed https://github.com/open-telemetry/opentelemetry-go-contrib/pull/6604/commits/4cdabcfd3dac9e2f2aaaf7c8d638d153f22fc37b

1. There is something wrong in the Makefile https://github.com/open-telemetry/opentelemetry-go-contrib/pull/6604/commits/4cdabcfd3dac9e2f2aaaf7c8d638d153f22fc37b (running too long)
2. Fuzzing should be executed for a long time so I do not think it make sense to have it in GitHub Actions anyway - OSS-Fuzz is the way

MrAlias (approved):

PR #6609, opened 2025-01-13
fix(otelgin): remove multipartform temporary file

### LINKED ISSUE #5946: Using otelgin.Middleware will cause temporary MultipartForm files not to be destroyed.
### Description

 Using otelgin.Middleware will cause temporary MultipartForm files not to be destroyed.

### Environment

- OS: MacOS
- Architecture: x86
- Go Version: 1.22
- `otelgin` version: 0.53.0

### Steps To Reproduce

1. Using this code start up gin http server
```
package main

import (
	"net/http"

	"github.com/gin-gonic/gin"
	"go.opentelemetry.io/contrib/instrumentation/github.com/gin-gonic/gin/otelgin"
)

func main() {
	engine := gin.Default()
	engine.MaxMultipartMemory = 1024 * 1024
	engine.Use(otelgin.Middleware("my-server"))

	engine.POST("/upload", func(c *gin.Context) {
		ff, err := c.FormFile("file")
		if err != nil {
			_ = c.AbortWithError(http.StatusInternalServerError, err)
			return
		}
		_ = ff
		c.JSON(http.StatusOK, map[string]interface{}{"c": 0})
	})

	engine.Run(":8624")
}
```
2. Upload a file to the http api
```
curl --request POST \
  --url http://127.0.0.1:8624/upload \
  --header 'Content-Type: multipart/form-data' \
  --header 'User-Agent: insomnia/9.3.2' \
  --form 'file=@xxx.mp4'
```

### Expected behavior

Upload a file larger than MaxMultipartMemory, and the HTTP server will generate multipart-xxxx temporary files in the system temporary directory. These files will be automatically destroyed when the request ends. However, if the otelgin.Middleware is used, the temporary files will not be destroyed.  

the HTTP server holds the original request, and using otelgin.Middleware replaces the request with [c.request = c.request.WithContext(ctx)](https://github.com/open-telemetry/opentelemetry-go-contrib/blob/cb45b36c145ff13e4ba76fd9e02ef39297c906b5/instrumentation/github.com/gin-gonic/gin/otelgin/gintrace.go#L79), causing subsequent c.FormFile reads to fall on the new request. As a result, the original request held by the HTTP server does not have the MultipartForm, preventing it from being destroyed.

Yiling-J:
Fix #5946

### DIFF
--- a/instrumentation/github.com/gin-gonic/gin/otelgin/gintrace.go
+++ b/instrumentation/github.com/gin-gonic/gin/otelgin/gintrace.go
@@ -85,6 +85,17 @@ func Middleware(service string, opts ...Option) gin.HandlerFunc {
 
 		// pass the span through the request context
 		c.Request = c.Request.WithContext(ctx)
+		defer func() {
+			// as we have created new http.Request object we need to make sure that temporary files created to hold MultipartForm
+			// files are cleaned up. This is done by http.Server at the end of request lifecycle but Server does not
+			// have reference to our new Request instance therefore it is our responsibility to fix the mess we caused.
+			//
+			// This means that when we are on returning path from handler middlewares up in chain from this middleware
+			// can not access these temporary files anymore because we deleted them here.
+			if c.Request.MultipartForm != nil {
+				c.Request.MultipartForm.RemoveAll()
+			}
+		}()
 
 		// serve the request to the next middleware
 		c.Next()
--- a/instrumentation/github.com/gin-gonic/gin/otelgin/test/gintrace_test.go
+++ b/instrumentation/github.com/gin-gonic/gin/otelgin/test/gintrace_test.go
@@ -6,8 +6,11 @@
 package test
 
 import (
+	"bytes"
 	"errors"
 	"html/template"
+	"io/fs"
+	"mime/multipart"
 	"net/http"
 	"net/http/httptest"
 	"strconv"
@@ -429,4 +432,45 @@ func TestWithGinFilter(t *testing.T) {
 		router.ServeHTTP(w, r)
 		assert.Len(t, sr.Ended(), 1)
 	})
+}
+
+func TestTemporaryFormFileRemove(t *testing.T) {
+	sr := tracetest.NewSpanRecorder()
+	provider := sdktrace.NewTracerProvider(sdktrace.WithSpanProcessor(sr))
+
+	router := gin.New()
+	router.MaxMultipartMemory = 1
+	router.Use(otelgin.Middleware("foobar", otelgin.WithTracerProvider(provider)))
+	var fileHeader *multipart.FileHeader
+	router.POST("/upload", func(c *gin.Context) {
+		ff, err := c.FormFile("file")
+		if err != nil {
+			_ = c.AbortWithError(http.StatusInternalServerError, err)
+			return
+		}
+		_ = ff
+		fileHeader = c.Request.MultipartForm.File["file"][0]
+		c.JSON(http.StatusOK, nil)
+	})
+
+	var body bytes.Buffer
+
+	mw := multipart.NewWriter(&body)
+	fw, err := mw.CreateFormFile("file", "file")
+	require.NoError(t, err)
+
+	_, err = fw.Write([]byte("hello world"))
+	require.NoError(t, err)
+	err = mw.Close()
+	require.NoError(t, err)
+	r := httptest.NewRequest("POST", "/upload", &body)
+	r.Header.Add("Content-Type", mw.FormDataContentType())
+	w := httptest.NewRecorder()
+
+	router.ServeHTTP(w, r)
+	assert.Len(t, sr.Ended(), 1)
+	require.Equal(t, http.StatusOK, w.Code)
+	_, err = fileHeader.Open()
+	require.ErrorIs(t, err, fs.ErrNotExist)
+
 }


instrumentation/github.com/gin-gonic/gin/otelgin/gintrace.go
```diff
@@ -85,6 +85,17 @@ func Middleware(service string, opts ...Option) gin.HandlerFunc {
...
 		// pass the span through the request context
 		c.Request = c.Request.WithContext(ctx)
+		defer func() {
+			// as we have created new http.Request object we need to make sure that temporary files created to hold MultipartForm
+			// files are cleaned up. This is done by http.Server at the end of request lifecycle but Server does not
+			// have reference to our new Request instance therefore it is our responsibility to fix the mess we caused.
+			//
+			// This means that when we are on returning path from handler middlewares up in chain from this middleware
+			// can not access these temporary files anymore because we deleted them here.
+			if c.Request.MultipartForm != nil {
```
flc1125, line 95: I'm wondering, if there's another middleware layer outside this one, would it be removed because of this?

instrumentation/github.com/gin-gonic/gin/otelgin/test/gintrace_test.go
```diff
@@ -430,3 +433,44 @@ func TestWithGinFilter(t *testing.T) {
...
+	router.MaxMultipartMemory = 1
+	router.Use(otelgin.Middleware("foobar", otelgin.WithTracerProvider(provider)))
+	var fileHeader *multipart.FileHeader
+	router.POST("/upload", func(c *gin.Context) {
+		ff, err := c.FormFile("file")
+		if err != nil {
+			_ = c.AbortWithError(http.StatusInternalServerError, err)
+			return
+		}
+		_ = ff
```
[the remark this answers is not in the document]
Yiling-J: The goal of the test is to ensure that temporary files are automatically removed, not to test multipart file parsing. The sole purpose of using `c.FormFile` is to trigger multipart parsing, which creates temporary files. I added a `fileHeader.Open()` test in this function to confirm that the temporary file is available. However, I think checking the value of `ff` is unnecessary.

instrumentation/github.com/gin-gonic/gin/otelgin/test/gintrace_test.go
```diff
@@ -430,3 +433,44 @@ func TestWithGinFilter(t *testing.T) {
...
+	sr := tracetest.NewSpanRecorder()
+	provider := sdktrace.NewTracerProvider(sdktrace.WithSpanProcessor(sr))
+
+	router := gin.New()
+	router.MaxMultipartMemory = 1
+	router.Use(otelgin.Middleware("foobar", otelgin.WithTracerProvider(provider)))
+	var fileHeader *multipart.FileHeader
+	router.POST("/upload", func(c *gin.Context) {
+		ff, err := c.FormFile("file")
+		if err != nil {
```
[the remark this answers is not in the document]
Yiling-J: I prefer using `require.NoError` because it ensures that no additional information distracts me when a test has already failed. I also searched the codebase and seems that both `require` and `assert` are used, and `require.NoError` appears more.

instrumentation/github.com/gin-gonic/gin/otelgin/gintrace.go
Yiling-J: It's hard to say, but in all cases, regardless of how many middlewares are chained, temporary files should be removed. I didn't check the return error of `RemoveAll` here, which aligns with the behavior in both Go's [`http` package](https://github.com/golang/go/blob/44a6f817ea0fbeb3ba4aa398794c4e80dba13b1e/src/net/http/server.go#L1727) and [echo-contrib](https://github.com/labstack/echo-contrib/pull/90). I believe each middleware should independently attempt to remove temporary files.

flc1125:
I have written 2 examples to facilitate the analysis of this issue.

I believe that gin falls into the first category, so I think the main logic of the modifications in this PR should be fine.

---

- The first example:

```go
package gin

import (
	"context"
	"net/http"
	"net/url"
	"testing"

)

type handler func(t *testing.T, req *http.Request)

var test1Handler handler = func(t *testing.T, req *http.Request) {
	t.Logf("req1.0: %p, req.URL: %p", req, req.URL)
}

var test2Handler handler = func(t *testing.T, req *http.Request) {
	t.Logf("req2.0: %p, req.URL: %p", req, req.URL)

	saveCtx := req.Context()
	defer func() {
		t.Logf("req2.3: %p, req.URL: %p", req, req.URL)
		req = req.WithContext(saveCtx)
		t.Logf("req2.4: %p, req.URL: %p", req, req.URL)
	}()

	defer func() {
		t.Logf("req2.1: %p, req.URL: %p", req, req.URL)
		req.URL = &url.URL{
			Scheme: "https",
		}
		t.Logf("req2.2: %p, req.URL: %p", req, req.URL)
	}()

	req = req.WithContext(context.WithValue(req.Context(), "key", "value"))
}

var test3Handler handler = func(t *testing.T, req *http.Request) {
	t.Logf("req3.0: %p, req.URL: %p", req, req.URL)
	t.Logf("req3.1: %v", req.Context().Value("key"))
}

func TestIssue_Middleware(t *testing.T) {
	req, err := http.NewRequest("POST", "/upload", nil)
	assert.NoError(t, err)
	req.URL = &url.URL{
		Scheme: "http",
	}

	t.Logf("req0.0: %p, req.URL: %p", req, req.URL)

	handlers := []handler{
		test1Handler,
		test2Handler,
		test3Handler,
	}
	for _, h := range handlers {
		h(t, req)
	}
}
```

output:

```
=== RUN   TestIssue_Middleware
    issue_test.go:51: req0.0: 0x1400049ba40, req.URL: 0x1400016ca20
    issue_test.go:15: req1.0: 0x1400049ba40, req.URL: 0x1400016ca20
    issue_test.go:19: req2.0: 0x1400049ba40, req.URL: 0x1400016ca20
    issue_test.go:29: req2.1: 0x1400049bb80, req.URL: 0x1400016ca20
    issue_test.go:33: req2.2: 0x1400049bb80, req.URL: 0x1400016cab0
    issue_test.go:23: req2.3: 0x1400049bb80, req.URL: 0x1400016cab0
    issue_test.go:25: req2.4: 0x1400049bcc0, req.URL: 0x1400016cab0
    issue_test.go:40: req3.0: 0x1400049ba40, req.URL: 0x1400016ca20
    issue_test.go:41: req3.1: <nil>
--- PASS: TestIssue_Middleware (0.00s)
```

- The secend example:

```go
package gin

import (
	"context"
	"net/http"
	"net/url"
	"testing"

)

type handler func(t *testing.T, req *http.Request) *http.Request

var test1Handler handler = func(t *testing.T, req *http.Request) *http.Request {
	t.Logf("req1.0: %p, req.URL: %p", req, req.URL)
	return req
}

var test2Handler handler = func(t *testing.T, req *http.Request) *http.Request {
	t.Logf("req2.0: %p, req.URL: %p", req, req.URL)

	saveCtx := req.Context()
	defer func() {
		t.Logf("req2.3: %p, req.URL: %p", req, req.URL)
		req = req.WithContext(saveCtx)
		t.Logf("req2.4: %p, req.URL: %p", req, req.URL)
	}()

	defer func() {
		t.Logf("req2.1: %p, req.URL: %p", req, req.URL)
		req.URL = &url.URL{
			Scheme: "https",
		}
		t.Logf("req2.2: %p, req.URL: %p", req, req.URL)
	}()

	req = req.WithContext(context.WithValue(req.Context(), "key", "value"))

	return req
}

var test3Handler handler = func(t *testing.T, req *http.Request) *http.Request {
	t.Logf("req3.0: %p, req.URL: %p", req, req.URL)
	t.Logf("req3.1: %v", req.Context().Value("key"))
	return req
}

func TestIssue_Middleware(t *testing.T) {
	req, err := http.NewRequest("POST", "/upload", nil)
	assert.NoError(t, err)
	req.URL = &url.URL{
		Scheme: "http",
	}

	t.Logf("req0.0: %p, req.URL: %p", req, req.URL)

	handlers := []handler{
		test1Handler,
		test2Handler,
		test3Handler,
	}
	for _, h := range handlers {
		req = h(t, req)
	}
}
```

output:

```
=== RUN   TestIssue_Middleware
    issue_test.go:55: req0.0: 0x140001b6500, req.URL: 0x14000130360
    issue_test.go:15: req1.0: 0x140001b6500, req.URL: 0x14000130360
    issue_test.go:20: req2.0: 0x140001b6500, req.URL: 0x14000130360
    issue_test.go:30: req2.1: 0x140001b6640, req.URL: 0x14000130360
    issue_test.go:34: req2.2: 0x140001b6640, req.URL: 0x140001303f0
    issue_test.go:24: req2.3: 0x140001b6640, req.URL: 0x140001303f0
    issue_test.go:26: req2.4: 0x140001b6780, req.URL: 0x140001303f0
    issue_test.go:43: req3.0: 0x140001b6640, req.URL: 0x140001303f0
    issue_test.go:44: req3.1: value
--- PASS: TestIssue_Middleware (0.00s)
PASS
```

instrumentation/github.com/gin-gonic/gin/otelgin/gintrace.go
flc1125: see: https://github.com/open-telemetry/opentelemetry-go-contrib/pull/6609#issuecomment-2587572600

instrumentation/github.com/gin-gonic/gin/otelgin/gintrace.go
```diff
@@ -85,6 +85,17 @@ func Middleware(service string, opts ...Option) gin.HandlerFunc {
...
+				_ = c.Request.MultipartForm.RemoveAll()
```
dmathieu, line 96: Rather than ignoring an error here, how about passing it to the otel error handler?

```suggestion
				err := c.Request.MultipartForm.RemoveAll()
				if err != nil {
					otel.Handle(err)
				}
```

instrumentation/github.com/gin-gonic/gin/otelgin/test/gintrace_test.go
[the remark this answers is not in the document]
dmathieu: I think what @flc1125 meant here is that you don't necessarily need to use `AbortWithError` (which also can return an error but ignores it), but instead call `require.NoError` here.

```suggestion
		require.NoError(err)
```

instrumentation/github.com/gin-gonic/gin/otelgin/gintrace.go
```diff
@@ -85,6 +85,17 @@ func Middleware(service string, opts ...Option) gin.HandlerFunc {
 
 		// pass the span through the request context
 		c.Request = c.Request.WithContext(ctx)
```
dmathieu, line 87: Wouldn't we be fixing this rather more easily with the following:

```suggestion
		c.Request = c.Request.Clone(ctx)
```

[WithContext](https://pkg.go.dev/net/http#Request.WithContext) performs a shallow clone, and therefore doesn't clone the multipart form.
However, [Clone](https://pkg.go.dev/net/http#Request.Clone) performs a deep copy, and makes an explicit clone of [MultipartForm](https://cs.opensource.google/go/go/+/refs/tags/go1.23.4:src/net/http/request.go;l=406).

Yiling-J:
I skipped the failed test on the Windows platform. This is consistent with how Go handles form cleanup tests (https://github.com/golang/go/blob/6da16013ba4444e0d71540f68279f0283a92d05d/src/net/http/serve_test.go#L7028), so I think it’s fine.

instrumentation/github.com/gin-gonic/gin/otelgin/gintrace.go
Yiling-J: I think this approach could work, but it feels a bit unusual to enforce form parsing in the middleware.

dmathieu: Having to force remove multipart files is a bit unusual too.

akats7: I'm probably more in favor of doing the initialization on the original request as well, explicitly handling clean up logic does seem a bit strange

akats7: @dmathieu maybe we should discuss this in the SIG since it might have implications in the other instrumentation modules

Yiling-J: @akats7 Another option is to save `c.Request` to a variable before calling `c.Request.WithContext(ctx)`. Then, in the defer function, if `c.Request.MultipartForm != nil`, we can reattach the `MultipartForm` to the original request. This way, there won’t be any automatic removing or parsing. If other middlewares handle the form correctly, this approach should work. 
In [http/server.go](https://github.com/golang/go/blob/40b3c0e58a0ae8dec4684a009bf3806769e0fc41/src/net/http/server.go#L2102), the `ServeHTTP` method is called first, which invokes the Gin handlers/middlewares, and then `finishRequest` is called, which removes the temporary files.

I haven’t tested yet, but I think it should work. However, I’m considering opening an issue in Go to get their suggestions. Since it’s closely related to the `http` package and the `WithContext` method is widely used by middlewares and frameworks, I think they might be able to suggest best practice for this situation.

Yiling-J: @akats7 It seems I can’t get any help from the Go team. I’m not in favor of parsing the form in the middleware, as this step should be explicitly done by the user. Parsing large forms can consume resources, and the `ParseMultipartForm` function has a `maxMemory` parameter that should also be decided by the user.

After giving this more thought over the past few days, I believe this is more of a Go design issue. The `WithContext` and `ParseMultipartForm` behavior is implicit, hard to notice, poorly documented, and lacks a clear best practice solution. I think the reattach approach might be better because it’s less aggressive and can likely be safely applied to other OpenTelemetry middlewares facing the same issue. The only downside, compared to `RemoveAll`, is that it relies on other middlewares to also handle `WithContext` and `ParseMultipartForm` correctly. If all middlewares do the right thing, `net/http` should eventually clean up the temporary files automatically.

flc1125:
@Yiling-J Hi, are you still willing to handle this issue?

According to the SIG discussion, it is recommended that middleware remove temporary resources.

cc @akats7

Yiling-J:
> are you still willing to handle this issue

@flc1125 Yes. But first, let’s reach an agreement on how we should fix it. As I proposed a while ago, I prefer the reattach approach: “copy the cloned `Request.MultipartForm` value back to the original request before returning,” as mentioned in https://github.com/golang/go/issues/74455#issuecomment-3099652232.

We should also decide whether to fix this in otelgin only, or to review other middlewares and address the issue there as well if it exists.

flc1125:
> let’s reach an agreement on how we should fix it

We tend to only need to add the following on the basis of the original main branch:

```go
c.Request.MultipartForm.RemoveAll()
```

> We should also decide whether to fix this in otelgin only, or to review other middlewares and address the issue there as well if it exists.

In fact, all middleware based on `*http.Request` has issues whenever a situation like `c.Request.WithContext(c.Request.Context())` occurs, so other components also need to be fixed.

However, for this current PR, we only need to fix otelgin to keep the PR as minimal as possible. For other middleware, if you are interested, you can also submit fixes in the form of minimal PRs for individual components.

Yiling-J:
@flc1125 `RemoveAll()` works for me. I think my original PR already aligns well with the goal, it looks like I just need to resolve the conflicts and make sure the tests pass. Let me confirm agin that we’re all in agreement on this approach. @dmathieu @akats7

akats7:
Sorry just got caught up on this thread. My only concern with RemoveAll is on the off chance that a middleware further up the chain still expect access to the form data. 

What about making it configurable with an Option, with the default behavior being to clean it up. That way the user can have more control over which middleware cleans up the form data.

I don't want to stall this issue further but just a thought.

cc. @flc1125 @dmathieu

flc1125:
> Sorry just got caught up on this thread. My only concern with RemoveAll is on the off chance that a middleware further up the chain still expect access to the form data.

Agree.

> What about making it configurable with an Option, with the default behavior being to clean it up. That way the user can have more control over which middleware cleans up the form data.

I would lean towards the solution proposed by the SIG: only support configurability when users explicitly request it. After all, enabling configurability after `RemoveAll()` is an operation that ensures both extensibility compatibility and backward compatibility.

@akats7 @dmathieu

akats7:
Ah gotcha, didn't see that that was mentioned. I'm good on holding off until it's explicitly requested.

akats7:
@Yiling-J I think we're good to resolve the conflicts and close this out, thanks!

Yiling-J:
I’ve resolved the conflicts, and the otelgin package tests passed on my machine @flc1125 @dmathieu @akats7

CHANGELOG.md
```diff
@@ -24,6 +24,10 @@ This project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0.htm
 
 - Improve performance by reducing allocations in the gRPC stats handler in `go.opentelemetry.io/contrib/instrumentation/google.golang.org/grpc/otelgrpc`. (#8035)
 
+### Fixed
+
+- Fix issue in `otelgin` middleware where temporary files for multipart/form-data were not being removed. (#6609)
```
```suggestion
- Fix issue in `go.opentelemetry.io/contrib/instrumentation/github.com/gin-gonic/gin/otelgin` middleware where temporary files for multipart/form-data were not being removed. (#6609)
```

flc1125:
Could you add the corresponding unit tests for this scenario?

> My only concern with RemoveAll is on the off chance that a middleware further up the chain still expect access to the form data.

instrumentation/github.com/gin-gonic/gin/otelgin/gin.go
```diff
@@ -85,6 +85,17 @@ func Middleware(service string, opts ...Option) gin.HandlerFunc {
 
 		// pass the span through the request context
 		c.Request = c.Request.WithContext(ctx)
```
flc1125: Close to reduce distractions.

instrumentation/github.com/gin-gonic/gin/otelgin/gin.go
```diff
@@ -85,6 +85,17 @@ func Middleware(service string, opts ...Option) gin.HandlerFunc {
...
+		defer func() {
+			// as we have created new http.Request object we need to make sure that temporary files created to hold MultipartForm
+			// files are cleaned up. This is done by http.Server at the end of request lifecycle but Server does not
+			// have reference to our new Request instance therefore it is our responsibility to fix the mess we caused.
+			//
+			// This means that when we are on returning path from handler middlewares up in chain from this middleware
+			// can not access these temporary files anymore because we deleted them here.
+			if c.Request.MultipartForm != nil {
```
flc1125: Close to reduce distractions.

Yiling-J:
> Could you add the corresponding unit tests for this scenario?
> 
> > My only concern with RemoveAll is on the off chance that a middleware further up the chain still expect access to the form data.

I’ve added more details to the test case, it now includes two extra middlewares, one upstream and one downstream, to demonstrate what happens when `RemoveAll` is called in the otel middleware. The behavior is correct to me, but let’s confirm that we all agree it’s the expected one.

instrumentation/github.com/gin-gonic/gin/otelgin/gin_test.go
```diff
@@ -479,6 +483,79 @@ func TestWithGinFilter(t *testing.T) {
...
+	// "file does not exist" error.
+
+	// MiddlewareA
+	router.Use(func(c *gin.Context) {
+		c.Next()
+		form, _ := c.MultipartForm()
+		files := form.File["files"]
+		require.Len(t, files, 1)
+		_, err := files[0].Open()
+		require.ErrorIs(t, err, fs.ErrNotExist)
```
flc1125, line 512: I don’t think this is the expected result; in any case—whether it’s middleware or something else—it should be able to retrieve the file again.

I think we may need to investigate this issue further.

flc1125: Based on this PR, I’ve provided a prototype reference: https://github.com/open-telemetry/opentelemetry-go-contrib/pull/8161

@dmathieu @akats7, take a look and see how it works if we adjust it according to the prototype.

akats7: @flc1125 can you clarify which behavior you’re referring to as incorrect, are you referring to 

```
require.ErrorIs(t, err, fs.ErrNotExist)
```

Yiling-J: I thought we had already agreed on the `RemoveAll` approach? Naturally, if you call RemoveAll, the temporary files are deleted and can’t be accessed afterward. Since this happens during the middleware’s return phase, I’d expect that most middlewares wouldn’t read the multipart form at that point anyway. I’m not sure if the implications of RemoveAll were discussed in the SIG meeting, but it might be worth confirming.

akats7: @flc1125 I’m still not clear on the result you’re referring to, the result aligns to my understanding of what we’ve discussed.

Also I’m not sure if your prototype fix would work, I think it still doesn’t address the core issue which is that when we originally set c.Request = c.Request.WithContext(ctx) with an unparsed form the http server won’t know to clean it up. I don’t think Cloning the request solves that.

Issue #6629, opened 2025-01-16
config: Handle null attributes as invalid


pellared:
`null` attribute should be considered as an invalid input for `ParseYAML` (and when parsing JSON as well) as we already have test cases like `invalid nil value`.

 _Originally posted by @pellared in [#6603](https://github.com/open-telemetry/opentelemetry-go-contrib/issues/6603#issuecomment-2586755582)_

CC @codeboten

mikeblum:
Can I pick this up? @pellared

mikeblum:
After digging pretty deeply into the YAML spec its unclear to me if we want to be writing custom parsers to error on different cases such as:

`attr: null`
`attr: []`
`attr: 0`
`attr:`
```
map:
  -
```
imo the current behaviour of `parseYAML` ignoring these attributes entirely is correct but I defer to @pellared on what the correct behavious should be.

pellared:
We should consistently "ignore" invalid values or return an error. I think that if something is not compliant with the scheme then it should return an error.

MrAlias:
> > its unclear to me if we want to be writing custom parsers to error on different cases such as:
> 
> This is my expectation.

Wouldn't this mean we would completely rewrite a yaml parsing library? That does not seem like something we want to maintain.

PR #6645, opened 2025-01-18
[GH-6629] config: Handle null attributes as invalid

mikeblum:


mikeblum:
as written the `omitempty` is allowing the YAML to parse without an error. Added a validate_config` test to factor out the `yaml:omitempty` issue.

https://github.com/open-telemetry/opentelemetry-go-contrib/blob/main/config/v0.3.0/generated_config.go#L427

config/testdata/invalid_empty.yaml
```diff
@@ -0,0 +1,9 @@
+# The file format version.
+file_format: "0.3"
+
+# Configure if the SDK is disabled or not. This is not required to be provided to ensure the SDK isn't disabled, the default value when this is not provided is for the SDK to be enabled.
+disabled: false
+
+resource:
+  attributes:
+    - name:
```
mikeblum, line 9:
```
attributes:
  -
```

is never reaching `(j *AttributeNameValue) UnmarshalYAML` but this diff trips the missing name value.

config/v0.3.0/config_yaml.go
```diff
@@ -31,6 +32,58 @@ func (j *AttributeNameValueType) UnmarshalYAML(unmarshal func(interface{}) error
...
+	}
+	if value, ok = raw["value"]; raw != nil && !ok {
+		return errors.New("field value in AttributeNameValue: required")
+	}
+	type Plain AttributeNameValue
+	plain := Plain{
+		Name:  name,
+		Value: value,
+	}
+	if plain.Type != nil && plain.Type.Value == "int" {
```
mikeblum, line 61: this is breaking the existing tests:
```
          	            	     Name: (string) (len=16) "double_array_key",
            	            	-    Type: (*config.AttributeNameValueType)({
            	            	-     Value: (string) (len=12) "double_array"
            	            	-    }),
            	            	+    Type: (*config.AttributeNameValueType)(<nil>),
            	            	     Value: ([]interface {}) (len=2) {
            	Test:       	TestParseYAML/valid_v0.3_config
```

how should the type get resolved here?

config/testdata/invalid_empty.yaml
mikeblum: thinking out loud I would expect something like this to get hit to mimic [`UnmarshalJSON`](https://github.com/codeboten/opentelemetry-go-contrib/blob/4f84afcdfa796c27db8eefb8529fe737010dd1fc/config/v0.3.0/config_json.go#L31)

```
func (j *AttributeNameValue) UnmarshalYAML(b []byte) error {
	var raw map[string]interface{}
	if err := yaml.Unmarshal(b, &raw); err != nil {
		return err
	}
	var name string
	var ok bool
	if _, ok = raw["name"]; raw != nil && !ok {
		return errors.New("field name in AttributeNameValue: required")
	}
	if name, ok = raw["name"].(string); !ok {
		return errors.New("yaml: cannot unmarshal field name in AttributeNameValue must be string")
	}
	if strings.TrimSpace(name) == "" {
		return errors.New("field name in AttributeNameValue: required")
	}
	if _, ok := raw["value"]; raw != nil && !ok {
		return errors.New("field value in AttributeNameValue: required")
	}
	type Plain AttributeNameValue
	var plain Plain
	if err := json.Unmarshal(b, &plain); err != nil {
		return err
	}
	if plain.Type != nil && plain.Type.Value == "int" {
		val, ok := plain.Value.(float64)
		if ok {
			plain.Value = int(val)
		}
	}
	if plain.Type != nil && plain.Type.Value == "int_array" {
		m, ok := plain.Value.([]interface{})
		if ok {
			var vals []interface{}
			for _, v := range m {
				val, ok := v.(float64)
				if ok {
					vals = append(vals, int(val))
				} else {
					vals = append(vals, val)
				}
			}
			plain.Value = vals
		}
	}

	*j = AttributeNameValue(plain)
	return nil
}
```

config/v0.3.0/config_yaml.go
```diff
@@ -7,9 +7,27 @@ import (
...
+	type alias OpenTelemetryConfiguration // Prevents infinite recursion
+	aux := &alias{}
+
+	// Decode into alias
+	if err := unmarshal(aux); err != nil {
+		return err
+	}
+
+	// Check for an empty attributes list
+	if len(aux.Resource.Attributes) == 0 {
```
mikeblum, line 22: this is just to demo surfacing an error on an empty resource attributes list. This in effect makes resource attributes required so I don't think this will work.

MrAlias:

Issue #6638, opened 2025-01-17
Migrate to `semconv` for `otelmux`


dmathieu:
Migrate the otelmux package to generate the `semconv` package, and use it for generating traces and metrics semantic conventions, like the `otelhttp` package does.

MrAlias:
@martinyonatann @TanishqPorwar : please coordinate among yourselves here and determine how to unify your efforts. Two separate PRs is not a viable solution.

Please let me know who this issues should be assigned to. Also, please outline your planned solution prior to any future PRs so we can ensure some level of consensus on the approach.

TanishqPorwar:
@MrAlias Thank you for pointing this out. I had a look at @martinyonatann's PR and noticed that it is quite comprehensive, but it involves implementing separate logic for `otelmux`. My intent with raising PR (#6653) was to propose an alternate approach that focuses on reusing `otelhttp`.

This reuse aligns with `semconv` out of the box and avoids duplicating logic, making it simpler to maintain and ensuring consistency across instrumentations. I thought it would be worth exploring.

@martinyonatann I’d like to understand your approach better to identify any synergies or areas where we can consolidate efforts. I’m happy to adapt or refine my PR based on our discussion.

dmathieu:
There are two things happening here.
The migration to the semconv package, and refactorings of otelmux itself, to use otelhttp.

The latter is interesting, and makes sense.
The former is what is tracked in this issue, and has some urgency since we want to drop support for old semconv versions.

We should switch otelmux to the semconv package with **no new features**. Then, we can look into making things better.

martinyonatann:
I currently have 3 PRs related to this effort:

1.  [Setup Body Wrapper Package ](https://github.com/open-telemetry/opentelemetry-go-contrib/pull/6650) 
2. [Migrate `semconv` and Implement the Package in `Tracing`](https://github.com/open-telemetry/opentelemetry-go-contrib/pull/6652)
3. [Migrate `semconv` and Implement the Package in `Metrics`](https://github.com/open-telemetry/opentelemetry-go-contrib/pull/6648)

These PRs are modular but interrelated. The body wrapper package (PR 1) is a prerequisite for metrics (PR 3).

I recommend reviewing and merging them sequentially, starting with PR 1, followed by PR 2, and then PR 3.
Please let me know if this approach works or if you’d prefer a different strategy. I'm happy to adjust based on the team's preferences.

dmathieu:
[...]

PR #6648, opened 2025-01-19
feat: support metrics for otelmux

martinyonatann:
Address issue #6638

PR #6653, opened 2025-01-21
[draft] Migrate to semconv for otelmux #6638

TanishqPorwar:
tried using `otelhttp.Handler` in `otelmux` to address #6638

### DIFF
--- a/instrumentation/github.com/gorilla/mux/otelmux/config.go
+++ b/instrumentation/github.com/gorilla/mux/otelmux/config.go
@@ -6,6 +6,8 @@ package otelmux // import "go.opentelemetry.io/contrib/instrumentation/github.co
 import (
 	"net/http"
 
+	"go.opentelemetry.io/otel/metric"
+
 	"go.opentelemetry.io/otel/propagation"
 	oteltrace "go.opentelemetry.io/otel/trace"
 )
@@ -13,6 +15,7 @@ import (
 // config is used to configure the mux middleware.
 type config struct {
 	TracerProvider    oteltrace.TracerProvider
+	meterProvider     metric.MeterProvider
 	Propagators       propagation.TextMapPropagator
 	spanNameFormatter func(string, *http.Request) string
 	PublicEndpoint    bool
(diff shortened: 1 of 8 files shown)


TanishqPorwar:
@dmathieu would it make sense to use `otelhttp` since it already adheres to `semconv`? I'll tidy up this PR if this approach aligns with the right direction.

dmathieu:
Yes, if we can get HTTP instrumentations to use otelhttp rather than reimplement things themselves, that's definitely better.

akats7:
> Yes, if we can get HTTP instrumentations to use otelhttp rather than reimplement things themselves, that's definitely better.

@dmathieu wasn't there talk of refactoring otelhttp, not sure to what extent the intended changes would be breaking

dmathieu:
Yes. But if/when that happens, we will make the appropriate changes to avoid breaking other instrumentations. Reusing otelhttp also makes it much easier to keep. things working the same way.

dmathieu:
@TanishqPorwar did you see you're working on this alongside @martinyonatann (https://github.com/open-telemetry/opentelemetry-go-contrib/pull/6648)?

TanishqPorwar:
@dmathieu thanks for pointing this out. My goal with this PR was to propose an alternate solution by reusing `otelhttp`.

This approach minimizes redundancy, adheres to `semconv`, and simplifies future maintenance. That said, I’m open to discussing how we can align our efforts. @martinyonatann, let me know your thoughts, and we can figure out the best way forward.

martinyonatann:
from my personal perspective, to avoid unintended changes to `otelhtttp`, i prefer to reimplement the `semconv` and `request` packages because that aligns with the purpose of creating the package template.

Reusing `otelhttp` is definitely is **nice-to-have**, and i agree that it's a goal worth pursuing. However, we'll need to carefully pan how we can reuse it across frameworks that are built on `net/http`, such as `mux`, `echo`, `gin`, etc.

Each framework has its nuances, so ensuring a consistent implementation while accommodating these differences is key. Let's discuss how we can create a roadmap or strategy for this, so we can make the transition as smooth and efficient as possible.

akats7:
I think for this current iteration and to first complete the semconv migration let’s go with the reimplementation, especially since a good amount of work has been completed. We can continue to evaluate the merits of reusing otelhttp but I agree there are additional considerations we need to make if we’re going to start using otelhttp as a base.

dmathieu:
Let's keep the discussion in a single place, #6638.

PR #6658, opened 2025-01-21
config: add support for Insecure

### LINKED ISSUE #6351: otlpHTTPLogExporter, otlpGRPCLogExporter, etc do not support fields like OTLP.Certificate

mattsains:
fixes #6351

Based on spec here: https://github.com/open-telemetry/opentelemetry-specification/blob/main/specification/protocol/exporter.md#configuration-options (search "Insecure")


config/v0.3.0/log.go
```diff
@@ -183,7 +183,7 @@ func otlpGRPCLogExporter(ctx context.Context, otlpConfig *OTLP) (sdklog.Exporter
 		} else {
 			opts = append(opts, otlploggrpc.WithEndpoint(*otlpConfig.Endpoint))
 		}
-		if u.Scheme == "http" {
+		if u.Scheme == "http" || (u.Scheme != "https" && otlpConfig.Insecure != nil && *otlpConfig.Insecure) {
```
dmathieu, line 186: Should we add tests?

mattsains: I would like to and spent some time on it already, but these options get double wrapped in structs with only private members so I was at a loss as to how to inspect the output of this method. Happy to take suggestions

pellared:
@codeboten, PTAL as a codeowner

config/v0.3.0/log.go
dmathieu: No, sorry, your reasoning makes sense.

dmathieu (approved):

mattsains:
@codeboten I think I need your approval to get this PR to merged

codeboten (approved):

Issue #4484, opened 2023-10-26
BUG: Prometheus Bridge Histogram Convert Statistics are incorrect


jamsonzan:
### this is use Prometheus bridge report to otel collecter，and read collecter export metrics
`
MyHistogramDurationMs_bucket{instance="abcdefghijkl",job="myservice",uri0="default",uri1="default",uri2="default",uri3="default",uri4="default",uri5="default",le="1"} 0
MyHistogramDurationMs_bucket{instance="abcdefghijkl",job="myservice",uri0="default",uri1="default",uri2="default",uri3="default",uri4="default",uri5="default",le="2"} 0
MyHistogramDurationMs_bucket{instance="abcdefghijkl",job="myservice",uri0="default",uri1="default",uri2="default",uri3="default",uri4="default",uri5="default",le="3"} 0
MyHistogramDurationMs_bucket{instance="abcdefghijkl",job="myservice",uri0="default",uri1="default",uri2="default",uri3="default",uri4="default",uri5="default",le="4"} 0
MyHistogramDurationMs_bucket{instance="abcdefghijkl",job="myservice",uri0="default",uri1="default",uri2="default",uri3="default",uri4="default",uri5="default",le="5"} 0
MyHistogramDurationMs_bucket{instance="abcdefghijkl",job="myservice",uri0="default",uri1="default",uri2="default",uri3="default",uri4="default",uri5="default",le="10"} 0
MyHistogramDurationMs_bucket{instance="abcdefghijkl",job="myservice",uri0="default",uri1="default",uri2="default",uri3="default",uri4="default",uri5="default",le="15"} 0
MyHistogramDurationMs_bucket{instance="abcdefghijkl",job="myservice",uri0="default",uri1="default",uri2="default",uri3="default",uri4="default",uri5="default",le="20"} 0
MyHistogramDurationMs_bucket{instance="abcdefghijkl",job="myservice",uri0="default",uri1="default",uri2="default",uri3="default",uri4="default",uri5="default",le="30"} 0
MyHistogramDurationMs_bucket{instance="abcdefghijkl",job="myservice",uri0="default",uri1="default",uri2="default",uri3="default",uri4="default",uri5="default",le="50"} 0
MyHistogramDurationMs_bucket{instance="abcdefghijkl",job="myservice",uri0="default",uri1="default",uri2="default",uri3="default",uri4="default",uri5="default",le="70"} 0
[...]

PR #6685, opened 2025-01-26
fix: convert Prometheus histograms to non-cumulative otel histogram

mehran-prs:
When we record a value using Prometheus histograms, it captures that value for that bucket and also previous buckets, but open telemetry histogram captures in a non-cumulative way, I mean it just captures for the matched bucket and doesn't add that value to the previous buckets. 
This issue causes getting wrong histogram values when we use Prometheus bridge to send our data to open telemetry collector.
Please take a look at issue #4484 as an example of this issue. 

Its fix was quick, so I made a PR, but let me know if I'm wrong or I should use the bridge in another way please.


dashpole (approved):
Thanks! good find.

dmathieu (approved):

dmathieu (commented):

MrAlias:
@mehran-prs thank you for the contribution. Please add a changelog entry so these changes can be accepted.

mehran-prs:
@dmathieu @MrAlias Sure. Added.

CHANGELOG.md
```diff
@@ -24,6 +24,7 @@ This project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0.htm
 ### Fixed
 
 - Use `context.Background()` as default context instead of nil in `go.opentelemetry.io/contrib/bridges/otellogr`. (#6527)
+- Convert Prometheus histogram to non-cumulative otel histogram in `go.opentelemetry.io/contrib/bridges/prometheus`. (#6685)
```
MrAlias, line 27:
```suggestion
- Convert Prometheus histogram to a delta histogram in `go.opentelemetry.io/contrib/bridges/prometheus`. (#6685)
```

bridges/prometheus/producer_test.go
```diff
@@ -229,6 +229,44 @@ func TestProduce(t *testing.T) {
...
+			expected: []metricdata.ScopeMetrics{{
+				Scope: instrumentation.Scope{
+					Name: scopeName,
+				},
+				Metrics: []metricdata.Metrics{
+					{
+						Name:        "test_histogram_metric",
+						Description: "A histogram metric for testing",
+						Data: metricdata.Histogram[float64]{
+							Temporality: metricdata.CumulativeTemporality,
```
MrAlias, line 254: Shouldn't this be delta temporality?

dashpole: No, prometheus always produces cumulative metrics. This is just changing how bucket counts are represented.  OTel 's bucket count is the number that fell into that specific bucket.  Promethues' bucket count is the number that fell into that bucket **or below**.

CHANGELOG.md
dashpole: I prefer the "non-cumulative" wording.  I would suggest
```suggestion
- Convert Prometheus histogram buckets to non-cumulative otel histogram buckets in `go.opentelemetry.io/contrib/bridges/prometheus`. (#6685)
```

mehran-prs: Thanks. Done

MrAlias (approved):

PR #6698, opened 2025-01-29
[chore] Update codespell target

MrAlias:
Copy of https://github.com/open-telemetry/opentelemetry-go/pull/6223

- Use the local user to install and run so that root does not own the artifacts.
- Move the python image into a place renovate can maintain it.


dashpole (approved):

MrAlias:
Merging early given this is for tooling and is a copy of an already accepted PR.

PR #6727, opened 2025-02-06
config: add support for configuring propagators

### LINKED ISSUE #6712: config: add support for configuring propagators
### Problem Statement

Configuring propagators via the config doesn't currently work, this needs to be done.

### Proposed Solution

Use the propagators from the configuration and set the propagators

#### Alternatives

This can be achieved via manually setting propagators

jpkrohling:
Fixes #6712


config/v0.3.0/propagation.go
```diff
@@ -0,0 +1,65 @@
...
+
+	if len(ps) == 0 {
+		return nil, nil
+	}
+
+	res := propagation.NewCompositeTextMapPropagator(ps...)
+	return res, nil
+}
+
+func propagatorByName(name string) (propagation.TextMapPropagator, error) {
```
pellared, line 48: _Not a blocker. This is likely to be tracked as a separate issue._

Should we not take advantage of https://pkg.go.dev/go.opentelemetry.io/contrib/propagators/autoprop? Similarly for exporters we could use https://pkg.go.dev/go.opentelemetry.io/contrib/exporters/autoexport.

I think that the config is supposed to be extensible and allow custom exporters, processors, propagators. See: https://github.com/open-telemetry/opentelemetry-specification/blob/main/specification/configuration/sdk.md#sdk-extension-components

config/v0.3.0/config.go
```diff
@@ -73,6 +80,7 @@ var noopSDK = SDK{
 	loggerProvider: nooplog.LoggerProvider{},
 	meterProvider:  noopmetric.MeterProvider{},
 	tracerProvider: nooptrace.TracerProvider{},
+	propagator:     nil,
```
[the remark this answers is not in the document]
jpkrohling: True, otherwise callers would have to check the value for nil, while an empty set is effectively noop.

config/v0.3.0/propagation.go
pellared: I agree. I created https://github.com/open-telemetry/opentelemetry-go-contrib/issues/6728. Resolving.

config/v0.3.0/config.go
```diff
@@ -103,10 +111,16 @@ func NewSDK(opts ...ConfigurationOption) (SDK, error) {
 		return noopSDK, err
 	}
 
+	propagator, err := propagator(o)
+	if err != nil {
+		return noopSDK, err
```
pellared, line 116: We need shutdown the providers that were already created so that we do not leak any resources (e.g. hanging goroutines).

pellared:
Changelog entry is missing.

config/v0.3.0/config.go
jpkrohling: Alright, I'll move it to the top then, as it's simpler.

config/v0.3.0/config_test.go
```diff
@@ -52,6 +55,7 @@ func TestNewSDK(t *testing.T) {
 			wantTracerProvider: &sdktrace.TracerProvider{},
 			wantMeterProvider:  &sdkmetric.MeterProvider{},
 			wantLoggerProvider: &sdklog.LoggerProvider{},
+			wantPropagators:    propagation.NewCompositeTextMapPropagator(),
```
pellared, line 58: This should have some configuration.

config/v0.3.0/propagation.go
```diff
@@ -0,0 +1,60 @@
...
+	"go.opentelemetry.io/contrib/propagators/aws/xray"
+	"go.opentelemetry.io/contrib/propagators/b3"
+	"go.opentelemetry.io/contrib/propagators/jaeger"
+	"go.opentelemetry.io/contrib/propagators/ot"
+	"go.opentelemetry.io/otel/propagation"
+)
+
+func propagator(cfg configOptions) (propagation.TextMapPropagator, error) {
+	if cfg.opentelemetryConfig.Propagator == nil {
+		return propagation.NewCompositeTextMapPropagator(), nil
```
pellared, line 18: Should we not default to `tracecontext,baggage`?
PTAL @open-telemetry/configuration-approvers

jpkrohling: I'm not so sure: that's not the default behavior of an unconfigured SDK, from what I remember. Having a default set of propagators would deviate from that behavior.

codeboten: This should default to the same as the env variables defined in the spec: https://github.com/open-telemetry/opentelemetry-specification/blob/main/specification/configuration/sdk-environment-variables.md#general-sdk-configuration

config/v0.3.0/propagation.go
```diff
@@ -0,0 +1,60 @@
...
+		ps = append(ps, p)
+	}
+	if rErr != nil {
+		return nil, rErr
+	}
+
+	return propagation.NewCompositeTextMapPropagator(ps...), nil
+}
+
+func propagatorByName(name string) (propagation.TextMapPropagator, error) {
```
codeboten, line 41: It might be possible to simplify the code here by using the `autoprop` package

codeboten (commented):
Thanks for taking on this work @jpkrohling! Added a couple of comments

config/v0.3.0/propagation.go
pellared: Given the second reviewer says the same I think this could be addressed 

Reference: https://github.com/open-telemetry/opentelemetry-go-contrib/pull/6727#discussion_r1944603362

jpkrohling: Done! I'm not sure the current code justifies a new function and new source file. On the other hand, it's aligned with how things are done for similar constructs. I think I prefer this way, but I don't have strong opinions.

jpkrohling:
If this looks good, I'll fix the merge conflicts.

CHANGELOG.md
```diff
@@ -18,6 +18,7 @@ This project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0.htm
 - Added metrics support, and emit all stable metrics from the [Semantic Conventions](https://github.com/open-telemetry/semantic-conventions/blob/main/docs/http/http-metrics.md) in `go.opentelemetry.io/contrib/instrumentation/github.com/gorilla/mux/otelmux`. (#6648)
 - Add support for configuring `Insecure` field for OTLP exporters in `go.opentelemetry.io/contrib/config`. (#6658)
 - Support for the `OTEL_HTTP_CLIENT_COMPATIBILITY_MODE=http/dup` environment variable in `instrumentation/net/http/httptrace/otelhttptrace` to emit attributes for both the v1.20.0 and v1.26.0 semantic conventions. (#6720)
+- Add support for configuring propagators. (#6727)
```
dmathieu, line 21:
```suggestion
- Add support for configuring propagators in `go.opentelemetry.io/contrib/config`. (#6727)
```

dmathieu (approved):

config/v0.3.0/propagation.go
```diff
@@ -0,0 +1,24 @@
...
+)
+
+func propagator(cfg configOptions) (propagation.TextMapPropagator, error) {
+	if cfg.opentelemetryConfig.Propagator == nil {
+		return autoprop.NewTextMapPropagator(), nil
+	}
+
+	names := make([]string, 0, len(cfg.opentelemetryConfig.Propagator.Composite))
+	for _, name := range cfg.opentelemetryConfig.Propagator.Composite {
+		if name != nil && *name != "" {
```
pellared, line 18: I think that a `null` name in JSON and YAML are invalid and should result in returning an error.

I also do not see a reason to ignore empty names.

config/v0.3.0/config_test.go
pellared, line 1: Can you also please update `v0.3.yaml` and `v0.3.json` with propagator configuration?
`v03OpenTelemetryConfig` variable would have to be updated accordingly.

config/go.mod
```diff
@@ -5,6 +5,11 @@ go 1.22.0
 require (
 	github.com/prometheus/client_golang v1.20.5
 	github.com/stretchr/testify v1.10.0
+	go.opentelemetry.io/contrib/propagators/autoprop v0.59.0
+	go.opentelemetry.io/contrib/propagators/aws v1.34.0
```
flc1125, line 9: Is the package missing a replace to introduce the configuration of the local package?

Other packages are similar.

config/go.mod
```diff
@@ -44,10 +49,14 @@ require (
...
+	go.uber.org/multierr v1.11.0 // indirect
 	golang.org/x/net v0.34.0 // indirect
 	golang.org/x/sys v0.30.0 // indirect
 	golang.org/x/text v0.22.0 // indirect
 	google.golang.org/genproto/googleapis/api v0.0.0-20250207221924-e9438ea467c6 // indirect
 	google.golang.org/genproto/googleapis/rpc v0.0.0-20250207221924-e9438ea467c6 // indirect
 	google.golang.org/protobuf v1.36.5 // indirect
 )
+
+// remove after v0.59.0 of autoprop
```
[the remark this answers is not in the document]
jpkrohling: I don't know, I'm not familiar with how things work in this repo. Do you have an example for me to follow?

propagators/autoprop/registry.go
```diff
@@ -124,7 +124,14 @@ func TextMapPropagator(names ...string) (propagation.TextMapPropagator, error) {
...
 	)
 
+	if len(names) == 0 {
+		return NewTextMapPropagator(), nil
+	}
+
 	for _, name := range names {
+		if len(name) == 0 {
+			continue
+		}
```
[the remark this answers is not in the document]
jpkrohling: Sure, I can revert that. As a user, I'd prefer the software I'm dealing with to be lenient instead of strict, but that's me.

config/v0.3.0/propagation.go
jpkrohling: Same as above: I think this makes it lenient for end-users (who are specifying it in a yaml file), but I'll follow the common practice for this repo.

config/v0.3.0/config_test.go
jpkrohling: > Can you also please update v0.3.yaml and v0.3.json with propagator configuration?

They were there before this PR, or perhaps I'm not seeing what you mean?

config/go.mod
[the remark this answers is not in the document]
jpkrohling: `make crosslink` added a few others, thanks for the pointer!

config/go.mod
jpkrohling: I ran crosslink which brought more replace statements. Is that what you had in mind?

flc1125: Although it may be a bit cumbersome. However, I'm thinking that if we don't do this, it means that the dependencies of the current package will always be the previous release version.

If we make some incompatible adjustments to the dependencies in the current unreleased version, our unit tests may not be able to detect them. The problem won't be found until after the official release.

config/v0.3.0/config_test.go
pellared: I mean these two files:
- `config/testdata/v0.3.yaml`
- `config/testdata/v0.3.json`

so that they include the propagators

config/go.mod
pellared: @flc1125, I can see replace for all these modules at the end of the file.

pellared: > I did see it. This is a comment, not a discussion that must be modified. 

I do not think it was clear. 

> Therefore, there is no "Resolved" button.

I do see a "Resolve conversation" button. Resolving now :wink:

config/v0.3.0/propagation_test.go
```diff
@@ -0,0 +1,183 @@
...
+	"go.opentelemetry.io/contrib/propagators/ot"
+	"go.opentelemetry.io/otel/propagation"
+)
+
+func TestPropagator(t *testing.T) {
+	tests := []struct {
+		name    string
+		cfg     configOptions
+		want    propagation.TextMapPropagator
+		wantErr bool
```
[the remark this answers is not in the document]
jpkrohling: I'll leave it as is for now, given that one of the errors is coming from a different package and the error isn't exported. I'd need a special condition for that error, testing the message itself, which is what we do already for all errors.

config/v0.3.0/config_test.go
jpkrohling: That's what I meant as well: 

https://github.com/open-telemetry/opentelemetry-go-contrib/blob/6780d3fc79127270f7a20ddc49243937f8433ec5/config/testdata/v0.3.yaml#L217C30-L220

https://github.com/open-telemetry/opentelemetry-go-contrib/blob/6780d3fc79127270f7a20ddc49243937f8433ec5/config/testdata/v0.3.json#L166-L176

config/v0.3.0/propagation.go
```diff
@@ -0,0 +1,41 @@
...
+	}
+
+	names := make([]string, 0, n)
+	for _, name := range cfg.opentelemetryConfig.Propagator.Composite {
+		if name == nil {
+			return nil, errInvalidPropagatorNil
+		}
+		if *name == "" {
+			return nil, errInvalidPropagatorEmpty
+		}
```
pellared, line 35: I do not see anything in the schema that disallows using empty name. See: https://github.com/open-telemetry/opentelemetry-configuration/blob/main/schema/propagator.json

`autoprop` supports empty names (similarly to `database/sql`).

codeboten: i opened an [issue](https://github.com/open-telemetry/opentelemetry-configuration/issues/172) in the configuration repo to discuss this, i submitted a PR to propose enforcing a minimum length on propagator configuration

MrAlias:

jpkrohling:
Folks, sorry for dropping the ball on this one. I'm closing, as I don't think I have time to merge the conflicts and follow-up on potential changes from the next PRs. Feel free to use the PR as inspiration for new PRs!

PR #6763, opened 2025-02-12
Remove otelhttp/test module

pellared:
Per https://github.com/open-telemetry/opentelemetry-go/pull/6286#discussion_r1946354996

When module `a` has a dependency is used only in tests then it does not come up as an indirect dependency for modules depending on `a`. 

I think this was not true before and that the Go module graph pruning has been improved at some point.

I think that it means that we can get rid of `test` modules as can use SDK in tests and examples as they do not come up as indirect dependencies in `go.mod`.

EDIT: The test modules were introduced when we were supporting Go 1.15. https://github.com/open-telemetry/opentelemetry-go-contrib/blob/fe2fb58536fff0f85173f56a9091bca9380cee3c/instrumentation/net/http/otelhttp/test/go.mod
Go module pruning was works for modules requiring at least Go 1.17. See: https://go.dev/ref/mod#graph-pruning and 
https://go.dev/doc/go1.17

I plan to do the same for all test modules in separate PRs.


dashpole (approved):

dmathieu (approved):

instrumentation/net/http/httptrace/otelhttptrace/example/go.sum
```diff
@@ -25,6 +25,8 @@ go.opentelemetry.io/otel/metric v1.34.0 h1:+eTR3U0MyfWjRDhmFMxe2SsW64QrZ84AOhvqS
 go.opentelemetry.io/otel/metric v1.34.0/go.mod h1:CEDrp0fy2D0MvkXE+dPV7cMi8tWZwX3dmaIhwPOaqHE=
 go.opentelemetry.io/otel/sdk v1.34.0 h1:95zS4k/2GOy069d321O8jWgYsW3MzVV+KuSPKp7Wr1A=
 go.opentelemetry.io/otel/sdk v1.34.0/go.mod h1:0e/pNiaMAqaykJGKbi+tSjWfNNHMTxoC9qANsCzbyxU=
+go.opentelemetry.io/otel/sdk/metric v1.34.0 h1:5CeK9ujjbFVL5c1PhLuStg1wxA7vQv7ce1EK0Gyvahk=
+go.opentelemetry.io/otel/sdk/metric v1.34.0/go.mod h1:jQ/r8Ze28zRKoNRdkjCZxfs6YvBTG1+YIqyFVFYec5w=
```
MrAlias, line 29: It seems like packages that import `otelhttp` are still taking a dependency on `go.opentelemetry.io/otel/sdk/metric`.

Why are we changing this? It is currently working, right?

pellared: This is not "real" a dependency.

go.sum has checksums of "everything" possible including unreachable code like test and tool dependencies 

> Why are we changing this? It is currently working, right?

Simpliify our codebase and number of modules

pellared: > I don't see that justifying us imposing this on our user-base. I can understand users complaining that they need to have reference to our SDK if they adopt this instrumentation.

They still won't have to reference our SDK.

> That is something we don't have to deal with currently and by adding work here we are causing ourselves more issues

Do you mean that it is a waste of our time and not worth the effort?

pellared: On the other hand, maybe it is nice that we are not even adding an SDK entry in `go.sum`. Just wondering it is not going too extreme. Should we then also get rid of `testify` :sweat_smile:

pellared: Notes from a conversation with @MrAlias.

We need to check what is the impact of `go.sum` entries. 

For sure it will be listed when `go list -m all`:

> Modules whose requirements have been pruned out still appear in the module graph and are still reported by go list -m all: their [selected versions](https://go.dev/ref/mod#glos-selected-version) are known and well-defined, and packages can be loaded from those modules (for example, as transitive dependencies of tests loaded from other modules).

Personally, I do not any tool which uses `go list -m all` to check the dependencies.

We should check if there are some static analysis tools that would report some issues because of it.
For sure it is fine for dependabot and renovate that operate on `go.mod` files.

I know that some are inspecting Go binaries using `go version -m $BINARY` which also has pruned content.

https://pkg.go.dev/golang.org/x/vuln/cmd/govulncheck results are even more pruned. From https://go.dev/blog/vuln:

> Govulncheck analyzes your codebase and only surfaces vulnerabilities that actually affect you, based on which functions in your code are transitively calling vulnerable functions.

Black Duck prunes unused dependencies https://documentation.blackduck.com/bundle/detect/page/packagemgrs/golang.html:

> - Runs go list -m and go mod graph, and parses the output of both to discover dependencies.
> - Runs go mod why to remove unused Go modules such as dependencies required by the build system or tests.

Snyk probably operates on `go.mod` files https://docs.snyk.io/supported-languages-package-managers-and-frameworks/go.

I do not think any tool uses `go.sum` or direct results of `go list -m all` as it would bring to many false positives.

We could merge this PR and wait e.g. 3 months to check if anybody would complain.

MrAlias: > Snyk probably operates on `go.mod` files https://docs.snyk.io/supported-languages-package-managers-and-frameworks/go.

I can confirm this was the way they it worked a few years ago at least :+1:

MrAlias: > We could merge this PR and wait e.g. 3 months to check if anybody would complain.

That sounds reasonable. In that time frame doing the same to other unstable modules would be okay, but let's hold off on doing the same for "stable" `test` modules.

If we find this has non-zero impact to compliance or other systems processes, we can revert it.

pellared: @open-telemetry/go-maintainers, does it sound reasonable? If nobody objects I plan to merge this PR on Tuesday my morning.

MrAlias (approved):

PR #6786, opened 2025-02-15
chore(otelgin): rename files for improved clarity and consistency

flc1125:
Same as them, adjust the naming to be more semantic. Moreover, currently, gin is not only about `trace` but also includes `metric`.

- https://github.com/open-telemetry/opentelemetry-go-contrib/tree/8b5220b6f51f51db8d05656b3c8ebce4e25ffbc5/instrumentation/github.com/labstack/echo/otelecho
- https://github.com/open-telemetry/opentelemetry-go-contrib/tree/8b5220b6f51f51db8d05656b3c8ebce4e25ffbc5/instrumentation/github.com/gorilla/mux/otelmux

> Additional note: This should be able to skip the `CHANGELOG`.


MrAlias (approved):
Not sure this is technically "the correct naming", but seems like an okay change.

flc1125:
> Not sure this is technically "the correct naming", but seems like an okay change.

It may not be the best name in terms of technology. More to unify it with other packages.

dmathieu (approved):

PR #6815, opened 2025-02-19
[config] Allow prometheus to configure IPv6 by using the standard net host port joining

bogdandrutu:


codeboten (commented):
Thanks @bogdandrutu, any tests that could be added?

i recall running into issues w/ ipv6 support in the collector when i was manually testing this previously.

sfc-gh-bdrutu:
>Thanks @bogdandrutu, any tests that could be added?

I really like when bugs are found, no unit-test at all for that code and the person who writes the fix has to also write first tests :))

CHANGELOG.md
```diff
@@ -41,6 +41,7 @@ The next release will require at least [Go 1.23].
 - Convert Prometheus histogram buckets to non-cumulative otel histogram buckets in `go.opentelemetry.io/contrib/bridges/prometheus`. (#6685)
 - Don't start spans that never end for filtered out gRPC stats handler in `go.opentelemetry.io/contrib/instrumentation/google.golang.org/grpc/otelgrpc`. (#6695)
 - Fix a possible nil dereference panic in `NewSDK` of `go.opentelemetry.io/contrib/config/v0.3.0`. (#6752)
+- Fix problem with using IPv6 for prometheus endpoint. (#6815)
```
dmathieu, line 44:
```suggestion
- Fix prometheus endpoint with an IPv6 address in `go.opentelemetry.io/contrib/config`. (#6815)
```

dashpole (approved):

dashpole:
Thanks @bogdandrutu!

MrAlias (approved):

config/v0.2.0/metric.go
```diff
@@ -298,6 +299,11 @@ func prometheusReader(ctx context.Context, prometheusConfig *Prometheus) (sdkmet
 	reg := prometheus.NewRegistry()
 	opts = append(opts, otelprom.WithRegisterer(reg))
 
+	reader, err := otelprom.New(opts...)
+	if err != nil {
+		return nil, fmt.Errorf("error creating otel prometheus exporter: %w", err)
```
pellared, line 304:
```suggestion
		return nil, fmt.Errorf("new prometheus exporter: %w", err)
```

pellared (approved):

config/v0.2.0/metric.go
bogdandrutu: I would defer this to a different PR.

codeboten (commented):
Not sure if this is expected behaviour or not, but testing this PR with the following config results in an error: 

```
meter_provider:
  readers:
    - pull:
        exporter:
          prometheus:
            host: "[::1]"
            port: 8888
```

The error message:

`binding address [[::1]]:8888 for Prometheus exporter: listen tcp: address [[::1]]:8888: missing port in address`

I think I ran into this w/ collector tests previously where both the notation wrappedi n square brackets and the one not wrapped was being used, causing me to add this code in the test utility https://github.com/open-telemetry/opentelemetry-collector/blob/f024b0d8d9afeb827139d82b7c5c4667635457d6/service/internal/promtest/server_util.go#L27

codeboten:
Some more details about ipv6 notation in the [RFC](https://www.ietf.org/rfc/rfc2732.txt)

codeboten (approved):

codeboten (approved):

dmathieu (approved):

Issue #6822, opened 2025-02-21
minsev: Update code to the latest FilterProcessor


pellared:
Sync after https://github.com/open-telemetry/opentelemetry-go/pull/6317

PR #6833, opened 2025-02-24
minsev: Update to use sdk/log.FilterProcessor

pellared:
Fixes https://github.com/open-telemetry/opentelemetry-go-contrib/issues/6822

Update after https://github.com/open-telemetry/opentelemetry-go/pull/6317

I also updated the example a little bid.


processors/minsev/example_test.go
```diff
@@ -10,39 +10,52 @@ import (
 	"strings"
 	"sync"
 
-	"go.opentelemetry.io/otel/log"
+	"go.opentelemetry.io/contrib/processors/minsev"
+	logapi "go.opentelemetry.io/otel/log"
+	"go.opentelemetry.io/otel/sdk/log"
```
dmathieu, line 15: Nit, but the (unwritten) convention in other packages is to use the default name for the API, and to name the SDK `logsdk`

dmathieu (approved):

processors/minsev/example_test.go
pellared: https://github.com/open-telemetry/opentelemetry-go-contrib/pull/6833/commits/1c28d9d2667fb66433759fec92073ab4633868a4

PR #6859, opened 2025-02-28
Design document for otelhttpconv package

dmathieu:
This provides a design document for a new instrumentation, `otelhttpconv` that would serve as a meta instrumentation for all HTTP packages to more easily share their behavior.


instrumentation/net/http/otelhttpconv/DESIGN.md
```diff
@@ -0,0 +1,139 @@
...
+internals (they don't all allow using `net/http.Handler`).
+This is causing a lot of duplication across instrumentations, and makes
+standardization harder.
+
+Folks have also expressed interest in being able to use the internal tools
+provided by the internal `semconvutil` package within their own
+instrumentations ([#4580](https://github.com/open-telemetry/opentelemetry-go-contrib/issues/4580)).
+This package aims to solve that by being publicly usable.
+
+This document outlines a proposal for a new instrumentation package called
```
pellared, line 16:
```suggestion
This document outlines a proposal for a new instrumentation module called
```

instrumentation/net/http/otelhttpconv/DESIGN.md
```diff
@@ -0,0 +1,139 @@
...
+`otelhttpconv` that will provide a way to reduce some of this code repetition,
+especially with the setup of spans and metrics and their attributes.
```
pellared, line 18: Do we want to depend on this module in other modules such as `otelhttp`, `otelmux` etc

If so then can you explicitly mention it in this document as well as add something about backwards compatibility? 

The reason why I did not want to come up with such module was that any breaking change of such module can break compilation of modules depending on this module.

Basically I am afraid of having problems like:
- https://github.com/open-telemetry/opentelemetry-go/issues/3846

dmathieu: I've added a section about that.

XSAM (commented):
Could you add an example for a certain existing HTTP instrumentation to describe how `otelhttpconv` is actually being used in HTTP instrumentation? This could also give other authors more clues about how to leverage `otelhttpconv` to reduce code repetition.

dmathieu:
I've added a sample implementation of an `http.Handler` to show what it would look like for implementations.
I'm happy to provide a full implementation of `otelhttpconv` as well, to make this code work. But I'd rather do that when we're nearing consensus to avoid having to change things too many times.

instrumentation/net/http/otelhttpconv/DESIGN.md
```diff
@@ -0,0 +1,229 @@
...
+One for client. and one for servers. Implementations can use one, the other or both.
+
+#### Shared between client and server
+
+```golang
+// RequestWrapper provides a layer on top of `*http.Response` that tracks
+// additional information such as bytes read etc.
+type ResponseWrapper interface {
+	http.ResponseWriter
+
```
boekkooi-impossiblecloud, line 62: Should a `Error() error` func be added (similar to [RespWriterWrapper.Error](https://github.com/open-telemetry/opentelemetry-go-contrib/blob/3ea6585a2223dd03cee660b55a81c85f2065cafb/internal/shared/request/resp_writer_wrapper.go.tmpl#L117))?

instrumentation/net/http/otelhttpconv/DESIGN.md
```diff
@@ -0,0 +1,229 @@
...
+	// Duration contains the duration of the HTTP request
+	Duration() time.Duration
+
+	// Duration contains the status code returned by the HTTP request
+	StatusCode() int
+
+	// BytesRead contains the amount of bytes read from the response's body
+	BytesRead() int64
```
boekkooi-impossiblecloud, line 70: I'm not sure but this method feel a little out of place for a `http.ResponseWriter`? 
Would this maybe require 2 interfaces? One for reading and one for writing?

dmathieu: I merged the `ResponseWriter` and `BodyWrapper` into one interface to reduce the surface area. But it may make more sense to keep them separate.

instrumentation/net/http/otelhttpconv/DESIGN.md
```diff
@@ -0,0 +1,240 @@
...
+	BytesWritten() int64
+
+	// Error contains the last error that occurred while writing
+	Error() error
+}
+
+// Server provides an interface for HTTP server instrumentations to set the
+// proper semantic convention attributes and metrics into their data.
+type Server interface {
+
```
XSAM, line 115:
```suggestion
```

nit

XSAM (approved):

instrumentation/net/http/otelhttpconv/DESIGN.md
```diff
@@ -0,0 +1,239 @@
...
+	ctx := m.propagators.Extract(r.Context(), propagation.HeaderCarrier(r.Header))
+
+	// We keep creating the span here, as that is something we want to do before
+	// the middleware stack is run
+	ctx, span := m.tracer.Start(ctx, fmt.Sprintf("%s %s", r.Method, r.Pattern))
+	defer span.End()
+
+	// NewResponseWrapper wraps additional data into the http.ResponseWriter,
+	// such as duration, status code and quantity of bytes read.
+	rww := otelhttpconv.NewResponseWrapper(w)
```
pellared, line 213: Isn't this coupled to `otelhttpconv.NewHTTPConv`?
Are we sure that it would work with other custom semantic conventions?

instrumentation/net/http/otelhttpconv/DESIGN.md
```diff
@@ -0,0 +1,239 @@
...
+attributes/metrics.
+
+## Design
+
+The proposed design aims to:
+
+* Work with every official instrumentation, and be available to external implementers.
+* Provide flexibility in its use, to allow folks the use of unstable semantic conventions if they wish.
+
+### Interfaces
```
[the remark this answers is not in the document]
dmathieu: For BodyWrapper, we may want to introduce structs. I just didn't didn't want to add that logic to the context of this PR.
I can expand to have that though.

For the other ones, having interfaces allows having extensions, for example to add unstable semantic conventions. Or possibly custom ones.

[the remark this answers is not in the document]
pellared: Can you please add a "Compatibility" section  explaining different scenarios? E.g. adding new stable functionality and adding new experimental functionality.

pellared (commented):
It would be nice to have a working prototype as a separate draft PR.

instrumentation/net/http/otelhttpconv/DESIGN.md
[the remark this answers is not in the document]
dmathieu: I have changed the implementation and added a compatibility section.

instrumentation/net/http/otelhttpconv/DESIGN.md
dmathieu: So, even with structs for the wrappers, I think this design makes sense.
Building a response wrapper from an `HTTPConv` would be strange, as then each implementations would need to know how to build a wrapper as well.

dmathieu:
> It would be nice to have a working prototype as a separate draft PR.

I intend to do that. But I'd rather have the spec stabilize first, to avoid rewriting it several times.

instrumentation/net/http/otelhttpconv/DESIGN.md
```diff
@@ -0,0 +1,239 @@
...
+	// BytesRead contains the amount of bytes read from the response's body
+	BytesRead() int64
+
+	// Error contains the last error that occurred while reading
+	Error() error
+}
+
+// Client provides an interface for HTTP client instrumentations to set the
+// proper semantic convention attributes and metrics into their data.
+type Client interface {
```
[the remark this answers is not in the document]
dmathieu: I've added an `Alternatives` section.

instrumentation/net/http/otelhttpconv/DESIGN.md
```diff
@@ -0,0 +1,283 @@
...
+passing optional parameters to be used within the `Record` method, such as the
+HTTP route.
+
+When the data those options provide is not specified, the related span
+attributes will not be set.
+
+### Compatibility
+
+These interfaces are kept relatively simple on purpose, so they can be used as
+foundations if folks wish to extend them with their own actual implementations.
```
pellared, line 154: What would be a problem to use a `struct`? They can still be used as foundations. The users can always create their own interfaces if they need to support different implementation.

instrumentation/net/http/otelhttpconv/DESIGN.md
```diff
@@ -0,0 +1,283 @@
...
+
+Once stable, we will not be able to change those interfaces anymore (adding,
+renaming or deleting methods), as that would break backwards compatibility with
+implementations.
```
pellared, line 158: Then what is our plan if we would like to add a new functionality?

instrumentation/net/http/otelhttpconv/DESIGN.md
```diff
@@ -0,0 +1,283 @@
...
+such as:
+
+* A proxy that would allow combining multiple implementations together.
+* An implementation that covers unstable semantic conventions.
+* Implementations for older semantic conventions.
+* Implementations that cover custom implementations if a customer doesn't wish to use semconvs.
+
+### Implementations
+
+We will provide one official implementation of the described interfaces.
```
pellared, line 170: What is the point of exposing an interface if we want to provide only one implementation?

instrumentation/net/http/otelhttpconv/DESIGN.md
dmathieu: Same as for the interfaces in the SDK. We either introduce a new interface, or we release a v2.

instrumentation/net/http/otelhttpconv/DESIGN.md
dmathieu: So that folks can provide their own implementation.
Also, I do mention that we may provide additional implementations later on, for unstable conventions for example.

instrumentation/net/http/otelhttpconv/DESIGN.md
dmathieu: How would you do that?
If we provide the following API:

```
otelhttp.NewTransport(http.DefaultTransport, otelhttp.WithHTTPConv(myCustomImplementation{}))
```

And `otelhttp.WithHTTPConv` takes a struct, you can't extend it. It has to be our implementation.

pellared: `otelhttp` would define the interface as this is the package that accepts input and not `otelhttpconv` which provides an implementation.

`otelhttp` should not expose `otelhttpconv` types

instrumentation/net/http/otelhttpconv/DESIGN.md
pellared: The places where we want to allow passing custom implementations should define the interface (e.g. `otelhttp`). `otelhttpconv` does not accept any custom implementation. 

Reference: https://go.dev/wiki/CodeReviewComments#interfaces

instrumentation/net/http/otelhttpconv/DESIGN.md
```diff
@@ -0,0 +1,283 @@
...
+// NewMiddleware returns a tracing and metrics instrumentation middleware.
+// The handler returned by the middleware wraps a handler
+// in a span named after the operation and enriches it with metrics.
+func NewMiddleware(operation string) func(http.Handler) http.Handler {
+	m := middleware{
+		operation:   operation,
+		tracer:      otel.Tracer("http"),
+		propagators: otel.GetTextMapPropagator(),
+		meter:       otel.Meter("http"),
+		httpconv:    otelhttpconv.NewHTTPConv(otel.Tracer("httpconv"), otel.Meter("httpconv")),
```
pellared, line 211: `NewHTTPConv` is not defined in the design.

instrumentation/net/http/otelhttpconv/DESIGN.md
pellared, line 154: Do we really want to support custom semconv implementations?
I think it is an overkill (and foodgun). Should we just focus on OTel SemConv?
People not happy with OTel SemConv can create their own instrumentation libraries. 
I propose that `otelhttpconv` has "concrete" implementations (functions, structs).
If we want to provide customization, I would rather add it on instrumentation library layer as this is the API that the user is interacting with.

dmathieu: I think being able to support custom semantic conventions is just a side effect of the modularity. The real interesting thing here would be to be able to support unstable semantic conventions, with no breaking changes on the code that supports stable ones.

instrumentation/net/http/otelhttpconv/DESIGN.md
dmathieu: Sorry, that seemed clear to me. It creates an instance of our official implementation of the interface.
Since the implementation itself isn't in the design, I didn't add this method.

instrumentation/net/http/otelhttpconv/DESIGN.md
pellared: I find it hard to evaluate without code and examples.

dmathieu: Let's discuss this during today's SIG meeting. We have different views here.

While otelhttp would definitely use this package, the idea here is that other packages which currently use the generated `semconv` package would too.
Then, they need the types exposed.

pellared: > Let's discuss this during today's SIG meeting. We have different views here.

> the idea here is that other packages which currently use the generated semconv package would too.

Sure but these can be `struct` and not `interface`.

PR #6887, opened 2025-03-05
fix OTEL_HTTP_CLIENT_COMPATIBILITY_MODE to OTEL_SEMCONV_STABILITY_OPT_IN in the changelog

dmathieu:
The changelog references the wrong environment variable.
It should be `OTEL_SEMCONV_STABILITY_OPT_IN`, this is what the semantic conventions spec mentions, and what the code does.

See
https://github.com/open-telemetry/opentelemetry-go-contrib/blob/870ad7ca411dcdc9b61aaedc43074c61daea9b50/internal/shared/semconv/env.go.tmpl#L24


pellared (approved):
Planning to merge as trivial.

PR #6892, opened 2025-03-06
Align jaeger probabilistic sampler with trace.TraceIDRatioBased

diurnalist:
currently the jaegerremotesampler has a custom probabilistic sampler that does not align with how the traceid_ratiobased sampler works. it only looks at the first 64 bits of the trace; the opentelemetry-go sampler looks at the last 64 bits, because there are cases when trace IDs having 64 bits are converted to otel trace IDs, which have 128 bits, and when that happens the first 64 bits are typically ignored/zero-d out.

rather than re-implement this logic, this just pulls in the traceid_ratiobased sampler so we re-use it, ensuring the decisions will be made similarly and predictably.


samplers/jaegerremote/sampler_test.go
```diff
@@ -34,26 +35,27 @@ const (
 	testFirstTimeOperationName = "firstTimeOp"
 
 	testDefaultSamplingProbability = 0.5
-	testMaxID                      = uint64(1) << 62
+	testMaxID                      = uint64(1) << 63
```
diurnalist, line 38: i think jaeger's old probabilistic sampler only looked at 63 bits

samplers/jaegerremote/sampler.go
```diff
@@ -39,25 +38,20 @@ const (
...
-// without generating a new random number, but simply calculating if traceID < (samplingRate * 2^63).
+// samplingRate, in the range between 0.0 and 1.0. it utilizes the SDK `trace.TraceIDRatioBased` sampler.
 func newProbabilisticSampler(samplingRate float64) *probabilisticSampler {
 	s := new(probabilisticSampler)
 	return s.init(samplingRate)
 }
 
 func (s *probabilisticSampler) init(samplingRate float64) *probabilisticSampler {
+	s.sampler = trace.TraceIDRatioBased(samplingRate)
 	s.samplingRate = math.Max(0.0, math.Min(samplingRate, 1.0))
```
yurishkuro, line 54:
```suggestion
	s.samplingRate = math.Max(0.0, math.Min(samplingRate, 1.0))
	s.sampler = trace.TraceIDRatioBased(s.samplingRate)
```

samplers/jaegerremote/sampler.go
```diff
@@ -67,24 +61,13 @@ func (s *probabilisticSampler) SamplingRate() float64 {
...
-		Tracestate: psc.TraceState(),
-	}
+	return s.sampler.ShouldSample(p)
 }
 
 // Equal compares with another sampler.
 func (s *probabilisticSampler) Equal(other trace.Sampler) bool {
 	if o, ok := other.(*probabilisticSampler); ok {
-		return s.samplingBoundary == o.samplingBoundary
+		return s.samplingRate == o.samplingRate
```
yurishkuro, line 70: float comparison should be done with an epsilon

diurnalist: ah yes, thanks for the reminder 🤦🏻

yurishkuro (approved):

CHANGELOG.md
```diff
@@ -8,6 +8,10 @@ This project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0.htm
 
 ## [Unreleased]
 
+### Changed
+
+- Jaeger remote sampler's probabilistic strategy now uses the same sampling algorithm as TraceIDRatioBased. (#6892)
```
```suggestion
- Jaeger remote sampler's probabilistic strategy now uses the same sampling algorithm as `trace.TraceIDRatioBased` in `go.opentelemetry.io/contrib/samplers/jaegerremote`. (#6892)
```

dmathieu (approved):

PR #6899, opened 2025-03-07
Switch the semconv migration default to new

dmathieu:
This switches the semantic conventions migration default from the old version to be the new one, and keeps the possibility to use `http/dup` to get both.

Note: we previously tested and documented an `old` config.
That worked because it used the default. However, that behavior is not documented in the [semconv migration spec](https://github.com/open-telemetry/semantic-conventions/blob/main/docs/http/README.md), and allowing to keep using the old values only when the next release will entirely drop their support looks like a bad idea.

This also removes the `v1.20.0_test.go` files. Those tests rely on the global server/client, but only tests the old behavior.
We can't run the old behavior only anymore, so running those would require a non-trivial refactoring that seems unneeded since this entire code will be removed in the next release.

Closes https://github.com/open-telemetry/opentelemetry-go-contrib/issues/6897


CHANGELOG.md
```diff
@@ -15,6 +15,15 @@ This project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0.htm
...
 
 - Jaeger remote sampler's probabilistic strategy now uses the same sampling algorithm as `trace.TraceIDRatioBased` in `go.opentelemetry.io/contrib/samplers/jaegerremote`. (#6892)
+- Switched the default for `OTEL_SEMCONV_STABILITY_OPT_IN` to emit the v1.26.0 semantic conventions by default in the following modules. (#6899)
+  - `go.opentelemetry.io/contrib/instrumentation/github.com/emicklei/go-restful/otelrestful`
+  - `go.opentelemetry.io/contrib/instrumentation/github.com/gin-gonic/gin/otelgin`
+  - `go.opentelemetry.io/contrib/instrumentation/github.com/gorilla/mux/otelmux`
+  - `go.opentelemetry.io/contrib/instrumentation/github.com/labstack/echo/otelecho`
+  - `go.opentelemetry.io/contrib/instrumentation/net/http/httptrace/otelhttptrace`
+  - `go.opentelemetry.io/contrib/instrumentation/net/http/otelhttp`
+  The `OTEL_SEMCONV_STABILITY_OPT_IN=http/dup` environment variable can be still used to emit both the v1.20.0 and v1.26.0 semantic conventions.
```
MrAlias, line 25: This renders with the last bullet point without a line break.

```suggestion

  The `OTEL_SEMCONV_STABILITY_OPT_IN=http/dup` environment variable can be still used to emit both the v1.20.0 and v1.26.0 semantic conventions.
```

MrAlias (approved):

AyseGokmen (commented):
The new convention `http.client.request.duration.seconds` might still be recording the metrics in milliseconds and with the histogram buckets that would fit the milliseconds use case. Can we please update them as well when switching the default to the new convention?

dmathieu:
Thank you. I have extracted your report into https://github.com/open-telemetry/opentelemetry-go-contrib/issues/6941

akats7 (approved):

CHANGELOG.md
```diff
@@ -20,6 +20,16 @@ This project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0.htm
...
 - Jaeger remote sampler's probabilistic strategy now uses the same sampling algorithm as `trace.TraceIDRatioBased` in `go.opentelemetry.io/contrib/samplers/jaegerremote`. (#6892)
+- Switched the default for `OTEL_SEMCONV_STABILITY_OPT_IN` to emit the v1.26.0 semantic conventions by default in the following modules. (#6899)
+  - `go.opentelemetry.io/contrib/instrumentation/github.com/emicklei/go-restful/otelrestful`
+  - `go.opentelemetry.io/contrib/instrumentation/github.com/gin-gonic/gin/otelgin`
+  - `go.opentelemetry.io/contrib/instrumentation/github.com/gorilla/mux/otelmux`
+  - `go.opentelemetry.io/contrib/instrumentation/github.com/labstack/echo/otelecho`
+  - `go.opentelemetry.io/contrib/instrumentation/net/http/httptrace/otelhttptrace`
+  - `go.opentelemetry.io/contrib/instrumentation/net/http/otelhttp`
+
+  The `OTEL_SEMCONV_STABILITY_OPT_IN=http/dup` environment variable can be still used to emit both the v1.20.0 and v1.26.0 semantic conventions.
```
pellared, line 31: Do I understand correctly that with this change it is NOT possible to emit only the 1.20.0 semantic conventions? Basically, this makes `OTEL_SEMCONV_STABILITY_OPT_IN=http` as default, right?

If so then, I think we should explicitly call it out. Especially, given it looks to be against https://opentelemetry.io/docs/specs/semconv/http/

pellared: > However, that behavior is not documented in the [semconv migration spec](https://github.com/open-telemetry/semantic-conventions/blob/main/docs/http/README.md),

Maybe, it should be documented? 

@open-telemetry/semconv-http-approvers, any recommendations?

dmathieu: Since we're only keeping this behavior for one release, I'm not sure allowing to emit only the 1.20.0 semantic conventions is a good idea. Folks need to migrate, and keeping the old behavior would let them stall.

For folks that ignored this, after the next release, they will automatically get the new semconvs.
If that breaks their behavior, they can get both while they migrate away their dashboards and alerts.

dmathieu: I've added a specific mention.

pellared (approved):

MrAlias:
@XSAM PTAL

XSAM (approved):

Issue #6193, opened 2024-10-04
Automatic route/pattern naming for `http.ServeMux`


nadiamoe:
### Problem Statement

Go recently added more sophisticated pattern matching to `http.ServeMux`, including matching methods and path variables. I think this patterns can make good span names: They are descriptive, but also bounded (unlike request paths).

Furthermore, in Go 1.23, `http.ServeMux` [populates `http.Request.Pattern`](https://github.com/golang/go/issues/66405), making the pattern that matched a request available for handlers and, more importantly, middlewares. It would be great if `otelhttp` provided, out of the box, an option to automatically name spans from `http.ServeMux` patterns.

### Proposed Solution

A middleware can be used to name spans using the mux pattern that matched:

```go
// spanNameFromPattern is a simple middleware that sets the name of the span in the request context to the pattern used
// to match this request.
func spanNameFromPattern(next http.Handler) http.Handler {
	return http.HandlerFunc(func(w http.ResponseWriter, r *http.Request) {
		// Call handler first, so http.ServeMux can populate r.Pattern
		next.ServeHTTP(w, r)
		// Set span name after the fact. As long as this middleware is used within otelhttp.Handler, the span should
		// still be open and thus renameable.
		trace.SpanFromContext(r.Context()).SetName(r.Pattern)
	})
}
```

I'm currently using this in this fashion:

```go
handler = otelhttp.NewHandler(
	// Instead of using a span name formatter, this middleware sets the span name using the mux pattern.
	spanNameFromPattern(handler),
	"http",
	// More options...
)
```

I think this could be a nice addition to the `otelhttp` package, so users don't need to write `spanNameFromPattern` on their own. We could fashion it as an option, which transparently adds this middleware:

```go
handler = otelhttp.NewHandler(
	handler,
	"http",
	otelhttp.WithSpanNameFromMuxPattern(), // Tentative name.
)
```

Or perhaps expose it as a middleware itself:

```go
handler = otelhttp.NewHandler(
	otelhttp.SpanNamer(handler), // VERY tentative name.
	"http",
)
```

Furthermore, I think some more sophistication could be added to the middleware, for example checking if the pattern already includes a method, and add it if it doesn't.

#### Alternatives

Have users write this middleware themselves, or set span names manually in application logic.

### Additional Context

I'd be happy to make a PR to `otelhttp` if you think this can be useful to have in there :)

dmathieu:
[...]

PR #6905, opened 2025-03-10
internal/shared/semconv: set http.route from Request.Pattern

axw:
If Request.Pattern is set (e.g. by net/http.ServeMux), then use it for setting the http.route attribute on server request spans.

Related to https://github.com/open-telemetry/opentelemetry-go-contrib/issues/6193

(But does not close that, since according to https://opentelemetry.io/docs/specs/semconv/http/http-spans/#name we SHOULD use `http.route` in the span name. I'm leaving that for another day.)


dmathieu:
semconvutil is for the 1.20.0 semantic conventions. It will disappear on the next release.
Don't you want to add that to `internal/shared/semconv` instead?

axw:
@dmathieu sounds like it - I wasn't aware of that :) I'll have a look tomorrow.

axw:
@dmathieu updated, I hope I'm doing the right thing now.

instrumentation/github.com/emicklei/go-restful/otelrestful/internal/semconv/httpconv_test.go
```diff
@@ -256,3 +258,18 @@ func TestCurrentHttpClient_MetricAttributes(t *testing.T) {
...
 	}
 }
+
+func TestRequestTraceAttrs_HTTPRoute(t *testing.T) {
+	req := httptest.NewRequest("GET", "/high/cardinality/path/abc123", nil)
+	req.Pattern = "/high/cardinality/path/{id}"
+
+	var found bool
+	for _, attr := range (CurrentHTTPServer{}).RequestTraceAttrs("", req) {
+		if attr.Key != semconvNew.HTTPRouteKey {
```
dmathieu, line 268: Nit: We should use static strings in tests.

CHANGELOG.md
```diff
@@ -8,6 +8,10 @@ This project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0.htm
 
 ## [Unreleased]
 
+### Added
+
+- `http.route` attribute to otelhttp server request spans, when `net/http.Request.Pattern` is set (e.g. typically by ServeMux)
```
```suggestion
- `http.route` attribute to otelhttp server request spans, when `net/http.Request.Pattern` is set in `go.opentelemetry.io/contrib/instrumentation/github.com/emicklei/go-restful/otelrestful`, `go.opentelemetry.io/contrib/instrumentation/github.com/gin-gonic/gin/otelgin`, `go.opentelemetry.io/contrib/instrumentation/github.com/gorilla/mux/otelmux`, `go.opentelemetry.io/contrib/instrumentation/github.com/labstack/echo/otelecho` and `go.opentelemetry.io/contrib/instrumentation/net/http/otelhttp`. (#6905)

instrumentation/github.com/emicklei/go-restful/otelrestful/internal/semconv/httpconv_test.go
axw: Thanks, much better.

dmathieu (approved):

akats7 (approved):
Looks like most of the middleware's are setting this value with their internal equivalent, in which case the order that the attributes are passed in the middleware matters, looks fine but just something to be aware of.

axw:
Thanks for pointing that out @akats7. I've checked all the in-tree instrumentation and the instrumentation-specific route attribute will take precedence.

dmathieu:
Not for this PR, but we should analyze the frameworks that have those instrumentation-specific route, and see if they could be replace to setting the request's Pattern.

PR #6913, opened 2025-03-11
otelhttp: Ignore informational header when persisting status

### LINKED ISSUE #6912: `RespWriterWrapper` incorrectly handles `100 Continue` response
### Description

`100 Continue` response can be issued many times before an actual response is return (eg. `200 OK`). In case of `otelhttp` these responses should be ignored.

When you look at the `http/server.go`: https://github.com/golang/go/blob/go1.24.1/src/net/http/server.go#L1212-L1228 you can notice that they only set `wroteHeader` variable or `status` when it is not informational header.
But here https://github.com/open-telemetry/opentelemetry-go-contrib/blob/main/instrumentation/net/http/otelhttp/internal/request/resp_writer_wrapper.go#L62-L84 we don't filter `100 Continue` response.

Therefore, when `100 Continue` is set and then there is no explicit `200 OK` reported, then final status code will be `100` and not `200` as it should be.

### Environment

- OS: MacOS
- Architecture: arch64
- Go Version: 1.24
- `otelhttp` version: v0.56.0

### Steps To Reproduce

1. Use following handler:
    ```go
    	targetHandler := http.HandlerFunc(func(w http.ResponseWriter, r *http.Request) {
    		w.WriteHeader(http.StatusContinue) // Explicitly inform client that we are ready for body.
    		
    		body, _ := io.ReadAll(r.Body)
    		w.Write(body)
    	})
    	http.Handle("/target", otelhttp.NewHandler(targetHandler, "target"))
    ```
1. Run the code.
1. Handler will report `100 Continue` header as final status code instead of `200 OK`.

### Expected behavior

`200 OK` status code to be reported even when there is explicit `100 Continue` done in the handler.

Likely we need to do similar thing that they do in `http/server.go` which is to simply ignore informational headers if they are issued before a regular response.

VirrageS:
Closes #6912


dmathieu:
Yes, please

instrumentation/net/http/otelhttp/internal/request/resp_writer_wrapper.go
```diff
@@ -61,7 +61,7 @@ func (w *RespWriterWrapper) Write(p []byte) (int, error) {
 
 // WriteHeader persists initial statusCode for span attribution.
 // All calls to WriteHeader will be propagated to the underlying ResponseWriter
-// and will persist the statusCode from the first call.
+// and will persist the statusCode from the first call (except for informational response status codes).
```
dmathieu, line 64: This code is auto-generated (see the top of the file).
You need to modify the template and run `make generate` locally.
https://github.com/open-telemetry/opentelemetry-go-contrib/tree/main/internal/shared/request

VirrageS: Good catch, don't know how I've missed that. Fixed and regenerated with `make generate`.

CHANGELOG.md
```diff
@@ -11,6 +11,7 @@ This project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0.htm
 ### Changed
 
 - Jaeger remote sampler's probabilistic strategy now uses the same sampling algorithm as `trace.TraceIDRatioBased` in `go.opentelemetry.io/contrib/samplers/jaegerremote`. (#6892)
+- Ignore informational response status codes (`100-199`) when processing the `writeHeader` call in `go.opentelemetry.io/contrib/instrumentation/net/http/otelhttp` and `go.opentelemetry.io/contrib/instrumentation/github.com/gorilla/mux/otelmux`. (#6913)
```
dmathieu, line 14:
```suggestion
- Ignore informational response status codes (`100-199`) when storing the HTTP status code in `go.opentelemetry.io/contrib/instrumentation/net/http/otelhttp` and `go.opentelemetry.io/contrib/instrumentation/github.com/gorilla/mux/otelmux`. (#6913)
```

Changelog entries should be user-facing. `writeHeader` is an internal method.

dmathieu (approved):

pellared:

dmathieu:
Yes, this is also still valid.

CHANGELOG.md
```diff
@@ -8,6 +8,10 @@ This project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0.htm
 
 ## [Unreleased]
 
+### Fixed
+
+- Ignore informational response status codes (`100-199`) when storing the HTTP status code in `go.opentelemetry.io/contrib/instrumentation/net/http/otelhttp` and `go.opentelemetry.io/contrib/instrumentation/github.com/gorilla/mux/otelmux`. (#6913)
```
pellared: @VirrageS, WDYT?

```suggestion
- Ignore informational response status codes (`100-199`) except `101 Switching Protocols` when storing the HTTP status code in `go.opentelemetry.io/contrib/instrumentation/net/http/otelhttp` and `go.opentelemetry.io/contrib/instrumentation/github.com/gorilla/mux/otelmux`. (#6913)
```

pellared:
@VirrageS, thank you for your contribution!

PR #6914, opened 2025-03-11
otelhttp: Remove custom wrapper after handling request

### LINKED ISSUE #6908: `otelhttp.NewHandler` body wrapper alters default behaviour of `Expect: 100-continue` header handling
### Description

`otelhttp.NewHandler` adds extra body wrapper on top of `r.Body`: https://github.com/open-telemetry/opentelemetry-go-contrib/blob/main/instrumentation/net/http/otelhttp/handler.go#L138-L144. This changes default behavior of handling `Expect: 100-continue` header by Go HTTP server.

Here: https://github.com/golang/go/blob/go1.24.1/src/net/http/server.go#L1418-L1420 since `r.Body` is not the `expectContinueReader` (it is our custom wrapper) this means that we won't trigger close after reply so instead of that we will go here: https://github.com/golang/go/blob/go1.24.1/src/net/http/server.go#L1461-L1462 and we will wait for the body without sending `100 Continue` response back (which means that client won't send us data until the specified timeout and thus whole operations takes seconds instead of milliseconds).

### Environment

- OS: MacOS
- Architecture: arch64
- Go Version: 1.24
- `otelhttp` version: v0.56.0

### Steps To Reproduce

1. This is toy example on the problem itself:
    ```go
    package main
    
    import (
    	"bytes"
    	"crypto/rand"
    	"fmt"
    	"io"
    	"net/http"
    	"time"
    )
    
    type customBodyWrapper struct {
    	rc io.ReadCloser
    }
    
    func (cbw customBodyWrapper) Read(p []byte) (n int, err error) {
    	return cbw.rc.Read(p)
    }
    
    func (cbw customBodyWrapper) Close() error {
    	return cbw.rc.Close()
    }
    
    func startRedirectServer() {
    	http.HandleFunc("/", func(w http.ResponseWriter, r *http.Request) {
    		r.Body = customBodyWrapper{rc: r.Body} // <- THIS IS THE PROBLEM. REMOVING THIS FIXES IT.
    		http.Redirect(w, r, "http://localhost:8082/target", http.StatusTemporaryRedirect)
    	})
    
    	fmt.Println("Redirect server listening on :8081")
    	go func() {
    		http.ListenAndServe(":8081", nil)
    	}()
    }
    
    func startTargetServer() {
    	http.HandleFunc("/target", func(w http.ResponseWriter, r *http.Request) {
    		body, _ := io.ReadAll(r.Body)
    		w.Write(body)
    	})
    
    	fmt.Println("Target server listening on :8082")
    	go func() {
    		http.ListenAndServe(":8082", nil)
    	}()
    }
    
    func sendClientRequest() {
    	client := &http.Client{
    		Transport: &http.Transport{
    			ExpectContinueTimeout: 5 * time.Second,
    		},
    	}
    
    	body := make([]byte, 5*1024*1024)
    	rand.Read(body)
    	
    	req, _ := http.NewRequest("PUT", "http://localhost:8081/", bytes.NewReader(body))
    	req.Header.Set("Expect", "100-continue")
    	req.ContentLength = int64(len(body))
    
    	resp, _ := client.Do(req)
    	defer resp.Body.Close()
    	responseBody, _ := io.ReadAll(resp.Body)
    	fmt.Printf("Code: %d, Response: %d\n", resp.StatusCode, len(responseBody))
    }
    
    func main() {
    	startRedirectServer()
    	startTargetServer()
    	time.Sleep(time.Second)
    	fmt.Println("")
    
    	sendClientRequest()
    }
    ```
1. Run this file.
1. See how long it takes with and without custom wrapper.

### Expected behavior

No extra delay when handling `Expect: 100-continue` headers.

### Further notes

In my opinion Go handling of this header is not ideal. Expecting specific type might not be a good idea... Not sure if we should instead file a ticket in Go to improve this...

... but in the meantime one of the thing that `otelhttp` library could do is something like:
```diff
	bw := request.NewBodyWrapper(r.Body, readRecordFunc)
	if r.Body != nil && r.Body != http.NoBody {
+		prevBody := r.Body
		r.Body = bw
+		defer func() { r.Body = prevBody }
	}
```
But not sure if this actually works properly and if it doesn't breaking something else instead :D

VirrageS:
Closes #6908


dmathieu:

VirrageS:
> Should we do this every time we use the body wrapper? https://github.com/search?q=repo%3Aopen-telemetry%2Fopentelemetry-go-contrib%20NewBodyWrapper&type=code

I wasn't sure in the beginning as I didn't know `gorilla/mux` that much. But from what I see it is simple muxer. So yes, I think we should update it too.

dmathieu:
Could you do that in this PR?

VirrageS:
Yeah, it is done already :)

CHANGELOG.md
```diff
@@ -11,6 +11,7 @@ This project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0.htm
 ### Changed
 
 - Jaeger remote sampler's probabilistic strategy now uses the same sampling algorithm as `trace.TraceIDRatioBased` in `go.opentelemetry.io/contrib/samplers/jaegerremote`. (#6892)
+- Remove `request.BodyWrapper` from `r.Body` after the request is processed to avoid issues with handling the `r.Body` by `http/server.go` (which checks the type of `r.Body`) in `go.opentelemetry.io/contrib/instrumentation/net/http/otelhttp` and `go.opentelemetry.io/contrib/instrumentation/github.com/gorilla/mux/otelmux`. (#6914)
```
dmathieu, line 14: Changelog entries should focus on the end-user impact, not internal method names.

```suggestion
- Remove the custom body wrapper from the request's body after the request is processed to allow body type comparisons with the original type `go.opentelemetry.io/contrib/instrumentation/net/http/otelhttp` and `go.opentelemetry.io/contrib/instrumentation/github.com/gorilla/mux/otelmux`. (#6914)
```

instrumentation/github.com/gorilla/mux/otelmux/mux.go
```diff
@@ -138,7 +138,12 @@ func (tw traceware) ServeHTTP(w http.ResponseWriter, r *http.Request) {
 	// ReadCloser fulfills a certain interface and it is indeed nil or NoBody.
 	bw := request.NewBodyWrapper(r.Body, readRecordFunc)
 	if r.Body != nil && r.Body != http.NoBody {
+		prevBody := r.Body
 		r.Body = bw
+
+		// Restore the original body after the request is processed to avoid issues
+		// with extra wrapper since `http/server.go` later checks type of `r.Body`.
+		defer func() { r.Body = prevBody }()
```
dmathieu, line 146: This code is missing tests.

dmathieu (approved):

pellared:

dmathieu:
I believe this is still valid, yes.

pellared:
@VirrageS, can you please sync with main and resolve the conflicts?

VirrageS:
@pellared done

dmathieu:
The failing tests seem to be due to your changes.

VirrageS:
Yeah, seems like the code changed since the last time I've touched it. Everything should pass now.

CHANGELOG.md
```diff
@@ -402,6 +402,7 @@ The next release will require at least [Go 1.24].
   - All `net.peer.*` and `net.host.*` attributes are now set to correct `server.*` attributes.
   - No `net.socket.*` attributes are set.
 - Only sample spans when `Sampled=1` in `go.opentelemetry.io/contrib/propagators/aws/xray`. (#7318)
+- Remove the custom body wrapper from the request's body after the request is processed to allow body type comparisons with the original type in `go.opentelemetry.io/contrib/instrumentation/net/http/otelhttp` and `go.opentelemetry.io/contrib/instrumentation/github.com/gorilla/mux/otelmux`. (#6914)
```
dmathieu: Could you move this higher up, within the unreleased section?

VirrageS: Hah, yes, ofc. I'm surprised that I haven't noticed this before. Thanks!

pellared (approved):

pellared:
@VirrageS, thanks for your contribution!

ash2k:
I suspect this might cause a data race: some handlers may spawn goroutines that access `r.Body` concurrently (read the field itself). I think a better approach might be to clone the `Request` object and pass that to the handler with a wrapped body. I.e. do not mutate the original object. Would that work?

PR #6937, opened 2025-03-15
internal/shared/semconv: remove redundant HTTP method from http.route

gaiaz-iusipov:
`[http.Request.Pattern]` may contain the name of the HTTP method, which is redundant in the value of the `http.route` attribute.

Related to https://github.com/open-telemetry/opentelemetry-go-contrib/issues/6193


instrumentation/github.com/emicklei/go-restful/otelrestful/internal/semconv/httpconv.go
```diff
@@ -124,7 +124,7 @@ func (n CurrentHTTPServer) RequestTraceAttrs(server string, req *http.Request) [
 		}
 	}
 
-	if useragent := req.UserAgent(); useragent != "" {
```
gaiaz-iusipov, line 127: The call to `req.UserAgent()` is duplicated here

instrumentation/github.com/emicklei/go-restful/otelrestful/internal/semconv/util.go
```diff
@@ -70,12 +70,19 @@ func requiredHTTPPort(https bool, port int) int { // nolint:revive
 }
 
 func serverClientIP(xForwardedFor string) string {
-	if idx := strings.Index(xForwardedFor, ","); idx >= 0 {
```
gaiaz-iusipov, line 73: It's better to use `strings.IndexByte` directly because `strings.Index` will call this function if a single-character string is passed

dmathieu: This is unrelated to the original subject of your PR. Could you split it into another one?

instrumentation/github.com/emicklei/go-restful/otelrestful/internal/semconv/httpconv.go
dmathieu: Please split this into another PR.

instrumentation/github.com/gin-gonic/gin/otelgin/internal/semconv/httpconv_test.go
```diff
@@ -259,16 +259,17 @@ func TestCurrentHttpClient_MetricAttributes(t *testing.T) {
 }
 
 func TestRequestTraceAttrs_HTTPRoute(t *testing.T) {
-	req := httptest.NewRequest("GET", "/high/cardinality/path/abc123", nil)
-	req.Pattern = "/high/cardinality/path/{id}"
+	req := httptest.NewRequest(http.MethodGet, "/path/abc123", nil)
+	req.Pattern = "GET /path/{id}"
```
dmathieu, line 263: We should test both a pattern with the method, and one without it.

akats7:
This is a good call, however this also leads to the question what about hostnames, since 1.22 also adds support for matching on the hostname. Do we want to filter that out of the route as well?

dmathieu:
Could you please add new commits rather than force pushing ?

gaiaz-iusipov:
> Could you please add new commits rather than force pushing ?

Sorry, from now on I will only push new commits.

gaiaz-iusipov:
> This is a good call, however this also leads to the question what about hostnames, since 1.22 also adds support for matching on the hostname. Do we want to filter that out of the route as well?

Thank you, I have added a test case with a domain in the string.

dmathieu (approved):

akats7 (approved):

PR #6938, opened 2025-03-16
otelhttp: Correctly set `url.scheme` attr in http client metrics.

wzy9607:
The current method of deciding `url.scheme` by `req.TLS` is incorrect, because `req.TLS` is ignored by `http.Client`.
<https://pkg.go.dev/net/http#Request>
```go
	// This field is ignored by the HTTP client.
	TLS *tls.ConnectionState
```
Therfore the current method will consider every request as http request, unless user manually set the otherwise useless `req.TLS` field.


instrumentation/github.com/emicklei/go-restful/otelrestful/internal/semconv/httpconv.go
```diff
@@ -510,7 +510,7 @@ func (n CurrentHTTPClient) MetricAttributes(req *http.Request, statusCode int, a
 	attributes = append(attributes,
 		semconvNew.HTTPRequestMethodKey.String(standardizeHTTPMethod(req.Method)),
 		semconvNew.ServerAddress(requestHost),
-		n.scheme(req.TLS != nil),
+		n.scheme(req.URL),
```
dmathieu, line 513: How about passing `req` here, so the method can fallback to the `req.TLS` value, and then `http` if the URL isn't set?

akats7: If we're passing the request directly we do also get the ability to account for proxies by checking the headers. Also leads to the question though of if a proxy or alb terminates TLS, whats the correct scheme to set.

dmathieu: > Also leads to the question though of if a proxy or alb terminates TLS, whats the correct scheme to set.

Let's leave that out for now. The [semantic conventions](https://github.com/open-telemetry/semantic-conventions/blob/main/docs/http/http-spans.md) also don't answer this clearly.

I have opened https://github.com/open-telemetry/semantic-conventions/issues/2009

wzy9607: I have added fallback to `req.TLS`.

instrumentation/github.com/emicklei/go-restful/otelrestful/internal/semconv/httpconv.go
```diff
@@ -558,11 +558,17 @@ func (n CurrentHTTPClient) TraceAttributes(host string) []attribute.KeyValue {
 	}
 }
 
-func (n CurrentHTTPClient) scheme(https bool) attribute.KeyValue { // nolint:revive
-	if https {
-		return semconvNew.URLScheme("https")
+func (n CurrentHTTPClient) scheme(req *http.Request) attribute.KeyValue {
+	if req.URL == nil {
```
dmathieu, line 562: This way of doing the condition looks weird. How about:

```golang
if req.URL != nil && req.URL.Scheme != "" {
	return semconvNew.URLScheme(req.URL.Scheme)
}
if req.TLS != nil {
	return semconvNew.URLScheme("https")
}
return semconvNew.URLScheme("http")
```

wzy9607: done

flc1125 (approved):

dmathieu (approved):

CHANGELOG.md
```diff
@@ -45,6 +45,7 @@ This project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0.htm
 - Improve performance by reducing allocations in the gRPC stats handler in `go.opentelemetry.io/contrib/instrumentation/google.golang.org/grpc/otelgrpc`. (#7186)
 - Update `http.route` attribute to support `request.Pattern` in `go.opentelemetry.io/contrib/instrumentation/github.com/gorilla/mux/otelmux`. (#7108)
 - Change the default span name to be `GET /path` so it complies with the HTTP semantic conventions in `go.opentelemetry.io/contrib/instrumentation/github.com/gin-gonic/gin/otelgin`. (#6381)
+- Set `url.Scheme` attribute to the request URL.Scheme when possible for http client metrics in `go.opentelemetry.io/contrib/instrumentation/net/http/otelhttp`. (#6938)
```
pellared, line 48:
```suggestion
- Set `url.scheme` attribute to the request URL.Scheme when possible for HTTP client metrics in `go.opentelemetry.io/contrib/instrumentation/net/http/otelhttp`. (#6938)
```

pellared (approved):

CHANGELOG.md
wzy9607: fixed

MrAlias (approved):

PR #6942, opened 2025-03-17
Turn milliseconds into seconds for new semconv latency histogram

### LINKED ISSUE #6941: semconv: Durations are reported in milliseconds, when they should be in seconds
The new convention `http.client.request.duration.seconds` might still be recording the metrics in milliseconds and with the histogram buckets that would fit the milliseconds use case. Can we please update them as well when switching the default to the new convention?

_Originally posted by @AyseGokmen in https://github.com/open-telemetry/opentelemetry-go-contrib/pull/6899#pullrequestreview-2689946152_

dmathieu:
This histogram expects seconds. But, since 1.20.0 expected milliseconds, that's the unit we use in `ElapsedTime`. So we need to convert them.

See https://github.com/open-telemetry/semantic-conventions/blob/6533b8a39e03e6925e080d5ca39234035cf87e70/docs/non-normative/http-migration.md#http-client-duration-metric

Closes #6941


dmathieu:
cc @AyseGokmen

AyseGokmen (approved):

akats7 (approved):

PR #6943, opened 2025-03-17
Fix close stale value to be relative

dmathieu:
It appears the close value is relative to the issue being marked as stale.
Which makes sense. Marking as stale updates the issue, so we'd have to wait for 2 more years before closing.

See https://github.com/actions/stale?tab=readme-ov-file#days-before-close
And https://github.com/open-telemetry/opentelemetry-go-contrib/actions/runs/13893660912/job/38869749363


pellared (approved):

dmathieu:
Merging as a trivial PR

MrAlias (approved):

PR #6951, opened 2025-03-18
otelrestful: use semconv package for request attributes and span status

dmathieu:
We missed this use of semconvutil which we need to remove to be able to finalize the semconv migration.

I chose not to add a changelog entry here because there is technically no visible change, since the default is still to use the old semantic conventions only.


pellared (approved):

dmathieu:
cc @XSAM for review.

XSAM (approved):

Issue #6919, opened 2025-03-12
Clean Route setup in HTTP instrumentations


dmathieu:
Now that we rely on the `request.Pattern` attribute to set the route attribute automatically (https://github.com/open-telemetry/opentelemetry-go-contrib/pull/6905), we should analyze all HTTP instrumentations that also set that attribute, and ensure the code can't be cleaned up to avoid repetition.

arshukla98:
Hi @dmathieu , I would like to take up this issue.

dmathieu:
This issue should actually bring multiple PRs, one for each instrumentation. I can add you to one of them. Which one would you like?

arshukla98:
Hi @dmathieu , I attempted to execute the code to check if the `http.route` attribute is being set, determine when it happens, and debug the process using the VSCode Debugger.

```
package main

import (
	"context"
	"fmt"
	"log"
	"net/http"

	"github.com/gin-gonic/gin"
	"go.opentelemetry.io/contrib/instrumentation/github.com/gin-gonic/gin/otelgin"
	"go.opentelemetry.io/otel"
	"go.opentelemetry.io/otel/exporters/stdout/stdouttrace"
	"go.opentelemetry.io/otel/sdk/trace"
)

func setupTracer() func() {
	exporter, err := stdouttrace.New(stdouttrace.WithPrettyPrint())
	if err != nil {
		log.Fatalf("Failed to create exporter: %v", err)
	}
	provider := trace.NewTracerProvider(trace.WithBatcher(exporter))
	otel.SetTracerProvider(provider)

	// Return a function to shutdown the provider with context
	return func() {
		ctx := context.Background()
		if err := provider.Shutdown(ctx); err != nil {
			log.Printf("Error shutting down tracer provider: %v", err)
		}
	}
}

func main() {
	cleanup := setupTracer()
	defer cleanup() // Ensure cleanup runs at the end
	
	r := gin.Default()
	r.Use(otelgin.Middleware("otelgin-server"))

	r.GET("/users/:name", func(c *gin.Context) {
		log.Println("Handler called with name:", c.Param("name"), "\nc.Request.Pattern:", c.Request.Pattern)
		c.String(http.StatusOK, "Hello, %s!", c.Param("name"))
	})

	log.Println("Server running on :8080")
	if err := r.Run(":8080"); err != nil {
		log.Fatalf("Failed to start server: %v", err)
	}
	fmt.Println("Server in Action")

}
```

[...]

PR #6980, opened 2025-03-24
Migrate to semconv for otelecho

Oloruntobi1:
Addresses the `otelecho` side comment in this issue:

https://github.com/open-telemetry/opentelemetry-go-contrib/issues/6919


instrumentation/github.com/labstack/echo/otelecho/echo.go
```diff
@@ -61,13 +62,10 @@ func Middleware(service string, opts ...Option) echo.MiddlewareFunc {
 			}()
 			ctx := cfg.Propagators.Extract(savedCtx, propagation.HeaderCarrier(request.Header))
 			opts := []oteltrace.SpanStartOption{
-				oteltrace.WithAttributes(semconvutil.HTTPServerRequest(service, request)...),
+				oteltrace.WithAttributes(hs.RequestTraceAttrs(service, request)...),
+				oteltrace.WithAttributes(hs.Route(c.Path())),
```
dmathieu, line 66: Isn't `Path()` the actual URL path (with dynamic ID) rather than the route?
Also, we already set this attribute automatically based on `request.Pattern`. If otelecho sets that value (which it should if it respects net/http), we don't need this at all.

instrumentation/github.com/labstack/echo/otelecho/echo_test.go
```diff
@@ -117,3 +119,37 @@ func TestSkipper(t *testing.T) {
 	router.ServeHTTP(w, r)
 	assert.Equal(t, http.StatusOK, w.Result().StatusCode, "should call the 'ping' handler") //nolint:bodyclose // False positive for httptest.ResponseRecorder: https://github.com/timakin/bodyclose/issues/59.
 }
+
+func TestHTTPRouteAttribute(t *testing.T) {
```
dmathieu, line 123: Until we bring everything back into the main package, this test should be in the `test` subfolder.

instrumentation/github.com/labstack/echo/otelecho/echo.go
Oloruntobi1: > Isn't `Path()` the actual URL path (with dynamic ID) rather than the route? 

No it isnt. It's the full route pattern. 

> If otelecho sets that value (which it should if it respects net/http), we don't need this at all.

Echo doesn’t set r.Pattern—it’d need a proposal to its maintainers, and that could take a while.

dmathieu: Can we set it ourselves then?

Oloruntobi1: Done.

https://github.com/labstack/echo/issues/2758

instrumentation/github.com/labstack/echo/otelecho/echo.go
```diff
@@ -60,14 +61,12 @@ func Middleware(service string, opts ...Option) echo.MiddlewareFunc {
 				c.SetRequest(request)
 			}()
 			ctx := cfg.Propagators.Extract(savedCtx, propagation.HeaderCarrier(request.Header))
+			request.Pattern = c.Path()
 			opts := []oteltrace.SpanStartOption{
-				oteltrace.WithAttributes(semconvutil.HTTPServerRequest(service, request)...),
+				oteltrace.WithAttributes(hs.RequestTraceAttrs(service, request)...),
+				oteltrace.WithAttributes(hs.Route(request.Pattern)),
```
dmathieu, line 67: We therefore don't need this anymore. The semconv package does that automatically.

```suggestion
```

dmathieu (commented):
You will need to add a changelog entry.

instrumentation/github.com/labstack/echo/otelecho/echo.go
Oloruntobi1: Thought so too but test was failing.

`RequestTraceAttrs(service, request)...)` isnt setting it. 

The `s.duplicate` flag is false here:

```
func (s HTTPServer) RequestTraceAttrs(server string, req *http.Request) []attribute.KeyValue {
	if s.duplicate {
		return append(OldHTTPServer{}.RequestTraceAttrs(server, req), CurrentHTTPServer{}.RequestTraceAttrs(server, req)...)
	}
	return OldHTTPServer{}.RequestTraceAttrs(server, req)
}
```

so it runs the code after the If statement. Which doesnt set it. Anything im missing?

instrumentation/github.com/labstack/echo/otelecho/test/echo_test.go
```diff
@@ -256,3 +256,37 @@ func TestSpanNameFormatter(t *testing.T) {
...
+	assert.Len(t, spans, 1, "expected one span")
+
+	found := false
+	for _, attr := range spans[0].Attributes() {
+		if attr.Key == "http.route" && attr.Value.AsString() == "/users/:id" {
+			found = true
+			break
+		}
+	}
+	assert.True(t, found, "http.route attribute not found or incorrect, got %v", spans[0].Attributes())
```
dmathieu, line 291: This doesn't scale to multiple attributes. If you look at other tests in the repository, you'll see we compare the full slice value right away.

instrumentation/github.com/labstack/echo/otelecho/echo.go
dmathieu: I'd recommend digging through the repository's issues to grasp the full context of this.
Especially the work in this project: https://github.com/orgs/open-telemetry/projects/87

instrumentation/github.com/labstack/echo/otelecho/test/echo_test.go
Oloruntobi1: Looked at the other test and yes this is superfluous. Removed it

dmathieu (approved):

dmathieu (changes requested):
This needs a changelog entry (and to resolve the conflict).

CHANGELOG.md
```diff
@@ -21,6 +21,7 @@ This project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0.htm
 ### Changed
 
 - Jaeger remote sampler's probabilistic strategy now uses the same sampling algorithm as `trace.TraceIDRatioBased` in `go.opentelemetry.io/contrib/samplers/jaegerremote`. (#6892)
+- - Migrated `otelecho` instrumentation to use OpenTelemetry semantic conventions (`semconv`) for HTTP attributes. (#6980)
```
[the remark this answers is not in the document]
Oloruntobi1: Okay. Done.

dmathieu:
Please run the test suite locally to ensure the CI doesn't fail.

Oloruntobi1:
> Please run the test suite locally to ensure the CI doesn't fail.

Yeah just did.

PR #6982, opened 2025-03-25
otelgin: Clean Route setup in otelgin HTTP instrumentation

### LINKED ISSUE #6976: otelgin: Clean Route setup in otelgin HTTP instrumentation
Hi, This issue is extracted from [Issue 6919](https://github.com/open-telemetry/opentelemetry-go-contrib/issues/6919)

arshukla98:
## Description:-

The intention of this PR is to resolve #6976 .

## Testing:-

The output of the following command

```
make test-race test-coverage lint
```

is captured in this pdf [20250401_CI_Test_Results.pdf](https://github.com/user-attachments/files/19546117/20250401_CI_Test_Results.pdf).

### DIFF
--- a/instrumentation/github.com/gin-gonic/gin/otelgin/gin.go
+++ b/instrumentation/github.com/gin-gonic/gin/otelgin/gin.go
@@ -86,9 +86,12 @@ func Middleware(service string, opts ...Option) gin.HandlerFunc {
 			HTTPClientIP: c.ClientIP(),
 		}
 
+		if c.Request.Pattern == "" {
+			c.Request.Pattern = c.FullPath()
+		}
+
 		opts := []oteltrace.SpanStartOption{
 			oteltrace.WithAttributes(hs.RequestTraceAttrs(service, c.Request, requestTraceAttrOpts)...),
-			oteltrace.WithAttributes(hs.Route(c.FullPath())),
 			oteltrace.WithSpanKind(oteltrace.SpanKindServer),
 		}
 		var spanName string


arshukla98:
Hi @dmathieu , I would like to request a PR review.

CHANGELOG.md
```diff
@@ -36,6 +36,7 @@ This project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0.htm
 
 - Drop support for [Go 1.22]. (#6853)
 - The deprecated `go.opentelemetry.io/contrib/config` package is removed, use `go.opentelemetry.io/contrib/otelconf` instead. (#6894)
+- The function call `hs.Route(c.FullPath())` is removed in `go.opentelemetry.io/contrib/instrumentation/github.com/gin-gonic/gin/otelgin/gin.go`. (#6982)
```
dmathieu, line 39: This shouldn't need a changelog entry, as there will be no user-facing change.

instrumentation/github.com/gin-gonic/gin/otelgin/test/gin_test.go
```diff
@@ -169,7 +169,45 @@ func TestTrace200(t *testing.T) {
...
+	assert.Empty(t, span.Events())
+	assert.Equal(t, codes.Unset, span.Status().Code)
+	assert.Empty(t, span.Status().Description)
+}
+
+func TestTrace200_HttpDup(t *testing.T) {
+	sr := tracetest.NewSpanRecorder()
+	provider := sdktrace.NewTracerProvider(sdktrace.WithSpanProcessor(sr))
+
+	t.Setenv("OTEL_SEMCONV_STABILITY_OPT_IN", "http/dup")
```
dmathieu, line 182: Please don't test the http/dup behavior. That's not the purpose of this PR.

instrumentation/github.com/gin-gonic/gin/otelgin/test/gin_test.go
```diff
@@ -169,7 +169,45 @@ func TestTrace200(t *testing.T) {
 	assert.Contains(t, attr, attribute.String("net.host.name", "foobar"))
 	assert.Contains(t, attr, attribute.Int("http.status_code", http.StatusOK))
 	assert.Contains(t, attr, attribute.String("http.method", "GET"))
-	assert.Contains(t, attr, attribute.String("http.route", "/user/:id"))
```
dmathieu, line 172: This assertions shouldn't be removed. We want to rely on `req.Pattern` rather than use the value ourselves. We don't want to remove the behavior.

dmathieu (changes requested):

instrumentation/github.com/gin-gonic/gin/otelgin/test/gin_test.go
```diff
@@ -169,7 +169,9 @@ func TestTrace200(t *testing.T) {
...
+	if r.Pattern != "" {
```
arshukla98, line 172: Hi @dmathieu , I added this condition because I observed that the `req.Pattern` field is an empty string, which means the `http.route` attribute wouldn't be included in the span attributes. 

## Screenshot for Test Debugging
image

I observed that the unit test fails without this condition, as seen in the following [GitHub Actions run](https://github.com/open-telemetry/opentelemetry-go-contrib/actions/runs/14305158374/job/40087367323). The corresponding error logs are provided below.

## Test Error Logs:-

```
go test -timeout 60s -race ./instrumentation/github.com/gin-gonic/gin/otelgin/test/...
--- FAIL: TestTrace200 (0.00s)
    gin_test.go:172: 
        	Error Trace:	/home/runner/work/opentelemetry-go-contrib/opentelemetry-go-contrib/instrumentation/github.com/gin-gonic/gin/otelgin/test/gin_test.go:172
        	Error:      	[]attribute.KeyValue{attribute.KeyValue{Key:"http.method", Value:attribute.Value{vtype:4, numeric:0x0, stringly:"GET", slice:interface {}(nil)}}, attribute.KeyValue{Key:"http.scheme", Value:attribute.Value{vtype:4, numeric:0x0, stringly:"http", slice:interface {}(nil)}}, attribute.KeyValue{Key:"[net.host.name](http://net.host.name/)", Value:attribute.Value{vtype:4, numeric:0x0, stringly:"foobar", slice:interface {}(nil)}}, attribute.KeyValue{Key:"net.sock.peer.addr", Value:attribute.Value{vtype:4, numeric:0x0, stringly:"192.0.2.1", slice:interface {}(nil)}}, attribute.KeyValue{Key:"net.sock.peer.port", Value:attribute.Value{vtype:2, numeric:0x4d2, stringly:"", slice:interface {}(nil)}}, attribute.KeyValue{Key:"http.client_ip", Value:attribute.Value{vtype:4, numeric:0x0, stringly:"192.0.2.1", slice:interface {}(nil)}}, attribute.KeyValue{Key:"[http.target](http://http.target/)", Value:attribute.Value{vtype:4, numeric:0x0, stringly:"/user/123", slice:interface {}(nil)}}, attribute.KeyValue{Key:"net.protocol.version", Value:attribute.Value{vtype:4, numeric:0x0, stringly:"1.1", slice:interface {}(nil)}}, attribute.KeyValue{Key:"http.status_code", Value:attribute.Value{vtype:2, numeric:0xc8, stringly:"", slice:interface {}(nil)}}} 
does not contain attribute.KeyValue{Key:"http.route", Value:attribute.Value{vtype:4, numeric:0x0, stringly:"/user/:id", slice:interface {}(nil)}}
        	Test:       	TestTrace200
FAIL
FAIL	[go.opentelemetry.io/contrib/instrumentation/github.com/gin-gonic/gin/otelgin/test](http://go.opentelemetry.io/contrib/instrumentation/github.com/gin-gonic/gin/otelgin/test)	0.020s
FAIL
make: *** [Makefile:261: test/./instrumentation/github.com/gin-gonic/gin/otelgin/test] Error 1
```

dmathieu: If `req.Pattern` is empty, we should set it. Your condition will never be executed.

akats7:
@dmathieu isn't the default use of Pattern only included in the new semconv version, which is still opt in? Wouldn't this break the default behavior?

dmathieu:
We can add support for the route to semconvutil.

akats7:
> We can add support for the route to semconvutil.

Are we dropping support for semconvutil in 1.36 or 1.37, should we just wait to merge the clean up PRs

dmathieu:
We're dropping semconvutil support in 1.37 (in 1.36, we default to the new semantic conventions).

instrumentation/github.com/gin-gonic/gin/otelgin/test/gin_test.go
arshukla98: Hi @dmathieu, should I set value of c.FullPath() in req.Pattern like this?

image

I am not sure what i should assign to req.Pattern and the location is correct or not?

arshukla98:
@dmathieu, I would like to request for a review.

dmathieu:
> isn't the default use of Pattern only included in the new semconv version, which is still opt in? Wouldn't this break the default behavior?

From what I can see, semconvutil has no support for routes at all.

arshukla98:
Hi @akats7 , I would like to request for a review.

akats7:
@dmathieu something does seem strange about setting the Pattern on the request explicitly

dmathieu:
I agree, let's not do that.
@arshukla98 I'm sorry, that therefore makes your PR unneeded anymore. I'm going to close it.

PR #7067, opened 2025-03-28
TestOT_Fields for ot propagator

derwesx:
1. Verifies Fields() returns correct header names and in the correct order.
2. Cross-check between Fields() and Inject() behavior.

### DIFF
--- a/propagators/ot/ot_propagator_test.go
+++ b/propagators/ot/ot_propagator_test.go
@@ -4,8 +4,11 @@
 package ot
 
 import (
+	"context"
 	"fmt"
 	"testing"
+
+	"go.opentelemetry.io/otel/propagation"
 
 	"github.com/stretchr/testify/assert"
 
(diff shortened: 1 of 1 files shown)


propagators/ot/ot_propagator_test.go
```diff
@@ -133,3 +136,26 @@ func TestOT_Extract(t *testing.T) {
...
 }
+
+func TestOT_Fields(t *testing.T) {
+	propagator := OT{}
+
+	want := []string{traceIDHeader, spanIDHeader, sampledHeader}
+
+	got := propagator.Fields()
+
+	assert.Equal(t, want, got, "Fields()\ngot:  %v\nwant: %v", got, want)
```
MrAlias, line 147: `assert` will format `got` and `want` in the message it logs. No need to duplicate.

```suggestion
	assert.Equal(t, want, got,)
```

MrAlias (commented):
Thanks for the contribution. Looks good overall, only issue is with test message.

propagators/ot/ot_propagator_test.go
derwesx: Good point

image

derwesx:
Fixed and squashed. Ready for checks.

MrAlias (approved):
Thanks for the contribution.

dmathieu (approved):

PR #7108, opened 2025-04-02
feat(otelmux): update `http.route` attribute to support `request.Pattern`

flc1125:
see https://github.com/open-telemetry/opentelemetry-go-contrib/issues/6919


akats7 (approved):
Unrelated to this PR, but based on that missing test coverage, it might be good to add a test case where GetPathTemplate returns an err, maybe using a path prefix or something

flc1125:
> Unrelated to this PR, but based on that missing test coverage, it might be good to add a test case where GetPathTemplate returns an err, maybe using a path prefix or something

I tried to cover this unit test, but because it's somewhat extreme, it's very difficult to trigger the execution that meets this part of the logic for coverage. So, I ignored it for now.

flc1125:
Ready for merge.

Issue #7053, opened 2025-03-27
Test the AWS XRAY propagator extraction


MrAlias:
Test the functionality provided:

https://github.com/open-telemetry/opentelemetry-go-contrib/blob/836e94b12225d049062e58eba916e5ac381591c5/propagators/aws/xray/propagator.go#L79-L89

The [`TestAwsXrayExtract`](https://github.com/open-telemetry/opentelemetry-go-contrib/blob/836e94b12225d049062e58eba916e5ac381591c5/propagators/aws/xray/propagator_test.go#L30) test should be updated to test the public method:

https://github.com/open-telemetry/opentelemetry-go-contrib/blob/836e94b12225d049062e58eba916e5ac381591c5/propagators/aws/xray/propagator_test.go#L79

Gaps in coverage for that test need to be identified and additional tests/test-cases need to be added.

PhilemonBrain:
@MrAlias can you assign this to me?

PR #7184, opened 2025-04-09
Test aws xray extract

PhilemonBrain:
Closes https://github.com/open-telemetry/opentelemetry-go-contrib/issues/7053

# X-Ray Propagator: Add unit tests for Extract method 

Adds comprehensive unit tests for the `Extract` method of the AWS X-Ray propagator to ensure correct handling of trace context extraction from HTTP headers.

The tests verify:
- Proper extraction of valid X-Ray headers with both sampling enabled and disabled
- Graceful handling of missing headers
- Proper error handling for malformed trace IDs
- Validation of trace ID version
- Behavior when required components like parent ID are missing

These tests complement the existing tests for the lower-level `extract` function and ensure the public `Extract` method works correctly within the OpenTelemetry context propagation system.


PhilemonBrain:
closes #7053

PhilemonBrain:
@akats7 @MrAlias could you help me with the "Skip Changelog" label? It appears I do not have rights to add the label

propagators/aws/xray/propagator_test.go
```diff
@@ -127,3 +213,107 @@ func TestPropagatorFields(t *testing.T) {
 	assert.Len(t, propagator.Fields(), 1, "Fields() should return exactly one field")
 	assert.Equal(t, []string{traceHeaderKey}, propagator.Fields())
 }
+
+func TestXrayExtract(t *testing.T) {
```
akats7, line 217: This seems to be a duplicate of what was added to TestAWSXrayExtract

propagators/aws/xray/propagator_test.go
```diff
@@ -28,69 +28,155 @@ var (
...
-		err          error
-	}{
-		{
-			xrayTraceID, parentID64Str, notSampled,
-			trace.SpanContextConfig{
-				TraceID:    traceID,
-				SpanID:     parentSpanID,
-				TraceFlags: traceFlagNone,
+	// Test the public Extract method
+	t.Run("Public Extract Method", func(t *testing.T) {
```
dmathieu, line 32: You're still testing things twice here, as subtests. Which is absolutely not the go way.

akats7: IMO you can probably just remove the tests for the private method entirely? The main functionality of the Public method is handled in the private, why don't you just combine test scenarios

PhilemonBrain: So the tests should be for the path where the helper function `extract` returns an error or sc.IsValid() returns false, right? ensuring that when `extract` fails, `Extract` fails gracefully and doesn't inject a bogus span into the context?

Please bear with me, I'm barely 5 months in Go and a few weeks of open telemetry, lol

PhilemonBrain: Okayy, got it!

PhilemonBrain:
@dmathieu @akats7 

I noticed something that might either be a bug or a misunderstanding on my end - The logic of the `parseTraceFlag` in `propagators/aws/xray/propagator.go`

The current logic implies that if the value is exactly 1 character and not "1" → return None, else → return Sampled. 

So in the case of `Sampled=maybe`, len("maybe") == 5 → not equal to 1, so it goes to the else block, and incorrectly returns trace.FlagsSampled, meaning it treats any non-"1" string of length ≠ 1 as sampled.

Is this how it should be? 

I assumed that only "1" should return → Sampled, while anything else (including "0", "maybe", "foo", etc.) → Not Sampled

If I'm mistaken, I'll adjust my tests accordingly, but please take a look and tell me what you think

PhilemonBrain:
@dmathieu @akats7

Please, this is a gentle reminder for re-review on this PR

propagators/aws/xray/propagator_test.go
```diff
@@ -6,90 +6,137 @@ package xray
...
-			errInvalidTraceIDVersion,
+			name:        "Missing Sampled",
+			headerVal:   "Root=1-abcdef12-1234567890abcdef12345678;Parent=1234567890abcdef",
+			wantValid:   true,
+			wantTraceID: "abcdef121234567890abcdef12345678",
+			wantSpanID:  "1234567890abcdef",
+			wantSampled: false,
+		},
+		// {
+		// 	name:        "Invalid Sampled value",
```
dmathieu, line 79: Why is this commented?

PhilemonBrain: Thats the point i'm making, it seems. It seems the logic of the `parseTraceFlag` in propagators/aws/xray/propagator.go is buggy

PhilemonBrain: The current logic implies that if the value is exactly 1 character and not "1" → return None, else → return Sampled.

So in the case of Sampled=maybe, len("maybe") == 5 → not equal to 1, so it goes to the else block, and incorrectly returns trace.FlagsSampled, meaning it treats any non-"1" string of length ≠ 1 as sampled.

Is this how it should be?

I assumed that only "1" should return → Sampled, while anything else (including "0", "maybe", "foo", etc.) → Not Sampled

If I'm mistaken, I'll adjust my tests accordingly

dmathieu: That makes sense, and does need to be fixed, but in another PR.
Having a commented out test case is quite confusing though. So how about removing it, and opening an issue to fix this after this PR is merged?

PhilemonBrain: Alright, that is better. I would remove the test and add it back in the PR for the bug

dmathieu: Can you open the issue as well?

dmathieu: https://github.com/open-telemetry/opentelemetry-go-contrib/issues/7262

dmathieu (approved):

PR #7192, opened 2025-04-10
otelhttp: handle request Pattern

### LINKED ISSUE #6193: Automatic route/pattern naming for `http.ServeMux`

dmathieu:
Closes #6193
Handles otelhttp for #6919 (@Ananyasinha13 I know you're assigned this, but it has been 2 weeks with no movement).


dmathieu:
This has the downside of running the span name formatter twice if there's a pattern.
I'm open to alternatives, but it would be good if they weren't modifying the public API.

dmathieu:
cc @seankhliao @axw

dmathieu:
I can add a comment to the WithSpanNameFormatter method for better awareness of that.

seankhliao (approved):
It's a cool trick,
it may be a bit fragile if someone uses some other middleware that clones the request (uses WithContext) between otelhttp and the mux, only the new copy will have the Pattern set.

Can we have a spanformatter func or option to use the pattern (a separate pr?)

seankhliao:
sure I can make one

seankhliao:
I was thinking something simple like https://github.com/open-telemetry/opentelemetry-go-contrib/pull/7197/commits/a44c60f65b97e0cda545763fd2f3d1598053ecf6 (currently based on top of this pr).

instrumentation/net/http/otelhttp/config.go
```diff
@@ -176,6 +176,10 @@ func WithMessageEvents(events ...event) Option {
 
 // WithSpanNameFormatter takes a function that will be called on every
 // request and the returned string will become the Span Name.
+//
+// When using `http.ServeMux` (or any middleware that sets `request.Pattern`),
```
pellared, line 180:
```suggestion
// When using [http.ServeMux] (or any middleware that sets the Pattern of [http.Request]),
```

instrumentation/net/http/otelhttp/config.go
```diff
@@ -176,6 +176,10 @@ func WithMessageEvents(events ...event) Option {
...
+// the span name formatter will run twice. Once when the span is created, and
+// once after the middleware, so the pattern can be used.
```
pellared, line 182:
```suggestion
// second time after the middleware, so the pattern can be used.
```

pellared (approved):

MrAlias:
@XSAM PTAL

MrAlias (approved):

dashpole (approved):

XSAM (approved):

PR #7210, opened 2025-04-11
POC: command attributes hook

prestonvasquez:


prestonvasquez:
@dmathieu Does this change seem reasonable? If so, I will clean it up, add tests, etc, and put it in ready for review.

PR #7229, opened 2025-04-15
build: add version check for go mod

flc1125:
This seems to be skippable CHANGELOG.

Usage:

```shell
MINIMUM_GO_VERSION="go 1.23.0" make verify-goversion
```


dmathieu:
This is a nice idea.
I would take a different approach though. In here, the go version is a constant within the test, which means an additional thing to change when we upgrade the Go version.

How about a new tool in `internal/tools` that does the same thing, with the expected Go version taken from an environment variable (or a CLI flag).
Kind of like what we do here: https://github.com/open-telemetry/opentelemetry-go/tree/main/internal/tools/verifyreadmes

Then, we can have a new CI task that uses the proper Go version.
We can add a `MINIMUM_GO_VERSION` in [ci.yml](https://github.com/open-telemetry/opentelemetry-go-contrib/blob/main/.github/workflows/ci.yml#L17). We'll have to update it on upgrades, but it's all in the same file.

flc1125:
I will study how the internal/tools directory operates. I'll first set this PR to draft mode.

dmathieu:
EXPECTED_GO_VERSION seems like a good idea.
Could you add tests to the Go file as well?

sahidvelji (changes requested):
@flc1125 I recommend using the `gomoddirectives` linter for this purpose.
https://golangci-lint.run/usage/linters/#gomoddirectives
https://github.com/ldez/gomoddirectives

```yaml
  settings:
    gomoddirectives:
      toolchain-forbidden: true
      go-version-pattern: "1.23.0"
```

PR #7233, opened 2025-04-16
otelconf: add tests for grpc exporters with different certificate configurations

### LINKED ISSUE #7222: otelconf: Increase test coverage for OTLP-related code
I just looked into how this could be tested but in this case it does not seem to be possible to do any kind of assertions on the internal attributes (i.e. the TLS config) of the created grpc exporters, as those are all unexported - also the current tests seem to only validate the type of the created exporters.
Currently I only see the option of adding a component test where in the tests we start a grpc server with different configurations (e.g. with/without TLS) and try to export logs/traces/metrics. If you'd like me to add this I'm happy to do so though (probably it would be better as a follow up to avoid bumping the complexity of this PR, but either way is fine with me, please let me know what you would prefer)

_Originally posted by @bacherfl in https://github.com/open-telemetry/opentelemetry-go-contrib/pull/6984#discussion_r2041926896_

bacherfl:
This PR adds component tests for the otlp grpc exporters for logs/metrics/traces. The tests work by starting a simple grpc server that implements the `Export` interface of the respective signals, and records what is being sent to it. This should server as a way to verify that the otlp exporters created within this package are able to communicate with the server.

Closes #7222


bacherfl:
@pellared this should be ready for review now - I'm not sure if a changelog entry is needed as this PR only adds tests.
The failing lint check seems to affect files that are not part of this PR, here I'm not sure how to address that

pellared:
> The failing lint check seems to affect files that are not part of this PR, here I'm not sure how to address that

From build failure

> Working tree is not clean, did you forget to run "make precommit"?

Guessing that `go mod tidy` run is missing as `modified:   otelconf/go.mod` :wink:

dmathieu (approved):

otelconf/v0.3.0/metric_test.go
```diff
@@ -1388,3 +1394,188 @@ func TestPrometheusIPv6(t *testing.T) {
...
+												Value: 1,
+											},
+										},
+									},
+								},
+							},
+						},
+					},
+				}))
+			}, 10*time.Second, 1*time.Second)
```
pellared, line 1504: Cannot we call it synchronously?

pellared (commented):
I think this can help you: https://github.com/signalfx/splunk-otel-go/blob/0e047ccc5cc160f2ff4a96e45065e8a3178cccbc/distro/otel_test.go#L763-L849

otelconf/v0.3.0/metric_test.go
bacherfl: theoretically there could be some timing issues, as the server is started in a separate routine. But I will try again with calling this synchronously (need to run it a couple of times to make sure we don't introduce some flakiness)

otelconf/v0.3.0/log_test.go
```diff
@@ -694,3 +704,217 @@ func TestLogProcessor(t *testing.T) {
...
+
+			col, err := newGRPCLogsCollector(n, tlsMode)
+			require.NoError(t, err)
+			defer col.srv.Stop()
+
+			exporter, err := otlpGRPCLogExporter(tt.args.ctx, tt.args.otlpConfig)
+			if tt.wantErr != "" {
+				require.Error(t, err)
+				require.Equal(t, tt.wantErr, err.Error())
+			} else {
```
pellared, line 793: Isn't this unnecessary?

otelconf/v0.3.0/log_test.go
```diff
@@ -694,3 +704,217 @@ func TestLogProcessor(t *testing.T) {
...
+		}
+		caCertPool := x509.NewCertPool()
+		caCertPool.AppendCertsFromPEM(caCert)
+		tlsCreds := credentials.NewTLS(&tls.Config{
+			Certificates: []tls.Certificate{cert},
+			ClientCAs:    caCertPool,
+			ClientAuth:   tls.RequireAndVerifyClientCert,
+		})
+		opts = append(opts, grpc.Creds(tlsCreds))
+	default:
```
pellared, line 899: How about adding an `func() []grpc.ServerOption` field to test cases struct so that we can get rid of conditional logic in tests and add this logic to the test cases?

otelconf/v0.3.0/metric_test.go
```diff
@@ -1388,3 +1394,177 @@ func TestPrometheusIPv6(t *testing.T) {
...
+						},
+					},
+				},
+			}))
+
+			// Ensure everything is flushed.
+			require.NoError(t, exporter.Shutdown(context.Background()))
+
+			// verify the reception of the sent data item
+			require.GreaterOrEqual(t, len(col.storage.data), 1)
```
pellared, line 1509: This is unnecessary. If `exporter.Export` has not returned an error it means that it succeeded. There is no need for `storage`.

otelconf/v0.3.0/metric_test.go
```diff
@@ -1388,3 +1394,177 @@ func TestPrometheusIPv6(t *testing.T) {
...
+			exporter, err := otlpGRPCMetricExporter(tt.args.ctx, tt.args.otlpConfig)
+			if tt.wantErr != "" {
+				require.Error(t, err)
+				require.Equal(t, tt.wantErr, err.Error())
+			} else {
+				require.NoError(t, err)
+			}
+
+			res, err := resource.New(context.Background())
+			require.NoError(t, err)
```
pellared, line 1483: Why?

otelconf/v0.3.0/log_test.go
bacherfl: yes, that was a copy-paste relict a previous test. i have removed that now

otelconf/v0.3.0/log_test.go
bacherfl: i feel like that will add quite some verbosity to the test cases, but I will give it a try

otelconf/v0.3.0/metric_test.go
bacherfl: after removing the `require.Eventually` one of the test runs ended up failing, so I have re-added it again. It seems like it is indeed needed, due to the reason outlined in the comment above

otelconf/v0.3.0/log_test.go
```diff
@@ -694,3 +704,163 @@ func TestLogProcessor(t *testing.T) {
...
+			require.NoError(t, exporter.Shutdown(context.Background()))
+		})
+	}
+}
+
+// grpcLogsCollector is an OTLP gRPC server that collects all requests it receives.
+type grpcLogsCollector struct {
+	collogpb.UnimplementedLogsServiceServer
+
+	listener net.Listener
```
pellared, line 837: This field seems unnecessary.

otelconf/v0.3.0/log_test.go
```diff
@@ -694,3 +704,163 @@ func TestLogProcessor(t *testing.T) {
...
+				return opts, nil
+			},
+		},
+	}
+	for _, tt := range tests {
+		t.Run(tt.name, func(t *testing.T) {
+			n, err := net.Listen("tcp", "localhost:0")
+			require.NoError(t, err)
+
+			tt.args.otlpConfig.Endpoint = ptr(strings.ReplaceAll(n.Addr().String(), "127.0.0.1", "localhost"))
```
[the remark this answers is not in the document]
bacherfl: yes, I added a comment for now indicating that this is a workaround - forgot to include that when I first stumbled over this

otelconf/v0.3.0/log_test.go
bacherfl: removed

otelconf/v0.3.0/log_test.go
```diff
@@ -694,3 +704,163 @@ func TestLogProcessor(t *testing.T) {
...
+			}
+
+			require.EventuallyWithT(t, func(collect *assert.CollectT) {
+				require.NoError(collect, exporter.Export(context.Background(), []sdklog.Record{
+					logFactory.NewRecord(),
+				}))
+			}, 10*time.Second, 1*time.Second)
+
+			// Ensure everything is flushed.
+			require.NoError(t, exporter.Shutdown(context.Background()))
```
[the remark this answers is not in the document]
bacherfl: removed

otelconf/v0.3.0/log_test.go
```diff
@@ -694,3 +704,154 @@ func TestLogProcessor(t *testing.T) {
...
+				return opts, nil
+			},
+		},
+	}
+	for _, tt := range tests {
+		t.Run(tt.name, func(t *testing.T) {
+			n, err := net.Listen("tcp", "localhost:0")
+			require.NoError(t, err)
+
+			// this is a workaround, as providing 127.0.0.1 resulted in an "invalid URI for request" error
```
pellared, line 803: Can you please create an issue for the bug it and reference it here?

pellared (approved):

otelconf/v0.3.0/log_test.go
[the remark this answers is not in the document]
pellared: Can you please create a bug for it?

otelconf/v0.3.0/log_test.go
bacherfl: sure, will do that today

dmathieu:
Can you fix the CI?

otelconf/v0.3.0/log_test.go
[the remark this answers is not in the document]
bacherfl: I just looked into this again in more detail - while passing e.g. `127.0.0.1:4318` does not work here, this does not seem to be an issue when used in a collector config, since the collector makes sure to prepend the scheme to the endpoint, before passing it to the `otelconf` function returning the metric/log/trace readers: https://github.com/open-telemetry/opentelemetry-collector/blob/main/service/telemetry/internal/migration/normalize.go#L8

Therefore, something like the following config can still be used.

```yaml
  telemetry:
    metrics:
      level: detailed
      readers:
        - periodic:
            interval: 5000
            timeout: 5000
            exporter:
              otlp:
                protocol: grpc
                temporality_preference: delta
                endpoint: 127.0.0.1:4317
```

The `invalid URI` error returned there is thus only limited to this particular scenario where we obtain the endpoint address from the listener created with `net.Listen()` - would you still consider this a bug in this case?

bacherfl:
> Can you fix the CI?

fixed

otelconf/v0.3.0/log_test.go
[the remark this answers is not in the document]
pellared: > since the collector makes sure to prepend the scheme to the endpoint

I guess the Collector does not use `otlpGRPCLogExporter` and just uses `otelconf` for parsing YAML and does not use `otelconf.SDK`.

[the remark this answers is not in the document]
bacherfl: ok thanks for clarifying. I have created the issue now: https://github.com/open-telemetry/opentelemetry-go-contrib/issues/7326

Issue #7269, opened 2025-04-25
Revert switching the HTTP semconv migration's default to new


dmathieu:
In a [SIG meeting](https://github.com/open-telemetry/opentelemetry-go/issues/6648#issuecomment-2823456539), we have decided to postpone the switch ot the HTTP semconv migration default value to `new` by one version, as there have been a few bugfixes, and changing the default right now is a bit harsh.

So we need to revert https://github.com/open-telemetry/opentelemetry-go-contrib/pull/6899

abhayporwals:
Hi, @dmathieu! I would like to help in reverting the #6899 and happy to resolve this asap. 

Does fixing this issue requires the reverting the files state before it was merging to #6899, right?

dmathieu:
It requires reverting to the logic how it was before that PR, yes. But we don't want to lose other related changes such as https://github.com/open-telemetry/opentelemetry-go-contrib/pull/7180.
So this is not a simple "git revert".

dmathieu:
This hasn't been moving for 2 weeks. I'll take over.

PR #7317, opened 2025-05-12
Revert the switch to default to dup semantic conventions for http packages

dmathieu:
This reverts #6899, as we've had several bug fixes we needed to bring into the main branch and waiting for an additional release will put less pressure on folks.

See SIG discussions: https://github.com/open-telemetry/opentelemetry-go/issues/6648
Closes https://github.com/open-telemetry/opentelemetry-go-contrib/issues/7269

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -26,13 +26,6 @@ This project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0.htm
 ### Changed
 
 - Jaeger remote sampler's probabilistic strategy now uses the same sampling algorithm as `trace.TraceIDRatioBased` in `go.opentelemetry.io/contrib/samplers/jaegerremote`. (#6892)
-- Switched the default for `OTEL_SEMCONV_STABILITY_OPT_IN` to emit the v1.26.0 semantic conventions by default in the following modules.
-  - `go.opentelemetry.io/contrib/instrumentation/github.com/emicklei/go-restful/otelrestful`
-  - `go.opentelemetry.io/contrib/instrumentation/github.com/gin-gonic/gin/otelgin`
-  - `go.opentelemetry.io/contrib/instrumentation/github.com/gorilla/mux/otelmux`
-  - `go.opentelemetry.io/contrib/instrumentation/github.com/labstack/echo/otelecho`
-  - `go.opentelemetry.io/contrib/instrumentation/net/http/httptrace/otelhttptrace`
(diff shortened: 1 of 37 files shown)


pellared (approved):

MrAlias (commented):
After this change will a user get both v1.20.0 and v1.26.0 attributes or will they only get v1.20.0 attributes by default?

dmathieu:
They will get v1.20.0 only by default, same as on the latest release. `http/dup` gives them both.

dmathieu:
There have been several bugfixes and issues discovered that will only be introduced with this new release.
Eg: https://github.com/open-telemetry/opentelemetry-go-contrib/pull/6942
But also https://github.com/open-telemetry/opentelemetry-go-contrib/pull/6951 (which is more problematic, as folks can't use `http/dup` at all right now).

Hence my proposal to postpone for one release (minor, not patch). I suggested that during a SIG meeting and folks are onboard with it, but I'm fine revisiting that too.

pellared:
> There have been several bugfixes and issues discovered that will only be introduced with this new release.

Sure, but I do not think it means that we cannot introduce also a change that was planned and announced. Users who care about these bugfixes would simply need to switch to use v1.26.0 sem conv or set `OTEL_SEMCONV_STABILITY_OPT_IN=http/dup`

> But also https://github.com/open-telemetry/opentelemetry-go-contrib/pull/6951 (which is more problematic, as folks can't use http/dup at all right now).

Is it only some issue for `otelrestful`? It looks like it works for `otelhttp` (https://github.com/open-telemetry/opentelemetry-go-contrib/pull/7321). Do we have some bug for it? Do we have any way to move forward with the issue and would postponing help in any way?

dmathieu:
Yes, it's only otelrestful which didn't use the semconv package at all.

pellared:
> Yes, it's only otelrestful which didn't use the semconv package at all.

I think it is a very minor issue. I wouldn't be supprised if you were the only one who noticed the issue. I do not expect that even having support for `http/dup` would help as (if I am not mistaken) nobody even filled an issue that `otelrestful` does not support `OTEL_SEMCONV_STABILITY_OPT_IN`.

Honestly speaking, I also consider `otelrestful` module a second-class citizen compared to `otelhttp`. I would not want to slow down our progress on `otelhttp` becasue of `otelrestful`.

dmathieu:
We have decided in the SIG meeting not to do this revert.

PR #7318, opened 2025-05-12
propagators/aws/xray: sample only when Sampled=1

### LINKED ISSUE #7262: parseTraceFlag erroneously marks non‑“1” values as sampled, causing over‑sampling
### Description  
`parseTraceFlag` sets the *sampled* bit for **any** value whose length ≠ 1, even when the flag is invalid (e.g. `Sampled=maybe`).  
```go
func parseTraceFlag(xraySampledFlag string) trace.TraceFlags {
    if len(xraySampledFlag) == 1 && xraySampledFlag != "1" {
        return traceFlagNone        // ok: explicit “0”
    }
    return trace.FlagsSampled       // ← bug: treats “maybe”, “foo”, “”, etc. as sampled
}
```  
According to [AWS X‑Ray header spec](https://docs.aws.amazon.com/xray/latest/devguide/xray-concepts.html#xray-concepts-tracingheader) the **only** value that should map to `trace.FlagsSampled` is `"1"`. Any other value (including `"0"` or garbage) must clear the flag.

Because of this, downstream OpenTelemetry SDKs mark spans as *sampled* when they should be *drop‑eligible*, inflating trace volume and potentially incurring unexpected cost.

---

### Environment
- **OS:** any (confirmed on macOS 14 & Ubuntu 22.04)
- **Architecture:** amd64 & arm64
- **Go version:** 1.22.2
- **Module:** `go.opentelemetry.io/contrib/propagators/aws/xray`
- **Version / commit:** v0.47.0 ( `2e1b0bb` )

---

### Steps To Reproduce
1. Create an inbound carrier with an invalid sampled flag:  
   ```go
   carrier := propagation.MapCarrier{
       "X-Amzn-Trace-Id": "Root=1-abcdef12-1234567890abcdef12345678;" +
                          "Parent=1234567890abcdef;" +
                          "Sampled=maybe",
   }
   ctx := xray.Propagator{}.Extract(context.Background(), carrier)
   sc  := trace.SpanContextFromContext(ctx)
   ```
2. Check the flag:  
   ```go
   fmt.Println(sc.IsValid(), sc.IsSampled())   // → true, **true**
   ```
3. Observe that `IsSampled()` is **true**, even though `"maybe"` is not legal.

---

### Expected Behavior
`parseTraceFlag("maybe")` (or any value other than `"1"`) should return `traceFlagNone`, so the extracted `SpanContext` is **not sampled**.

Proposed fix:
```go
func parseTraceFlag(xraySampledFlag string) trace.TraceFlags {
	if xraySampledFlag == isSampled {
		return trace.FlagsSampled
	}
	return traceFlagNone
}
```

This change:
- Honors the X‑Ray spec  
- Prevents accidental over‑sampling  
- Aligns with Otel propagator behavior for W3C `traceparent`

Happy to submit a PR if this approach is acceptable.

seankhliao:
Fixes #7262


propagators/aws/xray/propagator.go
```diff
@@ -65,7 +65,7 @@ func (xray Propagator) Inject(ctx context.Context, carrier propagation.TextMapCa
 		traceIDDelimiter + otTraceID[traceIDFirstPartLength:]
 	parentID := sc.SpanID()
 	samplingFlag := notSampled
-	if sc.TraceFlags() == traceFlagSampled {
+	if sc.TraceFlags().IsSampled() {
```
seankhliao, line 68: Technically unrelated, but this is more correct since TraceFlags is a bitmask

CHANGELOG.md
```diff
@@ -50,6 +50,7 @@ This project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0.htm
 - The semantic conventions have been upgraded from `v1.17.0` to `v1.30.0` in `go.opentelemetry.io/contrib/instrumentation/google.golang.org/grpc/otelgrpc`. (#7270)
   - All `net.peer.*` and `net.host.*` attributes are now set to correct `server.*` attributes.
   - No `net.socket.*` attributes are set.
+- AWS Xray propagator now only samples when `Sampled=1`. (#7262)
```
MrAlias, line 53: nit

```suggestion
- AWS X-Ray propagator now only samples when `Sampled=1`. (#7262)
```

propagators/aws/xray/propagator.go
```diff
@@ -163,10 +163,10 @@ func parseTraceID(xrayTraceID string) (trace.TraceID, error) {
 
 // parseTraceFlag returns a parsed trace flag.
 func parseTraceFlag(xraySampledFlag string) trace.TraceFlags {
-	if len(xraySampledFlag) == sampledFlagLength && xraySampledFlag != isSampled {
-		return traceFlagNone
+	if xraySampledFlag == isSampled {
```
MrAlias, line 166:
```suggestion
	// Use a direct comparison here (#7262).
	if xraySampledFlag == isSampled {
```

MrAlias (approved):

Issue #6266, opened 2024-10-17
Add tests to ensure the `otelhttp` instrumentation produce the latest semantic convention server metrics when `OTEL_SEMCONV_STABILITY_OPT_IN="http/dup"` is set


MrAlias:

dennisme:
Id also like to take this one too.

dmathieu:
This is now invalid, since we don't have the dup option anymore.

PR #7356, opened 2025-05-23
otelhttp: Add server-side test for OTEL_SEMCONV_STABILITY_OPT_IN=http/dup with new/old metrics

keisku:
https://github.com/open-telemetry/opentelemetry-go-contrib/issues/6266

### DIFF
--- a/instrumentation/net/http/otelhttp/handler_test.go
+++ b/instrumentation/net/http/otelhttp/handler_test.go
@@ -796,63 +796,154 @@ func TestHandlerWithMetricAttributesFn(t *testing.T) {
 }
 
 func TestHandlerWithSemConvStabilityOptIn(t *testing.T) {
+	newSpanAttrs := []attribute.KeyValue{
+		attribute.String("http.request.method", "GET"),
+		attribute.String("url.scheme", "http"),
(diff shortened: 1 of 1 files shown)


instrumentation/net/http/otelhttp/handler_test.go
```diff
@@ -796,63 +796,155 @@ func TestHandlerWithMetricAttributesFn(t *testing.T) {
...
+				DataPoints:  []metricdata.HistogramDataPoint[float64]{{Attributes: oldMetricAttrs}},
+			},
+		},
+	}
 	tests := []struct {
 		name                       string
 		semConvStabilityOptInValue string
-		expected                   []attribute.KeyValue
+		expectedSpanAttributes     []attribute.KeyValue
+		expectedMetrics            metricdata.ScopeMetrics
```
dmathieu, line 895:
```suggestion
		wantSpanAttributes     []attribute.KeyValue
		wantMetrics            metricdata.ScopeMetrics
```

Small change (I know it wasn't like this before, but it's a convention we used that must have fallen through the cracks here).

dashpole (approved):

MrAlias (approved):

PR #7060, opened 2025-03-28
Move `verify_examples.sh` to `tools/verify_examples.sh`

ElangoManikandan:
Resolve #7021

PR #7366, opened 2025-05-26
Fix verify_examples.sh to work from tools directory

### LINKED ISSUE #7353: verify_examples.sh does not work
```
$ ./tools/verify_examples.sh 
make: *** No rule to make target '/home/user/repos/opentelemetry-go-contrib/tools/.tools/gojq'.  Stop.
```

https://github.com/open-telemetry/opentelemetry-go-contrib/pull/7060 only moved the file without changing the script which was using relative paths. The script only worked when it was present in the root of the repository.

kayceenuel:
## Summary
Fixes #7353 - verify_examples.sh script was broken after the gojq tool was moved in #7060.

## Changes
- Replace `make` dependency with direct `go build` command for gojq tool
- Fix path resolution to correctly find `.tools` directory from repository root  
- Handle edge cases where go.mod files have no Replace sections
- Replace `printf` with `echo` to avoid Unicode escape issues in Git Bash
- Add directory existence checks for better error handling

## Testing
- Tested on Windows 10 with Git Bash
- Script now successfully builds all examples without errors
- Gracefully handles missing directories

## Before
Script failed with: `make: *** No rule to make target '.../.tools/gojq'. Stop.`

## After  
Script runs successfully and processes all examples.


pellared:
> after the gojq tool was moved

The tool has not been moved. It was the script that was moved.

kayceenuel:
> > after the gojq tool was moved
> 
> The tool has not been moved. It was the script that was moved.

Ugh you're right, I totally mixed that up in my head. The script moved, not the tool. My bad!

I'll fix the description so it actually makes sense.

kayceenuel:
> How about running the script in CI?

I was mostly focused on getting this to work locally, especially on Windows where make wasn’t available. Does the script actually run in CI right now? If it does, I can double-check that this change doesn’t break anything there

pellared:
> > How about running the script in CI?
> 
> I was mostly focused on getting this to work locally, especially on Windows where make wasn’t available. Does the script actually run in CI right now? If it does, I can double-check that this change doesn’t break anything there

@dmathieu, I think running this on CI should be done as a separate PR.

tools/verify_examples.sh
```diff
@@ -1,76 +1,104 @@
-#!/bin/bash
 
+#!/bin/bash
```
dmathieu: No need to start the file with an empty line.

pellared: Yes, the empty line should be before `# Copyright The OpenTelemetry Authors`

tools/verify_examples.sh
```diff
@@ -1,76 +1,104 @@
...
+	echo "  Build $ex in ${DIR_TMP}/${ex}"
 	(cd "${DIR_TMP}/${ex}" && \
 	 go build .)
 done
 
 # Cleanup
-printf "Remove copied files.\n"
-rm -rf $DIR_TMP
+echo "Remove copied files."
+rm -rf $DIR_TMP
```
pellared: Please EOL at the end of the file

```suggestion
rm -rf $DIR_TMP

```

pellared (approved):
LGTM apart from the nit comments regarding EOLs.
I tested it on Ubuntu and seems to work great.
 Thanks.

kayceenuel:
> ## Summary
> Fixes #7353 - verify_examples.sh script was broken after the gojq tool was moved in #7060.
> 
> ## Changes
> * Replace `make` dependency with direct `go build` command for gojq tool
> * Fix path resolution to correctly find `.tools` directory from repository root
> * Handle edge cases where go.mod files have no Replace sections
> * Replace `printf` with `echo` to avoid Unicode escape issues in Git Bash
> * Add directory existence checks for better error handling
> 
> ## Testing
> * Tested on Windows 10 with Git Bash
> * Script now successfully builds all examples without errors
> * Gracefully handles missing directories
> 
> ## Before
> Script failed with: `make: *** No rule to make target '.../.tools/gojq'. Stop.`
> 
> ## After
> Script runs successfully and processes all examples.

kayceenuel:
Updated. Reverted to using `make` as requested and kept the path fixes.

 Script now correctly resolves paths from tools/ directory  
Uses `make` for building gojq as preferred  
 Tested locally on Windows - processes all examples successfully

tools/verify_examples.sh
pellared: https://github.com/open-telemetry/opentelemetry-go-contrib/pull/7366/commits/8e87866d89f409c00924e9eb3ef141d3837a54d0

tools/verify_examples.sh
pellared: https://github.com/open-telemetry/opentelemetry-go-contrib/pull/7366/commits/8e87866d89f409c00924e9eb3ef141d3837a54d0

PR #7380, opened 2025-05-28
otelzap: refactor Core tests

pellared:
Use the new convenience of https://pkg.go.dev/go.opentelemetry.io/otel/log/logtest. 

The tests should be more readable and the assertions are more precise.
They also give better assertion messages on failures (diff).


khushijain21:
This looks great!

PR #7387, opened 2025-05-28
Upgrade semconv to 1.32.0 in otellogr

MrAlias:


pellared:
Merging as trivial and also same as https://github.com/open-telemetry/opentelemetry-go-contrib/pull/7389

PR #7393, opened 2025-05-28
Upgrade semconv to 1.32.0 for otelmongo

MrAlias:


MrAlias:
Keeping the term "latest" prevents constant manual intervention and the inevitability of becoming out of sync. I prefer to keep the terms as they are. 

Users with the desire for details on what version of semantic convention they are emitting can look at the produced telemetry scheme URL or the code they are reading this from .

pellared (approved):

prestonvasquez:
> Keeping the term "latest" prevents constant manual intervention and the inevitability of becoming out of sync. I prefer to keep the terms as they are.
> 
> Users with the desire for details on what version of semantic convention they are emitting can look at the produced telemetry scheme URL or the code they are reading this from .

@MrAlias I specifically added the word "latest" so you could just drop the update in. The current doc.go files specifies 1.26.0.

MrAlias:
> > Keeping the term "latest" prevents constant manual intervention and the inevitability of becoming out of sync. I prefer to keep the terms as they are.
> > Users with the desire for details on what version of semantic convention they are emitting can look at the produced telemetry scheme URL or the code they are reading this from .
> 
> @MrAlias I specifically added the word "latest" so you could just drop the update in. The current doc.go files specifies 1.26.0.

Ah, gotcha. I misunderstood. Sounds like we are in favor of the same result.

PR #7424, opened 2025-06-04
[DO NOT MERGE] minsev: Fix to filter out only in SeverityTrace1..SeverityFatal4 severity range

pellared:
Per https://github.com/open-telemetry/opentelemetry-specification/pull/4541

From https://github.com/open-telemetry/opentelemetry-specification/blob/bdcf53d55285ae42339f37908a24ba6f3ac167cb/specification/logs/data-model.md?plain=1#L418-L423:

>  In the contexts where severity participates in less-than / greater-than
comparisons `SeverityNumber` field should be used. `SeverityNumber` can be
compared to another `SeverityNumber` or to numbers in the 1..24 range (or to the
corresponding short names).

Currently the spec only describes how to compare values between 1 and 24.
All other values have undefined meaning.

The main reason is to NOT filter out records with `log.SeverityUndefined` severity which can be a common scenario.

I also did my best to improve the docs.


processors/minsev/minsev.go
```diff
@@ -57,26 +58,33 @@ var (
 	_ log.FilterProcessor = (*LogProcessor)(nil)
 )
 
-// OnEmit passes ctx and r to the [log.Processor] that p wraps if the severity
-// of record is greater than or equal to p.Minimum. Otherwise, record is
-// dropped.
+// OnEmit drops records with severity less than the one returned by [Severitier]
+// and inside the [logapi.SeverityTrace1]..[logapi.SeverityFatal4] range.
+// If the severity of record is greater than or equal to he one returned by [Severitier],
```
dmathieu:
```suggestion
// If the severity of record is greater than or equal to the one returned by [Severitier],
```

dmathieu (approved):

pellared:
Marking as blocked by specification as I do not want us to go back and forth until:
- https://github.com/open-telemetry/opentelemetry-specification/pull/4541 

and maybe

- https://github.com/open-telemetry/opentelemetry-specification/issues/4540

are resolved.

pellared:
From the SIG meetings discussions it seems that most prefer our current implementation: https://github.com/open-telemetry/opentelemetry-specification/pull/4541#discussion_r2138960382.

I will extract the Go docs improvements to a separate PR.

pellared:
Closed per https://github.com/open-telemetry/opentelemetry-specification/pull/4552

PR #7425, opened 2025-06-04
Remove internal/utils package from jaegersampler

dmathieu:
This package's name is too generic, and makes the latest revive unhappy: https://github.com/open-telemetry/opentelemetry-go-contrib/pull/7382

I have moved the `GetJSON` method into `testutils`. That's the only place this method is used. I also made it private, since it's not needed by other packages.

Now that the rate limiter is the only util thing, I've moved it to a `ratelimiter` package.


pellared (approved):

pellared (commented):
@yurishkuro, can you please take a look as a Code Owner?

Issue #7254, opened 2025-04-23
otelhttp: Cannot record error for spans when producing new semantic convention


XSAM:
### Description

`otelhttp` only uses `span.RecordError(err)` when `errType` is not empty

https://github.com/open-telemetry/opentelemetry-go-contrib/blob/b8572491f2d143672581d1c51693c928d376dff8/instrumentation/net/http/otelhttp/transport.go#L132-L140

But `ErrorType` method never returns an empty attribute.

https://github.com/open-telemetry/opentelemetry-go-contrib/blob/b8572491f2d143672581d1c51693c928d376dff8/instrumentation/net/http/httptrace/otelhttptrace/internal/semconv/httpconv.go#L443-L458

### Expected behavior

I don't have a preference on whether we should record `error.type` with `exception.type` at the same time. But the comment and the logic here seem conflicted.

dmathieu:
I don't know what was the intent there. But it feels like `ErroTypeOther` should be the case where we want to fallback to recording an error?

XSAM:
> But it feels like ErroTypeOther should be the case where we want to fallback to recording an error?

I don't find a way to make it return `ErrorTypeOther`. In most cases, it returns `semconv.ErrorTypeKey.String("*errors.errorString")`.

For instance, errors created by `errors.New()` like `driver.ErrBadConn`. The type is `*errors.errorString`.

I know it is difficult (or impossible) to get this via reflect, but do you think returning `semconv.ErrorTypeKey.String("database/sql/driver.ErrBadConn")` is the ideal result for `driver.ErrBadConn`?

@open-telemetry/go-maintainers What do you think about this issue?

XSAM:
More information: https://opentelemetry.io/docs/specs/otel/trace/exceptions/ (not in the semconv doc)

The spec about exceptions says

> An exception SHOULD be recorded as an Event on the span during which it occurred if and only if it remains unhandled **when the span ends and causes the span status to be set to ERROR**.

It even has a pseudo code that sets the `error.type` attribute for spans and calls `recordException` at the same time.

```java
Span span = myTracer.startSpan(/*...*/);
try {
  // Code that does the actual work which the Span represents
} catch (Throwable e) {
  span.recordException(e);
  span.setAttribute(AttributeKey.stringKey("error.type"), e.getClass().getCanonicalName())
  span.setStatus(StatusCode.ERROR, e.getMessage());
  throw e;
} finally {
  span.end();
}
```

[...]

PR #7456, opened 2025-06-10
Always set error type and record error in otelhttp transport

dmathieu:
Closes https://github.com/open-telemetry/opentelemetry-go-contrib/issues/7254

### DIFF
--- a/instrumentation/net/http/otelhttp/transport.go
+++ b/instrumentation/net/http/otelhttp/transport.go
@@ -130,15 +130,8 @@ func (t *Transport) RoundTrip(r *http.Request) (*http.Response, error) {
 
 	res, err := t.rt.RoundTrip(r)
 	if err != nil {
-		// set error type attribute if the error is part of the predefined
-		// error types.
-		// otherwise, record it as an exception
-		if errType := t.semconv.ErrorType(err); errType.Valid() {
-			span.SetAttributes(errType)
-		} else {
-			span.RecordError(err)
-		}
-
+		span.SetAttributes(t.semconv.ErrorType(err))
+		span.RecordError(err)
 		span.SetStatus(codes.Error, err.Error())
 		span.End()
 		return res, err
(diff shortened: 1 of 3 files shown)


dashpole (approved):

instrumentation/net/http/otelhttp/transport.go
```diff
@@ -130,15 +130,8 @@ func (t *Transport) RoundTrip(r *http.Request) (*http.Response, error) {
...
-		// error types.
-		// otherwise, record it as an exception
-		if errType := t.semconv.ErrorType(err); errType.Valid() {
-			span.SetAttributes(errType)
-		} else {
-			span.RecordError(err)
-		}
-
+		span.SetAttributes(t.semconv.ErrorType(err))
+		span.RecordError(err)
```
pellared, line 134: What is the value of recording the error as event if the type is set as attribute and value (`err.Error()`) as status description?

> An exception SHOULD be recorded as an Event on the span during which it occurred if and only if it remains unhandled when the span ends and causes the span status to be set to ERROR.

I argue that error result is not an exception. For me a `panic` that was not being recovered is an unhandled exception. We already cover this part of this specification here:

https://github.com/open-telemetry/opentelemetry-go/blob/bc531cb3f9ce9c328b12e473b6409784728dc608/sdk/trace/span.go#L465-L482

Additionally see: https://github.com/open-telemetry/semantic-conventions/blob/main/docs/general/recording-errors.md#recording-errors-on-spans

Therefore, we should not call `span.RecordError(err)`.

```suggestion
```

dmathieu: Do you want to chime in https://github.com/open-telemetry/opentelemetry-go-contrib/issues/7254?

pellared: https://github.com/open-telemetry/opentelemetry-go-contrib/issues/7254#issuecomment-2962485821

dmathieu:
Closing per the comment discussion.

PR #7483, opened 2025-06-18
fix: Stop EKS detector erroring outside of K8S

### LINKED ISSUE #7477: EKS detector errors outside of EKS/K8S
### Description

I'm adding the EKS detector to my code which in production, runs in/on EKS. However when I run my code locally on my machine, the detector errors, yet when I use the EC2 detector for example, (and I'm _not_ in/on an EC2 instance), it just silently returns no error.

### Environment

- OS: Linux
- Architecture: x86
- Go Version: 1.23.4
- `eks` version: 1.32

### Steps To Reproduce

Using this code:
```golang
r, err := resource.New(ctx, resource.WithFromEnv(), resource.WithDetectors(eks.NewResourceDetector()))
```

I get error `failed to create config: unable to load in-cluster configuration, KUBERNETES_SERVICE_HOST and KUBERNETES_SERVICE_PORT must be defined`.

The `Detect()` method is written to cater for not running inside K8S/EKS but the constructor calls `newK8sDetectorUtils()` which taints the detector from ever working.

### Expected behavior

I would expect the EKS detector to simply return the default `resource.Empty()`. I assume the `resource.WithDetectors()` option is intended to be used by adding as many different detectors for environments where your code _might_ run and they either return useful attributes or nothing.

bodgit:
This is a simple fix to stop the EKS detector erroring when run outside of Kubernetes.

Fixes #7477


dashpole:
Can you add a unit test?

dashpole (approved):

PR #7487, opened 2025-06-18
Ignore Azure endpoint in link check

MrAlias:
https://github.com/open-telemetry/opentelemetry-go-contrib/blob/96faa603bfa38f691ec6d519935f4ffc9aea4487/detectors/azure/azurevm/vm.go#L18


dmathieu:
Merging this as a trivial fix.

Issue #316, opened 2020-08-29
Runtime instrumentation: GC "total time spent" metric


jmacd:
I noticed a few things that I'd like to change about the GC timing metrics emitted by `instrumentation/runtime`. 

--

`runtime.go.gc.pause_total_ns` should not include a "_ns" unit suffix.
`runtime.go.gc.pause_ns` as well.

The main repository does not have a Unit declared for Nanoseconds. It's appropriate to report these in nanoseconds, since they are provided in nanoseconds, but I'd like to set the units being used so that downstream systems can normalize them into seconds.

--

There is a field `MemStats.GCCPUFraction` that is not being reported. This is a cumulative summary computed over the process lifetime, which the Metrics SIG has come to realize is rarely useful for direct monitoring. The comment reads:

```
    // GCCPUFraction is expressed as a number between 0 and 1,
    // where 0 means GC has consumed none of this program's CPU. A
    // program's available CPU time is defined as the integral of
    // GOMAXPROCS since the program started. That is, if
    // GOMAXPROCS is 2 and a program has been running for 10
    // seconds, its "available CPU" is 20 seconds. GCCPUFraction
    // does not include CPU time used for write barrier activity.
```

Therefore, the total time spent performing GC can be computed as follows:

```
	uptime := time.Since(processStartTime)
        gomaxprocs := float64(runtime.GOMAXPROCS(0))
	gcSeconds := memStats.GCCPUFraction * uptime.Seconds() * gomaxprocs
```

This `gcSeconds` is a number that we should be monitoring, as this can be compared directly with `process.cpu.usage`.

This works as long as GOMAXPROCS() does not change, and with more sophisticated logic it could approximately track changes in GOMAXPROCS() as well.

chrisleavoy:
I can take on adding a nannoseconds unit and updating the runtime instruments.

But GOMAXPROCS() indeed can change dynamically at runtime. It wasn't apparent to me how I could approximate a value for `gcSeconds` if maxprocs were to change after a long period at a different value.

skonto:
@jmacd is there are pointer for what the Metrics SIG has considered as useful for direct monitoring? I am planning to use the runtime plugin in [`knative/pkg](https://github.com/knative/pkg)` but it seems to me that some metrics are not exposed like `NextGC` and I am wondering if we should have like a full mode in case a user wants the raw dump as it comes from runtime.memstats.

jmacd:
[...]

PR #7499, opened 2025-06-23
Add runtimemetrics instrumentation

jmacd:
Revival of #316
Part of #2624 

My team has been looking at Golang runtime instrumentation and specifically are interested in access to runtime/metrics, which addresses the problem described in #316. As an example, with this instrumentation and any modern Go runtime, we get a `process.runtime.go.cpu.time` metric with attributes like

```
      class=gc,class2=mark,class3=assist
      class=gc,class2=mark,class3=dedicated
      class=gc,class2=mark,class3=idle
      class=gc,class2=pause
      class=scavenge,class2=assist
      class=scavenge,class2=background
      class=idle
      class=user
```

This allows you to construct graphs comparing the total for class=gc with the sub-total for class3=assist, which is an indicator of relatively bad GC performance.

This code was forked from https://github.com/lightstep/otel-launcher-go/tree/main/lightstep/instrumentation/runtime.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -11,6 +11,7 @@ This project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0.htm
 ### Added
 
 - Add the `WithPublicEndpoint` and `WithPublicEndpointFn` options to `go.opentelemetry.io/contrib/instrumentation/google.golang.org/grpc/otelgrpc`. (#7407)
+- Add new `instrumentation/runtimemetrics` providing a comprehensive view into Golang runtime/metrics instrumentation. ([#7499](https://github.com/open-telemetry/opentelemetry-go-contrib/pull/7499))
 
 ### Removed
 


dashpole:
I haven't reviewed closely, but have a few questions: How do you see this relating to the ones we defined in semconv: https://opentelemetry.io/docs/specs/semconv/runtime/go-metrics/?

* Are they complementary or supplementary?
* When would someone use one vs the other?
* Is there any way for us to achieve our goals with a single instrumentation library, or do we need both as separate libraries?

jmacd:
@dashpole I understand your questions, and they are ones I have tried to avoid. This library represents a literal translation from an existing thing into the OTel model, so these are not OTel semantic conventions. I would be glad to change the prefix used, if you think that will help.

> Are they complementary or supplementary?

I would call this instrumentation independent: definitely supplementary, possibly duplicative. I like the idea of a straightforward integration allowing us to expose the Go teams' metrics without these questions getting in the way.

> When would someone use one vs the other?

I would replace the instrumentation/runtime library with this one for my own uses. I do not care for the existing OTel semantic conventions, I care for what the Go-team gives me verbatim.

> Is there any way for us to achieve our goals with a single instrumentation library, or do we need both as separate libraries?

Probably. If someone could tell me what to do, I'll do it. (What should I do?)

jmacd:
To be clear, the new semantic conventions will be updated in the go-contrib/instrumentation/runtime package? 
That is all I wanted to ensure, so that we can have these metrics in the collector by default. 
Thank you!

PR #7513, opened 2025-06-26
Honor otelhttp.WithMessageEvents(otelhttp.ReadEvents) option with otelhttp.NewTransport(...)

jstangroome:
My primary goal is for http client requests to trace time-to-first-response-byte, and there is very likely a much more efficient method to capture just this detail. The time to the first response byte is important for large streaming responses where it is far less helpful to record only the moment after the entire response body has been completely read.

However, using the existing pattern established by `otelhttp.NewHandler(..., otelhttp.WithMessageEvents(otelhttp.ReadEvents))` for http server requests and extending it to also work with `otelhttp.NewTransport` the same information (and more) becomes available for a very small code change and no API change.

I looked for existing tests for Handler and WithMessageEvents to replicate to Transport but did not find any so I'm not sure what the preferred pattern would be for adding unit tests for this functionality.


dmathieu (commented):

instrumentation/net/http/otelhttp/transport.go
```diff
@@ -244,6 +248,10 @@ func (wb *wrappedBody) Read(b []byte) (int, error) {
 	// Record the number of bytes read
 	wb.read.Add(int64(n))
 
+	if wb.readEvent {
+		wb.span.AddEvent("read", trace.WithAttributes(ReadBytesKey.Int64(int64(n))))
```
dmathieu, line 247: A test for this code path would be nice too.

jstangroome: @dmathieu Test and changelog entry added

dmathieu:
The linter is now failing :)
Also, the changelog entry needs to be moved up to the unreleased section.

CHANGELOG.md
```diff
@@ -16,6 +16,10 @@ This project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0.htm
 - Add the `WithLoggerProviderOptions`, `WithMeterProviderOptions` and `WithTracerProviderOptions` options to `NewSDK` to allow passing custom options to providers in `go.opentelemetry.io/contrib/otelconf`. (#7552)
 - Added V2 version of AWS EC2 detector `go.opentelemetry.io/contrib/detectors/aws/ec2/v2` due to deprecation of `github.com/aws/aws-sdk-go`. (#6961)
 
+### Changed
+
+- `NewTransport()` in `go.opentelemetry.io/contrib/instrumentation/net/http/otelhttp` will now accept `WithMessageEvents(ReadEvents)` to enable span events to be emitted for each read of a HTTP response body.
```
dmathieu: This is an addition. It needs to be in the `Added` section.

```suggestion
- Handle the `WithMessageEvents` option for `Transport` in `go.opentelemetry.io/contrib/instrumentation/net/http/otelhttp`. (#7513)
```

dmathieu (commented):
LGTM overall.

instrumentation/net/http/otelhttp/transport.go
dmathieu, line 247: We should do this in the `readRecordFunc` method above, like we do in handler.

jstangroome: I looked at this approach originally, but the `readRecordFunc` in `Transport` is wrapped in a `atomic.Bool` to ensure it is only called once on `Read()` with `io.EOF` or on `Close()`, but the goal of this PR is to record every `Read()`.

I could change my implementation to follow the `readRecordFunc` pattern but I think we'd end up with an `everyReadFunc` and a `finalReadFunc`. The code in handler doesn't try to process Close/EOF as a special case so its `readRecordFunc` is called for every `Read`.

dmathieu (approved):

instrumentation/net/http/otelhttp/transport_test.go
```diff
@@ -617,6 +617,66 @@ func TestTransportErrorStatus(t *testing.T) {
...
+	defer func() { assert.NoError(t, res.Body.Close()) }()
+
+	_, err = io.ReadAll(res.Body)
+	require.NoError(t, err)
+
+	// Check span.
+	spans := spanRecorder.Ended()
+	if len(spans) != 1 {
+		t.Fatalf("expected 1 span; got: %d", len(spans))
+	}
```
pellared: Can we use `require.Len` instead?

instrumentation/net/http/otelhttp/transport_test.go
```diff
@@ -617,6 +617,66 @@ func TestTransportErrorStatus(t *testing.T) {
...
+	tr := NewTransport(
+		http.DefaultTransport,
+		WithTracerProvider(provider),
+		WithMessageEvents(ReadEvents),
+	)
+	c := http.Client{Transport: tr}
+	r, err := http.NewRequest(http.MethodGet, ts.URL, nil)
+	if err != nil {
+		t.Fatal(err)
+	}
```
pellared: why not `require.NoError`?

pellared:

dmathieu:
I believe this PR is still valid, yes. The conflicts need to be resolved and it need a second approval.

pellared:
I also see that my feedback has not been addressed.

ping @jstangroome

PR #7522, opened 2025-06-27
Add the `detectors/autodetect pacakge`

MrAlias:
There is user need to be able to define `resource.Detectors` dynamically at runtime[^1][^2]. This introduces a new module that handles this dynamic creation.

[^1]: https://github.com/open-telemetry/opentelemetry-go-instrumentation/pull/2438
[^2]: https://github.com/open-telemetry/opentelemetry-configuration/blob/04f4b6b07b112296d5e0028626c6089878bd195b/schema/resource.json#L79-L103


detectors/autodetect/autodetect.go
```diff
@@ -0,0 +1,325 @@
...
+	"go.opentelemetry.io/contrib/detectors/aws/eks"
+	"go.opentelemetry.io/contrib/detectors/aws/lambda"
+	"go.opentelemetry.io/contrib/detectors/azure/azurevm"
+	"go.opentelemetry.io/contrib/detectors/gcp"
+)
+
+var (
+	// IDAWSEC2 is the ID for the AWS EC2 detector that detects resource
+	// attributes on Amazon Web Services (AWS) EC2 instances (see
+	// ec2.NewResourceDetector for details).
```
pellared, line 28: These hyperlinks should work.

```suggestion
	// [ec2.NewResourceDetector] for details).
```

detectors/autodetect/autodetect.go
```diff
@@ -0,0 +1,325 @@
...
+
+	"go.opentelemetry.io/contrib/detectors/aws/ec2"
+	"go.opentelemetry.io/contrib/detectors/aws/ecs"
+	"go.opentelemetry.io/contrib/detectors/aws/eks"
+	"go.opentelemetry.io/contrib/detectors/aws/lambda"
+	"go.opentelemetry.io/contrib/detectors/azure/azurevm"
+	"go.opentelemetry.io/contrib/detectors/gcp"
+)
+
+var (
```
pellared, line 25: Maybe we can add a section for GoDoc? E.g.

```suggestion
// Resource detectors IDs that are provided out-of-the-box.
var (
```

detectors/autodetect/autodetect.go
```diff
@@ -0,0 +1,325 @@
...
+// ID represents the unique identifier of a resource detector.
+type ID string
+
+// Register registers a new resource detector function with the given ID.
+func Register(id ID, fn func() resource.Detector) {
+	registryMu.Lock()
+	defer registryMu.Unlock()
+	if _, exists := registry[id]; exists {
+		panic("detector already registered: " + id)
+	}
```
pellared, line 170: I know that this follow the design of `autoexport` and `autoprop` modules.

I just want to provide feedback that if there is a new "build-in" detector that is added it may conflict with something provided by the user. E.g. the user uses `v0.1.0` and calls `Register("cloudlflare", cfDetect)` and when `autodetect@v0.2.0` has added `cloudlflare` then such code would start to panic. I think such behavior may be OK for some users (e.g. that prefer to use upstream functionalities). However, I think whether we maybe should add an `Unregister` function?

Given this is not a stable module we do not need to worry much about my comment. Moreover, I like that it follows the same design as other `auto` modules.

detectors/autodetect/autodetect.go
```diff
@@ -0,0 +1,325 @@
...
+	registry[id] = fn
+}
+
+// Registered returns a sorted slice of all registered resource detector IDs.
+func Registered() []ID {
```
pellared, line 175: Maybe rename it to `Detectors` to follow https://pkg.go.dev/database/sql#Drivers?

detectors/autodetect/autodetect.go
```diff
@@ -0,0 +1,325 @@
...
+	return out.res, out.err
+}
+
+// detection is the result of a [resource.Detector] detection.
+type detection struct {
+	res *resource.Resource
+	err error
+}
+
+// doDetect runs all the detectors concurrently in their own goroutines. All
```
pellared, line 245: I think that we should document that registered detect functions are executed concurrently (maybe in `Detector` function?).

detectors/autodetect/example_base_test.go
pellared, line 1: Why is it in a separate file? We don't want to have the `MyDetector` defintions in the examples?

detectors/autodetect/example_base_test.go
```diff
@@ -0,0 +1,37 @@
...
+
+	iterator.Next()
+	_, _ = b.WriteString(string(iterator.Attribute().Key))
+
+	for iterator.Next() {
+		_, _ = b.WriteRune(' ')
+		_, _ = b.WriteString(string(iterator.Attribute().Key))
+	}
+	return b.String()
+}
```
pellared, line 37: Is this unused?

detectors/autodetect/example_cfg_test.go
```diff
@@ -0,0 +1,58 @@
...
+		"my.cfg.detector"
+	]
+}`)
+
+type Config struct {
+	Detectors []autodetect.ID `json:"detectors"`
+}
+
+func ExampleDetector() {
+	// This example shows how to parse resource.Detectors from a user define
```
pellared:
```suggestion
	// This example shows how to parse resource.Detectors from a user defined
```

detectors/autodetect/autodetect_test.go
```diff
@@ -0,0 +1,172 @@
...
+	Register(id, testFactory())
+
+	detector, err := Detector(id)
+	if err != nil {
+		t.Fatalf("unexpected error: %v", err)
+	}
+
+	c, ok := detector.(*composite)
+	if !ok {
+		t.Errorf("expected composite detector, got %T", detector)
```
pellared: "got" should be before "want".
See: https://go.dev/wiki/TestComments#got-before-want

Same comment applied to other assertions.

detectors/autodetect/autodetect_test.go
```diff
@@ -0,0 +1,172 @@
...
+
+	if len(c.detectors) != 1 {
+		t.Fatalf("expected 1 detector, got %d", len(c.detectors))
+	}
+
+	switch c.detectors[0].(type) {
+	case *testDetector:
+	default:
+		t.Errorf("expected testDetector, got %T", c.detectors[0])
+	}
```
pellared: I think it would be more readable if we do it something like

```suggestion
	got := c.detectors[0]
	if _, ok := got.(*testDetector); !ok {
		t.Errorf("got %T, want *testDetector")
	}
```

pellared (commented):
:+1:

detectors/autodetect/autodetect.go
MrAlias: Links are not rendered in `var` block comments.

detectors/autodetect/autodetect.go
MrAlias: Each variable is documented. This satisfies the documentation requirement for these values.

detectors/autodetect/autodetect.go
MrAlias: The panic is important to ensure malicious actors are not overriding behavior. An unregister would provide a silent way to circumvent this as well.

There is nothing stopping 3rd-party users from using an alternate registration name.

detectors/autodetect/autodetect.go
MrAlias: But, it's not returning detectors(?)

detectors/autodetect/autodetect.go
MrAlias: Doesn't that lock us into that model of execution and bleed the implementation details into the API?

detectors/autodetect/example_base_test.go
MrAlias: It is used in both examples. Which other file does it belong?

detectors/autodetect/example_base_test.go
MrAlias: No.

detectors/autodetect/autodetect_test.go
MrAlias: Huh? The desired detector type is not shown after the term "got", the one we actually received is.

detectors/autodetect/autodetect_test.go
MrAlias: These are equivalent.

detectors/autodetect/autodetect.go
pellared: I am not sure. I think it may be important if some detectors could not be run concurrently. At the same time I am not aware of such use case
 Feel free to resolve my comment.

detectors/autodetect/example_base_test.go
pellared: If I remember correctly the GoDoc would "render" these type into the example if everything would be in one file e.g. `example_test.go`. Right now the pkgsite does not render this definition

MrAlias: Correct. However, this definition is not critical to understanding the example, it is used as a token "my detector" 3rd-party detector and the contents of its definition are not important to the example.

And, again, which file should it go in if it is not defined here? I can only exist in one, so one example will not include its definition.

pellared: I was just thinking of merging all `example_*_test.go` fines into `example_test.go`.

Personally, I would prefer the example to have all code that is necessary to be compiled and run. However, I would not block the PR because this is just my opinion.

detectors/autodetect/autodetect_test.go
pellared: > Test outputs should output the actual value that the function returned before printing the value that was expected.

detectors/autodetect/autodetect_test.go
pellared: Yes, using a `switch` here is an overkill. 

Also I would prefer consistency as even in other place there already is:

```
c, ok := detector.(*composite)
```

detectors/autodetect/autodetect.go
pellared: https://pkg.go.dev/database/sql#Drivers is also not returning drivers (but their names).

PS. This is not a strong opinion.

detectors/autodetect/autodetect_test.go
MrAlias: Oh, I think I see what you're saying. :+1:

detectors/autodetect/autodetect.go
MrAlias: Keeping this linked to the registration concept (`func Register`) will couple related functionality and keep the function name specific to the functionality it performs and the data it returns. Adding this as `Detectors` means we need to clarify in that function's documentation what detectors will be returned (i.e. only the registered ones) and how it relates to the `Register` function.

detectors/autodetect/example_base_test.go
MrAlias: That approach really muddies up the example:

detectors/autodetect/autodetect_test.go
MrAlias: This is based on a style-guide that is not a part of our [contributing guidelines](https://github.com/open-telemetry/opentelemetry-go/blob/9af7382dceee4aa57de995141936c615c4d340b2/CONTRIBUTING.md) as far as I can tell. If you want to enforce this type of structure can you propose this to the SIG as a policy change and provide a linter to check this?

detectors/autodetect/example_base_test.go
pellared: 1. Probably if we inline `enc.Encode` then it would be less muddy and more readable.
2. I do not line that we see the `const envVar` and `init` here :vomiting_face: 

Let's keep it as it is.

detectors/autodetect/autodetect_test.go
pellared: This is not a blocking comment.
I still prefer this structure and output than using `testify` :wink:

dmathieu (approved):

PR #7553, opened 2025-07-07
otellogrus: set `SeverityText` field to logrus hook

huiyifyj:
The `Severity` field is used to represent the severity of the log, but it does not provide a human-readable representation of the log `level`.
The `SeverityText` [^1] field seems to be intended to provide a human-readable representation of the log level in otel's specification.

This PR is to set the `SeverityText` field based on the [logrus.Level] in the otellogrus hook.

[^1]: [OTEL Spec: Data Model for Logs #SeverityText](https://opentelemetry.io/docs/specs/otel/logs/data-model/#field-severitytext)

[logrus.Level]: https://pkg.go.dev/github.com/sirupsen/logrus#Level

### DIFF
--- a/bridges/otellogrus/hook.go
+++ b/bridges/otellogrus/hook.go
@@ -12,8 +12,7 @@
 //
 //   - Time is set as the Timestamp.
 //   - Message is set as the Body using a [log.StringValue].
-//   - Level is transformed and set as the Severity. The SeverityText is not
-//     set.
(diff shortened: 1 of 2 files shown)


dmathieu:
This looks good, but it needs a changelog entry.

huiyifyj:
> This only needs a changelog entry.

👌

dmathieu (approved):

PR #7556, opened 2025-07-07
otellambda: allow custom attributes

matbrito:
This PR introduces a new feature to the AWS Lambda OpenTelemetry Go instrumentation, allowing users to attach custom attributes to spans based on the Lambda event payload. This enhancement provides greater flexibility for enriching traces with application-specific context.

### New Option: WithCustomAttributes

* Added a new option function, WithCustomAttributes, which accepts an EventAttrExtractor (a function that extracts attributes from the event JSON).
* Updated the internal config struct to include a CustomAttributes field.
* The default extractor is a no-op, returning no attributes.
## Instrumentation Logic Update

* During span creation, the configured CustomAttributes extractor is called with the event payload.
* Any attributes returned by the extractor are appended to the span’s attributes.
## Unit Tests

* Added a new test, [TestWrapHandlerTracingWithCustomAttributes](vscode-file://vscode-app/usr/share/code/resources/app/out/vs/code/electron-sandbox/workbench/workbench.html), to verify that custom attributes extracted from the event are correctly attached to the span.
* The test uses a mock extractor and asserts that the expected custom attribute is present in the span.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -10,6 +10,8 @@ This project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0.htm
 
 ### Added
 
+- Add the `WithCustomAttributes` option to `go.opentelemetry.io/contrib/instrumentation/github.com/aws/aws-lambda-go/otellambda`. (#7556)
+
 - Add the unit `ns` to deprecated runtime metrics `process.runtime.go.gc.pause_total_ns` and `process.runtime.go.gc.pause_ns` in `go.opentelemetry.io/contrib/instrumentation/runtime`. (#7490)
 
 <!-- Released section -->
@@ -20,8 +22,6 @@ This project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0.htm
 ### Added
 
 - Add the `WithPublicEndpoint` and `WithPublicEndpointFn` options to `go.opentelemetry.io/contrib/instrumentation/google.golang.org/grpc/otelgrpc`. (#7407)
-
-- Add the `WithCustomAttributes` option to `go.opentelemetry.io/contrib/instrumentation/github.com/aws/aws-lambda-go/otellambda`. (#7556)
 
 ### Changed
 
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -21,6 +21,8 @@ This project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0.htm
 
 - Add the `WithPublicEndpoint` and `WithPublicEndpointFn` options to `go.opentelemetry.io/contrib/instrumentation/google.golang.org/grpc/otelgrpc`. (#7407)
 
+- Add the `WithCustomAttributes` option to `go.opentelemetry.io/contrib/instrumentation/github.com/aws/aws-lambda-go/otellambda`. (#7556)
+
 ### Changed
 
 - `go.opentelemetry.io/contrib/instrumentation/runtime` now produces the new metrics by default. Set `OTEL_GO_X_DEPRECATED_RUNTIME_METRICS=true` environment variable to additionally produce the deprecated metrics. (#7418)
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -10,7 +10,7 @@ This project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0.htm
 
 ### Added
 
-- Add the `WithCustomAttributes` option to `go.opentelemetry.io/contrib/instrumentation/github.com/aws/aws-lambda-go/otellambda`. (#7556)
+- Add the `WithTraceAttributesFn` option to `go.opentelemetry.io/contrib/instrumentation/github.com/aws/aws-lambda-go/otellambda`. (#7556)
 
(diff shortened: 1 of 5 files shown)


matbrito:
[#7557 ](https://github.com/open-telemetry/opentelemetry-go-contrib/issues/7557)

CHANGELOG.md
```diff
@@ -21,6 +21,8 @@ This project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0.htm
 
 - Add the `WithPublicEndpoint` and `WithPublicEndpointFn` options to `go.opentelemetry.io/contrib/instrumentation/google.golang.org/grpc/otelgrpc`. (#7407)
 
+- Add the `WithCustomAttributes` option to `go.opentelemetry.io/contrib/instrumentation/github.com/aws/aws-lambda-go/otellambda`. (#7556)
```
dmathieu: You will need to move this higher up, to the unreleased section.

instrumentation/github.com/aws/aws-lambda-go/otellambda/config.go
```diff
@@ -114,3 +129,10 @@ func WithPropagator(propagator propagation.TextMapPropagator) Option {
 		c.Propagator = propagator
 	})
 }
+
+// WithCustomAttributes configures a function that returns custom attributes.
+func WithCustomAttributes(eventAttrExtractor EventAttrExtractor) Option {
```
dmathieu: These attributes are for traces. If we add metrics, this method name becomes misleading.
We should name it `WithTraceAttributesFn`.

See otelhttp for a precedent: https://pkg.go.dev/go.opentelemetry.io/contrib/instrumentation/net/http/otelhttp#WithMetricAttributesFn

instrumentation/github.com/aws/aws-lambda-go/otellambda/README.md
```diff
@@ -78,6 +79,12 @@ func mockEventToCarrier(eventJSON []byte) propagation.TextMapCarrier{
 	return propagation.HeaderCarrier{someHeaderKey: []string{request.Headers[someHeaderKey]}}
 }
 
+func mockEventAttrExtractor(eventJSON []byte) []attribute.KeyValue {
+	var request mockRequest
```
dmathieu:
```suggestion
	var request mockHTTPRequest
```

MrAlias:
Thanks for your interest in the project. Please provide a complete enhancement issue in #7557 prior to proposing a change here. Design and motivation can be discussed there.

dmathieu:
@akats7 what do you think of this, as code owner?

instrumentation/github.com/aws/aws-lambda-go/otellambda/config.go
```diff
@@ -35,10 +36,18 @@ var _ Flusher = &noopFlusher{}
 // trace information is instead stored in the Lambda environment.
 type EventToCarrier func(eventJSON []byte) propagation.TextMapCarrier
 
+// EventAttrExtractor defines a function that extracts attributes
+// from the event JSON to be added to the span created by the instrumentation.
+type EventAttrExtractor func(eventJSON []byte) []attribute.KeyValue
```
akats7: Can we update the name to be more consistent with other packages, something like TraceAttributeFn

instrumentation/github.com/aws/aws-lambda-go/otellambda/config.go
```diff
@@ -80,6 +89,12 @@ type config struct {
 	// The default value of Propagator the global otel Propagator
 	// returned by otel.GetTextMapPropagator()
 	Propagator propagation.TextMapPropagator
+
+	// CustomAttributes is a function that returns custom attributes
+	// to be added to the span created by the instrumentation.
+	// The default value of CustomAttributes is nil, which means no custom attributes
+	// will be added to the span.
+	CustomAttributes EventAttrExtractor
```
akats7: Can we update the var name to also be more clear that this is a fn and not a list of attributes, TraceAttributeFn should be fine here as well

instrumentation/github.com/aws/aws-lambda-go/otellambda/lambda.go
```diff
@@ -58,6 +59,8 @@ func (i *instrumentor) tracingBegin(ctx context.Context, eventJSON []byte) (cont
 	spanName := os.Getenv("AWS_LAMBDA_FUNCTION_NAME")
 
 	var attributes []attribute.KeyValue
+	customAttrs := i.configuration.CustomAttributes(eventJSON)
+	attributes = append(attributes, customAttrs...)
```
akats7, line 63: Putting this before the lambda context prevents any of the default attributes from being overwritten, that definitely seems like the right behavior.

akats7:
This is a good update, definitely seems valuable

instrumentation/github.com/aws/aws-lambda-go/otellambda/config.go
matbrito: Sure, that makes sense

instrumentation/github.com/aws/aws-lambda-go/otellambda/lambda.go
matbrito: Totally agree with you. Wanted to ensure the preservation of those defaults attributes.

dmathieu (approved):

Issue #7326, opened 2025-05-15
[otelconf] Cannot use IP address for GRPC endpoint without prepending `http://`


bacherfl:
This issue is based of the discussion here: https://github.com/open-telemetry/opentelemetry-go-contrib/pull/7233#discussion_r2063328273

When providing the otlpConfig.Endpoint with an IP address, rather than a hostname, e.g. `127.0.0.1:4317` instead of `localhost:4317`, the creation of the OTLP exporter will fail due to the endpoint not being parsed successfully (the error is returned bu the `url.ParseRequestURI` here: https://github.com/open-telemetry/opentelemetry-go-contrib/blob/09c7ab5aed4aed07fabe778c99f8b68388c6aafd/otelconf/v0.3.0/log.go#L176

The collector prevents this by automatically prepending `http://` to the provided endpoint (https://github.com/open-telemetry/opentelemetry-collector/blob/main/service/telemetry/internal/migration/normalize.go#L8) before trying to create an exporter via `otelconf`, but other users of this library might run into this issue still

nithinputhenveettil:
@pellared: Can I take this up? I'm interested, as it’s related to #7558.

pellared:
@nithinputhenveettil, I assigned you. Thanks!

PR #7558, opened 2025-07-07
Fix building the mock collector's host:port address in otelconf tests

### LINKED ISSUE #7446: otelconf: Flaky test `Test_otlpGRPCTraceExporter`
Example run: https://github.com/open-telemetry/opentelemetry-go-contrib/actions/runs/15533686054/job/43727904638?pr=7445

```
go test -timeout 60s -short ./otelconf/...
go: downloading go.opentelemetry.io/otel/sdk v1.36.1-0.20250528182540-e25861aa7b3a
go: downloading go.opentelemetry.io/otel/sdk/log/logtest v0.0.0-20250521073539-a85ae98dcedc
?   	go.opentelemetry.io/contrib/otelconf	[no test files]
ok  	go.opentelemetry.io/contrib/otelconf/v0.2.0	0.169s
2025/06/09 11:49:34 reader is not registered
2025/06/09 11:49:34 reader is not registered
--- FAIL: Test_otlpGRPCTraceExporter (25.54s)
    --- FAIL: Test_otlpGRPCTraceExporter/with_TLS_config_and_client_key (10.81s)
        trace_test.go:968: 
            	Error Trace:	D:/a/opentelemetry-go-contrib/opentelemetry-go-contrib/otelconf/v0.3.0/trace_test.go:969
            	            				C:/hostedtoolcache/windows/go/1.23.0/x64/src/runtime/asm_386.s:1393
            	Error:      	Received unexpected error:
            	            	traces export: context deadline exceeded: rpc error: code = DeadlineExceeded desc = context deadline exceeded
        trace_test.go:968: 
            	Error Trace:	D:/a/opentelemetry-go-contrib/opentelemetry-go-contrib/otelconf/v0.3.0/trace_test.go:968
            	Error:      	Condition never satisfied
            	Test:       	Test_otlpGRPCTraceExporter/with_TLS_config_and_client_key
FAIL
FAIL	go.opentelemetry.io/contrib/otelconf/v0.3.0	31.766s
```

nithinputhenveettil:
Fixes #7446 

This PR addresses test flakiness likely caused by the listener binding to an IPv6 loopback ([::1]) on some Windows systems. The resulting URI ([::1]:port) caused issues like invalid URI for request and TLS hostname mismatches.

**Note :** I don’t have access to a Windows environment locally, so this is an untested fix on Windows.

### **Changes:**

1. Always construct the OTLP endpoint as localhost:<port> (instead of using n.Addr().String()), to avoid [::1] and ensure TLS cert hostname matches.
2. Added a ready channel to ensure the test waits until the gRPC server is fully ready before the client attempts to connect.
3. Added early checks to ensure that required certificate and key files exist and are accessible before starting the test.

Let me know how you feel about this approach. Open to suggestions or alternative directions if you'd prefer something different.


pellared:
You need to sign the CLA.

nithinputhenveettil:
@pellared : I have signed the CLA.

dmathieu:
This changes a lot of things. It checks certificates, increases and introduces timeouts. And you mention not having access to a windows machine.
How did you verify that everything you're changing is required?
Have you considered booting a virtual machine to get temporary access to a windows?

nithinputhenveettil:
Hi @dmathieu,
I primarily use a MacBook M4 Pro for development, with a few Linux VMs running on it, but I don’t have a Windows VM locally.

These changes were mainly introduced as I felt they would improve the current test setup. Since the tests were failing intermittently on Windows, I tried to reason about what might be causing the flakiness. My main suspicion is that `n.Addr().String()` can return either an IPv4 or IPv6 address. The existing logic that replaces `127.0.0.1` with `localhost` won't work properly if it returns an IPv6 address and I read that Windows tends to prefer IPv6 in many cases. So, I changed that part of the code to always construct the OTLP endpoint using `localhost:<port>`, which I believe should fix this flakiness on Windows.

Also, I noticed that the server was being started and then used immediately in the test, so I added a mechanism to wait for the server to be ready before continuing, to avoid any possible race conditions.

The other changes - such as slightly increasing the timeout and adding an early check for the local certificate/key files were added as precautionary improvements. The timeout increase isn’t strictly necessary and can be removed. The file check might be helpful in catching any local path resolution issues specific to Windows, in case that’s related to the failures.

Let me know your thoughts. I can remove the timeout and file check changes and keep just the first two fixes in the PR.
Also, now that the tests have passed on Windows in the CI pipeline, I can try to set up a Windows VM and run the tests 5–10 times to check if the flaky behaviour still exists.

Or if you're okay with it, we can merge the PR with the core fixes and keep the issue open for a few days to monitor if the problem comes back during CI runs.

Happy to go either way, open to your suggestions.

nithinputhenveettil:
> This changes a lot of things. It checks certificates, increases and introduces timeouts. And you mention not having access to a windows machine. How did you verify that everything you're changing is required? Have you considered booting a virtual machine to get temporary access to a windows?

@dmathieu : On second thought, I’ve removed the unnecessary timeout increases and the local files check (those would anyway throw an error if something goes wrong.) 

I’ve also updated the gRPC server readiness check to make it more proper and accurate.

I still feel the flakiness might be due to the IPv4/IPv6 address handling or possibly some server readiness issue.

What’s your take on the current code changes?

dmathieu:
Please don't force-push PRs. That breaks the GH review process.

nithinputhenveettil:
All other tests have passed except the changelog check. I don't think we need to add a changelog for this change, as it's only a modification in the test file. Please correct me if I'm mistaken.

otelconf/v0.3.0/trace_test.go
```diff
@@ -1010,12 +1005,29 @@ func Test_otlpGRPCTraceExporter(t *testing.T) {
 			require.NoError(t, err)
 
 			// this is a workaround, as providing 127.0.0.1 resulted in an "invalid URI for request" error
-			tt.args.otlpConfig.Endpoint = ptr(strings.ReplaceAll(n.Addr().String(), "127.0.0.1", "localhost"))
+			//
+			// Always use "localhost:PORT" for the OTLP endpoint, regardless of what n.Addr().String() returns.
+			// This avoids issues with IPv6 addresses (e.g., "[::1]:PORT")
+			_, port, err := net.SplitHostPort(n.Addr().String())
+			require.NoError(t, err)
+			tt.args.otlpConfig.Endpoint = ptr("localhost:" + port)
```
dmathieu, line 1013: This bit isn't about the flakiness. It's about ipv6 support, so it should be in its own PR.

otelconf/v0.3.0/trace_test.go
```diff
@@ -1040,6 +1052,22 @@ type grpcTraceCollector struct {
...
+	net.Listener
+	once  sync.Once
+	ready chan struct{}
+}
+
+// implements the net.Listener.Accept method.
+func (l *notifyListener) Accept() (net.Conn, error) {
+	l.once.Do(func() {
+		// signal server is ready
+		close(l.ready)
```
dmathieu: How does this guarantee the signal is indeed ready? We guarantee we close the channel only once, but we don't wait for the listener to actually accept.

otelconf/v0.3.0/trace_test.go
```diff
@@ -1010,12 +1005,29 @@ func Test_otlpGRPCTraceExporter(t *testing.T) {
...
 			serverOpts, err := tt.grpcServerOpts()
 			require.NoError(t, err)
 
-			startGRPCTraceCollector(t, n, serverOpts)
+			ready := make(chan struct{})
+			wrappedListener := &notifyListener{
+				Listener: n,
+				ready:    ready,
+			}
+			startGRPCTraceCollector(t, wrappedListener, serverOpts)
```
dmathieu: If the readiness is indeed the culprit, how about waiting before returning in `startGRPCTraceCollector` instead? That way we don't have to setup all this every time.

dmathieu (commented):
I ran this on a window 11 virtual machine, but couldn't reproduce the flakiness in either the main branch, or this one.

otelconf/v0.3.0/trace_test.go
pellared: Actually, I think that we need to check why he had `ptr(strings.ReplaceAll(n.Addr().String(), "127.0.0.1", "localhost"))` in the first place. Maybe we should simply add `dns://` prefix (https://pkg.go.dev/google.golang.org/grpc#NewClient) or maybe we have some bug in the OTLP exporters or the way we create them in `otelconf`? CC @bacherfl

otelconf/v0.3.0/trace_test.go
nithinputhenveettil: @dmathieu : Yes, you're right that `Accept()` doesn't mean a client has connected yet. But based on the [grpc-go/server.go#L844](https://github.com/grpc/grpc-go/blob/v1.73.0/server.go#L844), `Serve()` enters the loop and calls `Accept()` only after the server is fully set up and ready to listen.

So my thought is that, if we wrap the `Accept()` method and signal readiness the first time it's called, we can(somewhat) safely assume the server is ready to accept connections, even if no actual connection has arrived yet.

pellared: AFAIK this should not fix the flakiness issue as if I remember correctly even increasing the timeout was not helping. CC @bacherfl

pellared:
@bacherfl, PTAL

otelconf/v0.3.0/trace_test.go
nithinputhenveettil: @dmathieu :
`n.Addr()` can return either an IPv4 or an IPv6 address, right? So, on Windows, if it happens to return an IPv6 address, and the earlier code was expecting specifically 127.0.0.1 (IPv4), that could be a problem, especially since the code was then replacing that with localhost to run the tests.

This is just an assumption based on the random failures we’re seeing,  I’m not fully sure, but it’s possible that the Windows runner in CI sometimes returns IPv4 and other times IPv6.

So, instead of always depending on 127.0.0.1 being returned, I felt it’s safer to handle both cases more gracefully, and that’s why I made this change.

pellared: > This is just an assumption based on the random failures we’re seeing, I’m not fully sure, but it’s possible that the Windows runner in CI sometimes returns IPv4 and other times IPv6.

I think you may be correct here.

However, the question is still why we need to do any replacements and not just use the IPv4 or IPv6 addresses in the config? This looks more like a bug.

otelconf/v0.3.0/trace_test.go
nithinputhenveettil: @dmathieu : The startGRPCTraceCollector function calls srv.Serve in a separate goroutine. So even if we send the readiness signal before returning from that function, or if we signal it inside the same goroutine where Serve() is called, we still can't be completely sure that the server setup is finished and it's actually ready to accept connections.

That’s the main reason I moved the signal into the listener.Accept() method. Even though that doesn’t fully guarantee the server is 100% ready, it felt like the closest and most accurate point to indicate readiness.

In fact, in the first commit, I was sending the ready signal from the same goroutine where Serve() is called.

I don’t have any strong opinion on this, since the current implementation also doesn’t guarantee complete readiness. I’m okay with switching back to the earlier version if you feel that would be better.

otelconf/v0.3.0/trace_test.go
nithinputhenveettil: @pellared : Oh, do you think the IPv6 issue we discussed earlier could be the reason behind the flakiness?

I noticed that this test was skipped on Windows before, were those random failures happening more frequently earlier? Is that the reason why the test was disabled on Windows? Or was it just a few failures that led to it being skipped?

These failures seem very random, and the test itself looks quite straightforward when I read through the code. This PR mainly includes improvements based on a few guesses I had that might help reduce or fix those random failures.

If everyone feels these fixes may not fully solve the flakiness, I’ll try to set up a Windows VM and attempt to reproduce the issue again. 

cc : @dmathieu

otelconf/v0.3.0/trace_test.go
nithinputhenveettil: @pellared Yes, I agree with you on this, we can use IPv4 explicitly in the Listen, and I think that should also help fix the issue (if the flakiness is indeed due to that). I’ll go ahead and push a commit with this change, if you’re okay with it.

otelconf/v0.3.0/trace_test.go
pellared: >  Oh, do you think the IPv6 issue we discussed earlier could be the reason behind the flakiness?

Yes

> I noticed that this test was skipped on Windows before, were those random failures happening more frequently earlier? Is that the reason why the test was disabled on Windows? 

Yes. The tests were seen to be only failing on Windows. 

> I’ll try to set up a Windows VM and attempt to reproduce the issue again.

This would be great.

bacherfl: Good find regarding the IPv6 issue - as @pellared already mentioned, increasing the timeout did not help to fix the flakiness, so `net.Addr()` returning an IPv6 address instead of the expected `127.0.0.1` might indeed be the root cause for this

nithinputhenveettil: @pellared, @bacherfl : Great! I’ll go ahead and remove the unwanted/unasked code changes from this PR, including the readiness check. Then, I’ll rework the IPv4/IPv6 part based on the various comments and idea above. I’ll let you both know once I push the changes.

otelconf/v0.3.0/trace_test.go
dmathieu: The goroutine could signal when the srv is ready, and the main method waits for that signal to return (or timeout).
Exactly like what you're doing, except at the helper level.

otelconf/v0.3.0/trace_test.go
pellared: This looks to be connected with:
- https://github.com/open-telemetry/opentelemetry-go-contrib/issues/7326

otelconf/v0.3.0/trace_test.go
nithinputhenveettil: @pellared: I pushed a commit now to use `tcp4` for the connection to force `IPv4`. Even in this case, we need to manually construct the URL as `"localhost:<port>"`, because `n.Addr()` always returns `127.0.0.1`, not `localhost`. And even if we prepend "http://" to 127.0.0.1, the test will still fail (though `url.ParseRequestURI` will pass), since the certificates are created with CN as `localhost` and not `127.0.0.1`.

So the only option here is to manually construct the URL as `localhost:<port>`, or regenerate the certificates using 127.0.0.1, which would require additional changes in other tests as well.

I feel the current approach is better. Let me know what you all think.

cc : @bacherfl

otelconf/v0.3.0/trace_test.go
nithinputhenveettil: @dmathieu, @pellared  :  I’ve made the changes based on @dmathieu’s suggestion above (hope the current version matches your recommendation). Let me know if you feel anything needs to be removed or updated.

I hope it's okay to keep this readiness check, but happy to update if any of you feel otherwise.

otelconf/v0.3.0/trace_test.go
pellared: Makes sense!
Can you do the same changes in `metric_test.go` and `log_test.go`?
I think we remove the changes in `startGRPCTraceCollector` and the `notifyListener`, they seem unnecessary, rigth?.

nithinputhenveettil: @pellared : I’m slightly inclined towards keeping the readiness check 😄 But you’re right, since these are not directly related to the flakiness, I think we can remove those code changes. I will make the changes.

nithinputhenveettil: @pellared : Made changes in metric and log tests; cleaned up readiness check code. 👍🏽

otelconf/v0.3.0/trace_test.go
```diff
@@ -1006,15 +1000,20 @@ func Test_otlpGRPCTraceExporter(t *testing.T) {
...
+			// n.Addr() always returns 127.0.0.1 instead of localhost.
+			// But our certificate is created with CN as 'localhost', not '127.0.0.1'.
+			// So we have to manually form the endpoint as "localhost:<port>".
+			_, port, err := net.SplitHostPort(n.Addr().String())
+			require.NoError(t, err)
+			tt.args.otlpConfig.Endpoint = ptr("localhost:" + port)
 
 			serverOpts, err := tt.grpcServerOpts()
 			require.NoError(t, err)
-
```
pellared: nit: can you bring back this empty line? :smile:

pellared (approved):
Thanks for your contribution :medal_sports:

otelconf/v0.3.0/trace_test.go
nithinputhenveettil: Done 😁

dmathieu (approved):

PR #7584, opened 2025-07-15
Remove dup semconv support from HTTP packages

dmathieu:


dmathieu:
I know this is a quite large thing, and fixes some bits that were missed before (in echo and gin).

flc1125:
`instrumentation/github.com/labstack/echo/otelecho/echotest_test.go`

Is there an extra "test" in this filename.

instrumentation/net/http/otelhttp/transport.go
```diff
@@ -148,11 +148,7 @@ func (t *Transport) RoundTrip(r *http.Request) (*http.Response, error) {
 		}
 
 		if err == nil {
-			// For handling response bytes we leverage a callback when the client reads the http response
-			readRecordFunc := func(n int64) {
-				t.semconv.RecordResponseSize(ctx, n, metricOpts)
-			}
-
+			readRecordFunc := func(n int64) {}
```
flc1125, line 151: I'm thinking, if it's fixed, maybe move it out as a global variable to reduce GC.

internal/shared/semconv/env.go.tmpl
```diff
@@ -86,11 +67,7 @@ func (s HTTPServer) NetworkTransportAttr(network string) []attribute.KeyValue {
 //
 // If any of the fields in the ResponseTelemetry are not set the attribute will be omitted.
 func (s HTTPServer) ResponseTraceAttrs(resp ResponseTelemetry) []attribute.KeyValue {
-	attrs := CurrentHTTPServer{}.ResponseTraceAttrs(resp)
-	if s.duplicate {
-		return OldHTTPServer{}.ResponseTraceAttrs(resp, attrs)
-	}
-	return attrs
+	return CurrentHTTPServer{}.ResponseTraceAttrs(resp)
```
flc1125, line 70: `CurrentHTTPServer{}`

I think its situation is the same; it can be separated out. Even the variable names can be changed to lowercase.

Since the generated content is under the internal directory in the underlying package, adjusting to lowercase will still ensure backward compatibility.

flc1125 (changes requested):

flc1125:
No other issues, very nice.

instrumentation/net/http/otelhttp/transport.go
dmathieu: I don't know. We'll need this again very soon with #7513.

internal/shared/semconv/env.go.tmpl
dmathieu: As I mention in the description, the goal here is to do only the heavy lifting of removing the flag and support. I will refactor the template in another PR.

instrumentation/net/http/otelhttp/transport.go
flc1125: If there are other plans, please ignore it.😄

flc1125 (approved):

MrAlias (approved):

dmathieu:
> `instrumentation/github.com/labstack/echo/otelecho/echotest_test.go`
> 
> Is there an extra "test" in this filename.

There is both `echotest_test.go` and `echo_test.go`. We may want to make things nicer, but not in this PR.

flc1125:
> There is both `echotest_test.go` and `echo_test.go`. We may want to make things nicer, but not in this PR.

Yeah, you're right. We can focus on resolving just this current PR.

CHANGELOG.md
```diff
@@ -16,6 +16,16 @@ This project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0.htm
 - Add the `WithLoggerProviderOptions`, `WithMeterProviderOptions` and `WithTracerProviderOptions` options to `NewSDK` to allow passing custom options to providers in `go.opentelemetry.io/contrib/otelconf`. (#7552)
 - Added V2 version of AWS EC2 detector `go.opentelemetry.io/contrib/detectors/aws/ec2/v2` due to deprecation of `github.com/aws/aws-sdk-go`. (#6961)
 
+### Removed
+
+- Support for the `OTEL_SEMCONV_STABILITY_OPT_IN` environment variable, and support for semantic conventions v1.20.0 in the modules below. (#7584)
```
```suggestion
- Remove support for the `OTEL_SEMCONV_STABILITY_OPT_IN` environment variable as well as support for semantic conventions v1.20.0 in the modules below. (#7584)
```

PR #7625, opened 2025-07-28
fix: correct connection error and add new configuration

flc1125:
fix: https://github.com/open-telemetry/opentelemetry-go-contrib/actions/runs/16568952679/job/46855849594?pr=7624

The regular part: https://regex101.com/r/isexlM/1


.lycheeignore
```diff
@@ -2,3 +2,4 @@ http://localhost
 http://169.254.169.254/metadata/instance/compute
 https://10.0.0.0
 https://localhost:4317
+^file:\/\/\/home\/runner\/work\/opentelemetry-go-contrib\/opentelemetry-go-contrib\/samplers\/jaegerremote.+$
```
MrAlias: This will only work on the CI runner, it will not work locally.


```suggestion
localhost:14250
```

samplers/jaegerremote/sampler_remote.go
```diff
@@ -70,7 +70,7 @@ type samplerUpdater interface {
 // delegates to it for sampling decisions.
 type Sampler struct {
 	// These fields must be first in the struct because `sync/atomic` expects 64-bit alignment.
-	// Cf. https://github.com/uber/jaeger-client-go/issues/155, https://goo.gl/zW7dgq
+	// Cf. https://github.com/jaegertracing/jaeger-client-go/issues/155, https://goo.gl/zW7dgq
```
dmathieu: For another PR. `goo.gl` shortened links will soon not be available anymore.
See https://developers.googleblog.com/en/google-url-shortener-links-will-no-longer-be-available/

flc1125: Done, modified to the original link.

PR #7635, opened 2025-07-30
chore: several rules from go-critic

mmorel-35:
#### Description

Enable and fixes several rules from [gocritic](https://golangci-lint.run/usage/linters/#gocritic) linter

### DIFF
--- a/.golangci.yml
+++ b/.golangci.yml
@@ -61,7 +61,6 @@ linters:
         - exposedSyncMutex
         - hugeParam
         - paramTypeCombine
-        - preferStringWriter
         - rangeValCopy
         - typeDefFirst
         - typeUnparen
--- a/.golangci.yml
+++ b/.golangci.yml
@@ -62,7 +62,6 @@ linters:
         - hugeParam
         - paramTypeCombine
         - rangeValCopy
-        - typeDefFirst
         - typeUnparen
         - unnamedResult
         - unslice
--- a/.golangci.yml
+++ b/.golangci.yml
@@ -70,7 +70,6 @@ linters:
         - regexpMust
         - typeDefFirst
         - typeUnparen
-        - unlambda
         - unnamedResult
         - unslice
         - whyNoLint
--- a/.golangci.yml
+++ b/.golangci.yml
@@ -57,7 +57,6 @@ linters:
         - appendCombine
         - assignOp
         - elseif
-        - emptyStringTest
         - exitAfterDefer
         - exposedSyncMutex
         - httpNoBody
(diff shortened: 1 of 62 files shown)


.golangci.yml
```diff
@@ -74,13 +91,10 @@ linters:
       rules:
         # https://github.com/mgechev/revive/blob/master/RULES_DESCRIPTIONS.md#blank-imports
         - name: blank-imports
-          disabled: false
```
pellared: Aren't changes from this point to the bottom not related to `gocritic` but mostly `revive`? If so then could it be extracted to a separate PR?

mmorel-35: That's right , I removed it

mmorel-35: I created https://github.com/open-telemetry/opentelemetry-go-contrib/pull/7638 to fix it

detectors/aws/eks/detector.go
```diff
@@ -181,10 +181,7 @@ func (eksUtils eksDetectorUtils) getContainerID() (string, error) {
 	}
 
 	// is this going to stop working with 1.20 when Docker is deprecated?
-	r, err := regexp.Compile(`^.*/docker/(.+)$`)
-	if err != nil {
-		return "", err
-	}
+	r := regexp.MustCompile(`^.*/docker/(.+)$`)
```
MrAlias: Returning an error is preferred to panicking during this method call. If we want to use `MustCompile` can we define `r` as a file level var to have the panic occur on start instead of mid operation?

samplers/probability/consistent/statistical_test.go
```diff
@@ -150,7 +150,7 @@ func TestSamplerStatistics(t *testing.T) {
 
 				failed := countFailures(rand.NewSource(seed))
 
-				if failed != 1 && test.seedIndex < 0 {
+				if failed != 1 && test.seedIndex < 0 { //nolint:gocritic
```
MrAlias: Can you add a reason for the `nolint`?

e.g.


```suggestion
				if failed != 1 && test.seedIndex < 0 { //nolint:gocritic   // [some reason here]
```

detectors/aws/eks/detector.go
mmorel-35: I defined it as a file level var.

samplers/probability/consistent/statistical_test.go
mmorel-35: The suggested refactor was to use a switch case, then the break would not work. I did something else

pellared (approved):

flc1125 (approved):

dmathieu (approved):

MrAlias (approved):

PR #7666, opened 2025-08-05
Modernize

ash2k:
Use https://pkg.go.dev/golang.org/x/tools/gopls/internal/analysis/modernize to update code to new style.

This is similar to https://github.com/open-telemetry/opentelemetry-go/pull/7089.


sahidvelji:
The intrange and copyloopvar linters from golangci-lint could be added in order to detect code where the loop variable is copied or where the integer range loop is not used.
https://golangci-lint.run/usage/linters/

pellared (approved):

ash2k:
@sahidvelji Great suggestion! I've added the linters locally and they've found quite a few places to change. I don't want to increase the size of this PR further, so I'll open a follow up with this change.

flc1125 (approved):

MrAlias (approved):

PR #7682, opened 2025-08-07
Use v2 AWS EC2 detector in `autodetect`

MrAlias:
The v1 detector is going to be deprecated. This future-proofs the `autodetect` package.

This does not include a changelog as this package has yet to be released.


flc1125:
Association: https://github.com/open-telemetry/opentelemetry-go-contrib/pull/7694

PR #7694, opened 2025-08-08
chore: add detectors/aws/ec2/v2 to versions.yaml

### LINKED ISSUE #7683: Add `detectors/aws/ec2/v2` to `versions.yaml`
The module has been added since the last release but has not been added to our versioning system.

Update `versions.yaml` to include this `v2` module in the `experimental-detectors` module set.

Whitea029:
#### Description:

add detectors/aws/ec2/v2 to versions.yaml

#### Related Issue:

fix #7683


versions.yaml
```diff
@@ -15,6 +15,10 @@ module-sets:
       - go.opentelemetry.io/contrib/detectors/aws/ec2
       - go.opentelemetry.io/contrib/detectors/aws/ecs
       - go.opentelemetry.io/contrib/detectors/aws/eks
+  stable-v2:
+    version: v2.0.0
```
flc1125, line 19: I believe we don't need to migrate to the `v2.x` version, as it already has an actual `v2` directory.

I recommend following the current `v2` directory setup officially adopted by Go, which is still based on the `v1.x` foundation while adding v2 directory functionality.

The `v2.x` version, as I understand it, should represent a major milestone.

flc1125:
For reference: Created a reference document titled "Validation of Multi-Version Releases Based on v0.x," concluding that releasing a `v2.x` version is still necessary to comply with Go's dependency management.

https://github.com/flc1125/go-module-multi-version-validation?tab=readme-ov-file#verification-result

flc1125 (approved):

pellared (approved):

akats7 (approved):

PR #7707, opened 2025-08-11
feat: support empty option

darren-west:
### Summary
Support an `EmptyOption` so that options can easily be wrapped and support extensibility. 

e.g. 

```
type CustomOption struct{
	otelgrpc.EmptyOption
}
```go


MrAlias:
Thanks for your interest in the project. Please open an enhancement issue describing the proposal you would like to make. The code proposed here cannot be evaluated based on the lack of CLA fulfillment and if it were evaluated it likely is not achieving your desired solution.

Be sure to fill out the full enhancement issue template when you do open the issue. Thanks again for your interest in the project.

PR #7726, opened 2025-08-12
chore: enable exposedSyncMutex rule from go-critic

mmorel-35:
#### Description

Enable and fixes more rules from go-critic:

* https://go-critic.com/overview.html#exposedsyncmutex


propagators/aws/xray/idgenerator.go
```diff
@@ -19,16 +19,16 @@ import (
 
 // IDGenerator is used for generating a new traceID and spanID.
 type IDGenerator struct {
-	sync.Mutex
+	mu         sync.Mutex
```
flc1125, line 23: Since it's a stable [version](https://github.com/open-telemetry/opentelemetry-go-contrib/blob/8b2ef6949ee321f413d0fd606f44453ab63223c8/propagators/aws/version.go#L10), although this is an extremely low-probability event, the changes in this section are not backward-compatible adjustments.

We cannot ensure that there are applications that call `Lock()`/`Unlock()` methods.

---

I recommend not making changes here (i.e., resolve it with `nolint`).

samplers/jaegerremote/sampler_remote.go
```diff
@@ -72,7 +72,7 @@ type Sampler struct {
 	// Cf. https://github.com/jaegertracing/jaeger-client-go/issues/155, https://pkg.go.dev/sync/atomic#pkg-note-BUG
 	closed int64 // 0 - not closed, 1 - closed
 
-	sync.RWMutex // used to serialize access to samplerConfig.sampler
+	mu sync.RWMutex // used to serialize access to samplerConfig.sampler
```
flc1125, line 76: This part is also similar (not backward compatible), but what makes it special is that it is not yet a stable [version](https://github.com/open-telemetry/opentelemetry-go-contrib/blob/8b2ef6949ee321f413d0fd606f44453ab63223c8/samplers/jaegerremote/version.go#L8).

> For reference only, evaluation required. @open-telemetry/go-approvers 

PS: For reference only: I personally prefer to continue with the changes in this PR.

propagators/aws/xray/idgenerator.go
mmorel-35: I have another idea to suggest, what if the mutex methods are exposed with a depreciation comment so this help the user to known, they should not use it in a smooth way ?

propagators/aws/xray/idgenerator.go
```diff
@@ -19,16 +19,31 @@ import (
...
 	randSource *rand.Rand
 }
 
 var _ sdktrace.IDGenerator = &IDGenerator{}
 
+// Deprecated: mutex is internal and shall not be used.
+func (gen *IDGenerator) Lock() {
```
flc1125: This looks like a very good way.👍

samplers/jaegerremote/sampler_remote.go
```diff
@@ -95,11 +95,27 @@ func New(
 	return sampler
 }
 
+
+// Deprecated: mutex is internal and shall not be used.
```
flc1125, line 102: Since this is an experimental version, this section has been temporarily marked for now, pending feedback from other approvers.

flc1125 (approved):
Overall, it's very good, with an experimental version section (for reference only), pending decision.

propagators/aws/xray/idgenerator.go
```diff
@@ -19,16 +19,34 @@ import (
...
 type IDGenerator struct {
-	sync.Mutex
+	mu         sync.Mutex
 	randSource *rand.Rand
 }
 
 var _ sdktrace.IDGenerator = &IDGenerator{}
 
+// Lock is used to lock the IDGenerator for concurrent use.
+// Deprecated: mutex is internal and shall not be used.
```
flc1125:
```suggestion
// Lock is used to lock the IDGenerator for concurrent use.
//
// Deprecated: mutex is internal and shall not be used.
```

Deprecated A line needs to be retained, otherwise the deprecation identifier cannot be recognized by Go documentation and IDE. https://tip.golang.org/doc/comment#deprecations

propagators/aws/xray/idgenerator.go
```diff
@@ -19,16 +19,34 @@ import (
...
+func (gen *IDGenerator) Lock() {
+	gen.mu.Lock()
+}
+
+// TryLock attempts to lock the IDGenerator for concurrent use.
+// Deprecated: mutex is internal and shall not be used.
```
flc1125:
```suggestion
// TryLock attempts to lock the IDGenerator for concurrent use.
//
// Deprecated: mutex is internal and shall not be used.
```

propagators/aws/xray/idgenerator.go
```diff
@@ -19,16 +19,34 @@ import (
...
+func (gen *IDGenerator) TryLock() bool {
+	return gen.mu.TryLock()
+}
+
+// Unlock is used to unlock the IDGenerator for concurrent use.
+// Deprecated: mutex is internal and shall not be used.
```
flc1125:
```suggestion
// Unlock is used to unlock the IDGenerator for concurrent use.
//
// Deprecated: mutex is internal and shall not be used.
```

samplers/jaegerremote/sampler_remote.go
```diff
@@ -95,11 +95,29 @@ func New(
 	return sampler
 }
 
+// Lock is used to lock the Sampler for concurrent use.
+// Deprecated: mutex is internal and shall not be used.
```
flc1125:
```suggestion
// Lock is used to lock the Sampler for concurrent use.
//
// Deprecated: mutex is internal and shall not be used.
```

samplers/jaegerremote/sampler_remote.go
```diff
@@ -95,11 +95,29 @@ func New(
...
+func (s *Sampler) Lock() {
+	s.mu.Lock()
+}
+
+// TryLock attempts to lock the Sampler for concurrent use.
+// Deprecated: mutex is internal and shall not be used.
```
flc1125:
```suggestion
// TryLock attempts to lock the Sampler for concurrent use.
//
// Deprecated: mutex is internal and shall not be used.
```

samplers/jaegerremote/sampler_remote.go
```diff
@@ -95,11 +95,29 @@ func New(
...
+func (s *Sampler) TryLock() bool {
+	return s.mu.TryLock()
+}
+
+// Unlock is used to unlock the Sampler for concurrent use.
+// Deprecated: mutex is internal and shall not be used.
```
flc1125:
```suggestion
// Unlock is used to unlock the Sampler for concurrent use.
//
// Deprecated: mutex is internal and shall not be used.
```

propagators/aws/xray/idgenerator.go
mmorel-35: Thank you, I didn't know that :) !

samplers/jaegerremote/sampler_remote.go
MrAlias: +1 for removing the deprecated functions after they are released based on the fact this is not stable.

propagators/aws/xray/idgenerator.go
```diff
@@ -19,16 +19,37 @@ import (
...
 	randSource *rand.Rand
 }
 
 var _ sdktrace.IDGenerator = &IDGenerator{}
 
+// Lock is used to lock the IDGenerator for concurrent use.
+//
+// Deprecated: mutex is internal and shall not be used.
+func (gen *IDGenerator) Lock() {
+	gen.mu.Lock()
```
MrAlias: I think we can treat access to the underlying mutex as a bug, something we should never have exposed.

Given that, I think we can replace these methods with no-op methods that do nothing. It will resolve the bug but also ensure backwards compatibility.

akats7: Agreed, as long as the type signature doesn't change. I don't think there's a reason to continue exposing access to the mutex

pellared: ~~I am not convinced that doing a no-op is good. If people rely on these methods then they would have most probably runtime race conditions which may be far worse than getting an compilation error.~~

~~I suggest to keep it as it is for now and just remove the methods in the future release (we still need an issue to not forget about removing these methods).~~

Not valid comment: The module is stable. We cannot remove these methods in future.

pellared (commented):
Can you please add deperecation entries in the changelog?

propagators/aws/xray/idgenerator.go
akats7: The potential for misuse seems extremely low, we're not exposing the rand that the lock is for, so any usage of the lock would have to be pretty egregious

MrAlias: If users are relying on these methods to reach into the type I see it as they are relying on a bug to do something we don't want them to do. Keeping it as is continues this misbehavior.

pellared: I discussed it offline with Tyler.
The outcome is that we want to have "no-op + deprecation" for modules that are already stable as these methods should never be removed.

CHANGELOG.md
```diff
@@ -27,6 +27,7 @@ This project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0.htm
   - `go.opentelemetry.io/contrib/bridges/otelzap`
 - The `Severity` type from `go.opentelemetry.io/contrib/processors/minsev` now implements the `fmt.Stringer`, `encoding.TextMarshaler`, `encoding.TextUnmarshaler`, `encoding.TextAppender`, `json.Marshaler`, and `json.Unmarshaler` interfaces. (#7652)
 - The `SeverityVar` type from `go.opentelemetry.io/contrib/processors/minsev` now implements the `fmt.Stringer`, `encoding.TextMarshaler`, `encoding.TextUnmarshaler`, and `encoding.TextAppender` interfaces. (#7652)
+-- Deprecate direct exposure and usage of mutex objects via structs: introduce equivalent methods for locking/unlocking, and mark mutex fields as deprecated, specially for `go.opentelemetry.io/contrib/samplers/jaegerremote.Sampler` and `go.opentelemetry.io/contrib/propagators/aws/xray.IDGenerator`. (#7726)
```
pellared: This entry is not correctly formatted and I also think it should be in `Deprecated` section

pellared:
Please notice https://github.com/open-telemetry/opentelemetry-go/blob/main/CONTRIBUTING.md#how-to-send-pull-requests:

> Avoid rebasing and force-pushing to your branch to facilitate reviewing the pull request. Rewriting Git history makes it difficult to keep track of iterations during code review. All pull requests are squashed to a single commit upon merge to `main`.

pellared (approved):

dmathieu (approved):

CHANGELOG.md
```diff
@@ -57,6 +57,7 @@ The next release will require at least [Go 1.24].
 - `Extract` and `Inject` in `go.opentelemetry.io/contrib/instrumentation/google.golang.org/grpc/otelgrpc` are deprecated.
   These functions were initially exposed in the public API, but are now considered unnecessary. (#7689)
 - The `go.opentelemetry.io/contrib/detectors/aws/ec2` package is deprecated, use `go.opentelemetry.io/contrib/detectors/aws/ec2/v2` instead. (#7725)
+- Deprecate direct exposure and usage of mutex objects via structs: introduce equivalent methods for locking/unlocking, and mark mutex fields as deprecated, specially for `go.opentelemetry.io/contrib/samplers/jaegerremote.Sampler` and `go.opentelemetry.io/contrib/propagators/aws/xray.IDGenerator`. (#7726)
```
dmathieu: This needs to be moved higher up into the unreleased section of the changelog.

propagators/aws/xray/idgenerator.go
```diff
@@ -19,16 +19,36 @@ import (
 
 // IDGenerator is used for generating a new traceID and spanID.
 type IDGenerator struct {
-	sync.Mutex
+	mu         sync.Mutex
```
pellared, line 23: The field needs to be kept as removing an exported field is a breaking change.
The field should be marked as deprecated.

samplers/jaegerremote/sampler_remote.go
pellared, line 76: The field needs to be kept as removing an exported field is a breaking change.
The field should be marked as deprecated.

propagators/aws/xray/idgenerator.go
mmorel-35: Aren't the lign added between L28 and L47 here to handle deprecation ?

samplers/jaegerremote/sampler_remote.go
mmorel-35: Aren't the lign added between L98 and L117 here to handle deprecation ?

pellared: This only handles deprecation of the "embedded methods".
But after making this field not embedded this would no longer compile:

```go
package main

import "go.opentelemetry.io/contrib/samplers/jaegerremote"

func main() {
	s := new(jaegerremote.Sampler)
	s.RWMutex.Lock()
}

```

Thus this is a breaking change from Go perspective.
More: https://github.com/open-telemetry/opentelemetry-go/issues/6274

propagators/aws/xray/idgenerator.go
pellared: https://github.com/open-telemetry/opentelemetry-go-contrib/pull/7726/files#r2334121098

propagators/aws/xray/idgenerator.go
```diff
@@ -19,16 +19,37 @@ import (
 
 // IDGenerator is used for generating a new traceID and spanID.
 type IDGenerator struct {
-	sync.Mutex
+	sync.Mutex // Deprecated: mutex is internal and shall not be used.
+	mu         sync.Mutex
```
pellared, line 23: Shouldn't this be

```suggestion
	sync.Mutex // Deprecated: mutex is internal and shall not be used.
```

?

samplers/jaegerremote/sampler.go
```diff
@@ -209,6 +209,7 @@ func (*guaranteedThroughputProbabilisticSampler) Description() string {
 // perOperationSampler is a delegating sampler that applies guaranteedThroughputProbabilisticSampler
 // on a per-operation basis.
 type perOperationSampler struct {
+	sync.RWMutex // Deprecated: mutex is internal and shall not be used.
```
pellared: This is not needed for an internal struct.

mmorel-35: Indeed

samplers/jaegerremote/sampler_remote.go
```diff
@@ -71,8 +71,9 @@ type Sampler struct {
 	// These fields must be first in the struct because `sync/atomic` expects 64-bit alignment.
 	// Cf. https://github.com/jaegertracing/jaeger-client-go/issues/155, https://pkg.go.dev/sync/atomic#pkg-note-BUG
 	closed int64 // 0 - not closed, 1 - closed
-
-	sync.RWMutex // used to serialize access to samplerConfig.sampler
+	//nolint:gocritic // ongoing deletion
+	sync.RWMutex // Deprecated: mutex is internal and shall not be used.
+	mu sync.RWMutex // used to serialize access to samplerConfig.sampler
```
pellared, line 76:
```suggestion
	sync.RWMutex // Deprecated: mutex is internal and shall not be used.
```

propagators/aws/xray/idgenerator.go
pellared, line 23: Is there any reason for duplicating the fields?

```suggestion
	sync.Mutex // Deprecated: mutex is internal and shall not be used.
```

mmorel-35: I just wanted to fix exposedSyncMutex... this is a nightmare 😭. Please take it from here. I don’t want to here about this egg or chicken problem!

https://go-critic.com/overview.html#exposedsyncmutex

pellared: Cannot you add a `//nolint` that this is required for backwards compatibility?

PR #7727, opened 2025-08-12
chore: enable rangeValCopy rule from go-critic

mmorel-35:
#### Description

Enable and fixes more rules from go-critic:

* https://go-critic.com/overview.html#rangevalcopy

### DIFF
--- a/.golangci.yml
+++ b/.golangci.yml
@@ -59,7 +59,6 @@ linters:
         - exitAfterDefer
         - exposedSyncMutex
         - hugeParam
-        - rangeValCopy
         - unnamedResult
         - whyNoLint
       enable-all: true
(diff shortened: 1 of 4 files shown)


flc1125:
Thank you for your contribution, Could you briefly explain the benefits it can bring? Currently, it seems to make the code readability more complicated.

mmorel-35:
Hi ! 
Of course:

> ### rangeValCopy
> -----------------------------------------------------------------
> #### [ performance ]
>Detects loops that copy big objects during each iteration.
>
>Suggests to use index access or take address and make use pointer instead.


So if you prefer to keep it that way, there can be an exclusion but only for tests for example ?

flc1125:
> Hi ! Of course:
> 
> > ### rangeValCopy
> > 
> > #### [ performance ]
> > Detects loops that copy big objects during each iteration.
> > Suggests to use index access or take address and make use pointer instead.
> 
> So if you prefer to keep it that way, there can be an exclusion but only for tests for example ?

This looks like it would be better for performance.

flc1125 (approved):

MrAlias (approved):

PR #7728, opened 2025-08-12
chore: enable nolintlint and whyNoLint rule from go-critic

mmorel-35:
#### Description

Enable and fixes more rules from go-critic with require-explanation and require-specific from nolintlint

* https://go-critic.com/overview.html#whynolint
* https://golangci-lint.run/docs/linters/configuration/#nolintlint


samplers/probability/consistent/tracestate.go
```diff
@@ -92,7 +92,7 @@ func isUCAlpha(r byte) bool {
 	return r >= 'A' && r <= 'Z'
 }
 
-func parseOTelTraceState(ts string, isSampled bool) (otelTraceState, error) { // nolint: revive
+func parseOTelTraceState(ts string, isSampled bool) (otelTraceState, error) { //nolint:revive // ignore linter
```
pellared, line 95: Why we are ignoring the linter here?

samplers/probability/consistent/tracestate.go
```diff
@@ -201,7 +201,7 @@ func parseNumber(key, input string, maximum uint8) (uint8, error) {
 		return maximum + 1, parseError(key, strconv.ErrRange)
 	}
 	// `value` is strictly less then the uint8 maximum. This cast is safe.
-	return uint8(value), nil // nolint: gosec
+	return uint8(value), nil //nolint:gosec // ignore linter
```
pellared, line 204: You should move the comment from line 203.

pellared (commented):
Is there any reason to not use the https://golangci-lint.run/docs/linters/configuration/#nolintlint linter instead? Then we can also set `require-specific: true`.

mmorel-35:
@pellared ,
Both can be used. I can add it here if you want ?

pellared:
> I can add it here if you want ?

Go for it. Especially that there is no second approval yet.

pellared (approved):

pellared (approved):

prestonvasquez (approved):

Issue #7591, opened 2025-07-17
AWS Lambda detector: incorrect faas.max_memory value


sahidvelji:
### Description

According to the [faas semconv](https://opentelemetry.io/docs/specs/semconv/registry/attributes/faas/#faas-max-memory), faas.max_memory should be reported in bytes, not in MB.

The semconv explicitly states that 

> On AWS Lambda, the environment variable AWS_LAMBDA_FUNCTION_MEMORY_SIZE provides this information (which must be multiplied by 1,048,576).

https://opentelemetry.io/docs/specs/semconv/resource/faas/#:~:text=On%20AWS%20Lambda%2C%20the%20environment%20variable%20AWS_LAMBDA_FUNCTION_MEMORY_SIZE%20provides%20this%20information%20(which%20must%20be%20multiplied%20by%201%2C048%2C576).

- [In the implementation](https://github.com/open-telemetry/opentelemetry-go-contrib/blob/c1bbaea4dc2b671fa5d1599d53f6d883aab38bdf/detectors/aws/lambda/detector.go#L68), the maxMemory must be multiplied by 1,048,576 before appending to attrs.
- In the [AWS Lambda Detector README](https://github.com/open-telemetry/opentelemetry-go-contrib/tree/bridges/otelslog/v0.10.0/detectors/aws/lambda), the faas.max_memory example value is 128, which should instead be multiplied by 1,048,576.
- The [detector_test.go file](https://github.com/open-telemetry/opentelemetry-go-contrib/blob/c1bbaea4dc2b671fa5d1599d53f6d883aab38bdf/detectors/aws/lambda/detector_test.go#L32) should also be updated like the README.

Supporting documentation: 

> AWS_LAMBDA_FUNCTION_MEMORY_SIZE – The amount of memory available to the function in MB.

https://docs.aws.amazon.com/lambda/latest/dg/configuration-envvars.html#:~:text=AWS_LAMBDA_FUNCTION_MEMORY_SIZE%20%E2%80%93%20The%20amount%20of%20memory%20available%20to%20the%20function%20in%20MB.

dmathieu:
cc @akats7 as code owner.

ishaish103:
hey @akats7 
i would like to work on this issue, can you assign it to me?

PR #7745, opened 2025-08-14
Update lambda detector to use bytes instead of MB

akats7:
Fixes: #7591


detectors/aws/lambda/detector.go
```diff
@@ -23,6 +23,7 @@ const (
 	lambdaFunctionVersionEnvVar = "AWS_LAMBDA_FUNCTION_VERSION"
 	lambdaLogStreamNameEnvVar   = "AWS_LAMBDA_LOG_STREAM_NAME"
 	lambdaMemoryLimitEnvVar     = "AWS_LAMBDA_FUNCTION_MEMORY_SIZE"
+	MiB                         = 1 << 20
```
sahidvelji: Does this need to be an exported const?

akats7: Yeah probably not, I exported it since it could be helpful in tests as it was in otellambda but it’s fine to just have users redefine it

sahidvelji: If someone uses this detector, they wouldn't need to worry about the conversion since this package handles it.

akats7: Yeah it would be for tests, when defining the wantValue

flc1125: I tend to add data types.

For reference, the underlying Go language generally uses `int64`. Here is an example: https://github.com/golang/go/blob/a8564bd412d4495a6048f981d30d4d7abb1e45a7/src/runtime/mprof.go#L897

Secondly, most of the time, the underlying Go language returns such data in byte units.

flc1125: It is recommended to add some comments and include the addresses of some references.

> [4] faas.max_memory: It’s recommended to set this attribute since e.g. too little memory can easily stop a Java AWS Lambda function from working correctly. On AWS Lambda, the environment variable AWS_LAMBDA_FUNCTION_MEMORY_SIZE provides this information (which must be multiplied by 1,048,576).

https://opentelemetry.io/docs/specs/semconv/resource/faas/

---

Since it's part of the semantic convention, I think it can be used not only for unit testing but perhaps by users as well, so I suggest adding unit tests.

MrAlias: +1 for not exporting. We can redefine a const in tests.

akats7: The semconv package expects an int for the maxMemory attribute

akats7: I'll just leave it unexported for simplicity

flc1125 (approved):
There is a conflict that needs to be resolved.

dmathieu (approved):

CHANGELOG.md
```diff
@@ -24,6 +24,7 @@ The next release will require at least [Go 1.24].
 ### Changed
 
 - Change the default span name to be `GET /path` so it complies with the HTTP semantic conventions in `go.opentelemetry.io/contrib/instrumentation/github.com/gorilla/mux/otelmux`. (#7551)
+- Change the faas.max_memory unit to be bytes instead of MB to comply with the semantic conventions in `"go.opentelemetry.io/contrib/detectors/aws/lambda`. (#7745)
```
```suggestion
- Change the `faas.max_memory` unit to be bytes instead of MB to comply with the semantic conventions in `go.opentelemetry.io/contrib/detectors/aws/lambda`. (#7745)
```

pellared:
`TestWrapEndToEnd` fails

PR #7748, opened 2025-08-14
Clamp `minsev` `Severity.Severity()` conversion

### LINKED ISSUE #7662: Clamp `Severity` Method to Defined Range of `log.Severity`
Modifying the [`Severity` method](https://github.com/open-telemetry/opentelemetry-go-contrib/blob/8565a662cdd83ab3545ac4556b80ce320bdffb8f/processors/minsev/severity.go#L67-L73) in `severity.go` to return a clamped `log.Severity` value:

- If the severity is less than `SeverityTrace1`, it should return `log.SeverityTrace1`
- If the severity is greater than `SeverityFatal4`, it should return `log.SeverityFatal4`.

### Motivation:

It is possible for a user to create a `Severity` outside of the defined `const` values (e.g. `Severity(100)`). When this happens the user will get a `log.SeverityUndefined` returned from `Severity.Severity()`.

Assuming the severity levels a user expects are the same as `minsev` (lower severity is less severe, and greater severity is more severe), we can provide a closer approximation to what the user wants. Translating a severity outside allowed bounds of `log.Severity` to one at the closest bound will have a greater chance of achieving what the user wants instead of the current behavior.

### Proposed Change:

```go
// Severity returns the receiver translated to a [log.Severity].
//
// It implements [Severitier].
func (s Severity) Severity() log.Severity {
	switch {
	case s < SeverityTrace1:
		return log.SeverityTrace1
	case s > SeverityFatal4:
		return log.SeverityFatal4
	}

	const offset = int(log.SeverityTrace1) - int(SeverityTrace1)
	return log.Severity(int(s) + offset)
}
```

### Alternates Considered

Leave the method as is.

MrAlias:
Resolve #7662

### DIFF
--- a/processors/minsev/severity.go
+++ b/processors/minsev/severity.go
@@ -87,39 +87,21 @@ const (
 //
 // It implements [Severitier].
 func (s Severity) Severity() log.Severity {
-	// Unknown defaults to log.SeverityUndefined.
-	//
-	// TODO: return a clamped log.Severity. If s is less than
-	// SeverityTrace1, return log.SeverityTrace1, if s is greater than
-	// SeverityFatal4, return log.SeverityFatal4.
-	return translations[s]
-}
+	// Clamp to the defined range of log.Severity values. This provides a
+	// closer approximation for out-of-range values instead of returning
+	// log.SeverityUndefined.
+	switch {
+	case s < SeverityTrace1:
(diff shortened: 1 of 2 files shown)


pellared:
CC @sahidvelji

sahidvelji (approved):

PR #7786, opened 2025-08-22
Fix EKS detector container ID parsing for Kubernetes/containerd

alok87:
- Updated containerIDRegex to support both Docker and Kubernetes cgroup formats
- Modified getContainerID() to properly extract container ID from regex capture groups
- Fixes issue where EKS detector fails with 'cannot read containerID from file /proc/self/cgroup'
- Now supports cgroup paths like /kubepods/burstable/pod<pod-id>/<container-id>

Error I had encountered 
```
W0821 12:39:45.604746     127 telemetry.go:115] EKS detector failed: getContainerID() error: cannot read containerID from file /proc/self/cgroup
```

### DIFF
--- a/detectors/aws/eks/detector.go
+++ b/detectors/aws/eks/detector.go
@@ -55,8 +55,8 @@ var _ resource.Detector = (*resourceDetector)(nil)
 // Compile time assertion that eksDetectorUtils implements the detectorUtils interface.
 var _ detectorUtils = (*eksDetectorUtils)(nil)
 
-// is this going to stop working with 1.20 when Docker is deprecated?
-var containerIDRegex = regexp.MustCompile(`^.*/docker/(.+)$`)
+// Updated regex to support both Docker and Kubernetes/containerd cgroup formats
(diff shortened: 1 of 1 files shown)


pellared (commented):
Misses unit tests.

detectors/aws/eks/detector.go
```diff
@@ -55,8 +55,8 @@ var _ resource.Detector = (*resourceDetector)(nil)
 // Compile time assertion that eksDetectorUtils implements the detectorUtils interface.
 var _ detectorUtils = (*eksDetectorUtils)(nil)
 
-// is this going to stop working with 1.20 when Docker is deprecated?
-var containerIDRegex = regexp.MustCompile(`^.*/docker/(.+)$`)
+// Updated regex to support both Docker and Kubernetes/containerd cgroup formats
```
dmathieu, line 58: Comments shouldn't be about things that have been updated (GIT has history for that), but actual issues.

dmathieu (commented):
As @pellared mentioned, this needs tests.

PR #7808, opened 2025-08-27
Upgrade to go.opentelemetry.io/otel dependencies

MrAlias:
Resolve #7807 
Part of https://github.com/open-telemetry/opentelemetry-go/issues/7267

- Upgrade semconv packages to v1.37.0
- Handle deprecation of prometheus exporter options (ignore deprecation for configuration compatibility)

PR #7813, opened 2025-08-28
Cleanup remnants of the `http/dup` flag in shared semconv

dmathieu:
We removed support for producing old semconv a few weeks ago. This PR now cleans up the package to have a single struct that emits data rather than a `CurrentHTTPX` one used under the hood.


internal/shared/semconv/client.go.tmpl
```diff
@@ -0,0 +1,305 @@
...
+	"net/http"
+	"reflect"
+	"slices"
+	"strconv"
+	"strings"
+
+	"go.opentelemetry.io/otel/attribute"
+	"go.opentelemetry.io/otel/codes"
+	"go.opentelemetry.io/otel/metric"
+	"go.opentelemetry.io/otel/semconv/v1.36.0"
```
flc1125: Do we need to merge it together with the version that contains [this PR](https://github.com/open-telemetry/opentelemetry-go-contrib/pull/7808)?

dmathieu: We can do either this PR, or the other one first. But I think they are better as separate ones.

Issue #7685, opened 2025-08-07
Remove the deprecated `go.opentelemetry.io/contrib/detectors/aws/ec2` module


MrAlias:
Blocked by #7684 
Blocked by https://github.com/open-telemetry/opentelemetry-go-contrib/milestone/32

- [ ] Remove the module code
- [ ] Remove references to the module in `versions.yaml`

PR #7841, opened 2025-09-01
chore: remove deprecated EC2 detector

flc1125:
Resolve https://github.com/open-telemetry/opentelemetry-go-contrib/issues/7685


CHANGELOG.md
```diff
@@ -11,6 +11,7 @@ This project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0.htm
 ### Removed
 
 - Drop support for [Go 1.23]. (#7831)
+- Remove the deprecated module `go.opentelemetry.io/contrib/detectors/aws/ec2`, please use `go.opentelemetry.io/contrib/detectors/aws/ec2/v2` instead. (#7841)
```
dmathieu: How about mentioning that previous releases aren't removed. This module just won't be updated again.

flc1125: I made an adjustment and moved it to `Deprecated`. PTAL

CHANGELOG.md
```diff
@@ -8,6 +8,10 @@ This project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0.htm
 
 ## [Unreleased]
 
+### Deprecated
+
+- Deprecate the `go.opentelemetry.io/contrib/detectors/aws/ec2` module, please use `go.opentelemetry.io/contrib/detectors/aws/ec2/v2` instead. (#7841)
```
pellared: This should be "Remove" and in "Removed" section

CHANGELOG.md
pellared: I prefer the previous changlog entry

> How about mentioning that previous releases aren't removed.

This is rather obvious (the nature of Go modules). Moreover, I would be afraid that people would assume that we are  supporting these previously released modules which is not true.

CHANGELOG.md
pellared: See https://github.com/open-telemetry/opentelemetry-go-contrib/pull/7841#discussion_r2314477937

CHANGELOG.md
flc1125: For reference, since the previous version was already marked as deprecated, having this version exist as a removal might be more appropriate.

https://github.com/open-telemetry/opentelemetry-go-contrib/blob/35d7a691d0a4915d8068a50d191b3a6e4b1fa476/CHANGELOG.md?plain=1#L58C1-L59C1

dmathieu (approved):

pellared (approved):

PR #7845, opened 2025-09-02
zpages: deflake TestSpanProcessorFuzzer by aggregating latency samples

### LINKED ISSUE #7819: Flaky test in zpages span processor
### Description

We have a flaky test in the zpages module's span processor tests.

```
--- FAIL: TestSpanProcessorFuzzer (0.06s)
    spanprocessor_test.go:111: 
        	Error Trace:	D:/a/opentelemetry-go-contrib/opentelemetry-go-contrib/zpages/spanprocessor_test.go:111
        	Error:      	"0" is not greater than or equal to "1"
        	Test:       	TestSpanProcessorFuzzer
```

See https://github.com/open-telemetry/opentelemetry-go-contrib/actions/runs/17317518500/job/49163181772

tdn21:
## What
- Deflake `TestSpanProcessorFuzzer` by asserting on the aggregate count of latency samples across all buckets for testSpan1 and testSpan2, instead of checking a specific bucket index.
- Test-only change; no production code modified.
- Fixes #7819 

## Why
- Span durations are highly non-deterministic due to scheduling/timing variance. Asserting a particular bucket (e.g., index 1) is brittle and can be empty even when sampling is correct, causing flakes.
  - Rationale for aggregating across all buckets:
    - The fuzzer’s purpose is to stress concurrency and basic sampling behavior (start/end paths, active span cleanup, and that some non-error spans are sampled), not to verify distribution across specific latency buckets.
    - Detailed bucket-boundary behavior should be validated by separate, deterministic tests that control span duration (synthetic start/end times), not by this concurrency fuzzer.

### DIFF
--- a/zpages/spanprocessor_test.go
+++ b/zpages/spanprocessor_test.go
@@ -104,11 +104,21 @@ func TestSpanProcessorFuzzer(t *testing.T) {
 
 	assert.Empty(t, zsp.activeSpans("testSpan1"))
 	assert.GreaterOrEqual(t, len(zsp.errorSpans("testSpan1")), 1)
-	assert.GreaterOrEqual(t, len(zsp.spansByLatency("testSpan1", 1)), 1)
+	// Count latency samples across all buckets instead of a single bucket to avoid flakes
+	numLatencySamples1 := 0
+	for i := range defaultBoundaries.numBuckets() {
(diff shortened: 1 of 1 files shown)


pellared:
@tdn21, thank you for your contribution!

PR #7909, opened 2025-09-16
test: remove unused attribute results in HTTP server benchmarks

flc1125:
Remove unused variables.


dmathieu (approved):

instrumentation/net/http/otelhttp/internal/semconv/bench_test.go
```diff
@@ -43,6 +39,6 @@ func BenchmarkHTTPServerRequest(b *testing.B) {
 	b.ReportAllocs()
 	b.ResetTimer()
 	for range b.N {
-		benchHTTPServerRequestResults = serv.RequestTraceAttrs("", req, RequestTraceAttrsOpts{})
+		_ = serv.RequestTraceAttrs("", req, RequestTraceAttrsOpts{})
```
MrAlias, line 42: Does the compiler now optimize away this function call because the result is never used?

Having the result stored outside of the call scope ensures this optimization is not made. It is not clear that is still the case.

flc1125: This piece of knowledge is subtle and worth understanding.

After some research, I believe: regardless of whether there is potential for optimization by the optimizer, maintaining the current state can reduce the likelihood of optimizer-related issues.

PR #7977, opened 2025-10-03
otelconf: add support for propagators configuration

### LINKED ISSUE #6712: config: add support for configuring propagators

codeboten:
This is reviving https://github.com/open-telemetry/opentelemetry-go-contrib/pull/6727 to add support for configuration of propagators in otelconf.

Fixes #6712


otelconf/v0.3.0/propagator_test.go
```diff
@@ -0,0 +1,195 @@
...
+			name: "empty composite",
+			cfg: configOptions{
+				opentelemetryConfig: OpenTelemetryConfiguration{
+					Propagator: &Propagator{
+						Composite: []*string{},
+					},
+				},
+			},
+			want:    propagation.NewCompositeTextMapPropagator(propagation.TraceContext{}, propagation.Baggage{}),
+			wantErr: false,
```
pellared, line 142: Is this following the specification? Should it not be a noop propagator?

otelconf/v0.3.0/propagator_test.go
```diff
@@ -0,0 +1,195 @@
...
+			cfg: configOptions{
+				opentelemetryConfig: OpenTelemetryConfiguration{
+					Propagator: &Propagator{
+						Composite: []*string{ptr(""), ptr("tracecontext")},
+					},
+				},
+			},
+			want:    propagation.TraceContext{},
+			wantErr: false,
+		},
```
pellared, line 155: Wouldn't it mean that somebody explicitly set an empty string? Is not an error scenario? Shouldn't we failed the parsing if anything is wrong in the YAML file?

codeboten:
I revived this code from the original PR but after looking through the change in https://github.com/open-telemetry/opentelemetry-configuration/pull/187, i think it probably makes sense to wait for the 1.0.0-rc2 schema to be supported before bringing in more functionality. I'll focus on that PR instead. 

Will close this one for now and implement this once 1.0.0-rc2 is merged

PR #7986, opened 2025-10-07
feat(mongo): add SpanNameResolver for customizable span naming

Azer0s:
This PR adds a SpanNameResolver function to the monitor options in order to customize the span name to the users needs. This is especially relevant when dealing with multiple database instances, having to prefix spans, or generally needing to customize the span name.

### DIFF
--- a/instrumentation/go.mongodb.org/mongo-driver/v2/mongo/otelmongo/example_test.go
+++ b/instrumentation/go.mongodb.org/mongo-driver/v2/mongo/otelmongo/example_test.go
@@ -15,21 +15,6 @@ import (
 	"go.opentelemetry.io/contrib/instrumentation/go.mongodb.org/mongo-driver/v2/mongo/otelmongo"
 )
 
-func extractCollection(evt *event.CommandStartedEvent) string {
-	elt, err := evt.Command.IndexErr(0)
-	if err != nil {
-		return ""
-	}
-	if key, e := elt.KeyErr(); e == nil && key == evt.CommandName {
-		var v bson.RawValue
-		if v, e = elt.ValueErr(); e != nil || v.Type != bson.TypeString {
-			return ""
-		}
-		return v.StringValue()
-	}
-	return ""
-}
-
 func Example() {
 	// connect to MongoDB
 	opts := options.Client()
@@ -68,12 +53,10 @@ func ExampleWithSpanNameFormatter() {
 	opts := options.Client()
 	opts.Monitor = otelmongo.NewMonitor(
 		otelmongo.WithSpanNameFormatter(func(event *event.CommandStartedEvent) string {
-			collection := extractCollection(event)
-			if collection == "" {
-				return fmt.Sprintf("my-prefix-%s", event.CommandName)
-			}
-
-			return fmt.Sprintf("my-prefix-%s-%s", collection, event.CommandName)
+			// optionally, the collection name can be extracted for more
+			// descriptive span names; see the extractCollection helper function
+			// in the otelmongo package for an example of how to do this.
+			return fmt.Sprintf("my-prefix-%s", event.CommandName)
 		}),
 	)
 	opts.ApplyURI("mongodb://localhost:27017")


instrumentation/go.mongodb.org/mongo-driver/v2/mongo/otelmongo/config.go
```diff
@@ -48,6 +59,20 @@ func (o optionFunc) apply(c *config) {
 	o(c)
 }
 
+type SpanNameResolverFunc func(collection string, command string) string
+
+// WithSpanNameResolver specifies a function that resolves the span name given the
+// collection and command name. If none is specified, the default resolver is used,
+// which returns "<collection>.<command>" if the collection is non-empty, and just
+// "<command>" otherwise.
+func WithSpanNameResolver(resolver SpanNameResolverFunc) Option {
```
dmathieu: Could we use the same name as in other packages, [WithSpanNameFormatter](https://pkg.go.dev/go.opentelemetry.io/contrib/instrumentation/net/http/otelhttp#WithSpanNameFormatter)?

Azer0s: Done :)

dmathieu:

Azer0s:

I added one

CHANGELOG.md
```diff
@@ -33,6 +33,7 @@ The next release will require at least [Go 1.24].
 - Add the `WithTraceAttributeFn` option to `go.opentelemetry.io/contrib/instrumentation/github.com/aws/aws-lambda-go/otellambda`. (#7556)
 - Add support for HTTP server metrics in `go.opentelemetry.io/contrib/instrumentation/github.com/labstack/echo/otelecho`. (#7668)
 - Support testing of [Go 1.25]. (#7732)
+- Add a `WithSpanNameFormatter` option to otelmongo (#7986)
```
```suggestion
- Add a `WithSpanNameFormatter` option to `go.opentelemetry.io/contrib/instrumentation/go.mongodb.org/mongo-driver/v2/mongo/otelmongo`. (#7986)
```

dmathieu (approved):

Azer0s:
@dmathieu could you please merge? :)

dmathieu:
We require 2 approvals to merge.

Azer0s:
Ping @prestonvasquez :)

instrumentation/go.mongodb.org/mongo-driver/v2/mongo/otelmongo/config.go
```diff
@@ -48,6 +59,22 @@ func (o optionFunc) apply(c *config) {
 	o(c)
 }
 
+// SpanNameFormatterFunc is a function that resolves the span name given the
+// collection and command name.
+type SpanNameFormatterFunc func(collection, command string) string
```
flc1125: I cannot be certain, currently for reference only:

Why can't this part `func(collection, command string) string` be `func(collection string, evt *event.CommandStartedEvent) string`

cc @prestonvasquez

Azer0s: Another option would be to pass a formatter data struct like so:

```go
type FormatterData struct {
    *event.CommandStartedEvent
    CollectionName string
}
``` 

But tbh I don't really see what added benefits that would bring, but I am also not opposed to implementing it ☺️

dmathieu: I don't know enough about mongo to say what we need.
But with HTTP instrumentations, we pass the `*http.Request` object, so folks can use anything within that struct to build the span name.

Azer0s: @dmathieu should I change the formatter function to accept an event alongside the collection name?

Azer0s: I still feel like a formatter data struct is more extendable

dmathieu: @prestonvasquez is currently on leave. I have emailed his coworkers. Let's wait a bit for their input.

Azer0s: @dmathieu Hi! Are there any updates on this?

donkahlero:
I'd also would like to ask if there are any updates regarding this, as I am in need for this change, as well. Thx!

instrumentation/go.mongodb.org/mongo-driver/v2/mongo/otelmongo/config.go
prestonvasquez: @Azer0s @dmathieu @flc1125 

After comparing with analogous function signatures in `otelhttp` (op + `*http.Request`), `otelmux` (route name + `*http.Request`), and `otelgin` (`*gin.Context`, which contains a parsed route), I suggest that we use the following function signature:

```go
type SpanNameFormatterFunc func(collection string, evt *event.CommandStartedEvent) string
```

It’s conceivable that a user would want to include additional information:

```go
// Include database name.
WithSpanNameFormatter(func(collection string, evt *event.CommandStartedEvent) string {
	if collection != "" {
		return fmt.Sprintf("%s.%s.%s", evt.DatabaseName, collection, evt.CommandName)
	}
	return fmt.Sprintf("%s.%s", evt.DatabaseName, evt.CommandName)
})

// Differentiate by load balancer service.
WithSpanNameFormatter(func(collection string, evt *event.CommandStartedEvent) string {
	name := collection + "." evt.CommandName
	if evt.ServiceID != nil {
		name += fmt.Sprintf(" [service-%s]", evt.ServiceID.Hex())
	}	
	return name
})
```

Alternatively, [DRIVERS-2575](https://jira.mongodb.org/browse/DRIVERS-2575) prescribes that the collection name be included in `event.CommandStartedEvent`. Since the [specifications](https://github.com/mongodb/specifications/blob/master/source/command-logging-and-monitoring/command-logging-and-monitoring.md) are a minimal contract, the Go Driver team could prioritize this behavior for `otelmongo`. In which case, the function signature could be minimized:

```go
type SpanNameFormatterFunc func(evt *event.CommandStartedEvent) string
```

In my opinion, this would require additionally exporting the `extractCollections` function as a convenience until it can be deprecated in favor of the driver-specific solution.

For posterity, the event callbacks in the Go Driver pass a context. I suggest we do not model that behavior here as it implies it’s acceptable to use blocking operations while naming a span.

instrumentation/go.mongodb.org/mongo-driver/v2/mongo/otelmongo/config.go
```diff
@@ -48,6 +59,22 @@ func (o optionFunc) apply(c *config) {
...
+
+// WithSpanNameFormatter specifies a function that resolves the span name given the
+// collection and command name. If none is specified, the default resolver is used,
+// which returns "<collection>.<command>" if the collection is non-empty, and just
+// "<command>" otherwise.
+func WithSpanNameFormatter(resolver SpanNameFormatterFunc) Option {
```
prestonvasquez, line 72: [nit] Suggest adding an example to example_test.go illustrating the use case.

instrumentation/go.mongodb.org/mongo-driver/v2/mongo/otelmongo/config.go
Azer0s: @prestonvasquez done :)

instrumentation/go.mongodb.org/mongo-driver/v2/mongo/otelmongo/config.go
Azer0s: @prestonvasquez done :)

instrumentation/go.mongodb.org/mongo-driver/v2/mongo/otelmongo/config.go
dmathieu: If you expect the command should be available in the event soon, I'd go straight with the simpler method definition. Changing method definitions is tricky, as we can't do deprecations.

Azer0s: @dmathieu done

instrumentation/go.mongodb.org/mongo-driver/v2/mongo/otelmongo/example_test.go
```diff
@@ -45,3 +47,39 @@ func Example() {
...
+	// connect to MongoDB
+	opts := options.Client()
+	opts.Monitor = otelmongo.NewMonitor(
+		otelmongo.WithSpanNameFormatter(func(event *event.CommandStartedEvent) string {
+			collection, err := otelmongo.ExtractCollection(event)
+			if err != nil {
+				return fmt.Sprintf("my-prefix-%s-%s", collection, event.CommandName)
+			}
+
+			return fmt.Sprintf("my-prefix-%s", event.CommandName)
```
prestonvasquez: [docs] This logic appears inverted:

```suggestion
			collection, err := otelmongo.ExtractCollection(event)
			if err != nil || collection == "" {
				return fmt.Sprintf("my-prefix-%s", event.CommandName)
			}

			return fmt.Sprintf("my-prefix-%s-%s", collection, event.CommandName)
```

instrumentation/go.mongodb.org/mongo-driver/v2/mongo/otelmongo/mongo.go
```diff
@@ -101,11 +102,11 @@ func sanitizeCommand(command bson.Raw) string {
 	return string(b)
 }
 
-// extractCollection extracts the collection for the given mongodb command event.
+// ExtractCollection extracts the collection for the given mongodb command event.
```
prestonvasquez: [nit] Suggest noting that this function is a convenience API:

> ExtractCollection extracts the collection name from a MongoDB CommandStartedEvent. For CRUD operations, this is the first "\<operation\>": "\<collection\>" pair; meta-operations may omit it. This function is exported as a temporary convenience until DRIVERS-2575/GODRIVER-3688 adds the collection name to CommandStartedEvent in the Go Driver. This function returns the collection name or an error if no collection can be determined.

~~@dmathieu Is there any way we can mark this as experimental so it can be removed in the future? Otherwise, suggest also noting that it is either deprecated or will eventually be deprecated in favor of DRIVERS-2575.~~

Edit: since adding collection name to `event.CommandStartedEvent` will happen in a release >=2.4, we have to maintain ExtractCollection as non-deprecated for the lifetime of v2.

instrumentation/go.mongodb.org/mongo-driver/v2/mongo/otelmongo/mongo_test.go
```diff
@@ -24,6 +24,75 @@ import (
...
+func TestWithSpanNameFormatter(t *testing.T) {
+	tt := []struct {
+		title             string
+		operation         func(context.Context, *mongo.Database) (any, error)
+		expectedSpanName  string
+		mockResponses     []bson.D
+		SpanNameFormatter SpanNameFormatterFunc
+	}{
+		{
+			title: "insert",
```
prestonvasquez: [nit] Could we title this w/ the context of using the default formatter?

```suggestion
			title: "insert using default SpanNameFormatter",
```

instrumentation/go.mongodb.org/mongo-driver/v2/mongo/otelmongo/mongo_test.go
```diff
@@ -24,6 +24,75 @@ import (
...
+			operation: func(ctx context.Context, db *mongo.Database) (any, error) {
+				return db.Collection("test-collection").InsertOne(ctx, bson.D{{Key: "test-item", Value: "test-value"}})
+			},
+			expectedSpanName:  "test-collection.insert",
+			SpanNameFormatter: nil,
+			mockResponses:     []bson.D{{{Key: "ok", Value: 1}}},
+		},
+		{
+			title: "delete",
```
prestonvasquez: [nit] Could we title this w/ the context of using a non-default formatter?

```suggestion
			title: "delete with custom SpanNameFormatter",
```

instrumentation/go.mongodb.org/mongo-driver/v2/mongo/otelmongo/config.go
prestonvasquez: Here is the associated GODRIVER ticket: https://jira.mongodb.org/browse/GODRIVER-3688.

instrumentation/go.mongodb.org/mongo-driver/v2/mongo/otelmongo/example_test.go
Azer0s: How do you mean?

prestonvasquez: On L58 the collection is being added to the span name despite getting an error while extracting from the raw bson. It would always be empty, no? In this case I would expect to use `my-prefix-{commandName}`.

instrumentation/go.mongodb.org/mongo-driver/v2/mongo/otelmongo/mongo.go
Azer0s: Done :)

instrumentation/go.mongodb.org/mongo-driver/v2/mongo/otelmongo/mongo.go
```diff
@@ -101,11 +102,14 @@ func sanitizeCommand(command bson.Raw) string {
...
 // For CRUD operations, this is the first key/value string pair in the bson
 // document where key == "<operation>" (e.g. key == "insert").
 // For database meta-level operations, such a key may not exist.
-func extractCollection(evt *event.CommandStartedEvent) (string, error) {
+// This function is exported as a temporary convenience until DRIVERS-2575/GODRIVER-3688
+// adds the collection name to CommandStartedEvent in the Go Driver.
+// This function returns the collection name or an error if no collection can be determined.
+func ExtractCollection(evt *event.CommandStartedEvent) (string, error) {
```
dmathieu: Why does this need to be made public? There's no reason for us to expose this public API (if the only reason is until an issue is fixed on the mongo side, then folks can implement the same utility method on their end temporarily).

Azer0s: Because re-implementing this every time seems kind of error-prone. Deprecating this will be trivial and besides I don't really see a downside to exporting a utility function. Even less so if it's a temporary one.

dmathieu: Any public API is, by defintion, public. Deprecating may not be hard, but it's still maintenance work and we're rather short staffed.
This utility method has nothing to do with observability, so even more 👎 to having it.

Azer0s: Okay, I unexported it again

instrumentation/go.mongodb.org/mongo-driver/v2/mongo/otelmongo/example_test.go
```diff
@@ -45,3 +62,39 @@ func Example() {
 		panic(err)
 	}
 }
+
+func ExampleWithSpanNameFormatter() {
+	// connect to MongoDB
+	opts := options.Client()
+	opts.Monitor = otelmongo.NewMonitor(
+		otelmongo.WithSpanNameFormatter(func(event *event.CommandStartedEvent) string {
+			collection := extractCollection(event)
```
dmathieu: I'm fine with a comment explaining how one can retrieve the collection name. But duplicating the method for the sake of an example seems a bit too much.
The example doesn't need to use the collection name right?

Azer0s: Removed.

Azer0s:
@dmathieu the pipeline is currently failing (as far as I can see, nothing I broke) and I don't have the permission to merge or restart the pipeline. Could you do that please? :)

Azer0s:
@dmathieu I think we're ready to merge! 🥳

dmathieu:
We require 2 approvals to merge.

Azer0s:
@prestonvasquez could you approve?

instrumentation/go.mongodb.org/mongo-driver/v2/mongo/otelmongo/mongo.go
```diff
@@ -45,11 +43,14 @@ func (m *monitor) Started(ctx context.Context, evt *event.CommandStartedEvent) {
 	if !m.cfg.CommandAttributeDisabled {
 		attrs = append(attrs, semconv.DBQueryText(sanitizeCommand(evt.Command)))
 	}
-	if collection, err := extractCollection(evt); err == nil && collection != "" {
-		spanName = collection + "."
+
+	collection, err := extractCollection(evt)
```
flc1125, line 47: In fact, when using the default `SpanNameFormatter`, we executed the `extractCollection` method twice. Is there a better way?

~~Secondly, it seems that users cannot access the `collection` information. Unless it implements a similar logic itself.~~

dmathieu: See https://github.com/open-telemetry/opentelemetry-go-contrib/pull/7986#discussion_r2493659384
That helper method should be made available by the mongo SDK, not by us.

prestonvasquez (approved):
LGTM, thanks @Azer0s 👍

PR #7989, opened 2025-10-07
fix(otelecho): rename serviceName -> serverName to reflect parsed value

frederik-suerig:
The code in `instrumentation/github.com/labstack/echo/otelecho/internal/semconv/server.go` parses and uses a primary server identifier (host/header/default virtual host and optional port), not an OpenTelemetry `service.name`. Using `serviceName` was misleading and caused confusion between the HTTP server host and the OTEL service semantic attribute.

This change renames the parameter/field to `serverName` (and related symbols) to:
- accurately describe the value being parsed (server host/port),
- avoid accidental conflation with OTEL `service.name`,
- improve readability and maintainability.

No behavioral changes intended

### DIFF
--- a/instrumentation/github.com/labstack/echo/otelecho/echo.go
+++ b/instrumentation/github.com/labstack/echo/otelecho/echo.go
@@ -27,7 +27,7 @@ const (
 )
 
 // Middleware returns echo middleware which will trace incoming requests.
-func Middleware(service string, opts ...Option) echo.MiddlewareFunc {
+func Middleware(serverName string, opts ...Option) echo.MiddlewareFunc {
 	cfg := config{}
 	for _, opt := range opts {
 		opt.apply(&cfg)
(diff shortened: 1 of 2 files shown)


dmathieu:
What is the intent of the new test ?

frederik-suerig:
> What is the intent of the new test ?

@dmathieu 

I added the unit test because there were no tests covering this behavior yet. The test ensures the instrumentation emits the expected attributes.

I can take one of the following approaches; tell me which you prefer and I’ll update the PR accordingly:

1.  Extend the existing [`TestMetrics`](https://github.com/open-telemetry/opentelemetry-go-contrib/blob/c4aba22ceb2c013e0c3861ee5f338bd6d8c59290/instrumentation/github.com/labstack/echo/otelecho/echo_test.go#L128) in `echo_test.go` to also assert the `server.address` and `server.port` attributes.
2. Keep the new, separate unit test.
3. Remove the test if you believe it is redundant.

MrAlias:
What behavior are you talking about? You renamed a parameter.

dmathieu:
And the behavior you're testing comes from logic in the semconv package. So that test should be there. Otherwise, we can test everything twice.

frederik-suerig:
@MrAlias @dmathieu

Removed the tests. 

For clarification, I renamed the parameter from `service` to `server` because the code parses and uses a server host (filled from the Request Host header when empty), not an OpenTelemetry `service.name`. 
Using the same parameter name reduces confusion.

Issue #4588, opened 2023-11-17
Race condition during Gin context cancellation


yfeng997circle:
### Description

Using golang's race detector we identified a race condition with otelgin instrumentation. The race happens when 
- The read happens when Gin gonic, the go server, [tries to cancel the Gin context and checks if a request has the proper context](https://github.com/gin-gonic/gin/blob/a481ee2897af1e368de5c919fbeb21b89aa26fc7/context.go#L1184). 
- The write happens when opentelemetry, after instrumenting the http call with trace and span id, [reverts the request context to original](https://github.com/open-telemetry/opentelemetry-go-contrib/blob/release-otelhttp-v0.35.1/instrumentation/github.com/gin-gonic/gin/otelgin/gintrace.go#L60). 

Below we give a detailed race report. 

```
==================
WARNING: DATA RACE
Write at 0x00c001b7a920 by goroutine 61853:
  go.opentelemetry.io/contrib/instrumentation/github.com/gin-gonic/gin/otelgin.Middleware.func1.1()
      /home/runner/go/pkg/mod/go.opentelemetry.io/contrib/instrumentation/github.com/gin-gonic/gin/otelgin@v0.45.0/gintrace.go:68 +0x151
  runtime.deferreturn()
      /opt/hostedtoolcache/go/1.19.13/x64/src/runtime/panic.go:476 +0x32
  github.com/gin-gonic/gin.(*Context).Next()
      /home/runner/go/pkg/mod/github.com/gin-gonic/gin@v1.9.1/context.go:174 +0x14a
  github.com/circlefin/platform-common-go/v4/middleware.recoveryWithLogger.func1()
      /home/runner/go/pkg/mod/github.com/circlefin/platform-common-go/v4@v4.80.0/middleware/recovery.go:58 +0xae
  github.com/gin-gonic/gin.(*Context).Next()
      /home/runner/go/pkg/mod/github.com/gin-gonic/gin@v1.9.1/context.go:174 +0xb83
  github.com/gin-gonic/gin.(*Engine).handleHTTPRequest()
      /home/runner/go/pkg/mod/github.com/gin-gonic/gin@v1.9.1/gin.go:620 +0x779
  github.com/gin-gonic/gin.(*Engine).ServeHTTP()
      /home/runner/go/pkg/mod/github.com/gin-gonic/gin@v1.9.1/gin.go:576 +0x44b
  net/http.serverHandler.ServeHTTP()
[...]

PR #4577, opened 2023-11-13
otelgrpc: Fix data race in stats handlers when processing messages received and sent metrics

naphatkrit:
```
WARNING: DATA RACE
Read at 0x00c000bef1d8 by goroutine 890:
  go.opentelemetry.io/contrib/instrumentation/google.golang.org/grpc/otelgrpc.(*config).handleRPC()
      /go/pkg/mod/go.opentelemetry.io/contrib/instrumentation/google.golang.org/grpc/otelgrpc@v0.46.0/stats_handler.go:199 +0x190b
  go.opentelemetry.io/contrib/instrumentation/google.golang.org/grpc/otelgrpc.(*clientHandler).HandleRPC()
      /go/pkg/mod/go.opentelemetry.io/contrib/instrumentation/google.golang.org/grpc/otelgrpc@v0.46.0/stats_handler.go:124 +0x65
  google.golang.org/grpc.(*csAttempt).finish()
      /go/pkg/mod/google.golang.org/grpc@v1.59.0/stream.go:1171 +0x650
  google.golang.org/grpc.(*clientStream).finish()
      /go/pkg/mod/google.golang.org/grpc@v1.59.0/stream.go:988 +0x34b
  google.golang.org/grpc.(*clientStream).RecvMsg()
      /go/pkg/mod/google.golang.org/grpc@v1.59.0/stream.go:940 +0x2d2
```

PR #8014, opened 2025-10-14
otelgin: remove unneeded context save/restore to avoid data race

nicksherron:
There seems to be absolutely no reason to save and restore the request context in a deferred function. The way it was done before could (and did) lead to a race condition because when c.Next() is called, the request context could be changed by another goroutine, and then the deferred function would modify the request while another goroutine is using it. 

I can't for the life of me figure out what benefit the previous code was trying to achieve, other than maybe trying to minimize the context size in memory, but that seems unnecessary and not worth the risk of a data race. Furthermore, I'm not even sure that said memory savings would pay off because when running `go build -gcflags="-m"` against the current implementation you can see that context and request escapes to heap:
```
./gin.go:84:37: "nil context" escapes to heap
./gin.go:84:37: new(http.Request) escapes to heap
```
This is likely negligible but I figured I'd add it for more context, no pun intended 😉

Perhaps the goal was not that of performance/memory savings but instead to "restore" the context. If so, this is also incorrect because in Gin, there's no need to restore the context - each request gets its own *gin.Context .

Finally, this change actually more closely matches the [code cited](https://github.com/DataDog/dd-trace-go/blob/8fb554ff7cf694267f9077ae35e27ce4689ed8b6/contrib/gin-gonic/gin/gintrace.go) as the "inspiration" for the otelgin package. 

Without this change, we unfortunately cannot run our tests with the race detector enabled and is not production ready due to not being goroutine safe.

### DIFF
--- a/instrumentation/github.com/gin-gonic/gin/otelgin/gin_test.go
+++ b/instrumentation/github.com/gin-gonic/gin/otelgin/gin_test.go
@@ -630,8 +630,6 @@ func TestMetrics(t *testing.T) {
 // 3. The logging goroutine needs the trace context for correlation
 //
 // Regression test for #8014.
-//
-// Run with: go test -race -run TestMiddlewareContextRace -v
 func TestMiddlewareContextRace(t *testing.T) {
 	gin.SetMode(gin.TestMode)
 
@@ -684,7 +682,7 @@ func TestMiddlewareContextRace(t *testing.T) {
 	})
 
 	// Make a single request
-	req := httptest.NewRequest(http.MethodGet, "/test", nil)
+	req := httptest.NewRequest(http.MethodGet, "/test", http.NoBody)
 	w := httptest.NewRecorder()
 	router.ServeHTTP(w, req)
 


flc1125:
The current changes do not guarantee thread safety.

Additionally, I lean towards the view that for infrastructure-level components or frameworks, there is no need to consider the concurrency safety of `req.WithContext` and `req.Context()`.

Once these components introduce certain mechanisms for concurrency safety, it may lead to performance degradation.

For users pursuing high performance, such performance degradation is unacceptable. Moreover, once implemented, users have no choice but to use these mechanisms.

Therefore, I prefer to leave the decision of whether to ensure concurrency safety to the users themselves, and let them implement these mechanisms on their own.

nicksherron:
Just found an open 2 year old issue this would fix #4588

nicksherron:
Thanks for the quick response @flc1125 , but I would like to offer a rebuttal. 

## This PR Fixes an Existing Bug, Not Introducing New Safety Mechanisms

I respectfully disagree with the characterization that this PR "introduces mechanisms for concurrency safety." The opposite is true: this PR removes problematic code that causes a data race.

The current code has a real, reproducible race condition reported in issue #4588 (open for 2 years). The race detector fails consistently. This is not theoretical—it violates Go's memory model.

## The Defer Provides No Benefit

Let's analyze what the current code does:
```go
  savedCtx := c.Request.Context()
  defer func() {
      c.Request = c.Request.WithContext(savedCtx)  // ← Restores AFTER request completes
  }()
  ctx := cfg.Propagators.Extract(savedCtx, propagation.HeaderCarrier(c.Request.Header))
  // ...
  c.Request = c.Request.WithContext(ctx)  // ← Sets new context
  c.Next()  // ← Processes request
  // When we return, defer runs and "restores" the old context
```
  Question: Why restore the context after the request is complete?

  - After c.Next() returns, the request has been fully processed
  - Gin doesn't reuse *gin.Context instances across requests
  - Downstream handlers may spawn goroutines that read c.Request.Context()
  - The deferred write to c.Request races with those goroutine reads
  - There is no code that benefits from this restoration

## The Performance Argument is Backwards

The current code with the defer is slower than the fix:

  Current code overhead:
  - Allocates closure for defer
  - Saves context reference
  - Executes deferred function on return
  - Calls WithContext() twice (once to set, once to "restore")

  Fixed code:
  - No defer
  - No closure allocation
  - One fewer WithContext() call
  - No unnecessary context copy

 Benchmark comparison (removing defer should show improvement):
  -  Current: ~3 allocations per request
  -  Fixed:   ~2 allocations per request

## Users Cannot Fix This Themselves

 This is middleware code that users cannot work around without forking the library. The race happens inside otelgin.Middleware(). A user who wants to use otelgin AND pass race detection has no choice but to:

  1. Fork this library
  2. Apply this exact fix
  3. Maintain their fork indefinitely

  This is not "leaving the decision to users"—it's forcing users to fork or live with failing race detection.

## Precedent: This Project Has Fixed Similar Races Before

  From the git history:
  - #4577: "otelgrpc: Fix data race in stats handlers" (merged)
  - #881: "otelsarama: Fix WrapAsyncProducer race condition" (merged)

  These fixes were accepted because data races are bugs, not "concurrency mechanisms left to users."

## The Race Is Real and Reproducible

  From issue #4588:
  WARNING: DATA RACE
  Write at 0x... by goroutine X:
    [defer in Middleware]
  Read at 0x... by goroutine Y:
    [user handler accessing context]

  This is undefined behavior. The Go memory model states: "Programs that modify data being simultaneously accessed by multiple goroutines must serialize such access."

## The Fix is Correct

  The fix simply removes dead code:
```go
  // Before: Save and restore (unnecessary)
  savedCtx := c.Request.Context()
  defer func() { c.Request = c.Request.WithContext(savedCtx) }()
  ctx := cfg.Propagators.Extract(savedCtx, ...)

  // After: Use context directly (correct)
  reqCtx := c.Request.Context()
  ctx := cfg.Propagators.Extract(reqCtx, ...)
```
  The behavior is identical except:
  -  No race condition
  -  Better performance
  -  Simpler code
  -  Tests pass with -race

  Conclusion

  This is a straightforward bug fix that:
  1. Removes unnecessary code that causes a race
  2. Improves performance (fewer allocations, no defer overhead)
  3. Fixes a 2-year-old issue (#4588)
  4. Follows precedent from previous race fixes in this project
  5. Enables CI to enforce race-free code

  There is no downside to this change, and significant benefits. I respectfully request approval to merge.

akats7:
Hi @nicksherron, 

It's a fair question, I also don't really see the benefit of it, I'm going to assume it was added for hygiene to avoid leaking the modifications out of the intended request's middleware chain.

@flc1125 what are your thoughts, even if it doesn't provide thread safety it does seem valid

flc1125:
> I'm going to assume it was added for hygiene to avoid leaking the modifications out of the intended request's middleware chain.

I agree with this situation.

nicksherron:
I'm confused about this "doesn't provide thread safety" claim. I'm starring at a test failure due to data race when running `go test -race`  in one of my projects that clearly points the finger to the code I removed as being the culprit. When I do a `go mod -replace` using my change in this PR the race goes away.

flc1125:
> I'm confused about this "doesn't provide thread safety" claim. I'm starring at a test failure due to data race when running `go test -race` in one of my projects that clearly points the finger to the code I removed as being the culprit. When I do a `go mod -replace` using my change in this PR the race goes away.

I'm not quite sure about your specific case, so this is for reference only: Since `*http.Request` is often passed to subsequent logic, if there are concurrent modifications to it, it can lead to thread safety issues.

akats7:
> > I'm confused about this "doesn't provide thread safety" claim. I'm starring at a test failure due to data race when running `go test -race` in one of my projects that clearly points the finger to the code I removed as being the culprit. When I do a `go mod -replace` using my change in this PR the race goes away.
> 
> I'm not quite sure about your specific case, so this is for reference only: Since `*http.Request` is often passed to subsequent logic, if there are concurrent modifications to it, it can lead to thread safety issues.

 That’s essentially what’s happening here, this would be an example of subsequent logic that’s creating a data race. We can’t guarantee that others middleware’s won’t do something similar but I think this is a valid fix.

nicksherron:
The "hygiene to avoid leaking modifications" argument is backwards. The defer causes the race, it doesn't prevent leaking:

  1. There's nothing to leak - Gin doesn't reuse *gin.Context across requests. Each request gets a fresh context.
  2. The defer IS the race - When a handler spawns a goroutine that reads c.Request.Context(), the defer writes to c.Request while that goroutine reads from it. That's the data race.
  3. Proof:
  ```go
  router.POST("/test", func(c *gin.Context) {
      ctx := c.Request.Context()
      go func() {
          time.Sleep(1ms)
          _ = ctx.Value("key")  // READ
      }()
      c.JSON(200, "ok")
      // defer runs here and WRITES to c.Request ← RACE!
  })
  ```

The "concurrent modifications" concern is the problem we're fixing - The defer itself is the concurrent modification! It writes to c.Request after the handler returns while background goroutines may still be using it.

nicksherron:
Here's a test you can run to test this for yourself. It proves the current implementation is the culprit for data races.

https://gist.github.com/nicksherron/08dfbc599e0de0532cd4dc0d30f9f4a4 (moved to a gist to save the thread from a giant comment)

By the way I'm really blown away by your quick responses. I never see this kind of quick back and forth on a new PR in other oss projects. I'm really impressed and it's much appreciated.

nicksherron:
I can also provide a benchmark to prove the performance impact is actually net positive if you'd like @flc1125

flc1125:
> https://gist.github.com/nicksherron/08dfbc599e0de0532cd4dc0d30f9f4a4

<details>
<summary>This is a minimal case to examine the issue of data races:</summary>

```go
package otelgin

import (
	"context"
	"net/http/httptest"
	"sync"
	"testing"

	"github.com/gin-gonic/gin"
)

func TestConcurrency(t *testing.T) {
	t.Run("no defer", func(t *testing.T) {
		runConcurrency(t, noDeferMiddleware())
	})

	t.Run("have defer", func(t *testing.T) {
		runConcurrency(t, haveDeferMiddleware())
	})

	t.Run("no middleware", func(t *testing.T) {
		runConcurrency(t, nil)
	})
}

func runConcurrency(t *testing.T, middleware gin.HandlerFunc) {
	router := gin.New()
	if middleware != nil {
		router.Use(middleware)
	}
	router.POST("/test", func(c *gin.Context) {
		go func() {
			_ = c.Request.Context()
		}()
		c.String(200, "ok")
	})

	wg := sync.WaitGroup{}
	for i := 0; i < 1000; i++ {
		wg.Go(func() {
			req := httptest.NewRequest("POST", "/test", nil)
			w := httptest.NewRecorder()
			router.ServeHTTP(w, req)

			assert.Equal(t, 200, w.Code)
		})
	}
	wg.Wait()
}

func noDeferMiddleware() gin.HandlerFunc {
	return func(c *gin.Context) {
		c.Request.WithContext(
			context.WithValue(c.Request.Context(), "key1", "value1"),
		)
		c.Next()
	}
}

func haveDeferMiddleware() gin.HandlerFunc {
	return func(c *gin.Context) {
		origCtx := c.Request.Context()
		defer func() {
			c.Request = c.Request.WithContext(origCtx)
		}()
		c.Request.WithContext(
			context.WithValue(origCtx, "key1", "value1"),
		)
		c.Next()
	}
}
```
</details>



Whether `defer` is present or not, data races have always existed. My understanding of this is that the reason lies in the fact that `*http.Request` itself cannot guarantee thread safety, and this applies to any framework, including the native one as well.

Therefore, I'm not quite clear about what issue you are trying to verify.

[As mentioned before](https://github.com/open-telemetry/opentelemetry-go-contrib/issues/4588#issuecomment-2490995125), I was wondering what this piece of code is intended to handle—it doesn't seem to serve any purpose and appears to be removable.

```go
defer func() {
	c.Request = c.Request.WithContext(savedCtx)
}()
```

nicksherron:
> Whether `defer` is present or not, data races have always existed. My understanding of this is that the reason lies in the fact that `*http.Request` itself cannot guarantee thread safety, and this applies to any framework, including the native one as well.
> 
> Therefore, I'm not quite clear about what issue you are trying to verify.
> 
> [As mentioned before](https://github.com/open-telemetry/opentelemetry-go-contrib/issues/4588#issuecomment-2490995125), I was wondering what this piece of code is intended to handle—it doesn't seem to serve any purpose and appears to be removable.
> 
> ```go
> defer func() {
> 	c.Request = c.Request.WithContext(savedCtx)
> }()
> ```

I feel like we're talking in circles a little here and I don't know if I can explain it any better, but I'll try. 
I'm not claiming this change will make `*http.Request` impervious to data races. The disconnect seems to be thinking we're adding concurrency safety. **We're removing the code that breaks it**. 
I think there's a misunderstanding about when the defer runs. Let me clarify:

Timeline with OLD code (defer pattern):
```go
func Middleware() {
    savedCtx := c.Request.Context()
    defer func() {
        c.Request = c.Request.WithContext(savedCtx)  // ← Step 5: WRITE
    }()

    c.Request = c.Request.WithContext(ctx)  // Step 1
    c.Next()                                 // Step 2: All handlers/middleware run
    // Step 3: Handler spawned goroutine that reads c.Request.Context()
    // Step 4: Handler returns
    // Step 5: defer runs and WRITES to c.Request
    // Step 6: Background goroutine READS c.Request ← **RACE**!
}
```

The defer IS the concurrent modification. It writes to c.Request after handlers complete, while
background goroutines spawned by those handlers may still be reading it.

The fix removes this write:
```go
func Middleware() {
    reqCtx := c.Request.Context()
    ctx := cfg.Propagators.Extract(reqCtx, ...)

    c.Request = c.Request.WithContext(ctx)  // Step 1: Set once
    c.Next()                                 // Step 2: Handlers run
    // Step 3: Handler spawns goroutine
    // Step 4: Handler returns
    // Step 5: No defer, no write ← NO RACE
}
```
Proof: TestMiddleware_MinimalRaceRepro demonstrates exactly this:
- Handler spawns goroutine that accesses c.Request.Context()
- Handler returns
- Old code: defer writes → DATA RACE
- Fixed code: no write → PASS

Furthermore this isn't my "specific case" - this is how Go middleware is supposed to work. The defer breaks standard Go patterns that every Go developer uses.


Go's own documentation recommends this pattern

From https://pkg.go.dev/context:

>   "Incoming requests to a server should create a Context, and outgoing calls to servers should accept a Context. The chain of function calls between them must propagate the Context,
>   optionally replacing it with a derived Context..."
> 

Passing c.Request.Context() to goroutines is the recommended approach for:
  - Timeouts
  - Cancellation
  - Tracing (which is what otelgin does!)

Other popular Gin middleware use this exact pattern!

Example using an extremely common pattern in production:
```go

 func main() {
      router := gin.New()

      // Setup OpenTelemetry tracing (standard pattern)
      router.Use(otelgin.Middleware("my-service"))

      // Add async logging middleware (also standard pattern)
      router.Use(asyncLoggingMiddleware())

      router.POST("/api/orders", createOrder)
      router.Run(":8080")
  }

  // Async logging middleware - BREAKS with otelgin's defer pattern
  func asyncLoggingMiddleware() gin.HandlerFunc {
      logger := zap.NewProduction()

      return func(c *gin.Context) {
          start := time.Now()
          path := c.Request.URL.Path
          method := c.Request.Method

          // Process request
          c.Next()

          // Log asynchronously to avoid blocking response
          // This is THE standard pattern for production logging
          go func(ctx context.Context, statusCode int, duration time.Duration) {
              // Extract trace context for correlation
              traceID := span.SpanContext().TraceID().String()
              spanID := span.SpanContext().SpanID().String()

              // Write structured log with trace correlation
              logger.Info("http_request",
                  zap.String("method", method),
                  zap.String("path", path),
                  zap.Int("status", statusCode),
                  zap.Duration("duration", duration),
                  zap.String("trace_id", traceID),    // ← Links logs to traces
                  zap.String("span_id", spanID),
              )

              // Could also send to external logging service here
              // sendToDatadog(ctx, logEntry)
              // sendToCloudWatch(ctx, logEntry)
          }(c.Request.Context(), c.Writer.Status(), time.Since(start))
          //  ^^^^^^^^^^^^^^^^^^^^
          //  THIS RACES WITH otelgin's defer!
      }
  }

  func createOrder(c *gin.Context) {
      var order Order
      if err := c.BindJSON(&order); err != nil {
          c.JSON(400, gin.H{"error": err.Error()})
          return
      }

      // Process order...
      c.JSON(201, order)
  }

  ```

**Comment by nicksherron** [comment]
Hopefully that provides enough evidence to support my claim which is that **the spurious defer code is prone to data races when used in a project with typical gin server patterns** and this PR fixes it.

**Comment by dmathieu** [comment]
Could you turn your gist into a test that can run on CI, and add it to this PR?

**Comment by nicksherron** [comment]
> Could you turn your gist into a test that can run on CI, and add it to this PR?

Sure, I'd be happy to.

**Comment by flc1125** [comment]
> ```go
> }(c.Request.Context(), c.Writer.Status(), time.Since(start))
>           //  ^^^^^^^^^^^^^^^^^^^^
>           //  THIS RACES WITH otelgin's defer!
> ```

I believe my understanding aligns with yours. I approve this PR and agree with the changes.

I'd like to clarify that the primary value isn't about "fixing data race". Rather, it removes unnecessary code that has unintended side effects.

e.g., The defer restoration discards any context modifications made within the `otelgin` middleware chain (including downstream handlers).

**Comment by nicksherron** [comment]
> > ```go
> > }(c.Request.Context(), c.Writer.Status(), time.Since(start))
> >           //  ^^^^^^^^^^^^^^^^^^^^
> >           //  THIS RACES WITH otelgin's defer!
> > ```
> 
> I believe my understanding aligns with yours. I approve this PR and agree with the changes.
> 
> I'd like to clarify that the primary value isn't about "fixing data race". Rather, it removes unnecessary code that has unintended side effects.
> 
> e.g., The defer restoration discards any context modifications made within the `otelgin` middleware chain (including downstream handlers).

Sounds good, thank you for your time on this. I also added the test @dmathieu requested. If you want me to make any changes to it just let me know.

**Comment by dmathieu** [comment]
The intent of the defer is to remove the span from the context, so spans using the same context aren't childs of the created span once the middleware has finished running.

Imaging the following setup:

```golang
r := gin.New()
r.Use(myCustomMiddleware)
r.Use(otelgin.Middleware("my-server"))
	
r.GET("/", func(c *gin.Context) {
	c.HTML(http.StatusOK, "index.tmpl", gin.H{})
})
```

When `myCustomMiddleware` runs, the span has either not started yet, or is already finished.
If it's already finished, the context shouldn't have the span within it, as we're not within the trace anymore.

Right now, if you create a span within `myCustomMiddleware`, after the call to `c.Next()`, it won't be a child of the otelgin span.
After your change it is one.

This is what the defer prevents.
While I agree to fixing this race condition, the defer does have a use and I don't think we can just drop it.

(ideally, we should add a test for this too)

akats7:
> The intent of the defer is to remove the span from the context, so spans using the same context aren't childs of the created span once the middleware has finished running.
> 
> Imaging the following setup:
> 
> ```go
> r := gin.New()
> r.Use(myCustomMiddleware)
> r.Use(otelgin.Middleware("my-server"))
> 	
> r.GET("/", func(c *gin.Context) {
> 	c.HTML(http.StatusOK, "index.tmpl", gin.H{})
> })
> ```
> 
> When `myCustomMiddleware` runs, the span has either not started yet, or is already finished. If it's already finished, the context shouldn't have the span within it, as we're not within the trace anymore.
> 
> Right now, if you create a span within `myCustomMiddleware`, after the call to `c.Next()`, it won't be a child of the otelgin span. After your change it is one.
> 
> This is what the defer prevents. While I agree to fixing this race condition, the defer does have a use and I don't think we can just drop it.
> 
> (ideally, we should add a test for this too)

Ah that’s a really good observation

nicksherron:
> The intent of the defer is to remove the span from the context, so spans using the same context aren't childs of the created span once the middleware has finished running.
> 
> Imaging the following setup:
> 
> ```go
> r := gin.New()
> r.Use(myCustomMiddleware)
> r.Use(otelgin.Middleware("my-server"))
> 	
> r.GET("/", func(c *gin.Context) {
> 	c.HTML(http.StatusOK, "index.tmpl", gin.H{})
> })
> ```
> 
> When `myCustomMiddleware` runs, the span has either not started yet, or is already finished. If it's already finished, the context shouldn't have the span within it, as we're not within the trace anymore.
> 
> Right now, if you create a span within `myCustomMiddleware`, after the call to `c.Next()`, it won't be a child of the otelgin span. After your change it is one.
> 
> This is what the defer prevents. While I agree to fixing this race condition, the defer does have a use and I don't think we can just drop it.
> 
> (ideally, we should add a test for this too)

That's a really good theory about what the original code was intending to do and you made a really good point. However, I think there are some fundamental issues with this approach:

1. This is an anti-pattern for context.Context

The defer modifies c.Request.Context() after passing it forward, which violates Go's context conventions. From the https://pkg.go.dev/context:

> "Contexts form a tree... Successive calls to WithValue, WithCancel, WithDeadline, or WithTimeout derive new contexts."


Contexts are meant to flow forward, not be "restored" backward. The library shouldn't enforce this opinion.

2. Most users want otelgin early in the chain

I (and I assume most users) place otelgin as one of the first middleware specifically so child spans can be created from it:

r.Use(otelgin.Middleware("my-service"))  // ← First - creates root span
r.Use(loggingMiddleware)                  // ← Creates child spans
r.Use(authMiddleware)                     // ← Creates child spans
r.GET("/", handler)                       // ← Handler creates child spans

Common use cases that need the parent span:
- Database transactions with tracing
- Inter-service HTTP calls with trace propagation
- Structured logging with trace IDs
- Background jobs that need trace context

Your example breaks all of these - they'd have no parent span.

3. Users can control span isolation themselves 

If a user doesn't want their spans to be children of otelgin's span, they can do exactly what the defer does - but explicitly:
```go
var tracer = otel.Tracer(instrumentationName)

func myCustomMiddleware(c *gin.Context) {
    ctx, span := tracer.Start(c.Request.Context(), "myCustomMiddleware")
    defer span.End()

    c.Next() // runs otelgin and other middleware

    // Explicitly restore context if the user doesn't want child spans and since the user is in control of goroutines the onus is on them to do so safely (as it should be)  
    c.Request = c.Request.WithContext(ctx)
}
```
This gives users control without the library making opinionated (and race-prone) decisions.

Even if we agreed on the semantics, I'm not sure of a safe way of doing it from a library implementation.

Bottom line: The defer enforces an anti-pattern, breaks the common use case (otelgin as root span), prevents valid patterns (goroutines with trace context), and causes data races.
Users who need span isolation can manage it themselves without the library making unsafe assumptions.

What do you think? I'm happy to add documentation showing how users can achieve span isolation if they need it.

dmathieu:
I disagree with this being a violation of Go's conventions (in the context of the middleware pattern, we don't flow forward. At the end of a middleware, we go back to its parent).
I also disagree that folks should do this themselves. Especially since this has been the default behavior for many years now (and the same behavior exists for other instrumentations, such as otelecho).

I think this discussion would be better synchronously though. If you're able to, I'd recommend you to join our next SIG meeting (tomorrow at 4pm UTC).
https://github.com/open-telemetry/community/?tab=readme-ov-file#implementation-sigs

nicksherron:
> I disagree with this being a violation of Go's conventions (in the context of the middleware pattern, we don't flow forward. At the end of a middleware, we go back to its parent). I also disagree that folks should do this themselves. Especially since this has been the default behavior for many years now (and the same behavior exists for other instrumentations, such as otelecho).
> 
> I think this discussion would be better synchronously though. If you're able to, I'd recommend you to join our next SIG meeting (tomorrow at 4pm UTC). https://github.com/open-telemetry/community/?tab=readme-ov-file#implementation-sigs

Ok well one thing is for certain, you all have a much better understanding of the existing patterns and user expectations for this project than I do. It's getting quite late for me so I'll sleep on this and give your points their time to marinate. As far as the SIG meeting goes that's a great idea (and sounds fun too) - unfortunately that's a fairly busy time for me but I'll do my best to make it.

nicksherron:
> Especially since this has been the default behavior for many years now (and the same behavior exists for other instrumentations, such as otelecho).

This is a strong argument, I didn’t realize that was the case.  I wasn’t looking to add a breaking change and I’m glad you pointed it out. I’ll see what I can do to come up with something that maintains the existing semantics while also fixing the race.

dmathieu:
We have been discussing this in yesterday's SIG meeting, and have decided this fix is incorrect and the current behavior is the proper one.
See https://github.com/open-telemetry/opentelemetry-go-contrib/issues/4588#issuecomment-3414917592

PR #8025, opened 2025-10-16
fix(otelecho): add config option to skip global error handler call

### LINKED ISSUE #4419: [otelecho] Add option to skip calling global handler

asp3cto:
Add WithSkipErrorHandler configuration option to support Echo's native error handling pattern where handlers return errors instead of calling c.Error() explicitly.

This prevents double responses on errors propagated to the otelecho middleware.

Fixes #4419, related to #4420

### DIFF
--- a/instrumentation/github.com/labstack/echo/otelecho/config.go
+++ b/instrumentation/github.com/labstack/echo/otelecho/config.go
@@ -114,7 +114,7 @@ func WithEchoMetricAttributeFn(f EchoMetricAttributeFn) Option {
 // WARNING: If the passed function doesn't call `c.Error` and the global HTTPErrorHandler modifies the response,
 // the tracing span can contain invalid data.
 // If it calls `c.Error`, `HTTPErrorHandler` will be executed twice, but the span will have the actual response data.
-// To fix this, check response commitment status (c.Response().Committed) before writing to it.
+// To fix this, check the response commitment status with `c.Response().Committed` before modifying the response.
 func WithOnError(f OnErrorFn) Option {
 	return optionFunc(func(cfg *config) {
 		if f != nil {


dmathieu:
I'm not super fond of negative options (`WithoutHandleError`).
The option may make sense though (@scorpionknifes, as code owner, do you have an opinion?)

I wonder if we could make the option more powerful by having a `WithOnError(func (c echo.Context, err error))`
It means to keep the same behavior, folks would need to do this:

```golang
r.Use(otelecho.Middleware("my-server", WithOnError(func(c echo.Context, err error) {
  c.Error(err)
}))
```

And at the same time, as I write this, it seems a bit overfluous. If the child middleware returns and error, it's not for us to call `c.Error()`. We do telemetry, not a full app.
So entirely removing this may make more sense.

asp3cto:
Well, the reason i didn't remove `c.Error` completely is to keep backwards compatibility, even for an unstable package.
Waiting for @scorpionknifes.

dmathieu:
Unstable packages don't need to keep backwards compatibility, they are unstable :)

asp3cto:
@dmathieu i understand) @scorpionknifes has no activity since February, should we wait for some other review or i can fix to remove the call completely?

dmathieu:
I would remove the option (no push force please. Also, a new commit is more easily reverted).
I'll bring this to the SIG meeting later today. Feel free to join (https://github.com/open-telemetry/community/?tab=readme-ov-file#implementation-sigs)

asp3cto:
I see some tests are failing. The issue is that the response status code should be read by this middleware, but it can be set in the error handler, which runs after all middlewares. Looking for a workaround.

The only way i currently see is to do like in echo request logger (https://github.com/labstack/echo/blob/e644ff8f7bb01c694cacec3ad22a7471609ea106/middleware/request_logger.go#L337), but by doing so we will loose the status code.

dmathieu:
Ah! That would explain why `c.Error()` was called.

asp3cto:
I need some advice here :)
In this example:
```go
r := httptest.NewRequest("GET", "/ping", http.NoBody)
	w := httptest.NewRecorder()

	router := echo.New()
	router.Logger = log.New("echo")
	router.Use(middleware.RequestLoggerWithConfig(middleware.RequestLoggerConfig{
		LogValuesFunc: func(c echo.Context, v middleware.RequestLoggerValues) error {
			fmt.Println("Request:", c.Request().Method, c.Request().URL.Path, "->", v.Status)
			return nil
		},
		HandleError: false,
		LogStatus:   true,
	}))
	router.Use(otelecho.Middleware("foobar"))
	router.GET("/ping", func(_ echo.Context) error {
		return assert.AnError
	})

	handlerCalled := 0
	router.HTTPErrorHandler = func(err error, c echo.Context) {
		handlerCalled++
		fmt.Println("Error handler called:", err, "should be:", assert.AnError)
		err = c.NoContent(http.StatusTeapot)
		if err != nil {
			fmt.Println("Failed to write response:", err)
		}
	}

	router.ServeHTTP(w, r)
	fmt.Println("Response code:", w.Result().StatusCode, "should be:", http.StatusTeapot)
	fmt.Println("Handler called times:", handlerCalled, "should be: 1")
```

i tested the behaviour of the echo request logger. I found out that it logs `http.StatusOk` if `HandleError=true`, and `http. StatusTeapot` if `false`. So the issue exists because of echo error handling design, and there is a tradeoff in such middlewares: get valid response by calling `c.Error` two times (and process this in custom handler) or skip double call and loose the actual response data after the handler.

asp3cto:
https://github.com/labstack/echo/issues/1948
As i said, echo design issue) Based on the last message there, maybe it's a good idea to at least add a warning to README to use the `Commited` filter in custom error handlers.

flc1125:
> I wonder if we could make the option more powerful by having a `WithOnError(func (c echo.Context, err error))`

I agree with this approach; we only need to provide an interface for users to manage how errors are handled, or even choose not to handle errors at all.

Additionally, we may also need to clarify the purpose of the `echo.Error()` method, and determine whether otel should support it or if it should be left to echo to manage on its own.

asp3cto:
@flc1125 the main issue is that if the middleware doesn't call `c.Error` explicitly, we don't get the right response status code if it's overrided by the custom error handler.

dmathieu:
And if we call it, folks have issues as it's being called multiple times.
I think going to `WithOnError` seems the most logical path forward. We should also have a comment stating explicitly that if folks call `WithOnError`, their error endpoint may be called multiple times, or their span may have an invalid response status.

With this method, I think it makes sense to keep the default behavior similar to what we have today.

asp3cto:
So we need to have default `cfg.OnError = func(c echo.Context, err error) { c.Error(err) }` and add a comment like "If this function doesn't call `c.Error`, response status code will be invalid if it is overriden in `HTTPErrorHandler`, and if it does, your custom error handler should check the commitment status of the response"? Just to make it clear

asp3cto:
I have an idea to add a wrapper for` HTTPErrorHandler` to make sure telemetry is handled correctly on error. It means removing all recording of telemetry if we get an error and do it in the error handler, like `e.HTTPErrorHandler=otelecho.HTTPErrorHandler(customHandler)`

asp3cto:
@dmathieu any thoughts?

flc1125:
I can try to research it to figure out the function of `c.Error()`, but it might take some time.

asp3cto:
@flc1125 we've already figured it out, the call is needed to set the actual status code in the span if it was modified in custom HTTPErrorHandler

instrumentation/github.com/labstack/echo/otelecho/config.go
```diff
@@ -22,8 +22,13 @@ type config struct {
 	Skipper               middleware.Skipper
 	MetricAttributeFn     MetricAttributeFn
 	EchoMetricAttributeFn EchoMetricAttributeFn
+	OnError               func(echo.Context, error)
 }
 
+// DefaultOnError is the default function called when an error occurs during request processing.
+// The call to c.Error(err) ensures tracing data is correct, but forces the global HTTPErrorHandler to be called twice.
+var DefaultOnError = func(c echo.Context, err error) { c.Error(err) }
```
dmathieu: What's the need to make this public?

asp3cto: no real need actually :)

dmathieu (approved):

asp3cto:
currently waiting for owner review?

dmathieu:
Waiting for a second approver/maintainer review.

flc1125:
I have completed my research and here are my conclusions: **I lean towards removing the `c.Error()` logic**.

**The reasons are:**

I believe `c.Error` is used to handle errors caused by uncontrollable underlying system factors, similar to Go's `panic` mechanism.

Because of this, when we do not use `c.Error(echo.NewHTTPError(...))`, it defaults to returning a 500 status code error—isn't this precisely the typical underlying system error?

Furthermore, I think the role of middleware, unless it triggers errors from uncontrollable underlying factors, should be to pass the error returned by the handler unchanged, and this error will ultimately affect the final response status code.

**Impact of adjusting according to this method: (and the impact is significant)**

For projects relying on this component, when an error occurs, the status code will change from 500 to 200, unless we return an error via `echo.NewHTTPError(...)`.

If we decide to proceed with this, we need to document this change in the CHANGELOG.

asp3cto:
@flc1125 `c.Error` is also used for setting the correct status code in case we change it in the `HTTPErrorHandler`, see https://github.com/open-telemetry/opentelemetry-go-contrib/pull/8025#issuecomment-3410672292
Here out error handler always sets `StatusTeapot` if triggered, and if any middleware doesn't call `c.Error`, it can't read the actual response status code, because error handler always runs last.
That's why we decided to keep the default logic and also to give the user an opportunity to make a tradeoff with `WithOnError`.

flc1125:
@asp3cto I believe that https://github.com/labstack/echo/blob/e644ff8f7bb01c694cacec3ad22a7471609ea106/echo.go#L887 must always be used in conjunction with the logic at https://github.com/labstack/echo/blob/e644ff8f7bb01c694cacec3ad22a7471609ea106/echo.go#L424-L430. Any scenario where either is unconfigured or fails to implement similar logic will result in an inability to fully manage status codes.

asp3cto:
I understand, but the main issue is custom HTTPErrorHandler. Builtin middlewares do the same: https://github.com/labstack/echo/blob/e644ff8f7bb01c694cacec3ad22a7471609ea106/middleware/request_logger.go#L287

They also check if HandleError is set and call c.Error if it is true to get the actual status.

flc1125:
> They also check if HandleError is set and call c.Error if it is true to get the actual status.

If there is a precedent, I think this logic makes sense.

instrumentation/github.com/labstack/echo/otelecho/config.go
```diff
@@ -100,3 +103,20 @@ func WithEchoMetricAttributeFn(f EchoMetricAttributeFn) Option {
...
+// WithOnError specifies a function that is called when an error occurs during request processing.
+// If nil is provided, it does nothing.
+//
+// WARNING: If the passed function doesn't call c.Error and the global HTTPErrorHandler modifies the response,
+// the tracing span can contain invalid data.
+// If it calls c.Error, HTTPErrorHandler will be executed twice, but the span will be correct.
+// To fix this, check response commitment status (c.Response().Committed) before writing to it.
+func WithOnError(fn func(echo.Context, error)) Option {
+	if fn == nil {
+		fn = func(_ echo.Context, _ error) {}
```
flc1125:
```suggestion
		fn = func(echo.Context, error) {}
```

instrumentation/github.com/labstack/echo/otelecho/echo_test.go
```diff
@@ -371,3 +371,71 @@ func TestWithEchoMetricAttributeFn(t *testing.T) {
...
+// TestWithOnError verifies middleware error handling behavior:
+// - Default: calls global handler twice (middleware + echo framework)
+// - WithOnError(nil): suppresses middleware's call, global handler called once
+// - Custom with c.Error(): calls custom handler, then global handler (2x total)
+// - Custom without c.Error(): only custom handler called (1x total)
+func TestWithOnError(t *testing.T) {
+	tests := []struct {
+		name              string
+		opt               Option
+		errorHandlerCalls int
```
flc1125:
```suggestion
		wantHandlerCalled int
```

CHANGELOG.md
```diff
@@ -11,6 +11,7 @@ This project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0.htm
 ### Added
 
 - `ParseYAML` in `go.opentelemetry.io/contrib/otelconf` now supports environment variables substitution in the format `${[env:]VAR_NAME[:-defaultvalue]}`. (#6215)
+- WithOnError option for otelecho middleware in `go.opentelemetry.io/contrib/instrumentation/github.com/labstack/echo/otelecho` to specify the behavior when an error occurs. (#4420)
```
```suggestion
- WithOnError option for otelecho middleware in `go.opentelemetry.io/contrib/instrumentation/github.com/labstack/echo/otelecho` to specify the behavior when an error occurs. (#8025)
```

instrumentation/github.com/labstack/echo/otelecho/echo.go
```diff
@@ -28,7 +28,7 @@ const (
 
 // Middleware returns echo middleware which will trace incoming requests.
 func Middleware(serverName string, opts ...Option) echo.MiddlewareFunc {
-	cfg := config{}
+	cfg := config{OnError: defaultOnError}
```
flc1125: Can we uniformly process according to operations similar to the following:

```go
if cfg.TracerProvider == nil {
	cfg.TracerProvider = otel.GetTracerProvider()
}
```

instrumentation/github.com/labstack/echo/otelecho/config.go
```diff
@@ -100,3 +103,20 @@ func WithEchoMetricAttributeFn(f EchoMetricAttributeFn) Option {
 		cfg.EchoMetricAttributeFn = f
 	})
 }
+
+// WithOnError specifies a function that is called when an error occurs during request processing.
+// If nil is provided, it does nothing.
```
flc1125: In what scenarios would this exist? In most cases, understanding nil should involve invoking the default option.

> At least the other options are handled this way.

asp3cto: WithOnError(nil) is read like 'with nothing on error', so i think setting empty func here is more understandable

flc1125: I believe we should maintain a consistent understanding of `nil` parameters across all `Option`.

For this, I would rather accept an `EmptyOnError` variable to achieve it. However, I do not currently recommend adding it until there is feedback from other users indicating it is necessary.

If there is sufficient understanding that it is necessary, I hope you can provide more information.

instrumentation/github.com/labstack/echo/otelecho/config.go
```diff
@@ -22,8 +22,11 @@ type config struct {
 	Skipper               middleware.Skipper
 	MetricAttributeFn     MetricAttributeFn
 	EchoMetricAttributeFn EchoMetricAttributeFn
+	OnError               func(echo.Context, error)
 }
 
+var defaultOnError = func(c echo.Context, err error) { c.Error(err) }
```
flc1125: Missing doc

instrumentation/github.com/labstack/echo/otelecho/config.go
asp3cto: With this option, the middleware can be called in different ways:
```go
srv.Use(otelecho.Middleware()) // default behaviour: c.Error is called on error
srv.Use(otelecho.Middleware(otelecho.WithOnError(someCustomFunc))) // here we give the user the ability to control what should be done on error in this particular middleware
srv.Use(otelecho.Middleware(otelecho.WithOnError(nil))) // it's a special case: we can implement it to call defaultOnError or to do nothing
```

Looking at the last use case, i think it's more straightforward to keep the current logic in the pr: we just put an empty func to `config.OnError`

flc1125: > ```go
> srv.Use(otelecho.Middleware(otelecho.WithOnError(nil))) // it's a special case: we can implement it to call defaultOnError or to do nothing
> ```

What I want to express is the behavior in this situation, which should be understood consistently with the following part below.

```go
if cfg.OnError == nil {
	cfg.OnError = defaultOnError
}
```

I believe we should try to avoid situations where `nil` has multiple meanings, as this can cause confusion for both users and future maintainers.

I prefer to define it in this way:

```go
var NoopOnError = func(echo.Context, error) {}

srv.Use(otelecho.Middleware(otelecho.WithOnError(NoopOnError)))
```

However, with this method of adjustment, I would prefer to wait for feedback from other users before deciding whether to add it.——That is: we can temporarily remove the case of `nil`, and temporarily not support `NoopOnError`.

instrumentation/github.com/labstack/echo/otelecho/config.go
dmathieu: This is private. We don't necessarily need doc.

instrumentation/github.com/labstack/echo/otelecho/config.go
asp3cto: So what should we do on nil passed to the option?

flc1125: I recommend that this is enough:

```go
func WithOnError(fn func(echo.Context, error)) Option {
	return optionFunc(func(cfg *config) {
		if fn != nil {
			cfg.OnError = fn
		}
	})
}
```

instrumentation/github.com/labstack/echo/otelecho/config.go
flc1125: Removing doc isn't a big deal either. 😄

flc1125 (approved):

dmathieu:
I'm using the merge of this PR to call out that otelecho is lacking a code owner, and will be deprecated/removed in a few weeks if nobody steps up.
https://github.com/open-telemetry/opentelemetry-go-contrib/issues/8056

instrumentation/github.com/labstack/echo/otelecho/config.go
```diff
@@ -100,3 +108,17 @@ func WithEchoMetricAttributeFn(f EchoMetricAttributeFn) Option {
 		cfg.EchoMetricAttributeFn = f
 	})
 }
+
+// WithOnError specifies a function that is called when an error occurs during request processing.
+//
+// WARNING: If the passed function doesn't call c.Error and the global HTTPErrorHandler modifies the response,
```
dmathieu:
```suggestion
// WARNING: If the passed function doesn't call `c.Error` and the global HTTPErrorHandler modifies the response,
```

instrumentation/github.com/labstack/echo/otelecho/config.go
```diff
@@ -100,3 +108,17 @@ func WithEchoMetricAttributeFn(f EchoMetricAttributeFn) Option {
...
+// the tracing span can contain invalid data.
+// If it calls c.Error, HTTPErrorHandler will be executed twice, but the span will be correct.
```
dmathieu:
```suggestion
// If it calls `c.Error`, `HTTPErrorHandler` will be executed twice, but the span will have the actual response data.
```

instrumentation/github.com/labstack/echo/otelecho/echo_test.go
```diff
@@ -371,3 +371,71 @@ func TestWithEchoMetricAttributeFn(t *testing.T) {
 	assert.True(t, foundID, "echo param id attribute should be found")
 	assert.True(t, foundPath, "echo path attribute should be found")
 }
+
+// TestWithOnError verifies middleware error handling behavior:
+// - Default: calls global handler twice (middleware + echo framework)
+// - WithOnError(nil): default behavior (2 handler calls)
+// - Custom with c.Error(): calls custom handler, then global handler (2x total)
+// - Custom without c.Error(): only custom handler called (1x total)
```
dmathieu:
```suggestion
```

This kind of comment would easily become stale. And the tests are sufficiently self explanatory.

instrumentation/github.com/labstack/echo/otelecho/config.go
```diff
@@ -100,3 +108,17 @@ func WithEchoMetricAttributeFn(f EchoMetricAttributeFn) Option {
...
+// To fix this, check response commitment status (c.Response().Committed) before writing to it.
```
dmathieu:
```suggestion
// To fix this, check the response commitment status with `c.Response().Committed` before modifying the response.
```

dmathieu (approved):

PR #8026, opened 2025-10-16
otelconf: add support for 1.0.0 release candidate

codeboten:
This adds support for the first 1.0.0 release candidate of the OpenTelemetry configuration schema. The release candidate includes many breaking changes that were introduced in [v0.4.0](https://github.com/open-telemetry/opentelemetry-configuration/releases/tag/v0.4.0). I'm including the release notes for v0.4.0 in the description below. 

The implementation steps:
1. update `OPENTELEMETRY_CONFIGURATION_JSONSCHEMA_VERSION` to rc1, run make genjsonschema and commit the new generated code
2. copy over v0.3.0 implementation and update it as needed
3. added a test file with the kitchen-sink example and ensure the parsed output matches the input

I've added some TODOs to keep track of things that need to be implemented still. I'll open issues for these and include them in the comments in the code.

Fixes #7159

  <details>
  
  <summary>v0.4.0 release notes</summary>
This release has a significant number of breaking changes, new additions and fixes. The breaking changes reflect a higher degree of scrutiny on consistency and correct data modeling as we approach a stable release ([#161](https://github.com/open-telemetry/opentelemetry-configuration/issues/161), [#4374](https://github.com/open-telemetry/opentelemetry-specification/issues/4374)). We anticipate limited churn going forward. Additionally, once we do have a stable release, users can rely on the [strong compatibility guarantees](https://github.com/open-telemetry/opentelemetry-configuration?tab=readme-ov-file#stability-definition) we'll conform to in versions 1.0.0+.

### Schema

* **BREAKING:** Refactor otlp exporter to separate `otlp_http` and `otlp_grpc` exporters, add `_file` suffix to OTLP exporter certificate properties ([#146](https://github.com/open-telemetry/opentelemetry-configuration/pull/146), [#166](https://github.com/open-telemetry/opentelemetry-configuration/pull/166))

  <details>
  
  <summary>Migration steps</summary>
  
  ```yaml
  # Before
  tracer_provider:
    processors:
      - batch:
          exporter:
            otlp: 
              protocol: http/protobuf
              endpoint: http://localhost:4318/v1/traces
              certificate: /app/cert.pem
  ---
  # After
  tracer_provider:
    processors:
      - batch:
          exporter:
            otlp_http: 
              endpoint: http://localhost:4318/v1/traces
              certificate_file: /app/cert.pem
  ```
  </details>

[...]

PR #8031, opened 2025-10-17
otelconf: introduce support for v1.0.0-rc.2

codeboten:
This adds support for the second 1.0.0 release candidate of the OpenTelemetry configuration schema. The edits in this PR:

1. update `OPENTELEMETRY_CONFIGURATION_JSONSCHEMA_VERSION`
2. run `make genjsonschema`

First part of https://github.com/open-telemetry/opentelemetry-go-contrib/pull/8026


codeboten:
updated the candidate to rc2 since there's no point in introducing schema support for rc1 with rc2 already out

dmathieu:
I wonder if we want to name this module rcx. The package is unstable, so it can have breaking changes.
Couldn't we name `1.0.0`, and add any change that would happen before the final release into a single module?

codeboten:
> I wonder if we want to name this module rcx

i had the same question when i opened this pr. i guess one of the challenges is that anyone using 1.0.0rcx wouldn't be able ensure that any config they've written would work with updates to rcx, which would be problematic if someone just happens to pick up a dep update that includes a breaking change but isn't expecting it. 

at least with rc.2 and whatever subsequent versions, there would be *some* stability for end users of the package, even though we make no guarantees of stability at this point

codeboten:
Multiple RCs is more code though, but we could always mark the package deprecated as soon as the new rc is out.

codeboten:
Not sure if we want a changelog for this specific change as the generated code as-is cannot be used on its own

dashpole:
Can you remind me why we need to support multiple versions?  Could we not put the 1.0 (including RCs) in otelconf itself?

codeboten:
Needed the multiple versions to support backwards incompatible changes between versions in the collector. The collector supports both v0.2.0 and v0.3.0 configuration at this point, and my plan was to eventually only support 1.0 configuration, but it will need feature gates in the collector and deprecation warnings for end users.

dashpole:
Can we leave 0.2 and 0.3 as separate directories with their own module and put the 1.0 RCs (and 1.0) in the base /otelconf directory. Then eventually otel-go marks the 0.2 and 0.3 modules as deprecated and removes them, but gives the collector enough time to migrate?  That way, we eventually get back to the place of having just one stable entry-point to the otelconf library.  People complain about our semconv code structure a lot, so I would rather avoid that long-term if possible

codeboten:
@dashpole change has been made, the generated code is now in the root of the package. ~i'll update the makefile to change the destination of the generated code in a follow up PR~

also updated the makefile to generate the code in this location

dashpole (approved):

PR #8043, opened 2025-10-20
otelconf: add unmarshaling and validation for span & cardinality limits

codeboten:
Adds unmarshaler code for SpanLimits and CardinalityLimits which require validation to ensure fields are set following the boundaries set by the schema.

Part of breaking up https://github.com/open-telemetry/opentelemetry-go-contrib/pull/8026

### DIFF
--- a/otelconf/internal/kv/keyvalue.go
+++ b/otelconf/internal/kv/keyvalue.go
@@ -1,0 +1,48 @@
+// Copyright The OpenTelemetry Authors
+// SPDX-License-Identifier: Apache-2.0
+
+// Package kv contains function to translate name value pairs into
+// attribute.KeyValue.
+package kv // import "go.opentelemetry.io/contrib/otelconf/internal/kv"
+
+import (
+	"fmt"
+	"strconv"
+
+	"go.opentelemetry.io/otel/attribute"
+)
+
+func FromNameValue(k string, v any) attribute.KeyValue {
+	switch val := v.(type) {
+	case bool:
+		return attribute.Bool(k, val)
+	case int64:
+		return attribute.Int64(k, val)
+	case uint64:
+		return attribute.String(k, strconv.FormatUint(val, 10))
+	case float64:
+		return attribute.Float64(k, val)
+	case int8:
+		return attribute.Int64(k, int64(val))
+	case uint8:
+		return attribute.Int64(k, int64(val))
+	case int16:
+		return attribute.Int64(k, int64(val))
+	case uint16:
+		return attribute.Int64(k, int64(val))
+	case int32:
+		return attribute.Int64(k, int64(val))
+	case uint32:
+		return attribute.Int64(k, int64(val))
+	case float32:
+		return attribute.Float64(k, float64(val))
+	case int:
+		return attribute.Int(k, val)
+	case uint:
+		return attribute.String(k, strconv.FormatUint(uint64(val), 10))
+	case string:
+		return attribute.String(k, val)
+	default:
+		return attribute.String(k, fmt.Sprint(v))
+	}
+}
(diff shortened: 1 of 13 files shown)


codeboten:
> This looks good. But it seems it should have a changelog entry.

changelog added

otelconf/config_common.go
```diff
@@ -0,0 +1,58 @@
...
+	}
+	if plain.ObservableGauge != nil && 0 >= *plain.ObservableGauge {
+		return fmt.Errorf("field %s: must be > %v", "observable_gauge", 0)
+	}
+	if plain.ObservableUpDownCounter != nil && 0 >= *plain.ObservableUpDownCounter {
+		return fmt.Errorf("field %s: must be > %v", "observable_up_down_counter", 0)
+	}
+	if plain.UpDownCounter != nil && 0 >= *plain.UpDownCounter {
+		return fmt.Errorf("field %s: must be > %v", "up_down_counter", 0)
+	}
```
MrAlias, line 71: The repetitive use of `fmt.Errorf` with the same format string and the testing of error strings can likely be improved.

Given these all share the same error format and it would be advantageous to test the error type returned directly a new error type could be added here:

```go
type errBound struct {
	Field string
	Bound int
	Op    string
}

func (e *errBound) Error() string {
	return fmt.Sprintf("field %s: must be %s %d", e.Field, e.Op, e.Bound)
}

// newErrMin creates a new error indicating that the field must be greater than
// or equal to the bound.
func newErrMin(field string, bound int) error {
	return &errBound{Field: field, Bound: bound, Op: ">="}
}

// newErrExclMin creates a new error indicating that the field must be greater
// than the bound.
func newErrExclMin(field string, bound int) error {
	return &errBound{Field: field, Bound: bound, Op: ">"}
}
```

Using that new error type:

```suggestion
	if plain.Counter != nil && 0 >= *plain.Counter {
		return newErrExclMin("counter", 0)
	}
	if plain.Default != nil && 0 >= *plain.Default {
		return newErrExclMin("default", 0)
	}
	if plain.Gauge != nil && 0 >= *plain.Gauge {
		return newErrExclMin("gauge", 0)
	}
	if plain.Histogram != nil && 0 >= *plain.Histogram {
		return newErrExclMin("histogram", 0)
	}
	if plain.ObservableCounter != nil && 0 >= *plain.ObservableCounter {
		return newErrExclMin("observable_counter", 0)
	}
	if plain.ObservableGauge != nil && 0 >= *plain.ObservableGauge {
		return newErrExclMin("observable_gauge", 0)
	}
	if plain.ObservableUpDownCounter != nil && 0 >= *plain.ObservableUpDownCounter {
		return newErrExclMin("observable_up_down_counter", 0)
	}
	if plain.UpDownCounter != nil && 0 >= *plain.UpDownCounter {
		return newErrExclMin("up_down_counter", 0)
	}
```

And similar for below.

This will allow testing to use something like

```go
	[]struct {
		name       string
		yamlConfig []byte
		jsonConfig []byte
		wantErrT   error
	}
```

And the testcase will use the type system instead of string parsing:

```go
cl := CardinalityLimits{}
err := cl.UnmarshalJSON(tt.jsonConfig)
assert.ErrorIs(t, err, tt.wantErrT)

cl = CardinalityLimits{}
err = yaml.Unmarshal(tt.yamlConfig, &cl)
assert.ErrorIs(t, err, tt.wantErrT)
```

codeboten: Thanks for the suggestions, applied them, ptal

otelconf/config_common.go
```diff
@@ -0,0 +1,96 @@
...
+package otelconf // import "go.opentelemetry.io/contrib/otelconf"
+
+import (
+	"errors"
+	"fmt"
+)
+
+var (
+	errUnmarshalingCardinalityLimits = errors.New("unmarshaling error cardinality_limit")
+	errUnmarshalingSpanLimits        = errors.New("unmarshaling error span_limit")
```
MrAlias: nit: no need to include "error" in an error text.

```suggestion
	errUnmarshalingCardinalityLimits = errors.New("unmarshaling cardinality_limit")
	errUnmarshalingSpanLimits        = errors.New("unmarshaling span_limit")
```

PR #8049, opened 2025-10-21
 otelconf: add unmarshaling / validation for batch processors 

codeboten:
Follow up to https://github.com/open-telemetry/opentelemetry-go-contrib/pull/8043, adds unmarshaling for batch processors. Will rebase once #8043 is merged

Part of breaking up https://github.com/open-telemetry/opentelemetry-go-contrib/pull/8026

### DIFF
--- a/otelconf/config_common.go
+++ b/otelconf/config_common.go
@@ -47,7 +47,7 @@ type errUnmarshal struct {
 }
 
 func (e *errUnmarshal) Error() string {
-	return fmt.Sprintf("unmarshal error in %s", reflect.TypeOf(e.Object))
+	return fmt.Sprintf("unmarshal error in %T", e.Object)
 }
 
 func (e *errUnmarshal) Is(target error) bool {


dmathieu (approved):

otelconf/config_json.go
```diff
@@ -8,6 +8,126 @@ import (
...
+	var plain Plain
+	if err := json.Unmarshal(b, &plain); err != nil {
+		return err
+	}
+	// console can be nil, must check and set here
+	if _, ok := raw["console"]; ok {
+		plain.Console = ConsoleExporter{}
+	}
+	*j = PushMetricExporter(plain)
+	return nil
```
MrAlias: The double `Unmarshal` can be avoided by using a different type here that shadows the `Console` field.

```suggestion
	// Use a shadow struct with a RawMessage field to detect key presence.
	type Plain PushMetricExporter
	type shadow struct {
		Plain
		Console json.RawMessage `json:"console"`
	}
	var sh shadow
	if err := json.Unmarshal(b, &sh); err != nil {
		return err
	}
	// If key is present (even if empty object), ensure non-nil value.
	if sh.Console != nil {
		var c ConsoleExporter
		if err := json.Unmarshal(sh.Console, &c); err != nil {
			return err
		}
		sh.Plain.Console = c
	}
	*j = PushMetricExporter(sh.Plain)
	return nil
```

Similar for the `SpanExporter` and `LogRecordExporter`.

otelconf/config_json.go
```diff
@@ -8,6 +8,126 @@ import (
...
+	type Plain BatchLogRecordProcessor
+	var plain Plain
+	if err := json.Unmarshal(b, &plain); err != nil {
+		return errors.Join(errUnmarshalingBatchLogRecordProcessor, err)
+	}
+	if err := validateBatchLogRecordProcessor((*BatchLogRecordProcessor)(&plain)); err != nil {
+		return err
+	}
+	*j = BatchLogRecordProcessor(plain)
+	return nil
```
MrAlias: Similar here, the double `Unmarshal` can be avoided.

```suggestion
	type Plain BatchLogRecordProcessor
	type shadow struct {
		Plain
		Exporter json.RawMessage `json:"exporter"`
	}
	var sh shadow
	if err := json.Unmarshal(b, &sh); err != nil {
		return errors.Join(errUnmarshalingBatchLogRecordProcessor, err)
	}
	if sh.Exporter == nil {
		return newErrRequiredExporter("BatchLogRecordProcessor")
	}
	// Hydrate the exporter into the underlying field.
	if err := json.Unmarshal(sh.Exporter, &sh.Plain.Exporter); err != nil {
		return errors.Join(errUnmarshalingBatchLogRecordProcessor, err)
	}
	err := validateBatchLogRecordProcessor((*BatchLogRecordProcessor)(&sh.Plain))
	if err != nil {
		return err
	}
	*j = BatchLogRecordProcessor(sh.Plain)
	return nil
```

Similar for `BatchSpanProcessor` and `PeriodicMetricReader`.

otelconf/config_test.go
```diff
@@ -10,6 +10,246 @@ import (
...
+		},
+		{
+			name:       "valid with zero export_timeout",
+			jsonConfig: []byte(`{"exporter":{"console":{}},"export_timeout":0}`),
+			yamlConfig: []byte("exporter:\n  console: {}\nexport_timeout: 0"),
+		},
+		{
+			name:       "valid with zero schedule_delay",
+			jsonConfig: []byte(`{"exporter":{"console":{}},"schedule_delay":0}`),
+			yamlConfig: []byte("exporter:\n  console: {}\nschedule_delay: 0"),
```
MrAlias: The `PeriodicMetricReader` doesn't have `export_timeout` or `schedule_delay` fields. It has `timeout` and `interval` fields instead.

```suggestion
			jsonConfig: []byte(`{"exporter":{"console":{}},"timeout":0}`),
			yamlConfig: []byte("exporter:\n  console: {}\ntimeout: 0"),
		},
		{
			name:       "valid with zero schedule_delay",
			jsonConfig: []byte(`{"exporter":{"console":{}},"interval":0}`),
			yamlConfig: []byte("exporter:\n  console: {}\ninterval: 0"),
```

Or, is this a test for the default value of `0` being interpreted when unset?

otelconf/config_yaml.go
```diff
@@ -9,6 +9,126 @@ import (
...
+	var plain Plain
+	if err := node.Decode(&plain); err != nil {
+		return errors.Join(errUnmarshalingPeriodicMetricReader, err)
+	}
+	if err := validatePeriodicMetricReader((*PeriodicMetricReader)(&plain)); err != nil {
+		return err
+	}
+	*j = PeriodicMetricReader(plain)
+	return nil
+}
```
MrAlias, line 121: The duplicate decode can be avoided with an explicit node traversal:

```suggestion
// hasYAMLMapKey reports whether the provided mapping node contains the given
// key. It assumes the node is a mapping node and performs a linear scan of its
// key nodes.
func hasYAMLMapKey(node *yaml.Node, key string) bool {
	if node == nil || node.Kind != yaml.MappingNode {
		return false
	}
	for i := 0; i+1 < len(node.Content); i += 2 {
		if node.Content[i].Kind == yaml.ScalarNode && node.Content[i].Value == key {
			return true
		}
	}
	return false
}

// UnmarshalYAML implements yaml.Unmarshaler.
func (j *PeriodicMetricReader) UnmarshalYAML(node *yaml.Node) error {
	if !hasYAMLMapKey(node, "exporter") {
		return newErrRequiredExporter("PeriodicMetricReader")
	}
	type Plain PeriodicMetricReader
	var plain Plain
	if err := node.Decode(&plain); err != nil {
		return errors.Join(errUnmarshalingPeriodicMetricReader, err)
	}
	if err := validatePeriodicMetricReader((*PeriodicMetricReader)(&plain)); err != nil {
		return err
	}
	*j = PeriodicMetricReader(plain)
	return nil
}
```

I was able to verify this is an optimization using the following:

```go
func BenchmarkYAMLUnmarshalPeriodicMetricReader(b *testing.B) {
	yamlConfig := []byte(`exporter:
  console: {}
timeout: 5000
interval: 1000`)
	for b.Loop() {
		var v PeriodicMetricReader
		if err := yaml.Unmarshal(yamlConfig, &v); err != nil {
			b.Fatal(err)
		}
	}
}
```

```console
goos: linux
goarch: amd64
pkg: go.opentelemetry.io/contrib/otelconf
cpu: Intel(R) Core(TM) i7-8550U CPU @ 1.80GHz
                                    │ orig.bmark.result │          alt.bmark.result           │
                                    │      sec/op       │   sec/op     vs base                │
YAMLUnmarshalPeriodicMetricReader-8         16.64µ ± 5%   12.77µ ± 3%  -23.24% (p=0.000 n=10)

                                    │ orig.bmark.result │           alt.bmark.result           │
                                    │       B/op        │     B/op      vs base                │
YAMLUnmarshalPeriodicMetricReader-8       10.680Ki ± 0%   9.023Ki ± 0%  -15.51% (p=0.000 n=10)

                                    │ orig.bmark.result │          alt.bmark.result          │
                                    │     allocs/op     │ allocs/op   vs base                │
YAMLUnmarshalPeriodicMetricReader-8         114.00 ± 0%   83.00 ± 0%  -27.19% (p=0.000 n=10)
```

Similarly for the other yaml unmarshal functions, the `hasYAMLMapKey` function can be used to avoid to duplicate.

otelconf/config_json.go
codeboten: applied the feedback for all exporters 👍🏻

otelconf/config_yaml.go
codeboten: Applied the suggestion for unmarshaling the exporters and the processors 👍🏻

otelconf/config_json.go
codeboten: Done for all 3

codeboten:
Thanks for the suggestions @MrAlias, applied the suggestions PTAL

otelconf/config_common.go
```diff
@@ -31,6 +26,38 @@ func (e *errBound) Is(target error) bool {
...
+	}
+	return reflect.TypeOf(e.Object) == reflect.TypeOf(t.Object)
+}
+
+type errUnmarshal struct {
+	Object any
+}
+
+func (e *errUnmarshal) Error() string {
+	return fmt.Sprintf("unmarshal error in %s", reflect.TypeOf(e.Object))
```
MrAlias:
```suggestion
	return fmt.Sprintf("unmarshal error in %T", e.Object)
```

PR #8050, opened 2025-10-21
otelconf: refactor code to internal package

codeboten:
There is no functional change here, only moving the code as the 1.0.0 version will be able to reuse this code in the common location.

Part of #8026


codeboten:
Since there's no functional change i didn't think this needed a changelog

PR #8054, opened 2025-10-22
chore(sample): otel http server sample

naonao2323:


instrumentation/net/http/httptrace/otelhttptrace/example/server/server.go
```diff
@@ -59,6 +59,7 @@ func main() {
 	helloHandler := func(w http.ResponseWriter, req *http.Request) {
 		ctx := req.Context()
 		span := trace.SpanFromContext(ctx)
+		defer span.End()
```
flc1125, line 62: This is the span extracted from the context for upstream. Since `defer span.End()` has already been executed at the underlying level, there is no need to perform it here.

FYI: https://github.com/open-telemetry/opentelemetry-go-contrib/blob/ad58c3a23580c5e441928602b45a447e692af481/instrumentation/net/http/otelhttp/handler.go#L129

dmathieu:
Closing as invalid. See @flc1125's explanation.

Issue #3653, opened 2023-03-26
Missing sampler.type & sampler.param attributes/ tags in Jaeger samplers


ChenX1993:
## Background
Hi, thanks for anyone who contributed to jaeger remote sampler!    
I am trying to use jaeger remote sampler, but I notice in the currently implementation of each jaeger sampler, `sampler.type` and `sampler.param` attributes/ tags are not set, e.g in probabilisticSampler:
```
func (s *probabilisticSampler) ShouldSample(p trace.SamplingParameters) trace.SamplingResult {
       ...
	return trace.SamplingResult{
		Decision:   trace.Drop,
		Tracestate: psc.TraceState(),
	}
}
```

While in the deprecated jaeger-client-go, when a sampler is making the sampling decision, `sampler.type` and `sampler.param` tags are set, e.g 
```
func (s *ProbabilisticSampler) init(samplingRate float64) *ProbabilisticSampler {
	...
	s.tags = []Tag{
		{key: SamplerTypeTagKey, value: SamplerTypeProbabilistic},
		{key: SamplerParamTagKey, value: s.samplingRate},
	}
	return s
}

func (s *ProbabilisticSampler) IsSampled(id TraceID, operation string) (bool, []Tag) {
	return s.samplingBoundary >= id.Low&maxRandomNumber, s.tags
}
``` 
[Reference link](https://github.com/jaegertracing/jaeger-client-go/blob/master/sampler.go#L144)

These two sampler tags are essential to Jaeger [adaptive sampling](https://www.jaegertracing.io/docs/1.43/sampling/#adaptive-sampling) (more detailed explanation [here](https://medium.com/jaegertracing/adaptive-sampling-in-jaeger-50f336f4334)). 
In Jaeger backend (more specifically jaeger-collector), spans without sampler.type tag will not contribute to the span throughput which is needed for further adaptive probability calculation. 
```
// handleRootSpan returns a function that records throughput for root spans
func handleRootSpan(aggregator strategystore.Aggregator, logger *zap.Logger) ProcessSpan {
	return func(span *model.Span, tenant string) {
		....
		samplerType, samplerParam := span.GetSamplerParams(logger)
		if samplerType == "" {
			return
		}
		aggregator.RecordThroughput(service, span.OperationName, samplerType, samplerParam)
	}
}
```
[...]

PR #8073, opened 2025-10-26
feat: add sampling attributes for jaeger remote sampler

etilite:
### Which problem is this PR solving?

Resolves https://github.com/open-telemetry/opentelemetry-go-contrib/issues/3653
Resolves https://github.com/open-telemetry/opentelemetry-go-contrib/issues/691

### Description of the changes

- Added option `WithAttributesOn` which adds `jaeger.sampler.param` and `jaeger.sampler.type` attributes for sampled spans


flc1125:
cc code owner: @yurishkuro PTAL

dmathieu (changes requested):
Those attributes don't match anything defined in semantic conventions.

samplers/jaegerremote/sampler_remote_options.go
```diff
@@ -120,6 +120,11 @@ func newConfig(options ...Option) config {
 		MaxOperations:            c.posParams.MaxOperations,
 		OperationNameLateBinding: c.posParams.OperationNameLateBinding,
 	}}, c.updaters...)
+
+	if c.sampler == nil {
```
yurishkuro, line 124: isn't this always true?

samplers/jaegerremote/sampler.go
```diff
@@ -34,25 +35,55 @@ const (
...
+	samplerParamKey               = "sampler.param"
+)
+
+type samplerOptions struct {
+	attributesOn bool
+}
+
+type samplerOptionFunc func(*samplerOptions)
+
+func withAttributesOn() samplerOptionFunc {
```
yurishkuro: I would make then on by default, and have this option to turn off. The function needs to be public in order to be used from a different package.

```suggestion
func WithAttributesDisabled() samplerOptionFunc {
```

samplers/jaegerremote/sampler.go
```diff
@@ -34,25 +35,55 @@ const (
...
+
 // -----------------------
 
 // probabilisticSampler is a sampler that randomly samples a certain percentage
 // of traces.
 type probabilisticSampler struct {
 	samplingRate float64
 	sampler      trace.Sampler
+	attributes   []attribute.KeyValue
+	attributesOn bool
```
yurishkuro: I think this flag is only used at initialization, you don't really need to store it

samplers/jaegerremote/sampler.go
```diff
@@ -34,25 +35,55 @@ const (
 	defaultMaxOperations = 2000
 )
 
+const (
+	samplerTypeKey                = "sampler.type"
+	samplerTypeValueProbabilistic = "probabilistic"
+	samplerTypeValueRateLimiting  = "ratelimiting"
+	samplerParamKey               = "sampler.param"
```
yurishkuro: I think it would make sense to prefix with `jaeger.`, This may require small change in the backend to recognize, but it separates custom attributes into clear namespace.
```suggestion
	samplerTypeKey                = "jaeger.sampler.type"
	samplerParamKey               = "jaeger.sampler.param"
	samplerTypeValueProbabilistic = "probabilistic"
	samplerTypeValueRateLimiting  = "ratelimiting"
```

yurishkuro:
Thank you for this PR!  Originally when I looked at this problem there was no way in the API to pass back attributes (I think I even had an option issue about that).

samplers/jaegerremote/sampler_remote_options.go
etilite: This is false when option `WithInitialSampler()` is used with some non-nil sampler.

samplers/jaegerremote/sampler.go
etilite: I wasn't sure that we want to change current default behavior - jaegerremote does not set any attributes.

This is supposed to be an internal option for creating and updating samplers which constructors also aren't exported.

For external usage I added `WithAttributesOn` option.

samplers/jaegerremote/sampler.go
etilite: When `probabilisticSampler` or `ratelimitngSampler` are being updated its `init` method is also called, so I not sure if we can omit this flag.

etilite:
> Those attributes don't match anything defined in semantic conventions.

Can we do anything about it? Those attributes provide back-compatibility with jaeger-remote-sampler and can be turned only by an option. So yeah it is kinda jaeger attributes.

samplers/jaegerremote/sampler.go
etilite: Yeah, this is reasonable but I see some difficulties when using in mixed environments (jaeger and otel services) where will be different sampler attributes in that case. Not sure if backend can support more that one key.

etilite:
@dmathieu @yurishkuro If you’re okay with it, I could try a different approach - add two new options - functions that create attributes for probabilistic/ratelimiting samplers, so user can define their own attributes to solve this issue.
This solution does not interfere with the semantic convention.

samplers/jaegerremote/sampler.go
yurishkuro: this is a Jaeger-specific functionality and a Jaeger-specific sampler. Since it's not setting any attributes now then nothing is depending on them.

dmathieu:
Sorry, my assessment was wrong. As these are jaeger specific, they shouldn't need semantic conventions.

samplers/jaegerremote/sampler.go
etilite: So can we keep canonical keys? I did small research and using new keys will require some overhead on collector to transform these attributes.

etilite:
@yurishkuro I added new test in different package and answered on all comments. Review pls

flc1125:
> @yurishkuro I added new test in different package and answered on all comments. Review pls

We can fix the CI errors before entering the review to reduce some trivial checks. 😁

flc1125:
@etilite please add a CHANGELOG

flc1125:
@etilite There are currently some conflicts; please fix them.

etilite:
@flc1125 resolved conflicts but now ci/cd failing. Probably need just to retry these jobs, can't do it myself.

samplers/jaegerremote/sampler.go
dmathieu: Since those are jaeger specific, prefixing them with jaeger definitely makes sense.

etilite: Added prefix

samplers/jaegerremote/sampler.go
```diff
@@ -34,25 +35,55 @@ const (
...
 )
 
+const (
+	samplerTypeKey                = "jaeger.sampler.type"
+	samplerParamKey               = "jaeger.sampler.param"
+	samplerTypeValueProbabilistic = "probabilistic"
+	samplerTypeValueRateLimiting  = "ratelimiting"
+)
+
+type samplerOptions struct {
```
yurishkuro: this is redundant. The package already has the functional options mechanism
```
func New(
        serviceName string,
        opts ...Option,
) *Sampler {
        options := newConfig(opts...)
```
So `WithAttributesOn()` should be implemented as `Option` just like other options, and should set a flag on the `config` struct.

samplers/jaegerremote/sampler.go
yurishkuro: I do want to change the current behavior. The current behavior is useless, the sampler does not actually work with Jaeger's adaptive backend. And the change itself is additive, it's backwards compatible (especially since the new attributes are prefixed with `jaeger.`).

yurishkuro (changes requested):

samplers/jaegerremote/sampler.go
etilite: do we need an option to disable these attributes in that case? If not implementation will be simple and no need for internal sampler options too

samplers/jaegerremote/sampler.go
etilite: the idea behind these internal options is to have different behavior when creating samplers without need to change every test to add new argument in constructor.
I can use flag arg in sampler's constructors if it is ok with you.

samplers/jaegerremote/sampler.go
yurishkuro: Personally I would argue you don't even need the opt-out option, OTEL SDK is already extremely verbose in terms of how many attributes it captures, having two extra ones on root span only seems irrelevant. But for 100% backwards compat you would provide an option to opt-out.

samplers/jaegerremote/sampler.go
yurishkuro: > without need to change every test to add new argument in constructor.

a) why do you need to change every test? do they all compare attributes?
b) even if you need to, it seems like a trivial task for a Copilot

etilite: a) cause a lot of tests call these constructors
b) yeah but it is harder to review a lot of small changes
nvm I'll change to flag argument 👍

samplers/jaegerremote/sampler_remote.go
```diff
@@ -61,7 +61,7 @@ type samplingStrategyParser interface {
 //
 // Sampler invokes the updaters while holding a lock on the main sampler.
 type samplerUpdater interface {
-	Update(sampler trace.Sampler, strategy any) (modified trace.Sampler, err error)
+	Update(sampler trace.Sampler, strategy any, attributesDisabled bool) (modified trace.Sampler, err error)
```
yurishkuro: It is not necessary to change this interface to pass `attributesDisabled`. When individual updaters are created the value of this flag is known and can be captured in the updaters themselves. This way we separate configuration information from the execution information (only sampler and strategy are variable at runtime)

etilite: Good point! Can I remove `withUpdaters` option then? Cause it makes harder to properly create updaters and pass `attributesDisabled` to them. Anyway `withUpdaters` is not needed for tests now (can be safely removed) and seems to be an old arfifact

yurishkuro: `withUpdaters` is private, safe to remove if not needed. These APIs used to be public for extensibility.

yurishkuro (approved):

yurishkuro:
👏 

@etilite do you want to make the corresponding changes in Jaeger to recognize the prefixed attributes?

etilite:
👏

@yurishkuro I'm not familiar with Jaeger backend library so it will be nice if you can handle it

PR #8077, opened 2025-10-27
otelconf: adding remaining test files

codeboten:
Split from #8026. Tests that use these files will be added shortly.


codeboten:
Probably safe to skip changelog for this one

dmathieu (approved):

PR #8086, opened 2025-10-28
otelconf: add newResource internal func

codeboten:
This will be used via the NewSDK public function. Once the other pieces are in place, the public API will be added.

Internal API only, can probably skip changelog until the public api is introduced. Part of splitting https://github.com/open-telemetry/opentelemetry-go-contrib/pull/8026

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -17,6 +17,10 @@ This project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0.htm
 - Add unmarshaling and validation for `BatchLogRecordProcessor`, `BatchSpanProcessor`, and `PeriodicMetricReader` to v1.0.0 model in `go.opentelemetry.io/contrib/otelconf`. (#8049)
 - Add unmarshaling and validation for `TextMapPropagator` to v1.0.0 model in `go.opentelemetry.io/contrib/otelconf`. (#8052)
 
+### Changed
+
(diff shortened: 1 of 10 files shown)


otelconf/resource.go
```diff
@@ -0,0 +1,32 @@
...
+	if res == nil {
+		return resource.Default(), nil
+	}
+
+	r, ok := res.(*ResourceJson)
+	if !ok {
+		return nil, newErrInvalid("resource")
+	}
+
+	var attrs []attribute.KeyValue
```
MrAlias:
```suggestion
	attrs := make([]attribute.KeyValue, 0, len(r.Attributes))
```

otelconf/resource_test.go
```diff
@@ -0,0 +1,89 @@
...
+		},
+		{
+			name: "resource-with-attributes",
+			config: &ResourceJson{
+				Attributes: []AttributeNameValue{
+					{Name: "service.name", Value: "service-a"},
+				},
+			},
+			wantResource: resource.NewWithAttributes("",
+				semconv.ServiceName("service-a"),
```
MrAlias: These could drift. We don't want to test `semconv` stability here.

```suggestion
					{Name: semconv.ServiceNameKey, Value: "service-a"},
				},
			},
			wantResource: resource.NewWithAttributes("",
				semconv.ServiceName("service-a"),
```

Similar below.

codeboten: updated

PR #8089, opened 2025-10-28
otelconf: add logger provider internal funcs

codeboten:
This adds the internal functions to configure logger provider. No public API yet, part of splitting https://github.com/open-telemetry/opentelemetry-go-contrib/pull/8026

Most of this code is copied from v0.3.0, updated to support v1.0.0, and also updated to return more meaningfully typed errors. 

Follow up to https://github.com/open-telemetry/opentelemetry-go-contrib/pull/8086

### DIFF
--- a/bridges/prometheus/go.mod
+++ b/bridges/prometheus/go.mod
@@ -21,7 +21,7 @@ require (
 	github.com/munnerz/goautoneg v0.0.0-20191010083416-a7dc8b61c822 // indirect
 	github.com/pmezard/go-difflib v1.0.0 // indirect
 	github.com/prometheus/common v0.67.2 // indirect
-	github.com/prometheus/procfs v0.19.1 // indirect
+	github.com/prometheus/procfs v0.19.2 // indirect
 	go.opentelemetry.io/auto/sdk v1.2.1 // indirect
 	go.opentelemetry.io/otel/metric v1.38.0 // indirect
 	go.opentelemetry.io/otel/trace v1.38.0 // indirect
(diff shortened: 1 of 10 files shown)


otelconf/config_common.go
```diff
@@ -201,3 +223,27 @@ func validateSpanLimits(plain *SpanLimits) error {
...
+		// Parsing follows https://github.com/open-telemetry/opentelemetry-configuration/blob/568e5080816d40d75792eb754fc96bde09654159/schema/type_descriptions.yaml#L584.
+		headerslist, err := baggage.Parse(*headersList)
+		if err != nil {
+			return nil, errors.Join(newErrInvalid("invalid headers_list"), err)
+		}
+		for _, kv := range headerslist.Members() {
+			result[kv.Key()] = kv.Value()
+		}
+	}
+	// Headers take precedence over HeadersList, so this has to be after HeadersList is processed
```
MrAlias:
```suggestion
	// Headers take precedence over HeadersList, so this has to be after HeadersList is processed.
```

otelconf/config_common.go
```diff
@@ -201,3 +223,27 @@ func validateSpanLimits(plain *SpanLimits) error {
...
+	if len(headers) > 0 {
+		for _, kv := range headers {
+			if kv.Value != nil {
+				result[kv.Name] = *kv.Value
+			}
+		}
+	}
```
MrAlias: The `if` statement is not needed.

```suggestion
	for _, kv := range headers {
		if kv.Value != nil {
			result[kv.Name] = *kv.Value
		}
	}
```

otelconf/log.go
```diff
@@ -0,0 +1,236 @@
...
+)
+
+func loggerProvider(cfg configOptions, res *resource.Resource) (log.LoggerProvider, shutdownFunc, error) {
+	provider, ok := cfg.opentelemetryConfig.LoggerProvider.(*LoggerProviderJson)
+	if provider == nil {
+		return noop.NewLoggerProvider(), noopShutdown, nil
+	}
+	if !ok {
+		return noop.NewLoggerProvider(), noopShutdown, newErrInvalid("logger_provider")
+	}
```
MrAlias: `provider` will always be `nil` when `ok` is `false`.

```suggestion
	if !ok {
		return noop.NewLoggerProvider(), noopShutdown, newErrInvalid("logger_provider")
	}
	if provider == nil {
		return noop.NewLoggerProvider(), noopShutdown, nil
	}
```

dmathieu (approved):

PR #8094, opened 2025-10-29
otelconf: add meter provider internal funcs

codeboten:
This adds the internal functions to configure meter provider. No public API yet, part of splitting https://github.com/open-telemetry/opentelemetry-go-contrib/pull/8026

Most of this code is copied from v0.3.0, updated to support v1.0.0, and also updated to return more meaningfully typed errors.

Follow up to https://github.com/open-telemetry/opentelemetry-go-contrib/pull/8089

### DIFF
--- a/otelconf/metric.go
+++ b/otelconf/metric.go
@@ -335,6 +335,10 @@ func newIncludeExcludeFilter(lists *IncludeExclude) (attribute.Filter, error) {
 			return false
 		}
 
+		if len(included) == 0 {
+			return true
+		}
+
 		_, ok := included[kv.Key]
 		return ok
 	}, nil


otelconf/metric.go
```diff
@@ -0,0 +1,591 @@
...
+		if u.Path != "" {
+			opts = append(opts, otlpmetrichttp.WithURLPath(u.Path))
+		}
+	}
+	if otlpConfig.Compression != nil {
+		switch *otlpConfig.Compression {
+		case compressionGzip:
+			opts = append(opts, otlpmetrichttp.WithCompression(otlpmetrichttp.GzipCompression))
+		case compressionNone:
+			opts = append(opts, otlpmetrichttp.WithCompression(otlpmetrichttp.NoCompression))
```
dashpole, line 185: Is NoCompression the default, or is that only for grpc?

otelconf/metric.go
```diff
@@ -0,0 +1,591 @@
...
+	default:
+		return metricdata.CumulativeTemporality
+	}
+}
+
+// newIncludeExcludeFilter returns a Filter that includes attributes
+// in the include list and excludes attributes in the excludes list.
+// It returns an error if an attribute is in both lists
+//
+// If IncludeExclude is empty a include-all filter is returned.
```
MrAlias:
```suggestion
// If IncludeExclude is empty an include-all filter is returned.
```

otelconf/metric.go
```diff
@@ -0,0 +1,591 @@
...
+	return func(kv attribute.KeyValue) bool {
+		// check if a value is excluded first
+		if _, ok := excluded[kv.Key]; ok {
+			return false
+		}
+
+		if len(included) == 0 {
+			return true
+		}
+
```
MrAlias, line 341: nit: this is unneeded. Performance is equivalent or very slightly better without.

```suggestion
```

codeboten: i ended up reverting this change as it returns true when the map is empty but false if the key isn't present and the map is empty

otelconf/metric.go
codeboten: Right, i guess the default value for compression for http is also `NoCompression` based on the default value of the `Compression` enum. Probably good to be more specific than not in case that ever changes in the future.

Might be worth updating the grpc code to specifically set the compressor as well for consistency. Looking at the implementations for otlpgrpc for logs, metrics, and traces, the code does different things that end up setting the NoCompression option. Happy to follow this up in another PR if you're ok with it

dashpole: Whatever you decide is fine with me :)

MrAlias: https://github.com/open-telemetry/opentelemetry-specification/blob/main/specification/protocol/exporter.md#:~:text=If%20no%20compression%20value%20is%20explicitly%20specified%2C%20SIGs%20can%20default%20to%20the%20value%20they%20deem%20most%20useful%20among%20the%20supported%20options.%20This%20is%20especially%20important%20in%20the%20presence%20of%20technical%20constraints%2C%20e.g.%20directly%20sending%20telemetry%20data%20from%20mobile%20devices%20to%20backend%20servers.

> If no compression value is explicitly specified, SIGs can default to the value they deem most useful among the supported options. This is especially important in the presence of technical constraints, e.g. directly sending telemetry data from mobile devices to backend servers.

PR #8106, opened 2025-10-30
otelconf: add support for OTEL_EXPERIMENTAL_CONFIG_FILE

### LINKED ISSUE #8059: otelconf: add support for OTEL_EXPERIMENTAL_CONFIG_FILE variable
### Component

Config

### Problem Statement

As per the spec:

> Setting the [OTEL_EXPERIMENTAL_CONFIG_FILE](https://github.com/open-telemetry/opentelemetry-specification/blob/main/specification/configuration/sdk-environment-variables.md#declarative-configuration) environment variable (for languages that support it) provides users a convenient way to initialize OpenTelemetry components without needing to learn language-specific configuration details or use a large number of environment variables. The pattern for accessing the configured components and installing into instrumentation will vary by language. For example, the usage in Java might resemble:

https://github.com/open-telemetry/opentelemetry-specification/blob/main/specification/configuration/sdk.md#via-otel_experimental_config_file

### Proposed Solution

Open to suggestions of how this should be implemented to align w/ the rest of the Go SDK implementation

### Alternatives

_No response_

### Prior Art

_No response_

### Additional Context

_No response_

codeboten:
As discussed in #8059, adding an init function to support users passing in a
configuration file via the environment variables `OTEL_EXPERIMENTAL_CONFIG_FILE`.

Fixes #8059

~This will remain in draft until https://github.com/open-telemetry/opentelemetry-go-contrib/pull/8026 is merged~

### DIFF
--- a/otelconf/log_test.go
+++ b/otelconf/log_test.go
@@ -854,7 +854,7 @@ func Test_otlpGRPCLogExporter(t *testing.T) {
 			}
 
 			assert.EventuallyWithT(t, func(collect *assert.CollectT) {
-				assert.NoError(collect, exporter.Export(t.Context(), []sdklog.Record{
+				assert.NoError(collect, exporter.Export(context.Background(), []sdklog.Record{ //nolint:usetesting // required to avoid getting a canceled context.
 					logFactory.NewRecord(),
 				}))
 			}, 10*time.Second, 1*time.Second)
--- a/otelconf/metric_test.go
+++ b/otelconf/metric_test.go
@@ -1557,7 +1557,7 @@ func Test_otlpGRPCMetricExporter(t *testing.T) {
 			require.NoError(t, err)
 
 			assert.EventuallyWithT(t, func(collect *assert.CollectT) {
-				assert.NoError(collect, exporter.Export(t.Context(), &metricdata.ResourceMetrics{
+				assert.NoError(collect, exporter.Export(context.Background(), &metricdata.ResourceMetrics{ //nolint:usetesting // required to avoid getting a canceled context.
 					Resource: res,
 					ScopeMetrics: []metricdata.ScopeMetrics{
 						{
--- a/otelconf/trace_test.go
+++ b/otelconf/trace_test.go
@@ -1026,7 +1026,7 @@ func Test_otlpGRPCTraceExporter(t *testing.T) {
 			}
 
 			assert.EventuallyWithT(t, func(collect *assert.CollectT) {
-				assert.NoError(collect, exporter.ExportSpans(t.Context(), input.Snapshots()))
+				assert.NoError(collect, exporter.ExportSpans(context.Background(), input.Snapshots())) //nolint:usetesting // required to avoid getting a canceled context.
 			}, 10*time.Second, 1*time.Second)
 		})
 	}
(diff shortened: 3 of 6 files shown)


codeboten:
Hoping to get some feedback on this before i go through and add unit test coverage

otelconf/config.go
```diff
@@ -58,6 +62,50 @@ var noopSDK = SDK{
...
+	}
+
+	// Set the global providers.
+	otel.SetTracerProvider(s.TracerProvider())
+	otel.SetMeterProvider(s.MeterProvider())
+	global.SetLoggerProvider(s.LoggerProvider())
+	sdk = &s
+}
+
+// Shutdown calls the shutdown function of the global SDK instantiated if
```
MrAlias: This looks cut off.

otelconf/config.go
```diff
@@ -58,6 +62,50 @@ var noopSDK = SDK{
...
+func Shutdown(ctx context.Context) {
+	if sdk == nil {
+		return
+	}
+	if err := sdk.Shutdown(ctx); err != nil {
+		log.Fatal(err)
```
MrAlias: We do not want to crash a user application just because this fails. Should an error be logged instead of fatal?

otelconf/config.go
```diff
@@ -58,6 +62,50 @@ var noopSDK = SDK{
...
+	// Create SDK components with the parsed configuration.
+	s, err := NewSDK(WithOpenTelemetryConfiguration(*c))
+	if err != nil {
+		log.Fatal(err)
+	}
+
+	// Set the global providers.
+	otel.SetTracerProvider(s.TracerProvider())
+	otel.SetMeterProvider(s.MeterProvider())
+	global.SetLoggerProvider(s.LoggerProvider())
```
MrAlias: Doing something like this in `init` will cause side-effects from just importing this package, possibly conflicting with other `init`.

Why not provide integration with the explicit creation functions similar to what is done in [`autoprop`](https://pkg.go.dev/go.opentelemetry.io/contrib/propagators/autoprop#NewTextMapPropagator), [`autoexport`](https://pkg.go.dev/go.opentelemetry.io/contrib/exporters/autoexport#NewSpanExporter), ... packages that parse the environment variable and return a constructed type?

otelconf/config.go
codeboten: Updated

otelconf/config.go
codeboten: Updated the code to use this approach as it also aligns with the suggestion in https://github.com/open-telemetry/opentelemetry-specification/blob/main/specification/configuration/sdk.md#via-otel_experimental_config_file

otelconf/config.go
codeboten: code has been removed

codeboten:
Rebased on top of https://github.com/open-telemetry/opentelemetry-go-contrib/pull/8228 to make reviewing easier

otelconf/config.go
```diff
@@ -58,8 +61,33 @@ var noopSDK = SDK{
...
+	if err != nil {
+		return nil, err
+	}
+
+	// Create SDK components with the parsed configuration.
+	return WithOpenTelemetryConfiguration(*c), nil
+}
+
+// NewSDK creates SDK providers based on the configuration model. It checks the local environment and
+// uses the file set in the variable `OTEL_EXPERIMENTAL_CONFIG_FILE` to configure the SDK automatically.
```
MrAlias, line 81:
```suggestion
// uses the file set in the variable `OTEL_EXPERIMENTAL_CONFIG_FILE` to configure the SDK automatically.
// Any file defined by `OTEL_EXPERIMENTAL_CONFIG_FILE` will supersede all files passed with
// [WithOpenTelemetryConfiguration].
```

PR #8108, opened 2025-10-31
Added Test for http.Handler returned from zpages.NewTracezHandler

### LINKED ISSUE #7051: Test the `http.Handler` returned from `zpages.NewTracezHandler`
There is currently zero coverage of this code. The added tests need to comprehensively cover the functionality provided.

sonalgaud12:
Closes  #7051 

This PR introduces tests for the TracezHandler in the zpages package.
The tests validate the handler’s ability to correctly serve span data, handle query parameters

Result result:

image


CHANGELOG.md
```diff
@@ -16,6 +16,7 @@ This project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0.htm
 - Add unmarshaling and validation for `CardinalityLimits` and `SpanLimits` to v1.0.0 model in `go.opentelemetry.io/contrib/otelconf`. (#8043)
 - Add unmarshaling and validation for `BatchLogRecordProcessor`, `BatchSpanProcessor`, and `PeriodicMetricReader` to v1.0.0 model in `go.opentelemetry.io/contrib/otelconf`. (#8049)
 - Add unmarshaling and validation for `TextMapPropagator` to v1.0.0 model in `go.opentelemetry.io/contrib/otelconf`. (#8052)
+- Add Test for `http.Handler` returned from `zpages.NewTracezHandler`. (#8108)
```
dmathieu: No need for changelog entries in non user-facing changes.

zpages/tracez_test.go
```diff
@@ -0,0 +1,464 @@
...
+	body := w.Body.String()
+	if body == "" {
+		t.Error("expected non-empty response body")
+	}
+}
+
+func TestTracezHandler_ServeHTTP_WithRealSpans(t *testing.T) {
+	// a real tracer with our span processor
+	sp := NewSpanProcessor()
+	defer sp.Shutdown(context.Background())
```
dmathieu:
```suggestion
	defer func() {
		require.NoError(t, sp.Shutdown(context.Background()))
	}()
```

Same for the other lint issues.

dmathieu:
This introduces a lot of tests that seems to do similar things. Could they be merged into a single test (or maybe two) with subtests?

zpages/tracez_test.go
```diff
@@ -0,0 +1,258 @@
...
+			done <- true
+		}()
+	}
+
+	for range 5 {
+		<-done
+	}
+}
+
+func TestTracezHandler_Integration(t *testing.T) {
```
dmathieu: How is this test different from the first one? (apart from setting global config, which could cause issues with the following tests)

sonalgaud12: Thanks for the feedback, The test is meant to verify zPages behavior with the global tracer provider (otel.SetTracerProvider), unlike the earlier test which uses a local provider. I’ll update it to avoid any side effects.

zpages/tracez_test.go
```diff
@@ -0,0 +1,218 @@
...
+				}
+				for _, notWant := range tt.wantNotContain {
+					assert.NotContains(t, body, notWant)
+				}
+			}
+		})
+	}
+}
+
+func TestTracezHandler_ConcurrentRequests(t *testing.T) {
```
dmathieu:
```suggestion
func TestTracezHandler_ConcurrentSafe(t *testing.T) {
```

See https://github.com/open-telemetry/opentelemetry-go/blob/main/CONTRIBUTING.md#testing

zpages/tracez_test.go
```diff
@@ -0,0 +1,218 @@
...
+	ctx := t.Context()
+
+	for range 10 {
+		_, span := tracer.Start(ctx, "concurrent-span")
+		span.End()
+	}
+
+	handler := NewTracezHandler(sp)
+
+	done := make(chan bool, 5)
```
dmathieu: Rather than using a channel, it seems using a mutex would be clearer.
See https://github.com/open-telemetry/opentelemetry-go-contrib/blob/6342fe6da91e0e5d126ac69774dba3ccc3cf559a/propagators/autoprop/registry_test.go#L42 for an example.

PR #8129, opened 2025-11-05
refactor(otelgrpc): introduce `optionFunc` type to reduce boilerplate option structs

flc1125:
Use the `optionFunc` function to replace multiple option structures in order to simplify redundant and meaningless structures.


instrumentation/google.golang.org/grpc/otelgrpc/config.go
```diff
@@ -203,45 +154,42 @@ func (m messageEventsProviderOption) apply(c *config) {
...
+		}
+	})
 }
 
 // WithMetricAttributes returns an Option to add custom attributes to the metrics.
 func WithMetricAttributes(a ...attribute.KeyValue) Option {
-	return metricAttributesOption{a: a}
+	return optionFunc(func(c *config) {
+		if a != nil {
+			c.MetricAttributes = a
```
flc1125: For the reviewer's reference, to assess whether an adjustment is needed to: `c.MetricAttributes = append(c.MetricAttributes, a...)`

flc1125:
Currently, CI is another issue, but I cannot retry it.

instrumentation/google.golang.org/grpc/otelgrpc/config.go
dashpole: +1 to changing this to append.

flc1125: Done

Issue #6026, opened 2024-08-20
otelgrpc: Allow adding custom attributes to a request's metrics based on its payload


pembletonj2-sonrai:
### Problem Statement

It's not possible to add custom attributes to the metrics generated by `otelgrpc` based on the data in the request payloads.

### Proposed Solution

A new option can be added to pass a function which will be run on each payload received to generate metric attributes for that request. Example:

```go
otelgrpc.NewServerHandler(otelgrpc.WithMetricAttributesFromPayload(func(payload any) []attribute.KeyValue {
    data1 := extractSomeDataFromPayload(payload)
    data2 := extractOtherDataFromPayload(payload)
    return []attribute.KeyValue{
        attribute.String("key1", data1),
        attribute.String("key2", data2),
    }
}))
```

`key1` and `key2` will then be included in the metrics' attributes for this request.

#### Alternatives

We found an alternative way to add the needed attributes to spans. The span's attributes can be modified from an interceptor:

```go
span := trace.SpanFromContext(ctx)
span.SetAttributes(attribute.String("key", value))
```

The payload can also be accessed from an interceptor, so its data can be used to set a span's attribute. However, it's not possible to do this with metrics.

#### Prior Art

#3894 discusses setting constant values for metric/span attributes and a PR has been made for it: #5133. This issue builds on this by allowing custom attributes to be computed on a per-request basis rather than being set to a constant value.

dmathieu:
Similar solution for the otelhttp instrumentation: https://github.com/open-telemetry/opentelemetry-go-contrib/pull/5876

dhowden:
Looking at the HTTP implementation (#5876), feels like we can make a very similar change, but also passing the incoming `ctx` too (so that callers can extract metadata etc).

The associated option then might look like:

```golang
func WithMetricAttributesFn(metricAttributesFn func(ctx context.Context,  payload any) []attribute.KeyValue) Option {
// ...
}
```

I've made a sketch implementation here (in draft): https://github.com/open-telemetry/opentelemetry-go-contrib/pull/6092, mostly because the `grpc.StatsHandler` pattern makes it a bit messier.

scorpionknifes:
Took a quick look, just wondering, why not use `stats.RPCStats` or `*stats.InPayload`  instead of just passing payload `any`?
```go
MetricAttributesFn func(ctx context.Context, rpcStats stats.RPCStats) []attribute.KeyValue
[...]

PR #8135, opened 2025-11-05
otelgrpc: add labeler implementation to allow adding custom attributes on metrics

vlad-coman-hs:
Draft PR to address #6026
Implements the Labeler [pattern](https://github.com/open-telemetry/opentelemetry-go-contrib/blob/main/instrumentation/net/http/otelhttp/labeler.go) from `otelhttp`.

Unfortunately, this pattern does not allow us to add dynamic custom attributes on `rpc_server_request_size` due to the interceptors executing after the request size metric is recorded. I also applied the Labeler's attributes in the `InPayload` and `OutPayload` phases to have a visual example on what that looks like. Note however that this only adds the custom attributes on the `rpc_server_response_size`, `rpc_client_response_size` and `rpc_client_request_size` metrics.


vlad-coman-hs:
Closing this due to the issue mentioned in the description. Will open another PR dealing with this.

Issue #8131, opened 2025-11-05
otelhttp: make DefaultClient a non-mutable variable


dmathieu:
DefaultClient is a mutable global variable - replace with something immutable or remove entirely

KonstantinosTsitoumis:
Hello, may I pick this up?

KonstantinosTsitoumis:
Here is the PR, Lmk what you think:)
https://github.com/open-telemetry/opentelemetry-go-contrib/pull/8140

pellared:
I propose deprecating and later removing

MrAlias:
> I propose deprecating and later removing

Does this mean we should also remove the `Get`, `Post`, `PostForm`, and `Head` functions as well?

KonstantinosTsitoumis:
Hello

- Could you please clarify a few things so I can plan the next steps properly?
- What should happen after the deprecation phase?
- How long should the deprecation period last before removing the old implementation?
- Is there a preferred way to communicate the deprecation?

This is my first ever contribution, so apologies if I’m a bit slow.

Thanks!

dmathieu:
Since this is an unstable package, we only need to deprecate for one version before we clean up and remove the deprecated code.
Communicating through the changelog should be enough.

KonstantinosTsitoumis:
The first step of the deprecation process is implemented in the PR.

KonstantinosTsitoumis:
For the next change of removing the default client, did we have a specific reason that we wanted to have one instance of the client?? Should we create a new instance of client each time we want to do a request in each Post/Get etc func?

dmathieu:
If we remove the `Post`, `Get` etc, we shouldn't even need to have a public default client anymore.

KonstantinosTsitoumis:
So this means that we need to deprecate Post - Get etc funcs too?

MrAlias:
Only remaining task is to remove deprecated decls after the next release.

dmathieu:
This was removed in https://github.com/open-telemetry/opentelemetry-go-contrib/pull/8266

PR #8140, opened 2025-11-06
otelhttp: Deprecate DefaultClient

KonstantinosTsitoumis:
makes defaultClient private and creates a function that initialises it only once.

created for issue: [#8131](https://github.com/open-telemetry/opentelemetry-go-contrib/issues/8131)


flc1125:
As mentioned in: https://github.com/open-telemetry/opentelemetry-go-contrib/issues/8131#issuecomment-3499481534, we need to first mark it as deprecated before proceeding with the replacement. We need to give users a transition period.

I suggest that you discuss and confirm the specific details of the deprecation and replacement plan on the issue page so that we can proceed with the follow-up work.

PS: This issue is expected to be split into multiple phases for implementation.

pellared (changes requested):
From https://github.com/open-telemetry/opentelemetry-go-contrib/issues/8131#issuecomment-3499481534:

> I propose deprecating and later removing

Also https://github.com/open-telemetry/opentelemetry-go-contrib/pull/8140#issuecomment-3499925635

instrumentation/net/http/otelhttp/client.go
```diff
@@ -11,6 +11,8 @@ import (
 	"strings"
 )
 
+// Deprecated: This will be deleted in the future.
```
pellared: This is in a bad place.
Should be as the last section of Go Doc comments.

pellared: Please document what the user should do instead.

pellared: Missing changelog entry.

KonstantinosTsitoumis: I'm not sure what user needs to do next, I assumed that since we are going to delete this they need to initialise a client on theirselves.

instrumentation/net/http/otelhttp/client.go
```diff
@@ -14,6 +14,10 @@ import (
 // DefaultClient is the default Client and is used by Get, Head, Post and PostForm.
 // Please be careful of initialization order - for example, if you change
 // the global propagator, the DefaultClient might still be using the old one.
+//
+// Deprecated: DefaultClient will be removed in a future release.
```
flc1125:
```suggestion
// Deprecated: [DefaultClient] will be removed in a future release.
```

flc1125 (approved):
Blocked: Release plan

CHANGELOG.md
```diff
@@ -29,6 +29,8 @@ This project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0.htm
 - `WithRouteTag` in `go.opentelemetry.io/contrib/instrumentation/net/http/otelhttp` is deprecated.
   The route is already added automatically for spans.
   For metrics, the alternative is to use the `WithMetricAttributesFn` option. (#8117)
+- `DefaultClient` in `go.opentelemetry.io/contrib/instrumentation/net/http/otelhttp/client.go` is deprecated.
```
```suggestion
- `DefaultClient` in `go.opentelemetry.io/contrib/instrumentation/net/http/otelhttp` is deprecated.  
```

MrAlias:
There is not now way to set the client used by `Get`, `Post`, `PostForm`, and `Head` without using this deprecated value. An alternative should have been explored or those methods also handled in this process.

PR #8142, opened 2025-11-07
refactor(otelhttp): change Version() func to const Version string

### LINKED ISSUE #8134: otelhttp: `func Version() string` should just be `const Version string`

dogedede:
Fixes #8134
Changes the Version() definition in the otelhttp package from a function to a constant string for simplicity and consistency.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -36,7 +36,6 @@ This project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0.htm
 
 ### Changed
 
-- Change `Version()` function in `go.opentelemetry.io/contrib/instrumentation/net/http/otelhttp` to a `const Version` string. (#8142)
 - Improve performance by reducing allocations in the gRPC stats handler in `go.opentelemetry.io/contrib/instrumentation/google.golang.org/grpc/otelgrpc`. (#8035)
 - Export the `ReadEvents` and `WriteEvents` constants in `go.opentelemetry.io/contrib/instrumentation/net/http/otelhttp` so they can be used in `WithMessageEvents`. (#8153)
 - Switched the default for `OTEL_SEMCONV_STABILITY_OPT_IN` to emit the v1.37.0 semantic conventions by default in `go.opentelemetry.io/contrib/instrumentation/go.mongodb.org/mongo-driver/mongo/otelmongo`.


dogedede:
Apologies, I've updated the changelog.md.

CHANGELOG.md
```diff
@@ -1622,3 +1622,7 @@ First official tagged release of `contrib` repository.
 [Go 1.18]: https://go.dev/doc/go1.18
 
 [GO-2024-2687]: https://pkg.go.dev/vuln/GO-2024-2687
+
+### Changed
+
+- Change `Version()` function in `otelhttp` package to a `const Version` string. (#8134)
```
dmathieu: Changelog entries need to be at the beginning of the file. Not its end.

dmathieu: Buddy, look at the content of the file. You can surely see your last commit is also wrong.

dogedede: Sorry for the laughably silly mix-up , I’ve moved the content to the right place.

CHANGELOG.md
```diff
@@ -22,6 +22,7 @@ This project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0.htm
 
 ### Changed
 
+- Change `Version()` function in `otelhttp` package to a `const Version` string. (#8134)
```
flc1125: This is a non-backward-compatible change, and I tend to include such descriptions in the `CHANGELOG`.

instrumentation/net/http/otelhttp/version.go
```diff
@@ -4,7 +4,4 @@
 package otelhttp // import "go.opentelemetry.io/contrib/instrumentation/net/http/otelhttp"
 
 // Version is the current release version of the otelhttp instrumentation.
-func Version() string {
-	return "0.63.0"
-	// This string is updated by the pre_release.sh script during release
-}
+const Version = "0.63.0"
```
flc1125: I'm not quite sure, just for reference:

Can this value be updated when releasing? https://github.com/open-telemetry/opentelemetry-go-contrib/blob/8ef069f6370b8a04a642bac4ea8da5ed073ab5e8/versions.yaml#L30

cc @dmathieu

CHANGELOG.md
```suggestion
- Change `Version()` function in `otelhttp` package to a `const Version` string. (#8142)
```

instrumentation/net/http/otelhttp/version.go
dmathieu: The prerelease command finds any string that matches a semver version, and replaces it within the file. So yes, this should still work.

https://github.com/open-telemetry/opentelemetry-go-build-tools/blob/main/multimod/internal/prerelease/prerelease.go#L187-L206

@dogedede could you check? 
You should just have to update the version in `versions.yaml`, run make prerelease and check the file is properly updated.
See https://github.com/open-telemetry/opentelemetry-go-contrib/blob/main/RELEASING.md#update-module-set-versions

CHANGELOG.md
dmathieu: I agree, this should be more visible in the changelog.

instrumentation/net/http/otelhttp/version.go
flc1125: I conducted a test, and it can generate normally. Example: https://github.com/open-telemetry/opentelemetry-go-contrib/pull/8157/files#diff-a5698a2d4bebe20729ad27f4a8d806538ca8fff8444a0ebb7043193a6a7dfc01

dogedede: Thank you for your help! I spent a long time trying to run this myself due to my local configuration issues, and your example clarified a lot for me. I really appreciate it.

CHANGELOG.md
```diff
@@ -22,6 +22,7 @@ This project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0.htm
 
 ### Changed
 
+- Change `Version()` function in `otelhttp` package to a `const Version` string. (#8142)
```
```suggestion
- Change `Version()` function in `go.opentelemetry.io/contrib/instrumentation/net/http/otelhttp` to a `const Version` string. (#8142)
```

dogedede: Thank you for your guidance！

CHANGELOG.md
```diff
@@ -22,6 +22,7 @@ This project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0.htm
 
 ### Changed
 
+- Change `Version()` function in `go.opentelemetry.io/contrib/instrumentation/net/http/otelhttp` to a `const Version` string. (#8142)
```
dmathieu: @open-telemetry/go-approvers I think we could make it more visible this is a breaking change. But at the same time, doing something too prominent for the version string probably too much. WDYT?

flc1125: I prefer to only mark content or tags related to significant changes here.

flc1125: @dogedede Blocked here

dogedede: Hi @flc1125! Could you clarify what I should do here — should I add a (breaking change) label at the end of this changelog entry?

flc1125: Currently, there is no conclusion yet. However, from my personal perspective, I lean towards the need to add prominent information to alert users about the changes here.

Until a conclusion is reached, we can maintain the current situation.

pellared: I agree that "doing something too prominent for the version string probably too much". What we have now is good enough for me.

Side note: @dmathieu, do we even need to have this exported?

dmathieu: You mean do we even need to have the `Version` const provided? Every other instrumentation has it (as a function actually, not a string).

pellared: > You mean do we even need to have the Version const provided? 

Yes

> Every other instrumentation has it (as a function actually, not a string).

It does not mean that it is actually useful.
If it is then probably we should also export `ScopeName`.

EDIT:
I see that it is exported, ignore my proposal!
I think that the current proposal is good.

dmathieu: Actually, we should probably do the same with every other instrumentation (change from a function to a string).
Not in this PR.

CHANGELOG.md
dmathieu, line 10:
```suggestion

### Added

- Change `Version()` function in `go.opentelemetry.io/contrib/instrumentation/net/http/otelhttp` to a `const Version` string. (#8142)
```

CHANGELOG.md
```diff
@@ -33,6 +33,7 @@ This project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0.htm
...
```
```suggestion
```

dmathieu (approved):

PR #6747, opened 2025-02-10
feat(otelgin): add metrics recording for HTTP server

flc1125:
closed https://github.com/open-telemetry/opentelemetry-go-contrib/issues/5096
closed https://github.com/open-telemetry/opentelemetry-go-contrib/issues/2618

PR #7668, opened 2025-08-05
feat(otelecho):support http server metrics for echo instrumentation

Whitea029:
#### Description:

  - Implement `otelecho.WithServerMetrics()` option to enable HTTP server metrics
  - Add comprehensive test coverage for metrics functionality

#### Related Issue:

fix #7656

PR #8143, opened 2025-11-07
chore: update metrics support in instrumentation readme

jvllmr:
Updated the instrumentation package table to indicate that the Gin Gonic framework supports both metrics and traces. I initially thought that it does not support metrics until I saw that it was already implemented in #6747. 😅

### DIFF
--- a/instrumentation/runtime/doc.go
+++ b/instrumentation/runtime/doc.go
@@ -2,33 +2,4 @@
 // SPDX-License-Identifier: Apache-2.0
 
 // Package runtime implements the conventional runtime metrics specified by OpenTelemetry.
-//
-// The metric events produced are:
-//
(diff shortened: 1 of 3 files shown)


dmathieu (approved):

flc1125:
In fact, not only `otelgin`, but several other components also provide support for metrics.

Would you be willing to update the latest status of the other components together?

jvllmr:
> In fact, not only `otelgin`, but several other components also provide support for metrics.
> 
> Would you be willing to update the latest status of the other components together?

Thanks for the hint @flc1125. I have taken a look at the other instrumentation modules and update the PR. 👍🏻

jvllmr:
Related PRs are #6648 and #7668.

Issue #8133, opened 2025-11-05
otelhttp: Deprecate `WithPublicEndpoint` in favor of `WithPublicEndpointFn`


dmathieu:

n2p5:
Hi! I'm looking to make my first contribution to this repo and this issue caught my eye. I noticed that `WithPublicEndpoint` and `WithPublicEndpointFn` exist in:

```
otelhttp
otelgrpc
otelmux
otelrestful
```

Based on the name of this issue we're only interested in Deprecating `otelhttp`, but would it make sense to address this across these other areas as well?

flc1125:
> Based on the name of this issue we're only interested in Deprecating `otelhttp`, but would it make sense to address this across these other areas as well?

I think we can first resolve the issue with `otelhttp`, which can serve as a stability reference standard for other components in the future.

flc1125:
@n2p5 Assigned to you

n2p5:
We can close this issue. https://github.com/open-telemetry/opentelemetry-go-contrib/pull/8152 Has been merged!

dmathieu:
Let's keep it open until the next release, when we can remove the method entirely.

PR #8152, opened 2025-11-09
otelhttp: deprecate WithPublicEndpoint in favor of WithPublicEndpointFn (#8133)

n2p5:
> Note: This is my first contribution to `open-telemetry/oepntelemetry-go-contrib`. I found this issue from the "good first issue" label, so if I've messed anything up or I skipped a step, let me know!

This addresses Issue: https://github.com/open-telemetry/opentelemetry-go-contrib/issues/8133

`WithPublicEndpoint` now delegates to `WithPublicEndpointFn` to reduce code duplication and maintenance burden. The PublicEndpoint bool field has been removed from the config and middleware structs in favor of using only PublicEndpointFn.

This simplifies the implementation to a single code path while maintaining full backward compatibility.


n2p5:
Thank you, @dmathieu and @flc1125. Is there anything else you need me to do on my side for this? I merged the latest changes from `main` to ensure we don't have any code drift. However, other than that, this needs to be authorized to rerun the rest of the pipeline.

Issue #8141, opened 2025-11-07
[Feature]: Document go.schedule.duration metric emitted by NewProducer in package documentation


pellared:
### Component

Instrumentation: runtime

### Problem Statement

https://pkg.go.dev/go.opentelemetry.io/contrib/instrumentation/runtime does not document `go.schedule.duration` emitted via `NewProducer`

### Proposed Solution

Document in https://pkg.go.dev/go.opentelemetry.io/contrib/instrumentation/runtime metrics emitted via both `Start` and `NewProducer`. Maybe it would be better to document those on function docs level instead of package level so that the reader is aware what metrics are produced by which functionality.

tushar1977:
Hi, I’d love to work on this issue

pellared:
@tushar1977 , I assigned you

pellared:
Fixed by https://github.com/open-telemetry/opentelemetry-go-contrib/pull/8154

PR #8154, opened 2025-11-09
Added docs for go.schedule.duration

tushar1977:
Issue #8141 
Added documentation for ```go.schedule.duration```

### DIFF
--- a/instrumentation/runtime/producer.go
+++ b/instrumentation/runtime/producer.go
@@ -35,7 +35,7 @@ var _ metric.Producer = (*Producer)(nil)
 
 // NewProducer creates a Producer which provides precomputed histogram metrics from the go runtime.
 //
-// Metrics emitted by [NewProducer] include:
+// Metrics emitted by NewProducer include:
 //
 //	go.schedule.duration    s             The time goroutines have spent in the scheduler in a runnable state before actually running.
 func NewProducer(opts ...ProducerOption) *Producer {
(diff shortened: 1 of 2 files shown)


pellared:
What about

> Maybe it would be better to document those on function docs level instead of package level so that the reader is aware what metrics are produced by which functionality.

Or making separate sections on package level GoDoc?

tushar1977:
yea making seperate section on package level is good idea
I'll implement it

tushar1977:
@pellared please review and let me know if any changes are needed.

instrumentation/runtime/producer.go
```diff
@@ -34,6 +34,10 @@ type Producer struct {
 var _ metric.Producer = (*Producer)(nil)
 
 // NewProducer creates a Producer which provides precomputed histogram metrics from the go runtime.
+//
+// Metrics emitted by [NewProducer] include:
```
pellared: There is no need to self-reference.

```suggestion
// Metrics emitted by NewProducer include:
```

instrumentation/runtime/runtime.go
```diff
@@ -37,6 +37,34 @@ const (
 
 // Start initializes reporting of runtime metrics using the supplied config.
 // For goroutine scheduling metrics, additionally see [NewProducer].
+// Metrics emitted by [Start] includes:
```
pellared: For consistency

```suggestion
//
// Metrics emitted by Start includes:
```

pellared (approved):
Thank you!

PR #8162, opened 2025-11-11
Upgrade macos tests to latest

dmathieu:
GitHub is deprecating the macos-13 runners.
In fact, we should probably have upgraded much earlier.
https://github.com/actions/runner-images/issues/13046

See also https://github.com/open-telemetry/opentelemetry-go/pull/7597

### DIFF
--- a/otelconf/config_json.go
+++ b/otelconf/config_json.go
@@ -28,7 +28,7 @@ func (j *ConsoleExporter) UnmarshalJSON(b []byte) error {
 
 // UnmarshalJSON implements json.Unmarshaler.
 func (j *B3Propagator) UnmarshalJSON(b []byte) error {
-	type plain ConsoleExporter
+	type plain B3Propagator
 	var p plain
 	if err := json.Unmarshal(b, &p); err != nil {
 		return errors.Join(newErrUnmarshal(j), err)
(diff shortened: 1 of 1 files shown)


.github/workflows/ci.yml
```diff
@@ -108,7 +108,7 @@ jobs:
             arch: "386"
           - os: ubuntu-latest
             arch: amd64
-          - os: macos-13
+          - os: macos-15
```
codeboten: thoughts on using macos-latest here instead as is done for arm64?

dmathieu: That would make sense to me. I'd like to hear from other maintainers though.
cc @open-telemetry/go-maintainers

XSAM: I am fine with using `-latest`. But I am not sure we can use `macos-latest` for `amd64`, as GitHub does not mention this matrix on the doc.

https://docs.github.com/en/actions/reference/runners/github-hosted-runners

image

dmathieu: Tests are passing with macos-latest.

README.md
```diff
@@ -56,8 +56,8 @@ This project is tested on the following systems.
 | Ubuntu   | 1.24       | amd64        |
 | Ubuntu   | 1.25       | 386          |
 | Ubuntu   | 1.24       | 386          |
-| macOS 13 | 1.25       | amd64        |
-| macOS 13 | 1.24       | amd64        |
+| macOS 15 | 1.25       | amd64        |
+| macOS 15 | 1.24       | amd64        |
```
pellared: I am not sure we should list "15" given we use "latest" in the CI.

dmathieu: Good catch. I forgot to remove it.

Issue #8203, opened 2025-11-20
otelconf: provide strategy for supporting experimental features in schema


codeboten:
Capturing this in a new issue to not lose sight of wanting to provide a strategy for how experimental features in the configuration schema are supported in otelconf moving forward as discussed in the 20-Nov SIG call.

> Given this is an RC of v1 I want to ask what is the design/plan regarding backwards compatiblity and stability.
> 
> For instance `OpenTelemetryConfiguration` has `InstrumentationDevelopment` field. Should it be exported? This does not seem to be a stable field.
> 
> Do we want we eventually want to have `Instrumentation` instead of `InstrumentationDevelopment`?

That's a good question! The plan for the schema v1 is to include experimental support for:

* resource detection
* instrumentation
* otlp file exporter
* configurators for each signal
* prometheus exporter
* various samplers configuration (composite, probability)

The stability guarantee around these are near none (https://github.com/open-telemetry/opentelemetry-configuration/tree/main?tab=readme-ov-file#experimental-features). We could decide to not support these in the implementation at this time. 

Alternatively we could put these in an experimental module and only unmarshal anything marked with the `/development` suffix if an option is set (something like `WithExperimentalFeatures`)

Thoughts?

_Originally posted by @codeboten in https://github.com/open-telemetry/opentelemetry-go-contrib/issues/8026#issuecomment-3553515072_

codeboten:
I thought a bit more about what the strategy should look like here, providing individual options for things that would need to be set that are experimental would give us the flexibility to decide how much of the experimental functionality we want to provide an API for, but it would end up causing us to have to provide an API of some sort anyways so i'm not sure how much we would gain here.

codeboten:
I opened a PR that shows how we could include a `otelconf/x` package that supports experimental features. The change worth looking at is probably this one: https://github.com/open-telemetry/opentelemetry-go-contrib/pull/8205/commits/dfc76085c6fa33171e632c1cf5c8cfacc04375b6 as the others are really just everything thats in the 1.0.0 rc pull request.

This approach ends up creating a lot of duplicated code, but it is much simpler as most of the duplicated code is generated from the schema itself and updated via sed, so not a lot of manual work there. Not sure how much of the marshaling/unmarshaling code we could reuse across packages. I'd like to get more input from the SIG before i go down that route.

Pinging @pellared and @MrAlias for opinions as we talked about this in various conversations

MrAlias:
[...]

PR #8205, opened 2025-11-20
otelconf: experimental x package

codeboten:
This is an attempt at producing a separate `otelconf/x` package to support experimental features in the configuration schema. This is doing the bare minimum of using a `sed` patch to hide all experimental structs from the generated model code before moving it to the `otelconf` package. This has the upside of being pretty straight forward to do, but ends up with a lot of duplicate code.

Part of #8203


codeboten:
Closing this in favour of https://github.com/open-telemetry/opentelemetry-go-contrib/pull/8529

PR #8228, opened 2025-11-27
otelconf: update test files

codeboten:
Split off from #8106, updating the tls files to real test files.


codeboten:
no need for changelog

PR #8230, opened 2025-11-28
otelmongo: default to stable semconv

dmathieu:
otelmongo has had support for both 1.21.0 and the stabilized versions for quite a while now (4 releases, over a year: https://github.com/open-telemetry/opentelemetry-go-contrib/commit/04b152ded69a7af33da9877f6bc19ff903e4d73f).

So this switches the default behavior of the module to use the stable semconv.
Older semconvs can still be used with the `OTEL_SEMCONV_STABILITY_OPT_IN=database/dup` environment variable.

This matches what we did for the HTTP libraries.

PR #8231, opened 2025-11-28
Add linter checking we use the latest semconv version

dmathieu:
This is inspired by https://github.com/open-telemetry/opentelemetry-collector-contrib/pull/44606

Mongo is the only package with an older semconv. See #8230 to move forward with the upgrade.


.golangci.yml
```diff
@@ -54,6 +54,14 @@ linters:
             - pkg: crypto/md5
             - pkg: crypto/sha1
             - pkg: crypto/**/pkix
+        semconv:
+          list-mode: lax
+          deny:
+            - pkg: go.opentelemetry.io/otel/semconv
+              desc: Use go.opentelemetry.io/otel/semconv/v1.37.0 instead. If a newer semconv version has been released, update the depguard rule.
```
thompson-tomo: Same comment as I made on collector.

```suggestion
              desc: Use go.opentelemetry.io/otel/semconv/v1.38.0 instead. When a newer semconv version is released, it must be added to the depguard rule.
```
Or

```suggestion
              desc: Use go.opentelemetry.io/otel/semconv/v1.38.0 instead. If a newer semconv version has been released, update the depguard rule.
```

.golangci.yml
```diff
@@ -54,6 +54,14 @@ linters:
...
+          allow:
+            - go.opentelemetry.io/otel/semconv/v1.21.0 # still used by otelmongo
+            - go.opentelemetry.io/otel/semconv/v1.37.0
```
thompson-tomo, line 64:
```suggestion
            - go.opentelemetry.io/otel/semconv/v1.37.0
            - go.opentelemetry.io/otel/semconv/v1.38.0
```

dmathieu: Upgrading to 1.38 makes sense. But it shouldn't be within this PR.

codeboten (approved):

.golangci.yml
thompson-tomo: Agree and this just adds 1.38 as an allowed version.

dmathieu:
> Side note: I think it would be nice to have the process of bumping semconv (starting from generating new semconv) to be documented in CONTRIBUTING.md.

Yes. This seems like it should happen in core though.

dmathieu:
@open-telemetry/go-approvers this would welcome a second approval.

pellared:
> > Side note: I think it would be nice to have the process of bumping semconv (starting from generating new semconv) to be documented in CONTRIBUTING.md.
> 
> Yes. This seems like it should happen in core though.

I created https://github.com/open-telemetry/opentelemetry-go/issues/7697

dmathieu:
@open-telemetry/go-approvers ping for second approval :)

PR #8268, opened 2025-12-08
Remove the deprecated `WithRouteTag` option in `otelhttp`

MrAlias:


instrumentation/net/http/otelhttp/handler.go
```diff
@@ -219,21 +219,3 @@ func (h *middleware) metricAttributesFromRequest(r *http.Request) []attribute.Ke
 	}
 	return attributeForRequest
 }
-
-// WithRouteTag annotates spans and metrics with the provided route name
-// with HTTP route attribute.
-//
-// Deprecated: spans are automatically annotated with the route attribute.
-// To annotate metrics, use the [WithMetricAttributesFn] option.
```
dashpole, line 227: Non-blocking: If we are able to annotate spans automatically, why don't we annotate metrics automatically as well?

From https://opentelemetry.io/docs/specs/semconv/http/http-metrics/#metric-httpserverrequestduration for http.route:

> Conditionally Required If and only if it’s available

dashpole (approved):

instrumentation/net/http/otelhttp/handler.go
dmathieu: This was a bit different. It didn't allow annotating anything on the spans, it only added the `http.route` attribute.
This is from before the `Pattern` attribute was added to `http.Request`, so we had to manually set the route name.

We do also add the route tag onto metrics.
https://github.com/open-telemetry/opentelemetry-go-contrib/blob/main/instrumentation/net/http/otelhttp/internal/semconv/server.go#L399-L401

So technically, this deprecated comment appear to have been wrong.

pellared (approved):

instrumentation/net/http/otelhttp/handler.go
dashpole: Perfect, thanks.

MrAlias: No idea. I expect it should be possible.

MrAlias: If you would like to open an issue to track this I would support that.

I'm going to resolve the comment as it doesn't seem related to the removal being done in the PR.

PR #8299, opened 2025-12-16
otelhttptrace: fix false positive errors in otelhttptrace for http.receive span

### LINKED ISSUE #8216: [Bug]: false positive errors in otelhttptrace for http.receive span
### Component

Instrumentation: otelhttptrace

We use httptrace for our http client. Very frequently we observe traces being marked as "error" due to "http: putIdleConn: too many idle connections for host" in the http.receive span.

This error is a part of an internal mechanism in net/http to pool http connections. 
Transport keeps a per-host pool of idle keep-alive sockets.
In [tryPutIdleConn](https://github.com/golang/go/blob/master/src/net/http/transport.go#L1062), after a request finishes, the connection is returned to that pool. If the pool already holds MaxIdleConnsPerHost entries (default 2 if you don’t override it), the function returns errTooManyIdleHost, and the connection is immediately closed instead of being cached. 

### Expected behavior

This is not a client issue and not even an error(just a part of the pooling flow) and shouldn't mark traces as erroring

### Steps to Reproduce

1. Start a basic Go program with an http.Client using otelhhtp
```go
	client := http.DefaultClient
	client.Transport = otelhttp.NewTransport(
			http.DefaultTransport,
			otelhttp.WithClientTrace(func(ctx context.Context) *httptrace.ClientTrace {
				return otelhttptrace.NewClientTrace(ctx)
			}))
```
2. In a loop, make requests to a backend on another socket. Ensure the rate of in-flight requests is >2 (default per-host concurrency of http.Transport)
3. Observe the exported traces. Many requests should be flagged as "error"

### Operating System

Debian Linux

### Device Architecture

x86_64

### Go Version

1.23.0

### Component Version

v0.60.0

mitioshi:
Fixes https://github.com/open-telemetry/opentelemetry-go-contrib/issues/8216

## Problem

We use httptrace for our http client. Very frequently we observe traces being marked as "error" due to "http: putIdleConn: too many idle connections for host" in the http.receive span.

This error is a part of an internal mechanism in net/http to pool http connections.
Transport keeps a per-host pool of idle keep-alive sockets.
In [tryPutIdleConn](https://github.com/golang/go/blob/master/src/net/http/transport.go#L1062), after a request finishes, the connection is returned to that pool. If the pool already holds MaxIdleConnsPerHost entries (default 2 if you don’t override it), the function returns errTooManyIdleHost, and the connection is immediately closed instead of being cached.

This is not a client issue and not even an error(just a part of the pooling flow) and shouldn't mark traces as erroring

## Implementation
In this PR, we explicitly ignore the errors that can be passed by http.Transport to the PutIdleConn callback(which is used by ClientTrace). Since the errors are unexported, we have to match the error strings.

I've also added tests for two of three possible erorrs. The third one, errCloseIdle cannot be reliably reproduced in tests without introduing a flaky test, so I decided to omit the test for it.


mitioshi:
@dmathieu Since this PR has a pending CLA label, here's a FYI that I signed it. This PR can now be reviewed

instrumentation/net/http/httptrace/otelhttptrace/clienttrace.go
```diff
@@ -286,6 +286,32 @@ func (ct *clientTracer) gotConn(info httptrace.GotConnInfo) {
 }
 
 func (ct *clientTracer) putIdleConn(err error) {
+	// this error is passed to the PutIdleConn callback when http.Transport's HTTP 1.1 connection pool is full
+	// It's not an error per se, but rather a control flow error used for connection pooling logic in net/http internals
+	// corresponds to errTooManyIdleHost in net/http/internal/transport.go
+	const poolReturnErr = "too many idle connections"
```
dmathieu: Could you move those constants at the package level?

instrumentation/net/http/httptrace/otelhttptrace/clienttrace.go
```diff
@@ -286,6 +286,32 @@ func (ct *clientTracer) gotConn(info httptrace.GotConnInfo) {
...
+
+	if err == nil {
+		ct.end("http.receive", nil)
+		return
+	}
+
+	// non-nil err in this callback is one of the following net/http errors: errTooManyIdleHost, errBrokenConn, errCloseIdle
+	// All of those are internal. No reason to mark the span as error if we encounter them
+	// net/http doesn't export said errors, so we can only match against their strings
+	if strings.Contains(err.Error(), poolReturnErr) ||
```
dmathieu: Doing `ct.end` twice is a bit misleading.
How about checking for `err == nil` in this condition as well?

dmathieu:

instrumentation/net/http/httptrace/otelhttptrace/clienttrace.go
mitioshi: done

instrumentation/net/http/httptrace/otelhttptrace/clienttrace.go
mitioshi: done

mitioshi:

Added one in the recent commit

mitioshi:
@dmathieu FYI: I've implemented the changes proposed by code review

instrumentation/net/http/httptrace/otelhttptrace/clienttrace.go
```diff
@@ -22,6 +22,20 @@ import (
...
 
+// this error is passed to the PutIdleConn callback when http.Transport's HTTP 1.1 connection pool is full
+// It's not an error per se, but rather a control flow error used for connection pooling logic in net/http internals
+// corresponds to errTooManyIdleHost in net/http/internal/transport.go .
+const poolReturnErr = "too many idle connections"
+
+// The connection is already closed, typically due to an error, timeout, or cancellation.
+// This error is not visible to the user of http.Client/Transport.
+// corresponds to errConnBroken in net/http/internal/transport.go .
+const brokenConnErr = "connection is in bad state"
```
dmathieu, line 33: This error appears not to be tested?

mitioshi: > The third one, errCloseIdle cannot be reliably reproduced in tests without introduing a flaky test, so I decided to omit the test for it.

Yes, this is intentional. I spent an evening trying to write a test case to trigger this issue, but the best I could do was to reproduce this error "sometimes" when context is cancelled. The window is in milliseconds and makes it frustratingly hard for us to test this error. As we don't have access to internal code in net/http, I thought not covering this case is better than having a flaky test

dmathieu: While I agree, it also means that if for some reason, net/http changes the error content (they can, since the error is private), we won't know the behavior isn't working anymore.

mitioshi: that makes sense. However, this gets tricky. errConnBroken can only occur in a very specific scenario:
  1. Body must be fully read (bodyEOF = true in net/http)
  2. Underlying connection must NOT have seen EOF (sawEOF = false)
  3. Connection must be marked as broken (pc.closed != nil)

These conditions require the connection to be marked broken AFTER the body read completes but BEFORE isBroken() is checked . The window is so small, we can only hope we hit the race condition

I see three options here:
1. Remove the handling of errConnBroken from clienttrace.go. The error is very specific and won't happen often. This is still magnitudes better than not ignoring the other 2 errors
2. Accept errConnBroken not being tested.
3. Manually call PutIdleConn callback with errors.New containing the same error as errConnBroken in net/http. That would still have the problem you described
4. Run 100 iterations of racing body close and context cancel. That would make it more deterministic while still being a flaky test

I feel like the only good options are option 1 and 2. What do you think?

dmathieu: I agree that only 1 and 2 are good options. I'd like to hear other folks' opinions.

mitioshi: Who's opinion exactly are we expecting? Should we mention any other maintainers?

dmathieu: Ping @open-telemetry/go-approvers

dashpole: I think it is OK not to have test coverage of it.

dmathieu (approved):

CHANGELOG.md
```diff
@@ -18,6 +18,8 @@ This project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0.htm
 
 - Fix panic when passing nil `TracerProvider` or `MeterProvider` to `WithTracerProvider` or `WithMeterProvider` in `go.opentelemetry.io/contrib/instrumentation/google.golang.org/grpc/otelgrpc`. (#8323)
 - `Transport` in `go.opentelemetry.io/contrib/instrumentation/net/http/otelhttp` now supports reading request body multiple times for subsequent requests that reuse `http.Request`. (#8352)
+- Fix HTTP/1.1 connection pooling internal errors incorrectly marking spans as errors in `go.opentelemetry.io/contrib/instrumentation/net/http/httptrace/otelhttptrace`.
+  Internal connection pool errors (`errTooManyIdleHost`, `errConnBroken`, `errCloseIdle`) are now properly ignored and do not mark spans with error status.
```
```suggestion
- Fix HTTP/1.1 connection pooling internal errors (`errTooManyIdleHost`, `errConnBroken`, `errCloseIdle`) incorrectly marking spans as errors in `go.opentelemetry.io/contrib/instrumentation/net/http/httptrace/otelhttptrace`.  (#8299)
```

pellared:

dmathieu:
Yes, this is still valid.

instrumentation/net/http/httptrace/otelhttptrace/clienttracetest_test.go
```diff
@@ -571,3 +576,163 @@ func TestHTTPRequestWithExpect100Continue(t *testing.T) {
...
+func TestHTTPRequestWithClientTrace_PerHostIdlePoolExceeded(t *testing.T) {
+	sr := tracetest.NewSpanRecorder()
+	tp := trace.NewTracerProvider(trace.WithSpanProcessor(sr))
+	otel.SetTracerProvider(tp)
+	tr := tp.Tracer("httptrace/client")
+
+	shouldReturn := make(chan struct{})
+	reqsMade := atomic.Int32{}
+
+	// Mock http server
```
pellared, line 589:
```suggestion
```

instrumentation/net/http/httptrace/otelhttptrace/clienttracetest_test.go
```diff
@@ -571,3 +576,163 @@ func TestHTTPRequestWithExpect100Continue(t *testing.T) {
...
+	ts := httptest.NewServer(
+		http.HandlerFunc(func(http.ResponseWriter, *http.Request) {
+			reqsNew := reqsMade.Add(1)
+
+			if reqsNew > 1 {
+				close(shouldReturn)
+			}
+
+			<-shouldReturn
```
pellared, line 598: Can you please describe the idea behind this?

instrumentation/net/http/httptrace/otelhttptrace/clienttracetest_test.go
```diff
@@ -571,3 +576,163 @@ func TestHTTPRequestWithExpect100Continue(t *testing.T) {
...
+			MaxIdleConnsPerHost: 1,
+		},
+	}
+
+	wg := sync.WaitGroup{}
+
+	wg.Add(2)
+
+	mx := sync.Mutex{}
+	errors := make([]error, 0, 2)
```
pellared, line 616: why not using `make(chan error, 2)` for synchronization?

CHANGELOG.md
```diff
@@ -77,6 +77,7 @@ The next release will require at least [Go 1.25].
 
 - Fix panic when passing nil `TracerProvider` or `MeterProvider` to `WithTracerProvider` or `WithMeterProvider` in `go.opentelemetry.io/contrib/instrumentation/google.golang.org/grpc/otelgrpc`. (#8323)
 - `Transport` in `go.opentelemetry.io/contrib/instrumentation/net/http/otelhttp` now supports reading request body multiple times for subsequent requests that reuse `http.Request`. (#8352)
+- Fix HTTP/1.1 connection pooling internal errors (`errTooManyIdleHost`, `errConnBroken`, `errCloseIdle`) incorrectly marking spans as errors in `go.opentelemetry.io/contrib/instrumentation/net/http/httptrace/otelhttptrace`.  (#8299)
```
```suggestion
- Fix HTTP/1.1 connection pooling internal errors (`errTooManyIdleHost`, `errConnBroken`, `errCloseIdle`) incorrectly marking spans as errors in `go.opentelemetry.io/contrib/instrumentation/net/http/httptrace/otelhttptrace`. (#8299)
```

pellared:
@mitioshi, can you update this PR?

PR #8301, opened 2025-12-17
Clarifies what the dice example demonstrates

n9te9:
## Background

Dice example README didn't clearly state what users can learn from this example.
As a result, first-time users may be unsure about the learning focus of the dice example.

## What

This change adds a short description to clarify what this example demonstrates.

### DIFF
--- a/examples/dice/README.md
+++ b/examples/dice/README.md
@@ -4,6 +4,10 @@ This is the foundation example for [Getting Started](https://opentelemetry.io/do
 
 Below, you will see instructions on how to run this application, either with or without instrumentation.
 
+## What can you learn from this example?
+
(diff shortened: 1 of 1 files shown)


dmathieu:
Your statement isn't accurate. The otelhttp handler will do tracing too.

n9te9:
@dmathieu 
You're absolutely right - the otelhttp handler does create spans and performs tracing.
What I wanted to clarify in the README is not that tracing is absent, the example's **primary learning focus**, from a first-time user's perspective, is metrics instrumentation (for example, the custom dice.rolls metrics and exemplars).
The change is intended to make the learning intent of the example clearer, rather than to imply tracing is not involved.

PR #8323, opened 2025-12-20
fix(otelgrpc): prevent panic when nil TracerProvider or MeterProvider is passed

dims:
Commit 07504e453 removed nil checks from WithTracerProvider and WithMeterProvider, causing a panic when callers pass nil. This restores the checks so nil values are ignored and the global provider is used.

This showed up in some future-proof-ci jobs in kubernetes:
https://storage.googleapis.com/kubernetes-ci-logs/logs/ci-kubernetes-unit-dependencies/2002001097211777024/build-log.txt


dims:
cc @flc1125 and @dmathieu

dims:
xref: https://github.com/kubernetes/kubernetes/issues/135865

flc1125:
Perhaps we need to add a CHANGELOG.

PR #8339, opened 2025-12-28
feat: add MetricAttributesFn and SpanAttributesFn for dynamic attributes

mojixcoder:
Add support for dynamic span and metric attributes in `otelgrpc` via `WithSpanAttributesFn` and `WithMetricAttributesFn`.
These options allow extracting attributes from the context per request, enabling propagation of values injected by interceptors or middleware.


flc1125:
Thank you for your contribution. This change is somewhat similar to https://github.com/open-telemetry/opentelemetry-go-contrib/pull/8191.

mojixcoder:
Yeah didn’t see that. 
could add SpanAttributeFn but after this PR is merged

dmathieu:
SpanAttributesFn shouldn't be needed. You can retrieve the current span from the context and add attributes to it directly.

mojixcoder:
Ok thanks, So I'm closing the PR.

Issue #8272, opened 2025-12-09
Switch versions from functions to strings


dmathieu:
We should switch versions from functions to strings in every module, as we did in otelhttp: https://github.com/open-telemetry/opentelemetry-go-contrib/pull/8142

Please don't do one PR switching everything. This would better be as one PR per module.

* [ ] go.opentelemetry.io/contrib
* [ ] go.opentelemetry.io/contrib/tools
* [ ] go.opentelemetry.io/contrib/propagators/aws
* [ ] go.opentelemetry.io/contrib/propagators/ot
* [ ] go.opentelemetry.io/contrib/propagators/jaeger
* [ ] go.opentelemetry.io/contrib/propagators/b3
* [ ] go.opentelemetry.io/contrib/detectors/gcp
* [ ] go.opentelemetry.io/contrib/detectors/aws/ecs
* [ ] go.opentelemetry.io/contrib/detectors/aws/eks
* [ ] go.opentelemetry.io/contrib/detectors/aws/ec2/v2
* [ ] go.opentelemetry.io/contrib/bridges/prometheus
* [ ] go.opentelemetry.io/contrib/detectors/aws/lambda
* [ ] go.opentelemetry.io/contrib/exporters/autoexport
* [ ] go.opentelemetry.io/contrib/propagators/autoprop
* [ ] go.opentelemetry.io/contrib/propagators/opencensus
* [x] go.opentelemetry.io/contrib/instrumentation/net/http/otelhttp
* [ ] go.opentelemetry.io/contrib/instrumentation/net/http/httptrace/otelhttptrace
* [ ] go.opentelemetry.io/contrib/instrumentation/google.golang.org/grpc/otelgrpc
* [ ] go.opentelemetry.io/contrib/instrumentation/go.mongodb.org/mongo-driver/mongo/otelmongo
* [ ] go.opentelemetry.io/contrib/instrumentation/github.com/gorilla/mux/otelmux
* [ ] go.opentelemetry.io/contrib/instrumentation/github.com/gin-gonic/gin/otelgin
* [ ] go.opentelemetry.io/contrib/instrumentation/github.com/labstack/echo/otelecho
* [ ] go.opentelemetry.io/contrib/instrumentation/github.com/aws/aws-lambda-go/otellambda/xrayconfig
* [ ] go.opentelemetry.io/contrib/instrumentation/github.com/aws/aws-lambda-go/otellambda
[...]

PR #8341, opened 2025-12-29
refactor(otelgin): switch version from function to string

ElenaZvereva:
Change Version() func to const Version string in otelgin
Part of #8272


dmathieu (approved):

instrumentation/github.com/gin-gonic/gin/otelgin/version.go
```diff
@@ -4,7 +4,5 @@
 package otelgin // import "go.opentelemetry.io/contrib/instrumentation/github.com/gin-gonic/gin/otelgin"
 
 // Version is the current release version of the gin instrumentation.
-func Version() string {
-	return "0.64.0"
-	// This string is updated by the pre_release.sh script during release
-}
+// It is updated by the pre_release.sh script during release.
```
pellared: The is an internal comment. It should not be visible in pkg.godev. Use [`pkgsite`](https://pkg.go.dev/golang.org/x/pkgsite) for testing how the documentation renders.

PR #8346, opened 2026-01-04
Switch version from function to string in azurevm

sonalgaud12:
Change Version() func to const Version string in azurevm

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -24,7 +24,7 @@ This project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0.htm
 
 ### Changed
 
-- The `Version()` function in `go.opentelemetry.io/contrib/detectors/azure/azurevm` has been replaced by `const Version`. (#8346)
+- `go.opentelemetry.io/contrib/detectors/azure/azurevm` now uses semantic convention key constants with `.String()` method instead of helper functions for creating resource attributes. (#8346)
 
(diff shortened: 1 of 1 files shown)


detectors/azure/azurevm/vm.go
```diff
@@ -60,25 +60,25 @@ func (detector *ResourceDetector) Detect(ctx context.Context) (*resource.Resourc
 	}
 
 	if metadata.VMId != nil {
-		attributes = append(attributes, semconv.HostID(*metadata.VMId))
+		attributes = append(attributes, semconv.HostIDKey.String(*metadata.VMId))
```
flc1125, line 63: Can you explain why you want to adjust it this way?

For now, the functional style seems to be somewhat better.

sonalgaud12: This follows issue #8272 to switch from helper functions to key-based strings for semantic conventions. I see your point about the functional style being cleaner. I just implemented as per my understanding of the issue, so in case we prefer the current implementation, I am happy to close the PR

flc1125: I think you might have misunderstood. You can refer to https://github.com/open-telemetry/opentelemetry-go-contrib/pull/8142 to help you understand the purpose and requirements of #8272.

dmathieu (changes requested):
As @flc1125 mentions, this PR does the wrong change.

Issue #8258, opened 2025-12-08
[Bug]: Unable to reuse http.Request with body when client is instrumented with otelhttp.NewTransport


adomaskizogian:
### Component

Instrumentation: otelhttp

### Describe the issue you're facing

use case: retry http request on non 2xx status code or error

When reusing http.Request with a body that can be read only once, the subsequent request `Do(*http.Request)` fails with a formatted error (length and url may differ): `http://127.0.0.1:62343": net/http: HTTP/1.x transport connection broken: http: ContentLength=5 with Body length`

This suggests that the http client instrumentation attempts to ready body every time.

### Expected behavior

No errors.
Request is issued every time with body.

The example code does not fail when the http client is not instrumented.

### Steps to Reproduce

```go
import (
	"bytes"
	"io"
	"net/http"
	"net/http/httptest"
	"testing"

	"go.opentelemetry.io/contrib/instrumentation/net/http/otelhttp"
)

func Test_Reuse_Body_FAIL(t *testing.T) {
	var bodies []string
	server := httptest.NewServer(http.HandlerFunc(func(w http.ResponseWriter, r *http.Request) {
		body, err := io.ReadAll(r.Body)
		if err != nil {
			http.Error(w, "Failed to read body", http.StatusInternalServerError)
			return
		}
		defer r.Body.Close()
		bodies = append(bodies, string(body))

		w.WriteHeader(http.StatusOK)
	}))
	defer server.Close()

	httpClient := &http.Client{
		Transport: otelhttp.NewTransport(http.DefaultTransport),
	}

	body := bytes.NewBuffer([]byte("hello"))
	req, _ := http.NewRequestWithContext(t.Context(), http.MethodPost, server.URL, body)

	for range 2 {
		resp, err := httpClient.Do(req)
		if err != nil {
			t.Fatalf("failed to do http req: %v", err)
		}

		defer resp.Body.Close()
		io.Copy(io.Discard, resp.Body)
	}

	if len(bodies) != 2 || bodies[0] != bodies[1] {
		t.Fatal("test failed")
	}
}
```

### Potential Fix

I am not 100% sure this the right approach but here's my shot:
replace direct access to `Body` field with `GetBody()` calls.
on https://github.com/open-telemetry/opentelemetry-go-contrib/blob/a98be56f3747cab50e0ba0c32d74cf56fcba17fe/instrumentation/net/http/otelhttp/transport.go#L123

### Workaround
[...]

PR #8352, opened 2026-01-06
fix: get body using request.GetBody to enable request reuse

adomaskizogian:
fixes https://github.com/open-telemetry/opentelemetry-go-contrib/issues/8258

get http request body from `GetBody` if it is set, otherwise fallback to `Body` field.
This allows reusing the request body bytes if the same http request struct is reused on subsequent requests.


instrumentation/net/http/otelhttp/transport.go
```diff
@@ -117,11 +117,19 @@ func (t *Transport) RoundTrip(r *http.Request) (*http.Response, error) {
...
+	// GetBody is preferred over direct access to Body if the function is set.
+	// if resulting body is nil or is NoBody or err is not nil, we don't want to mutate the body as it
 	// will affect the identity of it in an unforeseeable way because we assert
 	// ReadCloser fulfills a certain interface and it is indeed nil or NoBody.
-	bw := request.NewBodyWrapper(r.Body, func(int64) {})
-	if r.Body != nil && r.Body != http.NoBody {
+	var body io.ReadCloser
+	var err error
+	if r.GetBody != nil {
+		body, err = r.GetBody()
```
dmathieu: We should handle `err` not being `nil` (and therefore, `body` being `nil`).

adomaskizogian: see 5daa015fd898dca10070f22a180e40b6b2565ce6
what do you think about terminating early request transport on GetBody error

dmathieu: I'm not sure we should skip the round trip if there's an error reading the body. That causes telemetry issues (so we maybe should record the error within the span). But it doesn't prevent the transport from running.

adomaskizogian: Okay, could you please then elaborate what do you have in mind 
> We should handle err not being nil (and therefore, body being nil).

As bodywritter is only overwritten when err is nil and body is not nil, is not `http.NoBody`
update: simplified the logic a bit on https://github.com/open-telemetry/opentelemetry-go-contrib/pull/8352/changes/2377ac948d12133a710d0b31414262e9333cef44

as for https://github.com/open-telemetry/opentelemetry-go-contrib/commit/5daa015fd898dca10070f22a180e40b6b2565ce6, I will revert

dmathieu: We could call `otel.Handle(err)`.

adomaskizogian: thanks, was not aware of this hook.

see d3055f87d14abcf62cd2c2664c22a9933db4ea49

CHANGELOG.md
```diff
@@ -16,6 +16,7 @@ This project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0.htm
 ### Fixed
 
 - Fix panic when passing nil `TracerProvider` or `MeterProvider` to `WithTracerProvider` or `WithMeterProvider` in `go.opentelemetry.io/contrib/instrumentation/google.golang.org/grpc/otelgrpc`. (#8323)
+- `Transport` in `go.opentelemetry.io/contrib/instrumentation/net/http/otelhttp` now supports reading request body multiple times for subsequent requests that reuse `http.Request`. (#8258)
```
```suggestion
- `Transport` in `go.opentelemetry.io/contrib/instrumentation/net/http/otelhttp` now supports reading request body multiple times for subsequent requests that reuse `http.Request`. (#8352)
```

instrumentation/net/http/otelhttp/transport.go
```diff
@@ -117,11 +118,24 @@ func (t *Transport) RoundTrip(r *http.Request) (*http.Response, error) {
...
+	// if resulting body is nil or is NoBody or GetBody returns an error, we don't want to mutate the body as it
 	// will affect the identity of it in an unforeseeable way because we assert
 	// ReadCloser fulfills a certain interface and it is indeed nil or NoBody.
-	bw := request.NewBodyWrapper(r.Body, func(int64) {})
-	if r.Body != nil && r.Body != http.NoBody {
+	var body io.ReadCloser
+	if r.GetBody != nil {
+		b, err := r.GetBody()
+		if err != nil {
+			otel.Handle(fmt.Errorf("http.Request GetBody returned an error: %w", err))
```
dmathieu, line 131: Shouldn't we still set a body here?
It's not ideal, as it means we can't reuse it when an error occured. But having `nil` seems worse.


```suggestion
			body = r.Body
			otel.Handle(fmt.Errorf("http.Request GetBody returned an error: %w", err))
```

dmathieu: Also, could we test the behavior where `GetBody` returns an error?
You should be able to do that by calling `RoundTrip` with a fake request on which you have overriden `GetBody`.

adomaskizogian: This is already covered, I believe. see: [TestHandleErrorOnGetBodyError](https://github.com/open-telemetry/opentelemetry-go-contrib/pull/8352/changes#diff-b8f73d827a9e8a1cee8752e1cedede43d81c24941e9d799916c13e2ccc646e5aR1161)

Do you have something else in mind?

CHANGELOG.md
adomaskizogian: good catch. I used the issue number, not the PR number.

instrumentation/net/http/otelhttp/transport.go
adomaskizogian: re: Shouldn't we still set a body here?

This is a great question to which I doubt I have an answer.
To me it seems that whenever the `GetBody` is set, the author intended it to be used as it's optional.

Should it fallback to `Body` field when `GetBody` returns an error? I need to think, what is a _nastier_ side effect. The fallback or missing body.

update: I'll set it to fallback to r.Body

dmathieu: Ah, sorry I missed that test.

adomaskizogian: implemented fallback to Body on GetBody error: https://github.com/open-telemetry/opentelemetry-go-contrib/pull/8352/changes/0b3beb39e49497b8bdaaee7f12d443f28ca9b081

instrumentation/net/http/otelhttp/transport.go
```diff
@@ -117,11 +120,22 @@ func (t *Transport) RoundTrip(r *http.Request) (*http.Response, error) {
...
+		} else {
+			body = b
+		}
+	}
+
+	bw := request.NewBodyWrapper(body, func(int64) {})
+	if body != nil && body != http.NoBody {
```
flc1125, line 138: I'm thinking that if we swap the order of these two lines, that is, first check and then assign, could we save some performance overhead?

However, this could be another PR. It won't be a blocking issue for this modification.

adomaskizogian: `bw` is also referenced here https://github.com/open-telemetry/opentelemetry-go-contrib/pull/8352/changes#diff-c9b3377bfdeb24f8c007a0e604dd5a65f5d7e8fa1e64baa1029beb0c8f81b0faR161

if it was never assigned, it always returns `0`
yeah it could save an allocation for requests with no body.

Happy to include it in this pr but imo this would come for a price of code clarity cause we need to know if the wrapper was created.

flc1125: Maybe it would be better to open a new one and handle it separately.

dmathieu: I agree. This improvement is independent from the fix in this PR, it should be in a separate one.

adomaskizogian: What do we do now? Do we proceed with this pr then and want me to propose an improvement to the bodywrapper initialization?

dmathieu: Let's keep this PR with the scope it currently has. Feel free to open another PR with the bodywrapper optimization.

instrumentation/net/http/otelhttp/transport.go
```diff
@@ -117,11 +120,22 @@ func (t *Transport) RoundTrip(r *http.Request) (*http.Response, error) {
...
+	// If the resulting body is nil or is NoBody, we don't want to mutate the body as it
 	// will affect the identity of it in an unforeseeable way because we assert
 	// ReadCloser fulfills a certain interface and it is indeed nil or NoBody.
-	bw := request.NewBodyWrapper(r.Body, func(int64) {})
-	if r.Body != nil && r.Body != http.NoBody {
+	body := r.Body
+	if r.GetBody != nil {
+		b, err := r.GetBody()
+		if err != nil {
+			otel.Handle(fmt.Errorf("http.Request GetBody returned an error: %w", err))
```
ash2k, line 131: Shouldn't this just return the error, as the stdlib does?

https://github.com/golang/go/blob/go1.25.6/src/net/http/client.go#L673

instrumentation/net/http/otelhttp/transport.go
```diff
@@ -117,11 +120,22 @@ func (t *Transport) RoundTrip(r *http.Request) (*http.Response, error) {
...
+		} else {
+			body = b
+		}
+	}
+
+	bw := request.NewBodyWrapper(body, func(int64) {})
```
ash2k, line 137: I think this should only be allocated if it is actually needed - to reduce RAM usage.

Also, this object is unconditionally used below - it shouldn't be if it is not used as a body.

instrumentation/net/http/otelhttp/transport.go
dmathieu: See https://github.com/open-telemetry/opentelemetry-go-contrib/pull/8352#discussion_r2668242463

ash2k: There are two issues, actually:

- If I use `transport := http.DefaultTransport` and set a breakpoint in the `GetBody` impl in that test, it is not triggered by the test execution. But it is triggered with the OTEL transport. This means OTEL transport calls `GetBody` when stdlib does not. stdlib only calls it if it needs another instance of the body (i.e. for second, third, etc attempts).
- stdlib returns the error from `GetBody`, it does not proceed. OTEL transport reports it and proceeds.

The unit test encodes incorrect behavior.

My understanding is that the code should:

- wrap the body that is set in the request, if it is not `nil` and not `NoBody`.
- wrap the `GetBody` func but not call it. The underlying transport will call it if necessary.

ash2k: https://github.com/open-telemetry/opentelemetry-go-contrib/pull/8532

PR #8356, opened 2026-01-07
Switch version from function to string in otelaws

jongwooo:
Change `Version()` func to `const Version` string in otelaws


dmathieu:
This needs a changelog entry.

dmathieu (approved):

PR #8358, opened 2026-01-07
Switch version from function to string in host

jongwooo:
Change `Version()` func to `const Version` string in host


dmathieu:
This needs a changelog entry.

PR #8360, opened 2026-01-07
Switch version from function to string in otelrestful

osullivandonal:
Change Version() func to const Version string in otelrestful

Part of #8272


dmathieu:
This needs a changelog entry.

osullivandonal:
> This needs a changelog entry.

Done!

dmathieu (approved):

Issue #8373, opened 2026-01-09
[Bug]: Dangerous metric labeling practices in example


eddyariki:
### Component

Instrumentation: otelecho

### Describe the issue you're facing

This section of test code can create high cardinality in metrics because RemoteAddr includes the client ip and port. 
`attribute.String("client.ip", r.RemoteAddr),`
https://pkg.go.dev/go.opentelemetry.io/contrib/instrumentation/github.com/labstack/echo/otelecho#example-Middleware

These ports are transient ports made for TCP connections. I believe we should not add potentially dangerous code snippets in the examples.

This example may be coming from this file.
https://github.com/open-telemetry/opentelemetry-go-contrib/blob/main/instrumentation/github.com/labstack/echo/otelecho/example_test.go#L112C3-L118C6

```
otelecho.WithMetricAttributeFn(func(r *http.Request) []attribute.KeyValue {
			// Add custom attributes from HTTP request
			return []attribute.KeyValue{
				attribute.String("client.ip", r.RemoteAddr),
				attribute.String("user.agent", r.UserAgent()),
			}
		}),
```

### Expected behavior

Attributes added following best practices of metric labels. 

### Steps to Reproduce

Visit: https://pkg.go.dev/go.opentelemetry.io/contrib/instrumentation/github.com/labstack/echo/otelecho#example-Middleware

### Operating System

Irrelevant

### Device Architecture

ARM32

### Go Version

1.25

### Component Version

otelecho v0.64.0

pellared:
@eddyariki, thanks a lot for creating the issue. Are you willing to fix it? It may be also worth to check if the same issue does not exist in other examples.

eddyariki:
@pellared Sure, let me submit a PR later.

dmathieu:
cc @flc1125 as code owner.

dmathieu:
As a side note, those two attributes also don't match semantic conventions. Your updated example should use valid semconv attributes.

eddyariki:
@dmathieu 

> As a side note, those two attributes also don't match semantic conventions. Your updated example should use valid semconv attributes. 

I think the intention with this snippet is to show case how custom attributes can be added via `otelecho.WithMetricAttributeFn` and `otelecho.WithEchoMetricAttributeFn`. 

I want to avoid introducing attributes that are already added by the middleware, so maybe it is indeed better to use a custom attribute name instead of the predefined ones in the semantic conventions. What do you think?

[...]

PR #8381, opened 2026-01-10
Use explicitly custom code snippet examples in otelecho

eddyariki:
Addresses https://github.com/open-telemetry/opentelemetry-go-contrib/issues/8373

- Adds comment in `echo_test.go` to explicitly recommend avoiding using high-cardinality attributes
- Replaces custom metrics example in `WithMetricAttributeFn`  and `WithEchoMetricAttributeFn`  with safer examples.
- Showcases behavior with duplicate metrics as well. 

Note: 
- Most standard metrics for http.Request and echo context are already added as attributes in the middleware, so I went with a more custom metric for the sake of high-lighting potential use-cases. 
- I checked other code snippets and only found `echo_test.go`'s dangerous pattern in their tests. I believe there are no other dangerous patterns in the core packages and instrumentation libraries.


instrumentation/github.com/labstack/echo/otelecho/example_test.go
```diff
@@ -112,15 +112,16 @@ func ExampleMiddleware_withMetrics() {
 		otelecho.WithMetricAttributeFn(func(r *http.Request) []attribute.KeyValue {
 			// Add custom attributes from HTTP request
 			return []attribute.KeyValue{
-				attribute.String("client.ip", r.RemoteAddr),
-				attribute.String("user.agent", r.UserAgent()),
+				attribute.Bool("custom.has_request_body", r.ContentLength > 0),
```
flc1125:
```suggestion
				attribute.Bool("custom.has_request_body", r.ContentLength != 0),
```

There is a case where `-1` just represents an unknown length, but it doesn't mean there is no length at all.

instrumentation/github.com/labstack/echo/otelecho/example_test.go
```diff
@@ -112,15 +112,16 @@ func ExampleMiddleware_withMetrics() {
...
+				attribute.Bool("custom.has_content_type", r.Header.Get("Content-Type") != ""),
 			}
 		}),
 		otelecho.WithEchoMetricAttributeFn(func(c echo.Context) []attribute.KeyValue {
 			// Add custom attributes from Echo context
+			// If attributes are duplicated between this method and `WithMetricAttributeFn`, the attributes in this method will be used.
 			return []attribute.KeyValue{
-				attribute.String("handler.path", c.Path()),
-				attribute.String("handler.method", c.Request().Method),
+				attribute.Bool("custom.has_request_body", c.Request().ContentLength > 0),
```
flc1125:
```suggestion
				attribute.Bool("custom.has_request_body", c.Request().ContentLength != 0),
```

instrumentation/github.com/labstack/echo/otelecho/example_test.go
eddyariki: Thanks for the catch. 

https://github.com/open-telemetry/opentelemetry-go-contrib/pull/8381/commits/a9790ed7de5f9d1a36ef957f1ff895de35867cfb

instrumentation/github.com/labstack/echo/otelecho/example_test.go
eddyariki: https://github.com/open-telemetry/opentelemetry-go-contrib/pull/8381#discussion_r2679118206

flc1125 (approved):

PR #8302, opened 2025-12-17
Switch version from function to string in otelhttptrace

ElenaZvereva:
refactor(otelhttptrace): change Version() func to const Version string
Part of #8272

PR #8383, opened 2026-01-12
Switch version from function to string in otelhttptrace

eddyariki:
Change Version() func to const Version string in otelhttptrace

Part of #8272


dmathieu:
Duplicate of https://github.com/open-telemetry/opentelemetry-go-contrib/pull/8302/

Issue #7470, opened 2025-06-16
Do not use `span.RecordError` for terminating errors


pellared:
## Why

> What is the value of recording the error as event if the type is set as attribute and value (`err.Error()`) as status description?
> 
>> An exception SHOULD be recorded as an Event on the span during which it occurred if and only if it remains unhandled when the span ends and causes the span status to be set to ERROR.
> 
> I argue that error result is not an exception. For me a `panic` that was not being recovered is an unhandled exception. We already cover this part of this specification here:
> 
> https://github.com/open-telemetry/opentelemetry-go/blob/bc531cb3f9ce9c328b12e473b6409784728dc608/sdk/trace/span.go#L465-L482
> 
> Additionally see: https://github.com/open-telemetry/semantic-conventions/blob/main/docs/general/recording-errors.md#recording-errors-on-spans
> 
> Therefore, we should not call `span.RecordError(err)`. 

 _Originally posted by @pellared in [#7254](https://github.com/open-telemetry/opentelemetry-go-contrib/issues/7254#issuecomment-2962485821)_

> My proposal is to not use `span.RecordError` for **terminating errors** per [recording errors on spans semantic conventions](https://github.com/open-telemetry/semantic-conventions/blob/main/docs/general/recording-errors.md#recording-errors-on-spans).
> 
> During the SIG meeting, it was also noted that `span.RecordError` captures a stack trace. However, since the stack trace reflects where the error was recorded, not where it originated, I don’t find it particularly valuable. Additionally, capturing and emitting a stack trace has a performance cost. In most languages with exception handling (such as Java or .NET), the stack trace is already included in the thrown exception so the cost of capturing the stack trace was already taken and the stack trace is taken in the place where the exception was thrown.
> 
>> @pellared does that mean span.RecordError should essentially never be called by instrumentation, since panic recovery is already handled by the SDK? Almost makes it seem like we should deprecate RecordError in the API,
> 
> I believe [`span.RecordError`](https://github.com/open-telemetry/opentelemetry-specification/blob/main/specification/trace/api.md#record-exception) remains valuable, particularly for **non-terminating errors**. A common use case is **retry logic**, where errors occur but don't immediately end the span. This is especially relevant in the context of **custom instrumentation**, where developers have full control and can explicitly record such intermediate errors.
> 
[...]

PR #8386, opened 2026-01-12
instrumentation: update error handling to set attributes instead of recording errors

pellared:
Fixes https://github.com/open-telemetry/opentelemetry-go-contrib/issues/7470
Fixes https://github.com/open-telemetry/opentelemetry-go-contrib/issues/7254

Related semconv PR: https://github.com/open-telemetry/semantic-conventions/pull/3256
Related spec PR: https://github.com/open-telemetry/opentelemetry-specification/pull/4825

### DIFF
--- a/tools/go.mod
+++ b/tools/go.mod
@@ -39,7 +39,7 @@ require (
 	github.com/Djarvur/go-err113 v0.1.1 // indirect
 	github.com/Masterminds/semver/v3 v3.4.0 // indirect
 	github.com/Microsoft/go-winio v0.6.2 // indirect
-	github.com/MirrexOne/unqueryvet v1.4.0 // indirect
+	github.com/MirrexOne/unqueryvet v1.5.0 // indirect
 	github.com/OpenPeeDeeP/depguard/v2 v2.2.1 // indirect
 	github.com/ProtonMail/go-crypto v1.3.0 // indirect
 	github.com/alecthomas/chroma/v2 v2.22.0 // indirect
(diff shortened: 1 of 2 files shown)


flc1125 (approved):

CHANGELOG.md
```diff
@@ -41,6 +41,8 @@ This project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0.htm
 - The `Version()` function in `go.opentelemetry.io/contrib/samplers/probability/consistent` has been replaced by `const Version`. (#8366)
 - The `Version()` function in `go.opentelemetry.io/contrib/instrumentation/net/http/httptrace/otelhttptrace` has been replaced by `const Version`. (#8302)
 - The `Version()` function in `go.opentelemetry.io/contrib/instrumentation/go.mongodb.org/mongo-driver/v2/mongo/otelmongo` has been replaced by `const Version`. (#8370)
+- Set `error.type` attribute instead of adding `exception` span events in `go.opentelemetry.io/contrib/instrumentation/net/http/otelhttp`. (#8386)
```
dmathieu, line 45: This should also mention go-restful, gingonic, mux, echo, httptrace.

dmathieu (commented):
🥳

CHANGELOG.md
pellared: Are you sure that these were calling `span.RecordError`? It looks like the changes are only in generated code which is unused by these packages. Or have I missed something?

dmathieu: Ah, good point!

pellared:
It would be nice to have a linter which not allows calling `RecordError`. However, we may deprecate this function in future and then this linter would no longer be needed (the existing linter would work). Reference: https://github.com/open-telemetry/opentelemetry-specification/pull/4825

PR #8395, opened 2026-01-14
chore(otelhttp): Remove ReadBytesKey, ReadErrorKey, WroteBytesKey, WriteErrorKey from public API 

### LINKED ISSUE #8130: Remove ReadBytesKey, ReadErrorKey, WroteBytesKey, WriteErrorKey from public API
These semantic conventions-esque constants should be upstreamed and removed from the package before stabilization.

Rxfa:
Closes #8130

### DIFF
--- a/instrumentation/net/http/otelhttp/common.go
+++ b/instrumentation/net/http/otelhttp/common.go
@@ -6,16 +6,7 @@ package otelhttp // import "go.opentelemetry.io/contrib/instrumentation/net/http
 import (
 	"net/http"
 
-	"go.opentelemetry.io/otel/attribute"
 	"go.opentelemetry.io/otel/trace"
-)
-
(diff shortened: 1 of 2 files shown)


Rxfa:
Please let me know if this is as [discussed](https://github.com/open-telemetry/opentelemetry-go-contrib/issues/8130#issuecomment-3744833670), or if further changes are required.

flc1125:
PTAL: https://github.com/open-telemetry/opentelemetry-go-contrib/issues/8130#issuecomment-3747263829

dmathieu:
I was also coming to say that 😁 
We should do this in two steps:

* Deprecate the old attributes, but keep them around and add the new attributes.
* Then, after at least one release has passed, remove the old attributes.

PR #8433, opened 2026-01-19
echo/v5

arielsrv:


dmathieu:
This includes breaking changes, meaning we'd drop support for echo v4.
We should do a new module, like we do for [mongo](https://github.com/open-telemetry/opentelemetry-go-contrib/tree/main/instrumentation/go.mongodb.org/mongo-driver) for example.

PR #8434, opened 2026-01-19
echo/v5

arielsrv:


flc1125:
Please refer to the comments in https://github.com/open-telemetry/opentelemetry-go-contrib/pull/8433#issuecomment-3767276788

Additionally, we need to pass the CLA.

dmathieu:
And we can't drop support for Go 1.24 until 1.26 is released, as we follow the [Go release policy](https://go.dev/doc/devel/release#policy).

arielsrv:
> And we can't drop support for Go 1.24 until 1.26 is released, as we follow the [Go release policy](https://go.dev/doc/devel/release#policy).

It’s fine — the problem is that the new version requires 1.25, so I’ll wait. Thanks!

dmathieu:
There can be exceptions for new modules. But this still ins't what this pr does.

flc1125:
> There can be exceptions for new modules

A stupid question: Does this mean we allow new modules to receive the 1.25.0 version? Or do we allow some exceptions, and what are these exceptions?

dmathieu:
We previously did an exception because the package only supported the latest version.
The exception is only at setup. It drops once there's a go upgrade.

flc1125:
@arielsrv Can you keep the v4 version and create a dedicated component based on v5?

We can find some ideas here: https://github.com/open-telemetry/opentelemetry-go-contrib/tree/main/instrumentation/go.mongodb.org/mongo-driver

arielsrv:
> @arielsrv Can you keep the v4 version and create a dedicated component based on v5?
> 
> We can find some ideas here: https://github.com/open-telemetry/opentelemetry-go-contrib/tree/main/instrumentation/go.mongodb.org/mongo-driver

Sorry, just to clarify — I’m not sure I understand.

Should we proceed with this approach in this PR, or should we keep the other PR [open](https://github.com/open-telemetry/opentelemetry-go-contrib/pull/8457) and continue from there?

dmathieu:
The other PR being on the right approach, I think we should continue from there. That doesn't prevent you from giving input.

arielsrv:
> The other PR being on the right approach, I think we should continue from there. That doesn't prevent you from giving input.

No problem! Let’s close this one :)

PR #8435, opened 2026-01-19
Deprecate ReadBytesKey, ReadErrorKey, WroteBytesKey and WriteErrorKey in favor of semconv keys

Rxfa:
Partial Fix for #8130


CHANGELOG.md
```diff
@@ -28,6 +28,10 @@ This project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0.htm
 - The deprecated `WithRouteTag` in `go.opentelemetry.io/contrib/instrumentation/net/http/otelhttp`.
   The route is already added automatically for spans.
   For metrics, the alternative is to use the `WithMetricAttributesFn` option. (#8268)
+- In `instrumentation/net/http/otelhttp`, the attributes `ReadBytesKey`, `ReadErrorKey`, `WroteBytesKey` and 
+  `WriteErrorKey` 
+  are deprecated and will be removed in a future release. Use `semconv.HTTPRequestBodySizeKey`, `semconv.
+  ErrorMessageKey`, `semconv.HTTPResponseBodySizeKey` and `semconv.ErrorMessageKey` respectively.
```
```suggestion
  ErrorMessageKey`, `semconv.HTTPResponseBodySizeKey` and `semconv.ErrorMessageKey` respectively. (#8435)
```

instrumentation/net/http/otelhttp/handler.go
```diff
@@ -144,7 +149,11 @@ func (h *middleware) serveHTTP(w http.ResponseWriter, r *http.Request, next http
 	writeRecordFunc := func(int64) {}
 	if h.writeEvent {
 		writeRecordFunc = func(n int64) {
-			span.AddEvent("write", trace.WithAttributes(WroteBytesKey.Int64(n)))
+			span.AddEvent("write", trace.WithAttributes(
+				WroteBytesKey.Int64(n),
+				semconv.HTTPResponseSizeKey.Int64(n),
```
flc1125:
```suggestion
				semconv.HTTPResponseSize(int(n)),
```

instrumentation/net/http/otelhttp/handler.go
```diff
@@ -129,7 +130,11 @@ func (h *middleware) serveHTTP(w http.ResponseWriter, r *http.Request, next http
 	readRecordFunc := func(int64) {}
 	if h.readEvent {
 		readRecordFunc = func(n int64) {
-			span.AddEvent("read", trace.WithAttributes(ReadBytesKey.Int64(n)))
+			span.AddEvent("read", trace.WithAttributes(
+				ReadBytesKey.Int64(n),
+				semconv.HTTPRequestBodySizeKey.Int64(n),
```
flc1125:
```suggestion
				semconv.HTTPRequestBodySize(int(n)),
```

flc1125 (commented):
Could you add some unit tests for the new attributes

CHANGELOG.md
```diff
@@ -28,6 +28,10 @@ This project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0.htm
 - The deprecated `WithRouteTag` in `go.opentelemetry.io/contrib/instrumentation/net/http/otelhttp`.
   The route is already added automatically for spans.
   For metrics, the alternative is to use the `WithMetricAttributesFn` option. (#8268)
+- In `instrumentation/net/http/otelhttp`, the attributes `ReadBytesKey`, `ReadErrorKey`, `WroteBytesKey` and
```
flc1125: I think this should be placed in the `Deprecated` section.

PS: Only when it is removed, is it considered `Removed` section

instrumentation/net/http/otelhttp/handler.go
Rxfa: I actually think I made a mistake here, shouldn't this be `semconv.HTTPResponseBodySize(int(n))`?

`semconv.HTTPResponseSize(int(n))` would be an attribute representing the total size of the response and not just the size of the body.

CHANGELOG.md
dmathieu: I think should even be two changelog entries.
One in the deprecated section that mentions the deprecated attributes, and one in the `Added` section that mentions the added ones.

flc1125: I agree.

instrumentation/net/http/otelhttp/handler_test.go
```diff
@@ -719,3 +720,58 @@ func BenchmarkHandlerServeHTTP(b *testing.B) {
...
+
+	require.NotEmpty(t, readAttrs)
+	require.NotEmpty(t, writeAttrs)
+
+	payloadSize := len(payload)
+	assert.Contains(t, readAttrs, ReadBytesKey.Int64(int64(payloadSize)))
+	assert.Contains(t, readAttrs, semconv.HTTPRequestBodySize(payloadSize))
+
+	assert.Contains(t, writeAttrs, WroteBytesKey.Int64(int64(payloadSize)))
+	assert.Contains(t, writeAttrs, semconv.HTTPResponseSize(payloadSize))
```
flc1125: I think it’s best to use an approach like `attribute.String("xxx", "xxx")` for validation here. The current situation is somewhat like using your own method to validate yourself. Once a property like `ReadBytesKey` is changed, it won’t be detected here.

flc1125 (approved):
After this adjustment, LGTM.

instrumentation/net/http/otelhttp/common.go
```diff
@@ -12,9 +12,16 @@ import (
 
 // Attribute keys that can be added to a span.
 const (
-	ReadBytesKey  = attribute.Key("http.read_bytes")  // if anything was read from the request body, the total number of bytes read
-	ReadErrorKey  = attribute.Key("http.read_error")  // If an error occurred while reading a request, the string of the error (io.EOF is not recorded)
+	// Deprecated: use semconv.HTTPRequestBodySizeKey instead.
+	ReadBytesKey = attribute.Key("http.read_bytes") // if anything was read from the request body, the total number of bytes read
```
pellared, line 16: How about?

```suggestion
	// Deprecated: use semconv.HTTPRequestBodySizeKey instead.
	ReadBytesKey = semconv.HTTPRequestBodySizeKey
```

Same comment applies to other consts

CHANGELOG.md
```diff
@@ -13,6 +13,8 @@ This project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0.htm
 - `WithMetricAttributesFn` option in `go.opentelemetry.io/contrib/instrumentation/google.golang.org/grpc/otelgrpc` to define dynamic attributes on auto-instrumented metrics. (#8191)
 - Add support for configuring propagators in `go.opentelemetry.io/contrib/otelconf`. (#8281)
 - Add `const Version` in `go.opentelemetry.io/contrib/bridges/prometheus`. (#8401)
+- Added semantic convention equivalents for legacy `otelhttp` read/write attributes, including request and response 
+  body size and error message attributes. (#8435)
```
pellared: This changelog entry is not clear

instrumentation/net/http/otelhttp/common.go
Rxfa: This has been previously discussed [here](https://github.com/open-telemetry/opentelemetry-go-contrib/issues/8130).

instrumentation/net/http/otelhttp/handler_test.go
Rxfa: Agreed, and changed the code accordingly.

CHANGELOG.md
Rxfa: Just changed it to try to improve clarity, LMK if it's better now.

instrumentation/net/http/otelhttp/common.go
dmathieu: Yes, and I said exactly what @pellared says here.

Rxfa: Apologies in that case, it seems like I misunderstood what was previously said then.

I do have a question now, though, since those attributes are actually different, wouldn't the whole idea of having a version that supports both the ones in place and the semantic-defined ones consist of leaving those constants as they are right now, while also including the semantic-defined attributes and removing the old constants afterwards? (I'm going off of what I understood at the time reading [this](https://github.com/open-telemetry/opentelemetry-go-contrib/issues/8130#issuecomment-3747263829) comment)

My understanding is that if we just change the values of the old constants, we will just have the semantic-defined constants instead of having both the semantic-defined constants and the old constants (thus breaking backwards compatibility).
I'm sorry if I'm missing something really obvious. I'm just trying to clear all this in my mind and find where the flaws are in my thought process.

dmathieu: Urgh actually, discard my last comment, it's me being wrong.
Those deprecated attributes do indeed nee to remain the same, since the semconv version changes their value. So using the semconv value breaks them.

MrAlias:
@Rxfa are you still able to work on this?

Rxfa:
> @Rxfa are you still able to work on this?

Hey, I'm still able to work on this.

I believe I addressed all feedback, and I'm awaiting further instructions if there's a reason for the PR not to be approved and merged.

flc1125:
@Rxfa Please resolve the conflicts and make sure the CI passes.

flc1125:
@Rxfa PTAL, The change log needs to be located below the "Unreleased" section.

instrumentation/net/http/otelhttp/common.go
pellared: @dmathieu, double-checking, does it mean that I can resolve the conversation?

pellared:
@Rxfa, sorry for inconvenience, but can I ask you again to resolve the conflicts and make sure the CI passes?

Rxfa:
> @Rxfa, sorry for inconvenience, but can I ask you again to resolve the conflicts and make sure the CI passes?

It should be fixed now.

CHANGELOG.md
```diff
@@ -10,7 +10,12 @@ This project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0.htm
 
 ### Added
 
-- Add environment variables propagation carrier in `go.opentelemetry.io/contrib/propagators/envcar`. (#8442)
+- Add environment variables propagation carrier in `go.opentelemetry.io/contrib/propagators/envcar`. (#8442e
```
```suggestion
- Add environment variables propagation carrier in `go.opentelemetry.io/contrib/propagators/envcar`. (#84420
```

CHANGELOG.md
```diff
@@ -10,7 +10,12 @@ This project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0.htm
 
 ### Added
 
-- Add environment variables propagation carrier in `go.opentelemetry.io/contrib/propagators/envcar`. (#8442)
+- Add environment variables propagation carrier in `go.opentelemetry.io/contrib/propagators/envcar`. (#84420
```
```suggestion
- Add environment variables propagation carrier in `go.opentelemetry.io/contrib/propagators/envcar`. (#8442)
```

dmathieu (approved):

instrumentation/net/http/otelhttp/common.go
```diff
@@ -12,9 +12,16 @@ import (
 
 // Attribute keys that can be added to a span.
 const (
-	ReadBytesKey  = attribute.Key("http.read_bytes")  // if anything was read from the request body, the total number of bytes read
-	ReadErrorKey  = attribute.Key("http.read_error")  // If an error occurred while reading a request, the string of the error (io.EOF is not recorded)
+	// Deprecated: use httpSemconv.HTTPRequestBodySizeKey instead.
```
dashpole: I don't see HTTPRequestBodySizeKey in any versions after v1.25.0 of semconv.  Is this the correct latest semconv to point people at?

dashpole: nit: people usually import this as `semconv` instead of httpSemconv.

dashpole: NVM.  Seems like github search just wasn't finding it.

CHANGELOG.md
```diff
@@ -11,6 +11,11 @@ This project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0.htm
 ### Added
 
 - Add environment variables propagation carrier in `go.opentelemetry.io/contrib/propagators/envcar`. (#8442)
+- Added semantic convention attribute keys `HTTPRequestBodySizeKey`,
```
pellared: This should be two entries, one "Changed" that we are emitting "x.y" instead of "a.b" per semantic conventions, second "Deprecated" about the deprecation of the constants

instrumentation/net/http/otelhttp/common.go
Rxfa: I did it this way to avoid a potential package name collision in [handler.go](https://github.com/open-telemetry/opentelemetry-go-contrib/pull/8435#diff-70a93fdc86adfd2f20a035bb596f6d9320d6fed7008eaefc6c7d7cde2b1fc97a)
LMK know if you think I should still have it imported as `semconv` and change the import below under a different alias
```go
// handler.go:14
semconv "go.opentelemetry.io/otel/semconv/v1.40.0"
```

dashpole: `httpSemconv` looks like a field name internal to the handler.  This deprecation comment should tell users of the otelhttp library where they can find the replacement (which is here: go.opentelemetry.io/otel/semconv/v1.40.0).

CHANGELOG.md
Rxfa: Regarding the entry under "Changed", you mean something similar to what can be seen [here](https://github.com/open-telemetry/opentelemetry-go-contrib/blob/b228c0f2f20121d34357da55a19ca92d4f67c1cd/CHANGELOG.md?plain=1#L110)?

instrumentation/net/http/otelhttp/common.go
Rxfa: You are right, that was changed by accident when doing some of the name changes around field names and package aliases
Just pushed a commit to fix it, please LMK what you think

dashpole: (this comment still seems unchanged)

Rxfa: Can you look at it again?

CHANGELOG.md
pellared: Similar to https://github.com/open-telemetry/opentelemetry-go-contrib/blob/b228c0f2f20121d34357da55a19ca92d4f67c1cd/CHANGELOG.md?plain=1#L121-L127

Rxfa: Just pushed a commit addressing this. Please LMK if it looks good now

PR #8442, opened 2026-01-21
envcar: add environment carrier

### LINKED ISSUE #8300: Support the environment variable propagation
With the [environment variable context propagation spec](https://opentelemetry.io/docs/specs/otel/context/env-carriers/) being added, we'd like to have its support added to Go. This is part of a larger effort to adding support in each of the currently supported languages.

Two example pull requests have been opened in the past that can be used for ideas in conjunction with the specification.

- https://github.com/open-telemetry/opentelemetry-go/pull/6778
- https://github.com/open-telemetry/opentelemetry-go/pull/2931

Joibel:
Fixes #8300 

Add a carrier implementing [environment carrier](https://opentelemetry.io/docs/specs/otel/context/env-carriers/) based on top of TextMapCarrier.

Allows the callers selection of mechanism for setting environment variables, compatible with `os.Setenv` - for users who'd like to set environment variables in spawned processes.

This is mostly a migration of @pellared's work in https://github.com/open-telemetry/opentelemetry-go/pull/6778 to go-contrib. I've added further test coverage to appease the coverage bot.

There are three example tests for how to use it.

~~This is incomplete due to ownership: I'd be happy to be CODEOWNER, or add @pellared as the owner.~~
I've set myself as CODEOWNER

### DIFF
--- a/propagators/envcar/carrier_test.go
+++ b/propagators/envcar/carrier_test.go
@@ -250,7 +250,7 @@ func TestConcurrentChildProcesses(t *testing.T) {
 	}
 }
 
-// Ensure Get and Keys to fetch from cache
+// Ensure Get and Keys to fetch from cache.
 func TestCarrierGetFetchOnce(t *testing.T) {
 	t.Setenv("TRACEPARENT", "myvalue")
 	c := envcar.Carrier{}


kamphaus:
I see that other propagators have example_test.go files.
Would it be interesting to add such for `envcar`?

I think two scenarios might be interesting:
* once with `os.Setenv`
* once to show how to launch a child process with different context than the current process

pellared:
> This is incomplete due to ownership: I'd be happy to be CODEOWNER, or add @pellared as the owner.

Go for it :+1:

propagators/envcar/carrier.go
```diff
@@ -0,0 +1,60 @@
...
+// This is useful for propagating values that are set in the environment
+// and need to be accessed by different processes or services.
+// The keys are uppercased to avoid case sensitivity issues across different
+// operating systems and environments.
+type Carrier struct {
+	// SetEnvFunc is a function that sets the environment variable.
+	// Usually, you want to set the environment variable for processes
+	// that are spawned by the current process.
+	// By default implementation, it does nothing.
+	SetEnvFunc func(key, value string) error
```
pellared: The result is unused.

```suggestion
	SetEnvFunc func(key, value string)
```

Joibel: This is written so that you can pass os.Setenv to it without wrapping it. I can change it, but my preference is that it retains this signature.

pellared: What is the use case of using `os.Setenv`?

I think the main use case is to spawn new processes (then `func` to populate a `map[string]string` can be used) or using nothing. 

Note that `os.Setenv` would work very badly in a concurrent scenario where multiple goroutines would spawn many new processes. 

By the way, I would like to have a test  and example how this can work in an "e2e" setup.

Wouldn't it require creating a new carrier for each goroutine?

CC @kamphaus

Joibel:
> > This is incomplete due to ownership: I'd be happy to be CODEOWNER, or add @pellared as the owner.
> 
> Go for it 👍

Done

propagators/envcar/carrier_example_test.go
```diff
@@ -0,0 +1,111 @@
...
+	fmt.Printf("Span ID: %s\n", spanCtx.SpanID())
+	fmt.Printf("Sampled: %t\n", spanCtx.IsSampled())
+	// Output:
+	// Trace ID: 0102030405060708090a0b0c0d0e0f10
+	// Span ID: 0102030405060708
+	// Sampled: true
+}
+
+// This example is a go program where we have a trace and we'd like to inject it
+// in our current environment variables.
```
pellared: What is the value of injecting it to the current environment variable? Note that it would also "destroy" the extraction if we would like to do it the second time. I think this is a bug-prone example.

pellared: This is against https://opentelemetry.io/docs/specs/otel/context/env-carriers/#environment-variable-immutability

> Once set for a process, environment variables SHOULD be treated as immutable within that process

CC @kamphaus

pellared (changes requested):
Let's please add a  test that ensures that it works correctly in a scenario when the application (code) spawns processes from a few concurrent goroutines and each process is created inside a new span (so that each of the new process receives  unique trace context).

Reference: https://opentelemetry.io/docs/specs/otel/context/env-carriers/#process-spawning

CODEOWNERS
```diff
@@ -64,6 +64,7 @@ processors/minsev                                                       @open-te
 propagators/autoprop/                                                   @open-telemetry/go-approvers @MrAlias
 propagators/aws/                                                        @open-telemetry/go-approvers @akats7
 propagators/b3/                                                         @open-telemetry/go-approvers @pellared
+propagators/envcar                                                      @open-telemetry/go-approvers @Joibel
```
pellared: to reduce the bus factor :wink: 

```suggestion
propagators/envcar                                                      @open-telemetry/go-approvers @Joibel @pellared
```

propagators/envcar/carrier.go
Joibel: Ok, I follow this now. In what I'd written elsewhere I had written but never used a os.Setenv path, and thought I was using it.

CHANGELOG.md
```diff
@@ -14,6 +14,7 @@ This project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0.htm
 - Add support for configuring propagators in `go.opentelemetry.io/contrib/otelconf`. (#8281)
 - Add `const Version` in `go.opentelemetry.io/contrib/bridges/prometheus`. (#8401)
 - Add `const Version` in `go.opentelemetry.io/contrib/otelconf`. (#8461)
+- Add environment carrier in `go.opentelemetry.io/contrib/propagators/envcar`. (#8300)
```
```suggestion
- Add environment variables propagation carrier in `go.opentelemetry.io/contrib/propagators/envcar`. (#8442)
```

propagators/envcar/carrier.go
pellared: I think we should even add a comment in `SetEnvFunc` to discourage using `os.Setenv` :wink:

Joibel: I've done that now.

propagators/envcar/carrier.go
```diff
@@ -0,0 +1,63 @@
...
+	// Usually, you want to set the environment variable for processes
+	// that are spawned by the current process.
+	// By default implementation, it does nothing.
+	// Using os.Setenv here is discouraged as the environment should
+	// be immutable:
+	// https://opentelemetry.io/docs/specs/otel/context/env-carriers/#environment-variable-immutability
+	SetEnvFunc func(key, value string)
+}
+
+var _ propagation.TextMapCarrier = Carrier{}
```
kamphaus:
```suggestion
// Compile time check that Carrier implements the TextMapCarrier.
var _ propagation.TextMapCarrier = Carrier{}
```

Suggestion copied and adapted from https://github.com/open-telemetry/opentelemetry-go/blob/3264bf171b1e6cd70f6be4a483f2bcb84eda6ccf/propagation/propagation.go#L49

propagators/envcar/carrier_example_test.go
kamphaus: The thought behind `Once set for a process, environment variables SHOULD be treated as immutable within that process` is to ensure that once a trace context has been extracted from environment variables it will remain consistent no matter if the environment variables are changed. Ie. extracting the tracecontext from environment variables a second time should give the same tracecontext as the first time, even if env vars were changed in between.

It would be good to have a test for this:
* extract tracecontext
* change env var TRACEPARENT
* extract tracecontext & assert it matches first result

What do you think would be the best way to achieve this? storing the first extracted context globally or in the carrier?

propagators/envcar/carrier.go
kamphaus: Great to see the test for multiple goroutines.

Since you would need to set a different `SetEnvFunc` for each spawned process, creating a separate carrier makes sense:
https://github.com/Joibel/opentelemetry-go-contrib/blob/166eab6e52a9a10ba878469f98429f3b8ed18017/propagators/envcar/carrier_test.go#L209-L213

I don't see a way around having one carrier per spawned process.

If you don't want to go with `SetEnvFunc` you could use a solution similar to [MapCarrier](https://github.com/open-telemetry/opentelemetry-go/blob/3264bf171b1e6cd70f6be4a483f2bcb84eda6ccf/propagation/propagation.go#L47) and have a method `GetEnv()` return `[]string` of encoded env vars to be used as the user sees fit for `cmd.Env`.
However you would still need a dedicated Carrier per spawned process since it needs to accumulate the env vars from the `prop.Inject`.
`SetEnvFunc` is more flexible while a `GetEnv()` accumulator might be more constrained.
I'm just writing down some thoughts.

I'm fine leaving it as `SetEnvFunc`.

propagators/envcar/carrier.go
```diff
@@ -0,0 +1,64 @@
...
+// Keys lists the keys stored in this carrier.
+// This returns all the keys in the environment variables.
+func (Carrier) Keys() []string {
+	keys := make([]string, 0, len(os.Environ()))
+	for _, kv := range os.Environ() {
+		kvPair := strings.SplitN(kv, "=", 2)
+		if len(kvPair) < 1 {
+			continue
+		}
+		keys = append(keys, strings.ToLower(kvPair[0]))
```
pellared: Why are we making it lowercase?

propagators/envcar/carrier.go
```diff
@@ -0,0 +1,64 @@
...
+// This is useful for propagating values that are set in the environment
+// and need to be accessed by different processes or services.
+// The keys are uppercased to avoid case sensitivity issues across different
+// operating systems and environments.
+type Carrier struct {
+	// SetEnvFunc is a function that sets the environment variable.
+	// Usually, you want to set the environment variable for processes
+	// that are spawned by the current process.
+	// By default implementation, it does nothing.
+	// Using os.Setenv here is discouraged as the environment should
```
pellared:
```suggestion
	// Using [os.Setenv] here is discouraged as the environment should
```

propagators/envcar/carrier.go
```diff
@@ -0,0 +1,64 @@
...
+// before being used to access the environment variables.
+// This is useful for propagating values that are set in the environment
+// and need to be accessed by different processes or services.
+// The keys are uppercased to avoid case sensitivity issues across different
+// operating systems and environments.
+type Carrier struct {
+	// SetEnvFunc is a function that sets the environment variable.
+	// Usually, you want to set the environment variable for processes
+	// that are spawned by the current process.
+	// By default implementation, it does nothing.
```
pellared: The grammar in this sentence does not feel correct (I know I have written it myself :rofl: )

propagators/envcar/carrier_test.go
```diff
@@ -0,0 +1,243 @@
...
+
+func TestConcurrentChildProcesses(t *testing.T) {
+	// Test that concurrent goroutines can each spawn child processes
+	// with their own unique trace context.
+	const numGoroutines = 10
+
+	type result struct {
+		index    int
+		expected string
+		actual   string
```
pellared:
```suggestion
		want string
		got   string
```

propagators/envcar/carrier.go
```diff
@@ -0,0 +1,64 @@
...
+
+// Keys lists the keys stored in this carrier.
+// This returns all the keys in the environment variables.
+func (Carrier) Keys() []string {
+	keys := make([]string, 0, len(os.Environ()))
+	for _, kv := range os.Environ() {
+		kvPair := strings.SplitN(kv, "=", 2)
+		if len(kvPair) < 1 {
+			continue
+		}
```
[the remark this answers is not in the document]
Joibel: Removed

propagators/envcar/doc.go
```diff
@@ -0,0 +1,6 @@
+// Copyright The OpenTelemetry Authors
+// SPDX-License-Identifier: Apache-2.0
+
+// Package envcar implements the Environment Carrier specification as documented
+// here https://opentelemetry.io/docs/specs/otel/context/env-carriers/
+package envcar // import "go.opentelemetry.io/contrib/propagators/envcar"
```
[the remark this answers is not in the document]
Joibel: Done

propagators/envcar/carrier_test.go
Joibel: Done

propagators/envcar/carrier.go
Joibel: Done

CODEOWNERS
```diff
@@ -64,6 +64,7 @@ processors/minsev                                                       @open-te
 propagators/autoprop/                                                   @open-telemetry/go-approvers @MrAlias
 propagators/aws/                                                        @open-telemetry/go-approvers @akats7
 propagators/b3/                                                         @open-telemetry/go-approvers @pellared
+propagators/envcar                                                      @open-telemetry/go-approvers @Joibel @pellared
```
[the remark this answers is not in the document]
Joibel: Done, although this file has both styles

propagators/envcar/go.mod
```diff
@@ -0,0 +1,19 @@
+module go.opentelemetry.io/contrib/propagators/envcar
+
+go 1.24.0
+
+require (
+	github.com/stretchr/testify v1.11.1
+	go.opentelemetry.io/otel v1.39.0
+	go.opentelemetry.io/otel/trace v1.39.0
```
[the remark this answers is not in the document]
Joibel: Done

propagators/envcar/carrier.go
Joibel: Updated

propagators/envcar/carrier.go
Joibel: Basically for case insensitivity.
The environment variable keys are all upper cased according to the spec, but I would expect that if I inject a key of  [`traceparentHeader`](https://github.com/open-telemetry/opentelemetry-go/blob/248da958375e4dfb4a1105645107be3ef04b1c59/propagation/trace_context.go#L18) which == `traceparent`, I'd be able to put that into an environment with this carrier using `Set` which upper case it and then pull it out using `Get` on the other side.

propagators/envcar/carrier_example_test.go
Joibel: I've added this, there is now a store on first read.

propagators/envcar/carrier.go
pellared: Nothing in the spec says to do so. If the carrier stores as uppercase then I would say that this is what `Keys` should return. What if the user injects via e.g. via a `X-My-VendorKey` key. Retruning it as `x-my-vendorkey` would be even more misleading.

propagators/envcar/carrier.go
```diff
@@ -0,0 +1,76 @@
...
+// before being used to access the environment variables.
+// This is useful for propagating values that are set in the environment
+// and need to be accessed by different processes or services.
+// The keys are uppercased to avoid case sensitivity issues across different
+// operating systems and environments.
+type Carrier struct {
+	// SetEnvFunc is the function that sets the environment variable.
+	// Usually, you want to set the environment variables for processes
+	// that are spawned by the current process.
+	// If you do not set this function, `Set` will do nothing.
```
pellared:
```suggestion
	// If you do not set this function, [Carrier.Set] will do nothing.
```

Please  also check how the docs are going to be rendered by using `pkgsite`. See: https://github.com/open-telemetry/opentelemetry-go/blob/main/CONTRIBUTING.md#documentation

propagators/envcar/carrier.go
```diff
@@ -0,0 +1,76 @@
...
+		environ := os.Environ()
+		c.values = make(map[string]string, len(environ))
+		for _, kv := range environ {
+			kvPair := strings.SplitN(kv, "=", 2)
+			c.values[strings.ToLower(kvPair[0])] = kvPair[1]
+		}
+	})
+}
+
+// Get returns the value associated with the passed key.
```
pellared: This should document the behavior that the the values are only read from env vars only once and future changes are not seen by the Carrier.

propagators/envcar/carrier.go
Joibel: Thanks, not used pkgsite before. This helps a lot!

Had to move this out of the struct internal comments to get linking working

propagators/envcar/carrier.go
Joibel: Added.

propagators/envcar/carrier.go
```diff
@@ -0,0 +1,83 @@
...
+	// SetEnvFunc is the function that sets the environment variable.
+	// Usually, you want to set the environment variables for processes
+	// that are spawned by the current process.
+	SetEnvFunc func(key, value string)
+	values     map[string]string
+	once       sync.Once
+}
+
+// Compile time check that Carrier implements the TextMapCarrier.
+var _ propagation.TextMapCarrier = &Carrier{}
```
pellared: nit

```suggestion
var _ propagation.TextMapCarrier = (*Carrier)(nil)
```

propagators/envcar/carrier.go
```diff
@@ -0,0 +1,83 @@
...
+
+// Fetch runs once on first access, and stores the environment in the
+// carrier by lower cased keys..
```
pellared:
```suggestion
// fetch runs once on first access, and stores the environment in the
// carrier by lower cased keys.
```

propagators/envcar/carrier.go
pellared: Side note: this is the main comment that prevents me from approving the PR.

@adrielp, PTAL and let us know if you think my feedback is wrong.

propagators/envcar/carrier_example_test.go
kamphaus: Is there a test for this?

pellared (changes requested):
Blocking because of  https://github.com/open-telemetry/opentelemetry-go-contrib/pull/8442#discussion_r2852238808

I like the direction it is going. Nice work so far :medal_sports:

propagators/envcar/carrier.go
```diff
@@ -0,0 +1,83 @@
...
+
+// fetch runs once on first access, and stores the environment in the
+// carrier by lower cased keys.
+func (c *Carrier) fetch() {
+	c.once.Do(func() {
+		environ := os.Environ()
+		c.values = make(map[string]string, len(environ))
+		for _, kv := range environ {
+			kvPair := strings.SplitN(kv, "=", 2)
+			c.values[strings.ToLower(kvPair[0])] = kvPair[1]
```
pellared: so that `Keys` returns exactly what is stored in the carrier

```suggestion
			c.values[kvPair[0]] = kvPair[1]
```

propagators/envcar/carrier.go
```diff
@@ -0,0 +1,83 @@
...
+	})
+}
+
+// Get returns the value associated with the passed key.
+// The first call to [Carrier.Get] or [Carrier.Keys] for a
+// given Carrier will read and store the values from the
+// environment and all future reads will be from that store.
+func (c *Carrier) Get(key string) string {
+	c.fetch()
+	return c.values[strings.ToLower(key)]
```
pellared: for consistently with `Set`

```suggestion
	return c.values[strings.ToUpper(key)]
```

I think this is also more inline with

> SHOULD use uppercase letters, digits, and underscores for maximum cross-platform compatibility

propagators/envcar/carrier.go
```diff
@@ -0,0 +1,83 @@
...
+}
+
+// Set stores the key-value pair in the environment variable.
+// The key is uppercased before being used to set the environment variable.
+// If SetEnvFunc is not set, this method does nothing.
+func (c *Carrier) Set(key, value string) {
+	if c.SetEnvFunc == nil {
+		return
+	}
+	k := strings.ToUpper(key)
```
pellared: From https://opentelemetry.io/docs/specs/otel/context/env-carriers/#name-restrictions

> MUST NOT include characters forbidden in environment variables per platform-specific restrictions

We do not meet this criteria. My proposal below should address it. 

> SHOULD use uppercase letters, digits, and underscores for maximum cross-platform compatibility

@adrielp, is there any reason why we are not more specific here?

Side nite: the spec should call out that these are ASCII letters and digits :wink: You do not want e.g. Japanese or Arabic letters and digits.

My proposal is to
- Leave `[A-Z][0-9]_` as it is
- `a-z` make uppercase
- replace all other characters with `_`

If this is acceptable then it should be defined in the spec.

propagators/envcar/carrier.go
Joibel: Yeah, happy with doing this. It makes sense.

propagators/envcar/carrier.go
Joibel: Yeah, this makes sense. I'll await Adriel's response.

propagators/envcar/carrier.go
Joibel: FYI: this has been changed now to store and return upper keys always.

propagators/envcar/carrier.go
```diff
@@ -0,0 +1,83 @@
...
+	SetEnvFunc func(key, value string)
+	values     map[string]string
+	once       sync.Once
+}
+
+// Compile time check that Carrier implements the TextMapCarrier.
+var _ propagation.TextMapCarrier = (*Carrier)(nil)
+
+// fetch runs once on first access, and stores the environment in the
+// carrier by lower cased keys.
```
pellared:
```suggestion
// carrier.
```

propagators/envcar/carrier.go
adrielp: Environment variables when In the environment or being set to the environment, should be uppercase `_` separated where applicable. With the carrier approach via TextMapPropagator, my understanding is that once the environment variable is read and normalized to lower case, it'll only be considered valid if the key name matches the header name for w/e propagator. See https://github.com/open-telemetry/opentelemetry-java/pull/8074#issuecomment-3955005909 for related conversation.

pellared: @adrielp, to make it clear. I do not want to limit the key to work only `traceparent`, `tracestate`, `baggage`. I want to provide some universal solution that can work for a bigger range of propagation formats. The technically the `key` can be any UTF-8 string and I want to define what happens if one uses e.g. `Mój Bagaż` as key, I think in such cases we can allow normalizing using the following rules

1. Leave [A-Z][0-9]_ as it is
2. [a-z] are uppercased
3. Replace all other characters with _

Then `Mój Bagaż` can be normalized to `M_J_BAGA_` - strange but it should be able work in most scenarios :sweat_smile: 


Note that OTel Java also does normalization (which they improperly call sanitization and covers less cases):
- https://github.com/open-telemetry/opentelemetry-java/blob/0ad755b00142956f5907c9c2aebed82202b50a52/api/incubator/src/main/java/io/opentelemetry/api/incubator/propagation/EnvironmentGetter.java#L80
- https://github.com/open-telemetry/opentelemetry-java/blob/0ad755b00142956f5907c9c2aebed82202b50a52/api/incubator/src/main/java/io/opentelemetry/api/incubator/propagation/EnvironmentSetter.java#L78

I can propose a spec PR if you think that it makes sense. This should also be inline with 

> Implementations MAY support additional propagation formats and SHOULD provide configuration options to override the default environment variable.

CC @Bhagirath00 @jack-berg @jkwatson

Joibel: I was hoping to catch up with Adriel today at the CICD semconv meeting, but he couldn't make it.

Can we refine the spec later and get this one merged if you're happy with it in other ways?

pellared: Let's do it the way I propose and we can work on the spec later.

> My proposal is to
> - Leave `[A-Z][0-9]_` as it is
> - `a-z` make uppercase
> - replace all other characters with `_`

We can also make refinements here later as well :wink:

Joibel: Ok, done.

propagators/envcar/carrier.go
```diff
@@ -0,0 +1,101 @@
...
+// The first call to [Carrier.Get] or [Carrier.Keys] for a
+// given Carrier will read and store the values from the
+// environment and all future reads will be from that store.
+func (c *Carrier) Get(key string) string {
+	c.fetch()
+	return c.values[upperWithUnderscores(key)]
+}
+
+// Set stores the key-value pair in the environment variable.
+// The key is uppercased before being used to set the environment variable.
```
pellared: Please update this. Maybe describe it in a way like "The key is normalized by ...".

propagators/envcar/carrier.go
```diff
@@ -0,0 +1,101 @@
...
+	k := upperWithUnderscores(key)
+	c.SetEnvFunc(k, value)
+}
+
+// Keys lists the keys stored in this carrier.
+// This returns all the keys in the environment variables.
+// The first call to [Carrier.Get] or [Carrier.Keys] for a
+// given Carrier will read and store the values from the
+// environment and all future reads will be from that store.
+func (c *Carrier) Keys() []string {
```
pellared, line 96: For clarity, I suggest explicitly stating that keys are currently returned without normalization. We may revisit this in the future, as I’m increasingly leaning toward having Keys return normalized values as well.

That said, I don’t want this to block the current PR. We can note that this behavior is subject to change and address it in a follow-up if needed.

propagators/envcar/carrier.go
```diff
@@ -0,0 +1,101 @@
...
+		default:
+			b.WriteByte('_')
+		}
+	}
+	return b.String()
+}
+
+// Carrier is a TextMapCarrier that uses the environment variables as a
+// storage medium for propagated key-value pairs. The keys are uppercased
+// before being used to access the environment variables.
```
pellared: Please update this. Maybe describe it in a way like "The key is normalized by ..."

propagators/envcar/carrier.go
```diff
@@ -0,0 +1,101 @@
...
+// as-is, a-z is uppercased, and all other characters are replaced with _.
+func upperWithUnderscores(s string) string {
+	var b strings.Builder
+	b.Grow(len(s))
+	for _, r := range s {
+		switch {
+		case r >= 'A' && r <= 'Z', r >= '0' && r <= '9', r == '_':
+			b.WriteByte(byte(r))
+		case r >= 'a' && r <= 'z':
+			b.WriteByte(byte(r - 'a' + 'A'))
```
pellared: (nit) This reads better for me. Feel free to disagree and ignore my suggestion if you think otherwise as this is subjective. 

```suggestion
			b.WriteByte(byte(r + 'A' - 'a'))
```

pellared (commented):
Few nit comments.

pellared:
@kamphaus, @adrielp, can you please take another look?

propagators/envcar/carrier.go
Joibel: Fine with that, no strong opinion here.

propagators/envcar/carrier.go
Joibel: Done

propagators/envcar/carrier.go
Joibel: Done

propagators/envcar/carrier.go
Joibel: I've changed the comments.

propagators/envcar/carrier.go
```diff
@@ -0,0 +1,103 @@
...
+// The first call to [Carrier.Get] or [Carrier.Keys] for a
+// given Carrier will read and store the values from the
+// environment and all future reads will be from that store.
+func (c *Carrier) Get(key string) string {
+	c.fetch()
+	return c.values[upperWithUnderscores(key)]
+}
+
+// Set stores the key-value pair in the environment variable.
+// The key is normalized by upperWithUnderscored before being
```
pellared: I think it is enough just to document the normalization in documentation for `Carrier` type.

```suggestion
// The key is normalized before being
```

propagators/envcar/carrier.go
```diff
@@ -0,0 +1,103 @@
...
+		c.values = make(map[string]string, len(environ))
+		for _, kv := range environ {
+			kvPair := strings.SplitN(kv, "=", 2)
+			c.values[kvPair[0]] = kvPair[1]
+		}
+	})
+}
+
+// Get returns the value associated with the passed key.
+// The first call to [Carrier.Get] or [Carrier.Keys] for a
```
pellared, line 69: We should document that the key is normalized.

propagators/envcar/carrier.go
```diff
@@ -0,0 +1,103 @@
...
+		default:
+			b = append(b, '_')
+		}
+	}
+	return string(b)
+}
+
+// Carrier is a TextMapCarrier that uses the environment variables as a
+// storage medium for propagated key-value pairs. The keys are normalised by
+// upperWithUnderscores before being used to access the environment variables.
```
pellared: Readers of https://pkg.go.dev/ will not know what `upperWithUnderscores` is.

propagators/envcar/carrier.go
Joibel: Done

propagators/envcar/carrier.go
Joibel: Added

propagators/envcar/carrier.go
Joibel: Makes sense, removed all references from the docs.

propagators/envcar/carrier.go
```diff
@@ -0,0 +1,103 @@
...
+	once       sync.Once
+}
+
+// Compile time check that Carrier implements the TextMapCarrier.
+var _ propagation.TextMapCarrier = (*Carrier)(nil)
+
+// fetch runs once on first access, and stores the environment in the
+// carrier.
+func (c *Carrier) fetch() {
+	c.once.Do(func() {
```
kamphaus, line 58: LGTM

One test I would add is to verify this fetch once functionality (but not blocking this PR):
```
func TestCarrierGetFetchOnce(t *testing.T) {
	t.Setenv("TRACEPARENT", "myvalue")
	c := envcar.Carrier{}
	assert.Equal(t, "myvalue", c.Get("TRACEPARENT"))
	t.Setenv("TRACEPARENT", "bad")
	// Assert a carrier instance reads the env vars only once:
	assert.Equal(t, "myvalue", c.Get("TRACEPARENT"))
}
```
(I haven't tried running this test code)

pellared: @Joibel, would you like to address it in this PR or do you want to do it in a separate one? Looking quickly I think the proposed test would work. I would just change the first `assert.Equal` into `require.Equal` :wink:

Joibel: I've added this, and tested `Keys` in the same test.

dmathieu:
I've invited you to the @open-telemetry/go-contributors team. This is required to let GH actions auto-request reviews from PRs opened against this module.

propagators/envcar/carrier_test.go
```diff
@@ -0,0 +1,265 @@
...
+
+	// Verify each goroutine's child process received the correct trace context.
+	for r := range results {
+		require.NoError(t, r.err, "goroutine %d failed to run child process", r.index)
+		assert.Equal(t, r.want, r.got,
+			"goroutine %d: child process received wrong trace context", r.index)
+	}
+}
+
+// Ensure Get and Keys to fetch from cache
```
dmathieu:
```suggestion
// Ensure Get and Keys to fetch from cache.
```

This should fix the linter.

Issue #8006, opened 2025-10-09
Retiring samplers/probability/consistent as obsolete


jmacd:
OpenTelemetry has new specifications that make the experimental sampler here:

https://github.com/open-telemetry/opentelemetry-go-contrib/tree/main/samplers/probability/consistent

obsolete. https://github.com/open-telemetry/opentelemetry-specification/pull/4673 removes the experimental specification it was based on.

https://github.com/open-telemetry/opentelemetry-go/issues/7485 describes the new specification that replaces this.

dmathieu:
For whoever would like to take this on, what's needed is to deprecate the [samplers/probability/consistent](https://github.com/open-telemetry/opentelemetry-go-contrib/tree/main/samplers/probability/consistent) package, then wait for one release before removing it from the codebase.

petern48:
I was about to pick this up, but then realized it looks like what's replacing this hasn't been implemented, to my understanding. I would've liked to include a message like "use ___ instead", but I don't see anything to fill in that blank at the moment. Should we put a hold on this until the new spec is implemented, or is there already a confirmed location in mind for it?

Ebitimiebelike:
Hi! I’d like to take on this issue. 
I understand the task is to deprecate the samplers/probability/consistent package 
and wait for one release before removal. 
Please let me know if this is okay or if there’s anything I should be aware of before starting.

sonalgaud12:
The deprecation comment belongs at the package level, not above every function. right? @dmathieu

dmathieu:
I think before we deprecate, we should answer the question above, as to what is the replacement.

sonalgaud12:
Make sense, I have commented on the issue listed above incase of any help I could provide for the same.

anarwal:
Hello, Does this issue still needs help? I can work on it if work is still needed.

PR #8447, opened 2026-01-22
samplers/probability/consistent: Deprecate package

sonalgaud12:
Deprecated the [samplers/probability/consistent](https://github.com/open-telemetry/opentelemetry-go-contrib/tree/main/samplers/probability/consistent) package

Related: #8006

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -47,7 +47,7 @@ This project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0.htm
 
 ### Deprecated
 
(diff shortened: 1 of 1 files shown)


sonalgaud12:
@dmathieu I was confused whether to add a deprecated comment in all the func cause in this case, we are trying to deprecate the whole package, not just a function. So tell me know what else to add or remove

dashpole:
I think you want to deprecate the module: https://go.dev/ref/mod#go-mod-file-module-deprecation and the package, but not all of the artifacts in it

PR #8456, opened 2026-01-22
otelgin: update error handling to set attributes instead of recording errors

### LINKED ISSUE #8441: otelgin: update error handling to set attributes instead of recording errors
Adjust position: https://github.com/open-telemetry/opentelemetry-go-contrib/blob/9bd144ac077475799793d0bad6e19f50d1aab5e1/instrumentation/github.com/gin-gonic/gin/otelgin/gin.go#L124C10-L124C21
Reference implementation: https://github.com/open-telemetry/opentelemetry-go-contrib/pull/8386

_Originally posted by @pellared in https://github.com/open-telemetry/opentelemetry-go-contrib/pull/8408#discussion_r2709096235_

Eshaan-byte:
## Changes

- `gin.go`: 
  - Added import for `go.opentelemetry.io/otel/semconv/v1.37.0`
  - Replaced `span.RecordError()` loop with `span.SetAttributes(otelsemconv.ErrorType(c.Errors[0].Err))`
  - Set `error.type` attribute for the first error when `c.Errors` is not empty
  - Multiple errors are still captured in the status description via `c.Errors.String()`

- `gin_test.go`:
  - Added import for `go.opentelemetry.io/otel/semconv/v1.37.0`
  - Updated `TestError` to verify `error.type` attribute instead of span events
  - Updated `TestSpanStatus` subtest to verify `error.type` attribute instead of span events

## Rationale

According to the [instrumentation README guidelines](https://github.com/open-telemetry/opentelemetry-go-contrib/blob/main/instrumentation/README.md#recording-errors), when an instrumented operation returns a non-nil error:

1. MUST set the span status to `codes.Error` with a description
2. SHOULD also set the `error.type` attribute
3. SHOULD NOT use `span.RecordError(err)` for this purpose

This change provides better consistency across OpenTelemetry instrumentation packages and and follows the semantic conventions for error handling.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -47,6 +47,7 @@ This project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0.htm
 - The `Version()` function in `go.opentelemetry.io/contrib/instrumentation/github.com/gin-gonic/gin/otelgin` has been replaced by `const Version`. (#8341)
 - Set `error.type` attribute instead of adding `exception` span events in `go.opentelemetry.io/contrib/instrumentation/net/http/otelhttp`. (#8386)
 - Set `error.type` attribute instead of adding `exception` span events in `go.opentelemetry.io/contrib/instrumentation/github.com/aws/aws-sdk-go-v2/otelaws`. (#8386)
+- Set `error.type` attribute instead of adding `exception` span events in `go.opentelemetry.io/contrib/instrumentation/github.com/gin-gonic/gin/otelgin`. (#8441)
 - The `Version()` function in `go.opentelemetry.io/contrib/instrumentation/go.mongodb.org/mongo-driver/mongo/otelmongo` hae been replaced by `const Version`. (#8340)
 - The `Version()` function in `go.opentelemetry.io/contrib/instrumentation/google.golang.org/grpc/otelgrpc` has been replaced by `const Version`. (#8317)
 
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -47,7 +47,7 @@ This project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0.htm
 - The `Version()` function in `go.opentelemetry.io/contrib/instrumentation/github.com/gin-gonic/gin/otelgin` has been replaced by `const Version`. (#8341)
 - Set `error.type` attribute instead of adding `exception` span events in `go.opentelemetry.io/contrib/instrumentation/net/http/otelhttp`. (#8386)
 - Set `error.type` attribute instead of adding `exception` span events in `go.opentelemetry.io/contrib/instrumentation/github.com/aws/aws-sdk-go-v2/otelaws`. (#8386)
-- Set `error.type` attribute instead of adding `exception` span events in `go.opentelemetry.io/contrib/instrumentation/github.com/gin-gonic/gin/otelgin`. (#8441)
+- Set `error.type` attribute instead of adding `exception` span events in `go.opentelemetry.io/contrib/instrumentation/github.com/gin-gonic/gin/otelgin`. (#8456)
 - The `Version()` function in `go.opentelemetry.io/contrib/instrumentation/go.mongodb.org/mongo-driver/mongo/otelmongo` hae been replaced by `const Version`. (#8340)
 - The `Version()` function in `go.opentelemetry.io/contrib/instrumentation/google.golang.org/grpc/otelgrpc` has been replaced by `const Version`. (#8317)
 
--- a/instrumentation/github.com/gin-gonic/gin/otelgin/gin.go
+++ b/instrumentation/github.com/gin-gonic/gin/otelgin/gin.go
@@ -15,9 +15,10 @@ import (
 	"go.opentelemetry.io/otel/codes"
 	"go.opentelemetry.io/otel/metric"
 	"go.opentelemetry.io/otel/propagation"
+	semconv "go.opentelemetry.io/otel/semconv/v1.37.0"
 	oteltrace "go.opentelemetry.io/otel/trace"
 
-	"go.opentelemetry.io/contrib/instrumentation/github.com/gin-gonic/gin/otelgin/internal/semconv"
+	intsemconv "go.opentelemetry.io/contrib/instrumentation/github.com/gin-gonic/gin/otelgin/internal/semconv"
 )
 
 const (
@@ -57,7 +58,7 @@ func Middleware(service string, opts ...Option) gin.HandlerFunc {
 		metric.WithInstrumentationVersion(Version),
 	)
 
-	sc := semconv.NewHTTPServer(meter)
+	sc := intsemconv.NewHTTPServer(meter)
 
 	return func(c *gin.Context) {
 		requestStartTime := time.Now()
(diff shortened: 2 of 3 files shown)


dmathieu:
This needs a changelog entry.

instrumentation/github.com/gin-gonic/gin/otelgin/gin.go
```diff
@@ -15,6 +15,7 @@ import (
 	"go.opentelemetry.io/otel/codes"
 	"go.opentelemetry.io/otel/metric"
 	"go.opentelemetry.io/otel/propagation"
+	otelsemconv "go.opentelemetry.io/otel/semconv/v1.37.0"
```
dmathieu: What semconv other than otel is there? We can require this as just `semconv`.


```suggestion
	"go.opentelemetry.io/otel/semconv/v1.37.0"
```

instrumentation/github.com/gin-gonic/gin/otelgin/gin_test.go
```diff
@@ -202,15 +203,9 @@ func TestError(t *testing.T) {
...
-	// verify the error events
-	events := span.Events()
-	require.Len(t, events, 2)
-	assert.Equal(t, "exception", events[0].Name)
-	assert.Contains(t, events[0].Attributes, attribute.String("exception.type", "*errors.errorString"))
-	assert.Contains(t, events[0].Attributes, attribute.String("exception.message", "oh no one"))
-	assert.Equal(t, "exception", events[1].Name)
-	assert.Contains(t, events[1].Attributes, attribute.String("exception.type", "*errors.errorString"))
-	assert.Contains(t, events[1].Attributes, attribute.String("exception.message", "oh no two"))
+	// verify the error.type attribute is set (using first error)
```
dmathieu: Those LLM comments provide nothing here.

instrumentation/github.com/gin-gonic/gin/otelgin/gin.go
```diff
@@ -120,9 +121,9 @@ func Middleware(service string, opts ...Option) gin.HandlerFunc {
 
 		if len(c.Errors) > 0 {
 			span.SetStatus(codes.Error, c.Errors.String())
-			for _, err := range c.Errors {
-				span.RecordError(err.Err)
-			}
+			// Set error.type attribute for the first error
+			// Multiple errors are already captured in the status description via c.Errors.String()
```
dmathieu:
```suggestion
			// Multiple errors are already captured in the status description via c.Errors.String()
```

While explaining why we only keep the first error's type, describing what a single-line call does is an LLM thing that really decreases readability.

CHANGELOG.md
```diff
@@ -46,6 +46,7 @@ This project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0.htm
 - The `Version()` function in `go.opentelemetry.io/contrib/instrumentation/github.com/gin-gonic/gin/otelgin` has been replaced by `const Version`. (#8341)
 - Set `error.type` attribute instead of adding `exception` span events in `go.opentelemetry.io/contrib/instrumentation/net/http/otelhttp`. (#8386)
 - Set `error.type` attribute instead of adding `exception` span events in `go.opentelemetry.io/contrib/instrumentation/github.com/aws/aws-sdk-go-v2/otelaws`. (#8386)
+- Set `error.type` attribute instead of adding `exception` span events in `go.opentelemetry.io/contrib/instrumentation/github.com/gin-gonic/gin/otelgin`. (#8441)
```
```suggestion
- Set `error.type` attribute instead of adding `exception` span events in `go.opentelemetry.io/contrib/instrumentation/github.com/gin-gonic/gin/otelgin`. (#8456)
```

dmathieu (approved):

instrumentation/github.com/gin-gonic/gin/otelgin/gin.go
```diff
@@ -113,16 +114,15 @@ func Middleware(service string, opts ...Option) gin.HandlerFunc {
...
 			WriteBytes: int64(c.Writer.Size()),
 		})...)
 
 		if len(c.Errors) > 0 {
 			span.SetStatus(codes.Error, c.Errors.String())
-			for _, err := range c.Errors {
-				span.RecordError(err.Err)
-			}
+			// Multiple errors are already captured in the status description via c.Errors.String()
+			span.SetAttributes(semconv.ErrorType(c.Errors[0].Err))
```
flc1125, line 125: This only records the type of the first error, but if there are multiple errors, some information may be hidden or ignored.

If we later wrap them using `errors.Join`, would it be a better choice when retrieving the types?

dmathieu: Semantic conventions don't allow joining multiple error types in this attribute.
https://github.com/open-telemetry/semantic-conventions/blob/c3be6481868d4e2ba330af4c0110026dc58b90e6/docs/registry/attributes/error.md#error-type

flc1125: I tested locally the output when using `errors.join`, I'm not quite sure if this is joining multiple error types.

```shell
(attribute.KeyValue) {
 Key: (attribute.Key) (len=10) "error.type",
 Value: (attribute.Value) {
  vtype: (attribute.Type) STRING,
  numeric: (uint64) 0,
  stringly: (string) (len=17) "*errors.joinError",
  slice: (interface {}) <nil>
 }
}
```

dmathieu: I'm not sure having a type of `*errors.joinError` is better than having the first error.

pellared: We would need a custom logic to join (e.g. with `, ` ) all `c.Errors` types.

dmathieu: Yes, but the behavior would need to be documented in semantic conventions, so we don't have different behavior for different languages.

pellared:

dmathieu:
Yes, this is still valid.

instrumentation/github.com/gin-gonic/gin/otelgin/gin.go
flc1125: ~~I agree with this approach: https://github.com/open-telemetry/opentelemetry-go-contrib/pull/8456#discussion_r2735732555~~

~~@Eshaan-byte, can you continue to solve it?~~

Please ignore, since it is to obtain `ErrorType`, in fact, combining errors into one or keeping them separate yields the same result, so the current situation can be maintained.

flc1125:
@Eshaan-byte Also please resolve the conflict.

flc1125 (approved):
Only resolve conflicts

pellared (approved):
Approving (assuming conflict is correctly resolved)

instrumentation/github.com/gin-gonic/gin/otelgin/gin.go
pellared: I prefer joining for multiple errors. Same comment here: https://github.com/open-telemetry/opentelemetry-go-contrib/pull/8977#discussion_r3268043313

pellared:
Closing this PR as it is abandoned

PR #8457, opened 2026-01-22
support github.com/labstack/echo/v5

vadimi:
This PR is based on https://github.com/open-telemetry/opentelemetry-go-contrib/pull/8434, but as stated in comments there are breaking changes in echo/v5 and it also requires go 1.25, so creating `go.opentelemetry.io/contrib/instrumentation/github.com/labstack/echo/v5/otelecho` to make sure echo/v4 instrumentation still works.


flc1125:
Thank you for your contribution. It is a duplicate of https://github.com/open-telemetry/opentelemetry-go-contrib/pull/8434. Considering the principle of priority, we will use the other PR as the baseline, unless no one follows up on that one.

dmathieu:
As this PR does it "the right way" (with a new component) contrarily to the other one, I'd be tempted to take this one as canonical.

This brings the question of [new components](https://github.com/open-telemetry/opentelemetry-go-contrib/blob/main/CONTRIBUTING.md#new-component).
Why should this new component be in this repo? Could it be in a repo owned by echo?
Will you be listed as a codeowner for this repository (with all that entitles, as documented in the contributing document)?

vadimi:
> Why should this new component be in this repo? Could it be in a repo owned by echo?

I've asked echo maintainer if it would makes sense to add the new otel package to one of the echo repos (https://github.com/labstack/echo-contrib/issues/135)

vadimi:
I think we can close this now, there is a new package in labstack org with open telemetry support:

https://github.com/labstack/echo-opentelemetry

PR #8461, opened 2026-01-24
Add version file for otelconf module

ElenaZvereva:
Adds `version.go` for `otelconf`
Part of #8272

### DIFF
--- a/otelconf/version.go
+++ b/otelconf/version.go
@@ -4,4 +4,4 @@
 package otelconf // import "go.opentelemetry.io/contrib/otelconf"
 
 // Version is the current release version of the otelconf module.
-const Version = "0.64.0"
+const Version = "0.19.0"


otelconf/version.go
```diff
@@ -0,0 +1,7 @@
+// Copyright The OpenTelemetry Authors
+// SPDX-License-Identifier: Apache-2.0
+
+package otelconf // import "go.opentelemetry.io/contrib/otelconf"
+
+// Version is the current release version of the otelconf module.
+const Version = "0.64.0"
```
flc1125: We need to refer to the values here: https://github.com/open-telemetry/opentelemetry-go-contrib/blob/f361d96bffd40da24bf574f5f1a03e454cd052ba/versions.yaml#L58-L61

MrAlias: That seems redundant. The `version.go` file is in the standard, top-level, location. The multimod tooling should identify it and manage the version without adding it to the version configuration file.

Why does it need to be added?

dmathieu: I think @flc1125 means the version number is wrong.

```suggestion
const Version = "0.19.0"
```

MrAlias: Oh, I gotcha. Makes sense :+1:

PR #8469, opened 2026-01-28
fix: correct function name and pretty print for all exporters

pellared:


pellared:
Merging as trivial. Needed for https://github.com/open-telemetry/opentelemetry.io/pull/9052

PR #8509, opened 2026-02-05
google.golang.org/grpc/otelgrpc: Fix for `elapsedTime` records an incorrect metric value

### LINKED ISSUE #8518: otelgrpc: duration metrics record milliseconds but declare unit as seconds
### Description

In v0.65.0, the gRPC duration metrics were renamed from `rpc.server.duration` / `rpc.client.duration` to `rpc.server.call.duration` / `rpc.client.call.duration`, and the declared unit was changed from `"ms"` to `"s"` (following the updated semantic conventions).

However, the actual value computation in `stats_handler.go` was **not updated** to match. It still divides by `time.Millisecond`:

```go
// stats_handler.go, in handleRPC(), case *stats.End:
elapsedTime := float64(rs.EndTime.Sub(rs.BeginTime)) / float64(time.Millisecond)
duration.Record(ctx, elapsedTime, recordOpts...)
```

Meanwhile the metric instrument declares its unit as seconds:

```go
// semconv/v1.39.0/rpcconv/metric.go
var newServerCallDurationOpts = []metric.Float64HistogramOption{
	metric.WithDescription("Measures the duration of inbound remote procedure calls (RPC)."),
	metric.WithUnit("s"),
}
```

This means all recorded duration values are **1000x too large** relative to the declared unit. A 100ms RPC records `100.0` into a metric that claims its unit is seconds.

### Expected behavior

The computation should divide by `time.Second` to produce values in seconds matching the declared unit:

```go
elapsedTime := float64(rs.EndTime.Sub(rs.BeginTime)) / float64(time.Second)
```

### Affected metrics

- `rpc.server.call.duration`
- `rpc.client.call.duration`

### Version

Introduced in v0.65.0. Also present on `main` as of 2026-02-06.

### Related issues

- #4547 (fixed nanoseconds → milliseconds)
- #4590 (tracked ms → s conversion, closed as stale without being implemented)

gaiaz-iusipov:
Fix `elapsedTime` being recorded as (ms * 1000) in `google.golang.org/grpc/otelgrpc`. For example, a value of 5ms is written as 5s instead of 0.005s.

Fix #8518

### DIFF
--- a/tools/go.mod
+++ b/tools/go.mod
@@ -91,7 +91,7 @@ require (
 	github.com/go-critic/go-critic v0.14.3 // indirect
 	github.com/go-git/gcfg v1.5.1-0.20230307220236-3a3c6141e376 // indirect
 	github.com/go-git/go-billy/v5 v5.7.0 // indirect
-	github.com/go-git/go-git/v5 v5.16.4 // indirect
+	github.com/go-git/go-git/v5 v5.16.5 // indirect
 	github.com/go-toolsmith/astcast v1.1.0 // indirect
 	github.com/go-toolsmith/astcopy v1.1.0 // indirect
 	github.com/go-toolsmith/astequal v1.2.0 // indirect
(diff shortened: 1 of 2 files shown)


dmathieu:
Please keep your changes atomic (a single thing per PR).

gaiaz-iusipov:
I will remove from this PR the changes that are not related to the gRPC tooling.

dmathieu:
For the record, this dates to https://github.com/open-telemetry/opentelemetry-go-contrib/pull/4548, before the stabilization of RPC attributes, and when the duration attribute was actually ms.
The semantic conventions upgrade missed the unit change.

CHANGELOG.md
```diff
@@ -14,6 +14,7 @@ This project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0.htm
 
 ### Fixed
 
+- Fixed `rpc.server.call.duration` being recorded 1000× too high. (#8509)
```
```suggestion
- Change the `rpc.server.call.duration` metric value from milliseconds to seconds in `go.opentelemetry.io/contrib/instrumentation/google.golang.org/grpc/otelgrpc`. (#8509)
```

gaiaz-iusipov: I changed the description because it seemed non-obvious to me, since both the metric name and its units indicate seconds, and only the value itself is incorrect.

dmathieu: 100x too high is also quite unclear.

PR #8511, opened 2026-02-05
internal/shared/semconv: Reduce allocations

gaiaz-iusipov:
```
$ go run golang.org/x/perf/cmd/benchstat@latest old.txt new.txt
goos: darwin
goarch: arm64
pkg: go.opentelemetry.io/contrib/instrumentation/net/http/otelhttp
cpu: Apple M3 Max
              │   old.txt   │              new.txt               │
              │   sec/op    │   sec/op     vs base               │
_RoundTrip-14   2.099µ ± 1%   2.026µ ± 4%  -3.48% (p=0.018 n=10)

              │   old.txt    │               new.txt               │
              │     B/op     │     B/op      vs base               │
_RoundTrip-14   8.442Ki ± 0%   7.963Ki ± 0%  -5.68% (p=0.000 n=10)

              │  old.txt   │              new.txt               │
              │ allocs/op  │ allocs/op   vs base                │
_RoundTrip-14   39.00 ± 0%   35.00 ± 0%  -10.26% (p=0.000 n=10)
```


dmathieu:
We have removed the duplicate metrics. So I think we could remove even more allocations by removing the `map[string]MetricOpts` and replacing it with a `MetricOpts`.

gaiaz-iusipov:
> We have removed the duplicate metrics. So I think we could remove even more allocations by removing the `map[string]MetricOpts` and replacing it with a `MetricOpts`.

Great. I didn’t initially make that change because it broke the function signature, but this option is, of course, much more preferable.

dmathieu:
We can change internal module's method signature, as long as we don't change the public ones.

CHANGELOG.md
```diff
@@ -17,6 +17,16 @@ This project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0.htm
...
 
+### Changed
+
+- Improve performance by reducing allocations in the modules below. (#8511)
+  - `go.opentelemetry.io/contrib/instrumentation/github.com/emicklei/go-restful/otelrestful`
+  - `go.opentelemetry.io/contrib/instrumentation/github.com/gin-gonic/gin/otelgin`
+  - `go.opentelemetry.io/contrib/instrumentation/github.com/gorilla/mux/otelmux`
+  - `go.opentelemetry.io/contrib/instrumentation/github.com/labstack/echo/otelecho`
+  - `go.opentelemetry.io/contrib/instrumentation/net/http/httptrace/otelhttptrace`
+  - `go.opentelemetry.io/contrib/instrumentation/net/http/otelhttp`
```
```suggestion
- Reduce allocations in the generated HTTP `internal/semconv` packages. (#8511)
  Updated modules include:
  - `go.opentelemetry.io/contrib/instrumentation/github.com/emicklei/go-restful/otelrestful`
  - `go.opentelemetry.io/contrib/instrumentation/github.com/gin-gonic/gin/otelgin`
  - `go.opentelemetry.io/contrib/instrumentation/github.com/gorilla/mux/otelmux`
  - `go.opentelemetry.io/contrib/instrumentation/github.com/labstack/echo/otelecho`
  - `go.opentelemetry.io/contrib/instrumentation/net/http/httptrace/otelhttptrace`
  - `go.opentelemetry.io/contrib/instrumentation/net/http/otelhttp`
```

flc1125:
Could you add some before-and-after performance comparison data? We can use [benchstat](https://pkg.go.dev/golang.org/x/perf/cmd/benchstat) to provide it.

gaiaz-iusipov:
If we compare the variants where the slice is allocated only once, then each call results in exactly one fewer allocation. I decided not to account for removing the map wrapper in the benchmark, I think that part is fairly obvious:

```go
func (n HTTPClient) RecordMetricsOld(ctx context.Context, md MetricData, opts MetricOpts) {
	n.requestBodySize.Inst().Record(ctx, md.RequestSize, opts.MeasurementOption())
	n.requestDuration.Inst().Record(ctx, md.ElapsedTime/1000, opts.MeasurementOption())
}

func (n HTTPClient) RecordMetricsNew(ctx context.Context, md MetricData, opts MetricOpts) {
	o := []metric.RecordOption{opts.MeasurementOption()}
	n.requestBodySize.Inst().Record(ctx, md.RequestSize, o...)
	n.requestDuration.Inst().Record(ctx, md.ElapsedTime, o...)
}
```
Here are the results:
```
go test . -bench=^Benchmark_RecordMetrics$ -run=^$ -benchmem -benchtime=10s
Benchmark_RecordMetrics/old-14         	444416276	        26.78 ns/op	      32 B/op	       2 allocs/op
Benchmark_RecordMetrics/new-14         	771929202	        15.49 ns/op	      16 B/op	       1 allocs/op
```

internal/shared/semconv/client.go.tmpl
```diff
@@ -247,22 +247,20 @@ func (o MetricOpts) AddOptions() metric.AddOption {
...
 	}
-
-	return opts
 }
 
-func (n HTTPClient) RecordMetrics(ctx context.Context, md MetricData, opts map[string]MetricOpts) {
-	n.requestBodySize.Inst().Record(ctx, md.RequestSize, opts["new"].MeasurementOption())
-	n.requestDuration.Inst().Record(ctx, md.ElapsedTime/1000, opts["new"].MeasurementOption())
+func (n HTTPClient) RecordMetrics(ctx context.Context, md MetricData, opts MetricOpts) {
+	o := []metric.RecordOption{opts.MeasurementOption()}
```
flc1125: If we use pool here, the performance improvement will be even greater.

```
❯ benchstat old.txt new.txt 
goos: darwin
goarch: arm64
pkg: go.opentelemetry.io/contrib/instrumentation/github.com/emicklei/go-restful/otelrestful/internal/semconv
cpu: Apple M3
                           │   old.txt   │               new.txt               │
                           │   sec/op    │   sec/op     vs base                │
HTTPClient_RecordMetrics-8   14.81n ± 0%   11.36n ± 1%  -23.32% (p=0.000 n=10)

                           │  old.txt   │              new.txt               │
                           │    B/op    │   B/op     vs base                 │
HTTPClient_RecordMetrics-8   16.00 ± 0%   0.00 ± 0%  -100.00% (p=0.000 n=10)

                           │  old.txt   │               new.txt               │
                           │ allocs/op  │ allocs/op   vs base                 │
HTTPClient_RecordMetrics-8   1.000 ± 0%   0.000 ± 0%  -100.00% (p=0.000 n=10)
```

gaiaz-iusipov: That’s a great point, I’ve added `sync.Pool`. @dmathieu please review once again.

pellared:
Can you please add results from [`benchstat`](https://pkg.go.dev/golang.org/x/perf/cmd/benchstat) to the PR description?

Issue #5428, opened 2024-04-25
Use a `kvBuffer` in logrus bridge


dmathieu:
Just a remark that we could use a `kvBuffer` like in `otelslog` but I prefer to do it as it is currently and focus on performance improvements in separate PRs.

_Originally posted by @pellared in https://github.com/open-telemetry/opentelemetry-go-contrib/pull/5356#discussion_r1577704970_

Rahuljagwani:
@pellared @dmathieu, I have opened a PR regarding this. Please review and let me know any changes/improvements.

MrAlias:
This issue asks for an implementation mechanism rather than a missing behavior. The direct conversion loop in `otellogrus` is simpler, and benchmarks do not show a performance gap that a `kvBuffer` fixes. In `otelslog`, the buffer is justified by concrete filtering, grouping, and cloning behavior that `otellogrus` does not currently need.

The issue title led a contributor to implement an abstraction with no present behavior or maintenance benefit. That was misleading project guidance. Closing as not planned; a `kvBuffer` should not be added unless a separately justified behavior requires it.

PR #5356, opened 2024-04-03
logrus logging bridge

dmathieu:
Closes #5193.

This adds the logrus log bridge

PR #8516, opened 2026-02-05
bridges/otellogrus: add kvBuffer abstraction for field conversion

Rahuljagwani:
Solves #5428 

This PR introduces a kvBuffer type to otellogrus for more consistent and maintainable field conversion, following the pattern established in otelslog.

Took reference from:
https://github.com/open-telemetry/opentelemetry-go-contrib/blob/main/bridges/otelslog/handler.go#L379


MrAlias (commented):
This doesn’t add any behavior or capability; it just wraps a simple slice append loop in a new type and likely adds an extra allocation. Unless there’s a real follow‑on use, I don’t see the value here.

Rahuljagwani:
Hi @MrAlias, the issue mentions about the [comment](https://github.com/open-telemetry/opentelemetry-go-contrib/pull/5356#discussion_r1577704970) of using `kvbuffer`.

I introduced this helper anticipating reuse for attribute filtering / normalization similar to  [Link](https://github.com/open-telemetry/opentelemetry-go-contrib/blob/main/bridges/otelslog/handler.go#L388). If that’s premature, should we ignore for now and introduce it later?

or

Do i need to write filter rules, nil checks etc.?

MrAlias:
Thanks for taking this on. After revisiting #5428, this change should not land. The `kvBuffer` in `otelslog` exists to support concrete filtering, grouping, and cloning behavior; here it only wraps the existing conversion loop, with no behavior change or allocation improvement.

The issue title made the abstraction itself sound like the desired outcome, which was misleading guidance from the project. I’m closing this PR rather than asking you to add speculative behavior that we have not defined or requested. No further changes are needed on this PR.

PR #8529, opened 2026-02-09
otelconf: move experimental types to otelconf/x

codeboten:
The otelconf package now only includes the subset of features that are
stable in the configuration schema. The package otelconf/x must be used
for anyone looking to utilize experimental types.

Follow up to https://github.com/open-telemetry/opentelemetry-go-contrib/pull/8505

Fixes https://github.com/open-telemetry/opentelemetry-go-contrib/issues/8203


codeboten:
Note to reviewers: the PR contains a lot of code, most of it is duplicated code. The `Makefile` genjsonschema now puts the initial generated code (including experimental features) in `otelconf/x`, then `generated_config.go` is processed by a second sed script to remove experimental features, the results of which land in the root of the `otelconf` package.

dashpole:
Is there any way to see the parts that are new compared to the base otelconf?  Or at least a way to search for them so I can pay more attention there

codeboten:
> Is there any way to see the parts that are new compared to the base otelconf? Or at least a way to search for them so I can pay more attention there

The changes in `otelconf` in this PR are all subtractions, you can look at the resulting diff. Or is what you're asking if there's a way to see the diff between `otelconf` and `otelconf/x` moving forward?

dashpole:
I see.  So all of the added lines of code are a copy without modification (besides being in a new module).

otelconf/remove_experimental_patch.sed
```diff
@@ -0,0 +1,69 @@
+#!/bin/sed -f
```
dashpole, line 1: Is this leftover from when you generated the /x package, or is this something that we are going to use going forward to manage the differences between otelconf and otelconf/x?

codeboten: This will be used moving forward, it works for now, open to suggestions to improve the flow here

PR #8532, opened 2026-02-10
HTTP instrumentations cleanup

ash2k:
Relates to https://github.com/open-telemetry/opentelemetry-go-contrib/issues/8258.

See discussion in https://github.com/open-telemetry/opentelemetry-go-contrib/pull/8352#discussion_r2756697418.

Fundamentally, I think reusing `http.Request` is not a supported use case. It's not a documented supported behavior, so, it only works when `Body` is set to a "rewindable" Go type. It's an edge case, an implementation detail that should not be relied upon by any code because it is not part of the API contract and, hence, can change. And it does change when we wrap the transport and that is ok (in my opinion anyway).

Here is why I think `http.Request` is not supposed to be reused:

- It is not a documented way to use it (as I said above)
- Body will be closed by the transport. If it was a body that is "closeable", then you cannot reuse the request object.
- The HTTP transport mutates the request to prepare it - e.g. [sets the `ContentLength` field](https://github.com/golang/go/blob/go1.25.7/src/net/http/client.go#L676) ([here](https://github.com/golang/go/blob/go1.25.7/src/net/http/request.go#L926-L940) too), [modifies `Header`](https://github.com/golang/go/blob/go1.25.7/src/net/http/client.go#L866) (and [here](https://github.com/golang/go/blob/go1.25.7/src/net/http/client.go#L693), [here](https://github.com/golang/go/blob/go1.25.7/src/net/http/client.go#L251)). This very strongly indicates that this is a create-use-discard object, a single-use object.

Hence, I propose this alternative implementation. It matches the behavior of stdlib - uses `Body`, then `GetBody` (if set, etc). See commits. I refactored things a bit in separate commits, I hope it's ok.

WDYT? @adomaskizogian @dmathieu

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -16,6 +16,7 @@ The next release will require at least [Go 1.25].
 - `WithSpanKind` option in `go.opentelemetry.io/contrib/instrumentation/google.golang.org/grpc/otelgrpc` to override the default span kind. (#8506)
 - Add `const Version` in `go.opentelemetry.io/contrib/bridges/otelzap`. (#8544)
 - Support testing of [Go 1.26]. (#8549)
+- Add `const Version` in `go.opentelemetry.io/contrib/detectors/autodetect`. (#8555)
 - Add `const Version` in `go.opentelemetry.io/contrib/detectors/azure/azurevm`. (#8553)
 - Add `const Version` in `go.opentelemetry.io/contrib/processors/baggagecopy`. (#8557)
 - Add `const Version` in `go.opentelemetry.io/contrib/detectors/aws/lambda`. (#8510)
@@ -37,6 +38,14 @@ The next release will require at least [Go 1.25].
 
 ### Changed
 
+- Reduce allocations in the generated HTTP `internal/semconv` packages. (#8511)
+  Updated modules include:
+  - `go.opentelemetry.io/contrib/instrumentation/github.com/emicklei/go-restful/otelrestful`
+  - `go.opentelemetry.io/contrib/instrumentation/github.com/gin-gonic/gin/otelgin`
+  - `go.opentelemetry.io/contrib/instrumentation/github.com/gorilla/mux/otelmux`
+  - `go.opentelemetry.io/contrib/instrumentation/github.com/labstack/echo/otelecho`
+  - `go.opentelemetry.io/contrib/instrumentation/net/http/httptrace/otelhttptrace`
+  - `go.opentelemetry.io/contrib/instrumentation/net/http/otelhttp`
 - Updated the configuration schema used in `go.opentelemetry.io/contrib/otelconf` to [rc.3](https://github.com/open-telemetry/opentelemetry-configuration/releases/tag/v1.0.0-rc.3). (#8505)
 
 ### Deprecated
--- a/detectors/autodetect/go.mod
+++ b/detectors/autodetect/go.mod
@@ -3,6 +3,7 @@ module go.opentelemetry.io/contrib/detectors/autodetect
 go 1.24.0
 
 require (
+	github.com/stretchr/testify v1.11.1
 	go.opentelemetry.io/contrib/detectors/aws/ec2/v2 v2.2.0
 	go.opentelemetry.io/contrib/detectors/aws/ecs v1.40.0
 	go.opentelemetry.io/contrib/detectors/aws/eks v1.40.0
@@ -58,6 +59,7 @@ require (
 	github.com/modern-go/concurrent v0.0.0-20180306012644-bacd9c7ef1dd // indirect
 	github.com/modern-go/reflect2 v1.0.3-0.20250322232337-35a7c28c31ee // indirect
 	github.com/munnerz/goautoneg v0.0.0-20191010083416-a7dc8b61c822 // indirect
+	github.com/pmezard/go-difflib v1.0.1-0.20181226105442-5d4384ee4fb2 // indirect
 	github.com/x448/float16 v0.8.4 // indirect
 	go.opentelemetry.io/auto/sdk v1.2.1 // indirect
 	go.opentelemetry.io/otel/metric v1.40.0 // indirect
@@ -73,6 +75,7 @@ require (
 	google.golang.org/protobuf v1.36.11 // indirect
 	gopkg.in/evanphx/json-patch.v4 v4.13.0 // indirect
 	gopkg.in/inf.v0 v0.9.1 // indirect
+	gopkg.in/yaml.v3 v3.0.1 // indirect
 	k8s.io/api v0.34.3 // indirect
 	k8s.io/apimachinery v0.34.3 // indirect
 	k8s.io/client-go v0.34.3 // indirect
--- a/detectors/autodetect/version.go
+++ b/detectors/autodetect/version.go
@@ -1,0 +1,7 @@
+// Copyright The OpenTelemetry Authors
+// SPDX-License-Identifier: Apache-2.0
+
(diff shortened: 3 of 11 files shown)


instrumentation/net/http/otelhttp/internal/semconv/server.go
```diff
@@ -245,10 +246,8 @@ type MetricAttributes struct {
 }
 
 type MetricData struct {
-	RequestSize int64
-
-	// The request duration, in milliseconds
-	ElapsedTime float64
+	RequestSize     int64
+	RequestDuration time.Duration
```
ash2k, line 250: I think this is much cleaner. Type-safe and decoupled from how the data should be used in a metric/trace (i.e. what unit to use).

instrumentation/net/http/otelhttp/internal/semconv/server.go
```diff
@@ -397,7 +396,7 @@ func (n HTTPServer) MetricAttributes(server string, req *http.Request, statusCod
 	}
 
 	if route != "" {
-        attributes = append(attributes, semconv.HTTPRoute(route))
-    }
+		attributes = append(attributes, semconv.HTTPRoute(route))
+	}
```
ash2k, line 394: Spaces -> tabs. Is there no formatter setup in CI?

instrumentation/net/http/otelhttp/internal/semconv/util.go
```diff
@@ -125,3 +126,8 @@ func standardizeHTTPMethod(method string) string {
 	}
 	return method
 }
+
+func durationToSeconds(d time.Duration) float64 {
```
ash2k, line 130: I've seen this code a lot of time in OTEL libs. Should be in some helper library?

instrumentation/net/http/otelhttp/handler.go
```diff
@@ -176,6 +176,7 @@ func (h *middleware) serveHTTP(w http.ResponseWriter, r *http.Request, next http
 
 	r = r.WithContext(ctx)
 	next.ServeHTTP(w, r)
+	requestDuration := time.Since(requestStartTime)
```
ash2k: Measure closer for more precision. In other places too.

instrumentation/net/http/otelhttp/transport.go
```diff
@@ -144,32 +144,23 @@ func (t *Transport) RoundTrip(r *http.Request) (*http.Response, error) {
 	res, err := t.rt.RoundTrip(r)
 	requestDuration := time.Since(requestStartTime)
 
-	// Defer metrics recording function to record the metrics on error or no error.
-	defer func() {
```
ash2k, line 149: I think this `defer` complicates thing more than simplifies them.

instrumentation/net/http/otelhttp/handler.go
```diff
@@ -193,16 +193,14 @@ func (h *middleware) serveHTTP(w http.ResponseWriter, r *http.Request, next http
 		WriteError: rww.Error(),
 	})...)
 
-	metricAttributes := semconv.MetricAttributes{
-		Req:                  r,
-		StatusCode:           statusCode,
-		AdditionalAttributes: append(labeler.Get(), h.metricAttributesFromRequest(r)...),
-	}
-
 	h.semconv.RecordMetrics(ctx, semconv.ServerMetricData{
```
ash2k, line 196: I think this is more readable vs bits-in-vars approach. WDYT? Below too.

instrumentation/net/http/otelhttp/transport.go
```diff
@@ -118,23 +115,26 @@ func (t *Transport) RoundTrip(r *http.Request) (*http.Response, error) {
...
+	if r.GetBody != nil {
+		originalGetBody := r.GetBody
+		r.GetBody = func() (io.ReadCloser, error) {
+			b, err := originalGetBody()
+			if err != nil {
+				lastBW = nil // the underlying transport will fail to make a retry request, hence, record no data.
+				return nil, err
+			}
+			return maybeWrapBody(b), nil
+		}
```
ash2k, line 133: I think the above approach is much cleaner:
- Errors are propagated to the caller
- behavior is consistent with stdlib
- data is recorded for body from `GetBody` too - current code does not instrument the returned body.

instrumentation/net/http/otelhttp/transport_test.go
```diff
@@ -1118,88 +1117,80 @@ func BenchmarkTransportRoundTrip(b *testing.B) {
 }
 
 func TestRequestBodyReuse(t *testing.T) {
```
ash2k, line 1120: In these tests we make the HTTP client use the `Body` and `GetBody` by responding with a redirect.

instrumentation/net/http/otelhttp/transport_test.go
```diff
@@ -1118,88 +1117,80 @@ func BenchmarkTransportRoundTrip(b *testing.B) {
...
+	var (
+		receivedBody1, receivedBody2 string
+		requests                     int
+	)
 	server := httptest.NewServer(http.HandlerFunc(func(w http.ResponseWriter, r *http.Request) {
-		body, err := io.ReadAll(r.Body)
+		b, err := io.ReadAll(r.Body)
 		assert.NoError(t, err)
 
-		defer r.Body.Close()
```
ash2k, line 1126: Unnecessary on the server side. See docs.

instrumentation/net/http/otelhttp/transport_test.go
```diff
@@ -1118,88 +1117,80 @@ func BenchmarkTransportRoundTrip(b *testing.B) {
...
-		bodies = append(bodies, string(body))
-
-		w.WriteHeader(http.StatusOK)
```
ash2k, line 1129: This is the default already. Shorter tests are nicer.

dmathieu:
Could you split this in two PRs? One for the cleanup, and another for the bugfix. Also, the bugfix will need a changelog entry.

ash2k:
@dmathieu I've dropped the bugfix commit. I'll open a new PR once this is merged (hopefully). I was getting linter failures in CI, but locally it works fine. Let's see if CI passes now.

ash2k:
Figured out what's going on. Fixing.

dmathieu:
The `internal/semconv` package is based on shared templates.
https://github.com/open-telemetry/opentelemetry-go-contrib/tree/main/internal/shared/semconv

They need to be updated, and you should run `make generate` to rebuild everything.

ash2k:
@dmathieu I've fixed the issues. The failing tests seem unrelated.

instrumentation/github.com/gin-gonic/gin/otelgin/gin.go
```diff
@@ -60,8 +60,6 @@ func Middleware(service string, opts ...Option) gin.HandlerFunc {
 	sc := semconv.NewHTTPServer(meter)
 
 	return func(c *gin.Context) {
-		requestStartTime := time.Now()
```
dmathieu, line 63: This is done early on purpose, to account for the filters time.

ash2k: Ack, I've moved it back. FWIW IMO timing should measure just the thing that is being wrapped, but it's not the point of this PR.

dmathieu (approved):

ash2k:
This is ready to go. Needs to be merged so that I can open a follow up with the bugfix.

dmathieu:
This isn't ready to go. We require 2 approvals.

pellared:
@ash2k, are you able to resolve the conflicts? I want to merge this PR :wink:

PR #8560, opened 2026-02-14
otelconf: fail closed when OTLP HTTP exporter uses http endpoint with TLS/mTLS config

sandy2008:
## Summary
This PR fixes a TLS/mTLS misconfiguration gap in `otelconf` OTLP HTTP exporters.

- Added shared helpers to detect HTTP exporter TLS config (`ca_file`, `cert_file`, `key_file`) and validate endpoint scheme.
- Updated HTTP exporters for traces, metrics, and logs to require `https` when any TLS/mTLS config is set.
- Added regression tests for all three exporters.

## Problem
Previously, if users configured:
- a non-HTTPS endpoint (for example `http://...`, `unix:...`, or no-scheme endpoint), and
- TLS/mTLS cert settings

the exporter could still proceed in ways that did not use TLS, so client certs were not applied. This failed open and could silently weaken transport security.

## Fix
Now we fail closed:
- If TLS/mTLS files are configured and endpoint scheme is not `https`, exporter creation returns:
  - `invalid config: tls configuration`
  - with detail: `tls configuration requires an https endpoint`

## Scope
Updated in:
- trace HTTP exporter
- metric HTTP exporter
- log HTTP exporter
- shared config helper

## Tests
Added/updated tests to assert validation behavior for:
- `http://...` + TLS config -> error
- existing TLS-related success/failure cases updated to use explicit `https://...` where appropriate

Coverage includes:
- trace
- metric
- log

Validation run:
- `go test ./...` in `otelconf` passes.

## Behavior impact
- `https://...` + TLS/mTLS: unchanged (works).
- non-HTTPS endpoint without TLS config: unchanged.
- non-HTTPS endpoint + TLS/mTLS config: now returns a validation error (intended).


dmathieu:
How about making that part of the exporters themselves? Using certificates with protocols that don't support them could return errors even for folks who are not using otelconf.

Also, I would put the check the other way around. If there are certificates, check that the protocol is indeed https.
Using a socker connection to send data can't use certifi ates either.

sandy2008:
> How about making that part of the exporters themselves? Using certificates with protocols that don't support them could return errors even for folks who are not using otelconf.
> 
> Also, I would put the check the other way around. If there are certificates, check that the protocol is indeed https. Using a socker connection to send data can't use certifi ates either.

@dmathieu You’re absolutely right on both points.

I updated this PR to do the check in that direction: if any TLS/mTLS cert config is present (ca_file, cert_file, or key_file), the endpoint scheme must be https. So http, unix/socket, and no-scheme endpoints now fail fast with a TLS config error.

I also agree this validation should ideally live in the OTLP HTTP exporters themselves so non-otelconf users get the same protection. I kept the guard in otelconf for immediate safety here, and I can open a follow-up upstream issue/PR to add the same behavior in the exporter packages.

dmathieu:
Both this package and the exporters and maintained by the same folks, and released on the same schedule. So doing this here first doesn't really make sense.

sandy2008:
> Both this package and the exporters and maintained by the same folks, and released on the same schedule. So doing this here first doesn't really make sense.

@dmathieu Thank you, I made https://github.com/open-telemetry/opentelemetry-go/pull/7914

PR #8565, opened 2026-02-16
google.golang.org/grpc/otelgrpc: Use canonical string status codes

### LINKED ISSUE #8543: [Bug]: `otelgrpc` uses PascalCase status codes, spec uses SCREAMING_SNAKE_CASE
### Component

Instrumentation: otelgrpc

The [spec](https://opentelemetry.io/docs/specs/semconv/rpc/grpc/) shows  `rpc.response.status_code` as being `DEADLINE_EXCEEDED`, etc.

`otelgrpc` uses [`(Code).String()`](https://github.com/grpc/grpc-go/blob/v1.78.0/codes/code_string.go#L31) and thus produces `DeadlineExceeded`, etc.

### Expected behavior

`otelgrpc` is consistent with the spec. It will be problematic if you get different casing for this attribute in different languages (although it looks like Go might be first to move to the new semconv, it looks like at least Python and JS are still producing the deprecated `rpc.grpc.status_code` attribute).

### Steps to Reproduce

Look at spans with anything other than `OK` status.

### Operating System

macOS

### Device Architecture

ARM64

### Go Version

1.25

### Component Version

otelgrpc v0.65.0

haines:
Fixes #8543


instrumentation/google.golang.org/grpc/otelgrpc/stats_handler.go
```diff
@@ -362,3 +363,44 @@ func (c *config) handleRPC(
...
+		return "UNIMPLEMENTED"
+	case grpc_codes.Internal:
+		return "INTERNAL"
+	case grpc_codes.Unavailable:
+		return "UNAVAILABLE"
+	case grpc_codes.DataLoss:
+		return "DATA_LOSS"
+	case grpc_codes.Unauthenticated:
+		return "UNAUTHENTICATED"
+	default:
```
dmathieu, line 403: This default case doesn't appear to be tested.

pellared:
There is also a failure from the linter

instrumentation/google.golang.org/grpc/otelgrpc/interceptor.go
```diff
@@ -44,14 +44,19 @@ func serverAddrAttrs(hostport string) []attribute.KeyValue {
 // code of Unset and an empty message.
 func serverStatus(grpcStatus *status.Status) (codes.Code, string) {
```
dmathieu, line 45: This change is wrong. It changes the status we set on the span, which we don't want do in this PR.

dmathieu: Also, as a side note, the comment would now be entirely wrong.

haines: Shouldn't a code outside the range of known gRPC codes be recorded as an error? I think it should be treated the same as `Unknown`.

I don't think the comment is entirely wrong, but I can change it to something like 

```go
// If the gRPC status code is Unknown, DeadlineExceeded, Unimplemented,
// Internal, Unavailable, DataLoss, or an invalid code, it returns a span
// status code of Error and the message from the gRPC status. Otherwise, it
// returns a span status code of Unset and an empty message.
```

if you want to be specific about the handling of invalid codes.

dmathieu: Possibly. But that's not what this PR is about. If we want to reconsider this behavior, it should be its own PR.

dmathieu (approved):

instrumentation/google.golang.org/grpc/otelgrpc/stats_handler.go
```diff
@@ -362,3 +363,44 @@ func (c *config) handleRPC(
 		return
 	}
 }
+
+func canonicalString(code grpc_codes.Code) string {
+	switch code {
+	case grpc_codes.OK:
+		return "OK"
+	case grpc_codes.Canceled:
+		return "CANCELLED"
```
bincyber, line 372: Should not this be spelled the same way as `grpc_codes.Canceled`? @haines

dmathieu: The string representation of this error does have two L in the grpc spec.

https://github.com/grpc/grpc/blob/v1.75.0/doc/statuscodes.md

haines: And this is what is used as the canonical string internally in the Go gRPC implementation as well, unfortunately they named the constant using the alternate spelling which makes things confusing: https://github.com/grpc/grpc-go/blob/b71c26202050d6a6c528e42f96a2712baa74eb61/codes/code_string.go#L77

Issue #8115, opened 2025-11-01
[Bug]: Flaky test in `otelconf/v0.2.0`


MrAlias:
https://github.com/open-telemetry/opentelemetry-go-contrib/actions/runs/18991376824/job/54244892209?pr=8113

```console
--- FAIL: Test_otlpGRPCLogExporter (30.01s)
    --- FAIL: Test_otlpGRPCLogExporter/no_TLS_config (10.00s)
        log_test.go:882: 
            	Error Trace:	/Users/runner/work/opentelemetry-go-contrib/opentelemetry-go-contrib/otelconf/v0.3.0/log_test.go:883
            	            				/Users/runner/hostedtoolcache/go/1.25.0/x64/src/runtime/asm_amd64.s:1693
            	Error:      	Received unexpected error:
            	            	exporter export timeout: rpc error: code = DeadlineExceeded desc = context deadline exceeded
        log_test.go:882: 
            	Error Trace:	/Users/runner/work/opentelemetry-go-contrib/opentelemetry-go-contrib/otelconf/v0.3.0/log_test.go:882
            	Error:      	Condition never satisfied
            	Test:       	Test_otlpGRPCLogExporter/no_TLS_config
    --- FAIL: Test_otlpGRPCLogExporter/with_TLS_config (10.00s)
        log_test.go:882: 
            	Error Trace:	/Users/runner/work/opentelemetry-go-contrib/opentelemetry-go-contrib/otelconf/v0.3.0/log_test.go:883
            	            				/Users/runner/hostedtoolcache/go/1.25.0/x64/src/runtime/asm_amd64.s:1693
            	Error:      	Received unexpected error:
            	            	exporter export timeout: rpc error: code = DeadlineExceeded desc = context deadline exceeded
        log_test.go:882: 
            	Error Trace:	/Users/runner/work/opentelemetry-go-contrib/opentelemetry-go-contrib/otelconf/v0.3.0/log_test.go:882
            	Error:      	Condition never satisfied
            	Test:       	Test_otlpGRPCLogExporter/with_TLS_config
    --- FAIL: Test_otlpGRPCLogExporter/with_TLS_config_and_client_key (10.00s)
        log_test.go:882: 
            	Error Trace:	/Users/runner/work/opentelemetry-go-contrib/opentelemetry-go-contrib/otelconf/v0.3.0/log_test.go:883
[...]

PR #8574, opened 2026-02-18
Use the actual subtest's context rather than the main test one

dmathieu:
Those tests share a context, which is passed as the subtest's arguments.
We really don't need to do that, as we always use `t.Context()` anyway.

Also, using the main test's context could be the reason we're seeing the flaky failures described in https://github.com/open-telemetry/opentelemetry-go-contrib/issues/8115, as if another subtest canceled the context, all the remaining ones wouldn't be able to use it.
With this change, each subtest gets its own context.

This also force-stops the collector that gracefully doing so. This is a test environment, we don't really care about finishing ongoing requests, and would rather stop the server quickly.

### DIFF
--- a/otelconf/log_test.go
+++ b/otelconf/log_test.go
@@ -834,7 +834,7 @@ func Test_otlpGRPCLogExporter(t *testing.T) {
 
 			startGRPCLogsCollector(t, n, serverOpts)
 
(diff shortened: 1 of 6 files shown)


dmathieu:
Not the right path it seems.

PR #8578, opened 2026-02-18
otelconf: host detector should include os as well

codeboten:
As per the specification, the host resource detector should also include os attributes.


codeboten:
> We are missing test(s) that would protect us from a regression bug.

test added, thanks for the suggestion @pellared

otelconf/resource_test.go
```diff
@@ -81,3 +81,81 @@ func TestNewResource(t *testing.T) {
 		})
 	}
 }
+
+func TestResourceOptsWitDetectors(t *testing.T) {
```
pellared:
```suggestion
func TestResourceOptsWithDetectors(t *testing.T) {
```

otelconf/resource_test.go
```diff
@@ -81,3 +81,81 @@ func TestNewResource(t *testing.T) {
...
+			_, ok = attrMap[semconv.OSTypeKey]
+			require.Equal(t, tt.wantOSAttributes, ok)
+
+			// Check for host.id attribute
+			_, ok = attrMap[semconv.HostIDKey]
+			require.Equal(t, tt.wantHostIDAttribute, ok)
+
+			// Check for process.pid attribute
+			_, ok = attrMap[semconv.ProcessPIDKey]
+			require.Equal(t, tt.wantProcessAttribute, ok)
```
pellared: From https://go.dev/doc/effective_go#maps

> A set can be implemented as a map with value type bool. 

simplified, better assertion messages, and do not fail on first failed assertion

```suggestion
			attrSet := make(map[attribute.Key]bool)
			for _, attr := range attrs {
				attrSet[attr.Key] = true
			}

			assert.Equal(t, tt.wantHostAttributes, attrSet[semconv.HostNameKey], "should have host.name attribute")
			assert.Equal(t, tt.wantOSAttributes, attrSet[semconv.OSTypeKey], "should have os.type attribute (from WithOS()")
			assert.Equal(t, tt.wantHostIDAttribute, attrSet[semconv.HostIDKey], "should have host.id attribute")
			assert.Equal(t, tt.wantProcessAttribute, attrSet[semconv.ProcessPIDKey], "should have process.pid attribute")
```

PR #8581, opened 2026-02-19
otelconf: remove host ID detector in host configuration

codeboten:
The host ID detector returns an error in a containerized environment. My suggestion with this PR is to remove it. An option here, would be to update the `WithHost` detector to include the host ID as well, and return a partial error, which the configuration package could check for. Maybe someone who is more familar with the resource detectors and their implementations could chime in here. I suppose we could also add another option WithHostAndHostID or something like it, but that seems redundant.


CHANGELOG.md
```diff
@@ -29,6 +29,7 @@ The next release will require at least [Go 1.25].
 ### Removed
 
 - Removed the deprecated zipkin exporter example in `go.opentelemetry.io/contrib/examples/zipkin`. (#8501)
+- Host ID resource detector has been removed when configuring the `host` resource detector in `go.opentelemetry.io/contrib/otelconf`. (#)
```
codeboten:
```suggestion
- Host ID resource detector has been removed when configuring the `host` resource detector in `go.opentelemetry.io/contrib/otelconf`. (#8581)
```

PR #8595, opened 2026-02-24
otelconf: use translation strategy instead of deprecated funcs

codeboten:
This fixes an issue in the otelconf package that was preventing the Collector from upgrading to the latest version of the library. The issue is that the prometheus exporter uses a translation strategy that could not be changed with the existing configuration options, and this changed the output from the exporter, breaking backwards compat for the collector when upgrading to otelconf v0.20.0. See https://github.com/open-telemetry/opentelemetry-collector/pull/14602#pullrequestreview-3822709128 for more details.

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -38,7 +38,7 @@ jobs:
       with:
         path: .tools
         key: ${{ runner.os }}-${{ env.cache-name }}-${{ env.DEFAULT_GO_VERSION }}-${{ hashFiles('./tools/**') }}
-      # The step below is needed to not rebuild all the build tools.
+    # The step below is needed to not rebuild all the build tools.
     - name: Set tools/go.mod timestamp
       run: |
         filename="tools/go.mod"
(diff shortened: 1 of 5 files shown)


otelconf/v0.2.0/metric.go
```diff
@@ -272,12 +273,13 @@ func prometheusReader(ctx context.Context, prometheusConfig *Prometheus) (sdkmet
 	if prometheusConfig.WithoutScopeInfo != nil && *prometheusConfig.WithoutScopeInfo {
 		opts = append(opts, otelprom.WithoutScopeInfo())
 	}
-	if prometheusConfig.WithoutTypeSuffix != nil && *prometheusConfig.WithoutTypeSuffix {
-		opts = append(opts, otelprom.WithoutCounterSuffixes()) //nolint:staticcheck // WithoutCounterSuffixes is deprecated, but we still need it for backwards compatibility.
-	}
-	if prometheusConfig.WithoutUnits != nil && *prometheusConfig.WithoutUnits {
-		opts = append(opts, otelprom.WithoutUnits()) //nolint:staticcheck // WithouTypeSuffix is deprecated, but we still need it for backwards compatibility.
+
+	if prometheusConfig.WithoutTypeSuffix != nil && *prometheusConfig.WithoutTypeSuffix && prometheusConfig.WithoutUnits != nil && *prometheusConfig.WithoutUnits {
```
MikeGoldsmith, line 277: Instead of this chained and duplicated series of conditions, could we introduce a utility func that all metric implementations could use? ie this check is duplicated in 0.2.0 & 0.3.0 and might be used in later implementations too

MikeGoldsmith (approved):

otelconf/v0.2.0/metric.go
codeboten: later implementations don't have those fields, and the 0.2.0 and 0.3.0 will be removed once the stable version is released

codeboten:
I dont think the lint check failure is related to changes in this pr

Issue #8524, opened 2026-02-09
[Bug]: `otehttp.NewTransport` causes requests to hang when server responds without consuming the request body in v0.65.0


akhilerm:
### Component

Instrumentation: otelhttp

### Describe the issue you're facing

containerd project uses a request implementation in which request body is via an io.Pipe. When the server responds without consuming the body of the request, the subsequent request hangs. 

Bisected the changes and found that failures are seen after https://github.com/open-telemetry/opentelemetry-go-contrib/pull/8352

### Expected behavior

Adding tracing should not cause the request to hang. 

### Steps to Reproduce

The below code has been referred from  https://github.com/containerd/containerd/blob/bde439b02feea42936feb03bc40759b5427a42b8/core/remotes/docker/pusher.go#L70 to reproduce the issue
```go
package otelhttp

import (
	"context"
	"io"
	"net/http"
	"net/http/httptest"
	"sync/atomic"
	"testing"
	"time"

	"go.opentelemetry.io/contrib/instrumentation/net/http/otelhttp"
)

func TestOtelHTTPHangReproduction(t *testing.T) {
	var requestCount atomic.Int32

	server := httptest.NewServer(http.HandlerFunc(func(w http.ResponseWriter, r *http.Request) {
		count := requestCount.Add(1)
		if count == 1 {
			// the first request returns an error
			w.WriteHeader(http.StatusInternalServerError)
			return
		}
		io.Copy(io.Discard, r.Body)
		w.WriteHeader(http.StatusOK)
	}))
	defer server.Close()

	ctx, cancel := context.WithTimeout(context.Background(), 5*time.Second)
	defer cancel()

	type request struct {
		url  string
		body func() (io.ReadCloser, error)
		size int64
	}

	var pipeWriter *io.PipeWriter

	req := &request{
		url:  server.URL,
		size: 4,
		body: func() (io.ReadCloser, error) {
			pr, pw := io.Pipe()
			pipeWriter = pw
			return pr, nil
		},
	}

	doRequest := func() error {
		body, err := req.body()
		if err != nil {
			return err
		}

		httpReq, err := http.NewRequestWithContext(ctx, http.MethodPut, req.url, body)
		if err != nil {
			return err
		}
		httpReq.GetBody = req.body

		client := &http.Client{}

		// this modifies the transport
		client.Transport = otelhttp.NewTransport(http.DefaultTransport)

		go func() {
			pipeWriter.Write([]byte("test"))
[...]

PR #8618, opened 2026-03-03
fix(otelhttp): make Body handling in Transport consistent with stdlib

ash2k:
First request uses `Body`, following requests use `Body` if it can be rewound, which is only possible for well-known types that stdlib knows about. Otherwise, `GetBody()` is used, when available.

This is a follow up for https://github.com/open-telemetry/opentelemetry-go-contrib/pull/8532.

Fixes https://github.com/open-telemetry/opentelemetry-go-contrib/issues/8524.


instrumentation/net/http/otelhttp/transport.go
```diff
@@ -119,23 +116,26 @@ func (t *Transport) RoundTrip(r *http.Request) (*http.Response, error) {
...
-	// will affect the identity of it in an unforeseeable way because we assert
-	// ReadCloser fulfills a certain interface and it is indeed nil or NoBody.
-	body := r.Body
-	if r.GetBody != nil {
-		b, err := r.GetBody()
-		if err != nil {
-			otel.Handle(fmt.Errorf("http.Request GetBody returned an error: %w", err))
-		} else {
-			body = b
+	var lastBW *request.BodyWrapper // records the last body wrapper. Can be nil.
```
pellared:
```suggestion
	var lastBW *request.BodyWrapper // Records the last body wrapper. Can be nil.
```

instrumentation/net/http/otelhttp/transport.go
```diff
@@ -119,23 +116,26 @@ func (t *Transport) RoundTrip(r *http.Request) (*http.Response, error) {
...
-	bw := request.NewBodyWrapper(body, func(int64) {})
-	if body != nil && body != http.NoBody {
-		r.Body = bw
+	r.Body = maybeWrapBody(r.Body)
+	if r.GetBody != nil {
+		originalGetBody := r.GetBody
+		r.GetBody = func() (io.ReadCloser, error) {
+			b, err := originalGetBody()
+			if err != nil {
+				lastBW = nil // the underlying transport will fail to make a retry request, hence, record no data.
```
pellared:
```suggestion
				lastBW = nil // The underlying transport will fail to make a retry request, hence, record no data.
```

ash2k:
@pellared Done!

CHANGELOG.md
```diff
@@ -8,6 +8,10 @@ This project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0.htm
 
 ## [Unreleased]
 
+### Fixed
+
+- Make `Body` / `GetBody()` behavior consistent with stdlib in `go.opentelemetry.io/contrib/instrumentation/net/http/otelhttp`.
```
```suggestion
- Make `Body` handling in `Transport` consistent with stdlib in `go.opentelemetry.io/contrib/instrumentation/net/http/otelhttp`. (#8618)
```

PR #8632, opened 2026-03-04
feat(net/http/otelhttp): Add `http.route` attribute

gaiaz-iusipov:
Fixes https://github.com/open-telemetry/opentelemetry-go-contrib/issues/8633

In this pull request, I add recording of the `http.route` metric attribute for `net/http/otelhttp`.

This PR does not affect any instrumentation other than `otelhttp`, because the check `if route == "" && req.Pattern != ""` is triggered only in `otelhttp`.

### DIFF
--- a/instrumentation/github.com/gin-gonic/gin/otelgin/internal/semconv/server_test.go
+++ b/instrumentation/github.com/gin-gonic/gin/otelgin/internal/semconv/server_test.go
@@ -18,13 +18,14 @@ import (
 )
 
 func TestHTTPServer_MetricAttributes(t *testing.T) {
-	defaultRequest, err := http.NewRequestWithContext(t.Context(), http.MethodGet, "http://example.com/path?query=test", http.NoBody)
-	require.NoError(t, err)
+	defaultRequest := httptest.NewRequestWithContext(t.Context(), http.MethodGet, "http://example.com/path?query=test", http.NoBody)
+	reqWithPattern := httptest.NewRequestWithContext(t.Context(), http.MethodGet, "/path/abc123", http.NoBody)
+	reqWithPattern.Pattern = "/path/abc123"
 
 	tests := []struct {
 		name                 string
 		server               string
-		reqFunc              func(t *testing.T) *http.Request
+		req                  *http.Request
 		statusCode           int
 		route                string
 		additionalAttributes []attribute.KeyValue
@@ -33,7 +34,7 @@ func TestHTTPServer_MetricAttributes(t *testing.T) {
 		{
 			name:                 "routine testing",
 			server:               "",
-			reqFunc:              func(_ *testing.T) *http.Request { return defaultRequest },
+			req:                  defaultRequest,
 			statusCode:           200,
 			route:                "",
 			additionalAttributes: []attribute.KeyValue{attribute.String("test", "test")},
@@ -53,7 +54,7 @@ func TestHTTPServer_MetricAttributes(t *testing.T) {
 		{
 			name:                 "use server address",
 			server:               "example.com:9999",
-			reqFunc:              func(_ *testing.T) *http.Request { return defaultRequest },
+			req:                  defaultRequest,
 			statusCode:           200,
 			route:                "/path/${id}",
 			additionalAttributes: nil,
@@ -73,11 +74,7 @@ func TestHTTPServer_MetricAttributes(t *testing.T) {
 		},
 		{
 			name: "use route from request pattern",
-			reqFunc: func(t *testing.T) *http.Request {
-				req := httptest.NewRequestWithContext(t.Context(), http.MethodGet, "/path/abc123", http.NoBody)
-				req.Pattern = "/path/abc123"
-				return req
-			},
+			req:  reqWithPattern,
 			wantFunc: func(t *testing.T, attrs []attribute.KeyValue) {
 				require.Len(t, attrs, 6)
 				assert.ElementsMatch(t, []attribute.KeyValue{
@@ -94,8 +91,7 @@ func TestHTTPServer_MetricAttributes(t *testing.T) {
 
 	for _, tt := range tests {
 		t.Run(tt.name, func(t *testing.T) {
-			req := tt.reqFunc(t)
-			got := HTTPServer{}.MetricAttributes(tt.server, req, tt.statusCode, tt.route, tt.additionalAttributes)
+			got := HTTPServer{}.MetricAttributes(tt.server, tt.req, tt.statusCode, tt.route, tt.additionalAttributes)
 			tt.wantFunc(t, got)
 		})
 	}
--- a/instrumentation/github.com/gorilla/mux/otelmux/internal/semconv/server_test.go
+++ b/instrumentation/github.com/gorilla/mux/otelmux/internal/semconv/server_test.go
@@ -18,13 +18,14 @@ import (
 )
 
 func TestHTTPServer_MetricAttributes(t *testing.T) {
-	defaultRequest, err := http.NewRequestWithContext(t.Context(), http.MethodGet, "http://example.com/path?query=test", http.NoBody)
-	require.NoError(t, err)
+	defaultRequest := httptest.NewRequestWithContext(t.Context(), http.MethodGet, "http://example.com/path?query=test", http.NoBody)
+	reqWithPattern := httptest.NewRequestWithContext(t.Context(), http.MethodGet, "/path/abc123", http.NoBody)
+	reqWithPattern.Pattern = "/path/abc123"
 
 	tests := []struct {
 		name                 string
 		server               string
-		reqFunc              func(t *testing.T) *http.Request
+		req                  *http.Request
 		statusCode           int
 		route                string
 		additionalAttributes []attribute.KeyValue
@@ -33,7 +34,7 @@ func TestHTTPServer_MetricAttributes(t *testing.T) {
 		{
 			name:                 "routine testing",
 			server:               "",
-			reqFunc:              func(_ *testing.T) *http.Request { return defaultRequest },
+			req:                  defaultRequest,
 			statusCode:           200,
 			route:                "",
 			additionalAttributes: []attribute.KeyValue{attribute.String("test", "test")},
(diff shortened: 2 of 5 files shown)


pellared (changes requested):
Can you please create an issue first?

From https://github.com/open-telemetry/opentelemetry-go-contrib/blob/main/CONTRIBUTING.md#contributing-code:

> It's recommended that you signal your intention to contribute in the issue tracker, either by [filing a new issue](https://github.com/open-telemetry/opentelemetry-go-contrib/issues/new) or by claiming an [existing one](https://github.com/open-telemetry/opentelemetry-go-contrib/issues).

dmathieu:
Could this be done with a Labeler?

gaiaz-iusipov:
I created an issue where I tried to explain my motivation: https://github.com/open-telemetry/opentelemetry-go-contrib/issues/8633

gaiaz-iusipov:
I changed the logic so that the attribute is recorded by default, with an option to disable it. There is also still the question of moving the `routeFromRequest` function to the `semconv` package. Let’s discuss the final approach in the issue.

gaiaz-iusipov:
> Could this be done with a Labeler?

Sure it could, but @pellared pointed out in the issue that this attribute should always (when it is available) be recorded by default.

dmathieu:
Sure, though this is not what that PR was doing when I left my comment.
Now: the attribute can be removed with a view. Why do we need the option?

gaiaz-iusipov:
> Sure, though this is not what that PR was doing when I left my comment. Now: the attribute can be removed with a view. Why do we need the option?

So it turns out the option is not needed at all. What about moving the `routeFromRequest` function to `semconv.HTTPServer` and making it public since its logic is duplicated?

internal/shared/semconv/util.go.tmpl
```diff
@@ -57,6 +57,15 @@ func SplitHostPort(hostport string) (host string, port int) {
 	return host, int(p) //nolint:gosec  // Byte size checked 16 above.
 }
 
+// RouteFromPattern extracts the HTTP route path from a pattern string.
+// If the pattern does not contain '/', an empty string is returned.
+func RouteFromPattern(pattern string) string {
```
dmathieu: We don't need to make this "public" (the quotes are because the package is private, so it's no public).
This attribute can be added within the `RecordMetrics` method.

instrumentation/net/http/otelhttp/handler_test.go
```diff
@@ -110,64 +110,54 @@ func TestHandlerBasics(t *testing.T) {
 	reader := sdkmetric.NewManualReader()
 	meterProvider := sdkmetric.NewMeterProvider(sdkmetric.WithReader(reader))
 
-	h := NewHandler(
+	handler := NewHandler(
```
dmathieu: Please keep cosmetic changes out.

CHANGELOG.md
```diff
@@ -8,6 +8,10 @@ This project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0.htm
 
 ## [Unreleased]
 
+### Added
+
+- Add the `http.route` metric attribute to `go.opentelemetry.io/contrib/instrumentation/net/http/otelhttp` (#8632)
```
flc1125: In fact, it’s not just `otelhttp`; there are also several other components.

flc1125 (approved):
Some small adjustments.

CHANGELOG.md
gaiaz-iusipov: Despite the shared code generation, the behavior of other components was not affected. Different components work differently, and in a couple of places the `http.route` value is passed via `additionalAttributes` instead of using the `MetricAttributes.Route` variable. I will change this in another PR.

gaiaz-iusipov:
I think this PR is ready to merge.

dmathieu:
I'm waiting for @pellared's review, since he requested changes.

internal/shared/semconv/server.go.tmpl
```diff
@@ -364,7 +364,9 @@ func (n HTTPServer) MetricAttributes(server string, req *http.Request, statusCod
 	if statusCode > 0 {
 		num++
 	}
-
+	if route == "" && req.Pattern != "" {
+		route = httpRoute(req.Pattern)
+	}
```
pellared, line 369: This change does not seem to be needed.
Why not simply setting `Route: r.Pattern,` here: 

https://github.com/open-telemetry/opentelemetry-go-contrib/blob/a0bbd72e0f5659391736516758ec6151dd24bd16/instrumentation/net/http/otelhttp/handler.go#L199-L203

This will be also safer as the current change can also affect other instrumentation libraries.

internal/shared/semconv/server_test.go.tmpl
```diff
@@ -24,7 +24,7 @@ func TestHTTPServer_MetricAttributes(t *testing.T) {
 	tests := []struct {
 		name                 string
 		server               string
-		req                  *http.Request
+		reqFunc              func(t *testing.T) *http.Request
```
pellared: This change seems ennecessary. You can create `reqWithPattern` above `tests := []struct {`

CHANGELOG.md
```diff
@@ -17,6 +17,7 @@ This project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0.htm
 - Added support for `attribute_count_limit` and `attribute_value_length_limit` in logger provider configuration in `go.opentelemetry.io/contrib/otelconf`. (#8686)
 - Added support for `server.address` and `server.port` attributes in `go.opentelemetry.io/contrib/instrumentation/google.golang.org/grpc/otelgrpc`. (#8723)
 - Support `OTEL_SEMCONV_STABILITY_OPT_IN` in `go.opentelemetry.io/contrib/instrumentation/google.golang.org/grpc/otelgrpc`. Supported values are `rpc` (default), `rpc/dup` and `rpc/old`. (#8726)
+- Add the `http.route` metric attribute to `go.opentelemetry.io/contrib/instrumentation/net/http/otelhttp` (#8632)
```
```suggestion
- Add the `http.route` metric attribute to `go.opentelemetry.io/contrib/instrumentation/net/http/otelhttp`. (#8632)
```

internal/shared/semconv/server.go.tmpl
gaiaz-iusipov: [it was originally done that way](https://github.com/open-telemetry/opentelemetry-go-contrib/pull/8632/changes/2936802804710cc36472ab2fbbba05abc9ebca00#diff-70a93fdc86adfd2f20a035bb596f6d9320d6fed7008eaefc6c7d7cde2b1fc97a), but [in this comment](https://github.com/open-telemetry/opentelemetry-go-contrib/pull/8632#discussion_r2910612607) I was asked to resolve the value of routePath inside the `semconv` package.

CHANGELOG.md
gaiaz-iusipov: Fixed

internal/shared/semconv/server_test.go.tmpl
gaiaz-iusipov: Fixed

internal/shared/semconv/server.go.tmpl
pellared: None of these comments seem related. The code was different at that point e.g. there was a `routeFromRequest` function and `WithMetricRouteAttribute` option which are no longer present.

gaiaz-iusipov:
> Can you please update the issue and PR description?

@pellared thanks for the review, I will update the issue description and this PR.

internal/shared/semconv/server.go.tmpl
gaiaz-iusipov: There are no more options in this PR.

The remaining question is that to set the value through the `MetricAttributes.Route` variable, as it was originally done, I need to use the `httpRoute` function from the `semconv` package. That is why I asked whether I should make this function public or duplicate it in the `otelhttp` package. A third option was suggested, where I do not use `MetricAttributes.Route` at all and instead apply the `httpRoute` function directly in the `semconv` package. I do not like this approach because I consider the instrumentation to be agnostic, but I included the link to the comment above.

pellared: My bad. I have missed the invocation of `httpRoute` function!

>  I do not like this approach because I consider the instrumentation to be agnostic, but I included the link to the comment above.

I am also worried about this. @dmathieu, can you please comment on this? How are we sure that it does not affect other instrumentation libraries such as `otelmux`?

EDIT: Also see https://github.com/open-telemetry/opentelemetry-go-contrib/pull/8632#discussion_r3022209084

internal/shared/semconv/server.go.tmpl
```diff
@@ -364,7 +364,9 @@ func (n HTTPServer) MetricAttributes(server string, req *http.Request, statusCod
 	if statusCode > 0 {
 		num++
 	}
-
+	if route == "" && req.Pattern != "" {
```
pellared, line 367: As far as I see, `route` is always empty. 
What do you think of removing the `Route` field from `MetricAttributes` and `route` parameter from this function?
Then this can be simplified to

```go
route := httpRoute(req.Pattern)
if route != "" {
  num++
}
```

gaiaz-iusipov: Here are the places where `MetricAttributes.Route` is used:

* instrumentation/github.com/gin-gonic/gin/otelgin/gin.go:143
* instrumentation/github.com/gorilla/mux/otelmux/mux.go:190
* instrumentation/github.com/labstack/echo/otelecho/echo.go:124

internal/shared/semconv/server.go.tmpl
gaiaz-iusipov: Sure I added a note that this PR does not affect any instrumentation other than `otelhttp`.

internal/shared/semconv/server.go.tmpl
pellared: Thanks. Somehow `gopls` missed this for me :shrug: 

Then, I think it makes even more sense to move the `routeFromRequest` to `otelhttp`...

gaiaz-iusipov: It is because the logic for obtaining the route path differs between `otelhttp` and the other instrumentations that I wanted to bring everything to a consistent form.

dmathieu: I think setting the route based on the pattern, unless it's been set another way first makes sense.
We shouldn't have to handle `req.Pattern` in every package, but packages can override the behavior with their own route if they need.

Issue #8638, opened 2026-03-05
[Bug]: otelgrpc server and client call.duration metrics do not have ExplicitBucketBoundaries specified as per somconv


wzy9607:
### Component

Instrumentation: otelgrpc

### Describe the issue you're facing

Currently, neither otelgrpc nor otel/semconv/v1.39.0/rpcconv have set ExplicitBucketBoundaries for rpc call.duration metrics, resulting the default boundaries for histogram being used by the metrics. 

### Expected behavior

Metrics should be specified with ExplicitBucketBoundaries as per [somconv](https://opentelemetry.io/docs/specs/semconv/rpc/rpc-metrics/)

### Steps to Reproduce

Check the exported result of rpc.server.call.duration and rpc.client.call.duration.

### Operating System

not relevant 

### Device Architecture

x86_64

### Go Version

1.25

### Component Version

v0.66.0

dmathieu:
cc @dashpole as code owner.

dashpole:
This is definitely something we should fix ASAP.

dashpole:
Fix: https://github.com/open-telemetry/opentelemetry-go-contrib/pull/8642

dashpole:
Lets make sure this is in the next release cc @open-telemetry/go-maintainers

PR #8481, opened 2026-01-30
Upgrade `semconv` to `v1.39.0`

MrAlias:
Resolve #8404 

Upgrade `go.opentelemetry.io/otel/semconv` to `v1.39.0`, including updates across all instrumentation and detector modules.

- The semantic conventions v1.39.0 release introduces breaking changes, including:
  - `rpc` span and metric attributes have been renamed to align with naming guidelines:
    - `rpc.system` → `rpc.system.name` (values: `grpc`, `grpc_web`, `connectrpc`, `thrift`, `dubbo`, etc.)
    - `rpc.method` and `rpc.service` have been merged into a fully-qualified `rpc.method` attribute
    - `rpc.client|server.duration` → `rpc.client|server.call.duration` (unit changed to seconds)
    - `rpc.grpc.request.metadata`/`rpc.grpc.response.metadata` → `rpc.request.metadata`/`rpc.response.metadata`
    - `rpc.grpc.status_code` → deprecated in favor of `rpc.response.status_code`
    - `rpc.jsonrpc.request_id` → `jsonrpc.request.id`
    - `rpc.jsonrpc.version` → `jsonrpc.protocol.version`
  - `system` and `process` metrics:
    - `*.linux.memory` metrics renamed to `*.memory.linux`
    - `system.process.status` → `process.state`
    - `system.paging.type` and `process.paging.fault_type` → `system.paging.fault.type`
  - `peer.service` attribute has been deprecated in favor of `service.peer.name`

  See [semantic-conventions v1.39.0 release](https://github.com/open-telemetry/semantic-conventions/releases/tag/v1.39.0) and [v1.38.0 release](https://github.com/open-telemetry/semantic-conventions/releases/tag/v1.38.0) for complete breaking change details.
- Updated modules include:instrumentation/github.com/aws/aws-sdk-go-v2/otelaws/attributes.go
  - `go.opentelemetry.io/contrib/bridges/otellogr`
  - `go.opentelemetry.io/contrib/bridges/otellogrus`
  - `go.opentelemetry.io/contrib/bridges/otelslog`
  - `go.opentelemetry.io/contrib/bridges/otelzap`
  - `go.opentelemetry.io/contrib/instrumentation/net/http/otelhttp`
    - Uses `service.peer.name` attribute (deprecated `peer.service`) in example
  - `go.opentelemetry.io/contrib/instrumentation/net/http/httptrace/otelhttptrace`
[...]

PR #8642, opened 2026-03-05
Use correct bucket boundaries for otelgrpc client and server histograms

dashpole:
Fixes https://github.com/open-telemetry/opentelemetry-go-contrib/issues/8638

It follows:

> This metric SHOULD be specified with [ExplicitBucketBoundaries](https://github.com/open-telemetry/opentelemetry-specification/blob/v1.50.0/specification/metrics/api.md#instrument-advisory-parameters) of [ 0.005, 0.01, 0.025, 0.05, 0.075, 0.1, 0.25, 0.5, 0.75, 1, 2.5, 5, 7.5, 10 ].

The issue was introduced in https://github.com/open-telemetry/opentelemetry-go-contrib/pull/8481, where the unit changed from `ms` to `s` and we didn't catch it.

This impacts the v1.40.0 and v1.41.0 releases.

It is hard to unit test histogram bucket boundaries because metricdatatest.IgnoreValue() also ignores the bucket boundaries. It seems like we could make IgnoreValue not ignore histogram buckets (since those are constant), but that would probably break a lot of existing tests. We don't test bucket boundaries for otelhttp either.

### DIFF
--- a/.github/workflows/ossf-scorecard.yml
+++ b/.github/workflows/ossf-scorecard.yml
@@ -42,6 +42,6 @@ jobs:
       # Upload the results to GitHub's code scanning dashboard (optional).
       # Commenting out will disable upload of results to your repo's Code Scanning dashboard
       - name: "Upload to code-scanning"
-        uses: github/codeql-action/upload-sarif@c793b717bc78562f491db7b0e93a3a178b099162 # v4.32.5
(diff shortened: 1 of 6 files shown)


pellared (approved):

dmathieu:
Merging this without the wait for 24h so we can get this into the upcoming release.

Issue #8659, opened 2026-03-10
[Feature]: Expose the effective SDK resource built from otelconf configuration


iblancasa:
### Component

AutoExporter

### Problem Statement

When otelconf builds an SDK from declarative configuration, downstream users have no supported way to read back the final resource that the SDK is actually using.

This creates a problem for projects that need to surface resource attributes outside the SDK providers (for example, converting resource data into another internal representation).

### Proposed Solution

Expose stable public method to access the effective SDK resource produced by otelconf after configuration is applied.

dmathieu:
cc @pellared @codeboten as code owners.

iblancasa:
I set a wrong component. Sorry.

iblancasa:
Crosslink https://github.com/open-telemetry/opentelemetry-collector/pull/14412#discussion_r2864780421

codeboten:
This feels like something that should be proposed at the spec level, to ensure all implementations have a mechanism to report the effective resource as produced by declarative config. Ideally this spec would also provide answers for questions like this https://github.com/open-telemetry/opentelemetry-go-contrib/pull/8660#discussion_r2915976743

iblancasa:
> This feels like something that should be proposed at the spec level, to ensure all implementations have a mechanism to report the effective resource as produced by declarative config. Ideally this spec would also provide answers for questions like this [#8660 (comment)](https://github.com/open-telemetry/opentelemetry-go-contrib/pull/8660#discussion_r2915976743)

Ok, created https://github.com/open-telemetry/opentelemetry-specification/issues/4948

PR #8660, opened 2026-03-10
Add SDK.Resource() to otelconf/v0.3.0

iblancasa:
Fixes https://github.com/open-telemetry/opentelemetry-go-contrib/issues/8659


otelconf/v0.3.0/config.go
```diff
@@ -70,6 +72,11 @@ func (s *SDK) LoggerProvider() log.LoggerProvider {
 	return s.loggerProvider
 }
 
+// Resource returns the resolved SDK resource configured in this SDK.
+func (s *SDK) Resource() *sdkresource.Resource {
+	return s.resource
```
flc1125: I think we need to confirm whether this `Resource` is read-only or can be modified externally.

iblancasa: `sdkresource.Resource` is immutable, so returning `*sdkresource.Resource` is read-only from the caller's perspective.

flc1125: This is a minimized use case that has been externally modified:

```go
func TestSDK_Resource(t *testing.T) {
	sdk := &SDK{
		resource: sdkresource.NewSchemaless(
			attribute.Bool("key", true),
		),
	}

	res := sdk.Resource()

	assert.ElementsMatch(t, []attribute.KeyValue{attribute.Bool("key", true)}, res.Attributes())

	newRes := sdkresource.NewSchemaless(
		attribute.Bool("key", false),
	)
	*res = *newRes

	assert.ElementsMatch(t, []attribute.KeyValue{attribute.Bool("key", false)}, res.Attributes())
	assert.ElementsMatch(t, []attribute.KeyValue{attribute.Bool("key", false)}, sdk.Resource().Attributes())
}
```

flc1125: Without considering the specification, would it be better to return it like this: 

`func (s *SDK) Resource() sdkresource.Resource {`

iblancasa: Oh, I was totally wrong. I mixed up the upstream immutability contract with Go's pointer aliasing behavior.

iblancasa: How about keeping `func (s *SDK) Resource() *sdkresource.Resource` and returning a defensive copy instead of switching to `func (s *SDK) Resource() sdkresource.Resource`? Returning by value  avoids aliasing too, but `sdkresource.Resource` is typically used as `*resource.Resource`, right? And also its methods are defined on pointer receivers.

CHANGELOG.md
```diff
@@ -10,6 +10,7 @@ This project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0.htm
 
 ### Added
 
+- Add `Resource` method to `SDK` in `go.opentelemetry.io/contrib/otelconf/v0.3.0` to expose the resolved SDK resource from declarative configuration. (#8659)
```
```suggestion
- Add `Resource` method to `SDK` in `go.opentelemetry.io/contrib/otelconf/v0.3.0` to expose the resolved SDK resource from declarative configuration. (#8660)
```

flc1125 (approved):

otelconf/v0.3.0/config.go
```diff
@@ -70,6 +72,15 @@ func (s *SDK) LoggerProvider() log.LoggerProvider {
 	return s.loggerProvider
 }
 
+// Resource returns a copy of the resolved SDK resource configured in this SDK.
```
dmathieu, line 75: I would document why we return a copy, to emphasize that the resource needs to be considered immutable.

iblancasa: I added a note. Can you check if something else should be added? Thanks!

dmathieu (approved):

dmathieu:
We shouldn't merge this until https://github.com/open-telemetry/opentelemetry-specification/issues/4948 has been agreed upon.

otelconf/v0.3.0/config.go
jade-guiton-dd: Is `sdkresource.Resource` not already immutable without a (shallow) copy? I don't see any exported fields or methods that could be modified

iblancasa: > Is `sdkresource.Resource` not already immutable without a (shallow) copy? I don't see any exported fields or methods that could be modified

The concern is not exported mutators, but aliasing the SDK-owned pointer. If we returned the internal `*sdkresource.Resource` directly, a caller could still replace the value via pointer assignment (if I'm not mistaken).

jade-guiton-dd: Ah I see, that makes sense.

iblancasa:
https://github.com/open-telemetry/opentelemetry-specification/issues/4948 has been merged. I think we are ok to merge this one.

Issue #5087, opened 2024-02-13
Potentially incorrect number of read bytes in otelhttp client instrumentation


ash2k:
### Description

From https://github.com/open-telemetry/opentelemetry-go-contrib/issues/4895 and https://github.com/open-telemetry/opentelemetry-go-contrib/pull/5080:

> Using atomic.Int64 will fix the data race but the logical race will still be there i.e. the number of read bytes will be incorrect because the reader goroutine might have not finished reading. My understanding is the value can only be read after Close() has been called on the body. And in that case a plain int64 should be fine.

> I think this fix may lead to incorrect metric reported, see my comment here https://github.com/open-telemetry/opentelemetry-go-contrib/issues/4895#issuecomment-1931787280. This can be checked by having a reader for the body that is slow+a big body and a server that starts responding without consuming all of the incoming request's body.

### Environment

- Architecture: [e.g. x86, i386]
- Go Version: [e.g. 1.15]
- `otelhttp` version: [e.g. v0.14.0, 3c7face]

### Steps To Reproduce

1. Using this code ...
2. Run ...
3. See error ...

### Expected behavior

Number of **reported** bytes read always matches the **actual** number of bytes read.

Xio-Shark:
I’d like to work on this issue.

My understanding is that the reported number of read bytes in `otelhttp` client instrumentation can still be incorrect because the body may still be read asynchronously, so simply changing the counter type does not fully address the logical race.

I plan to first reproduce the issue with a regression test, then investigate where the byte count is finalized and reported, and update the implementation so the reported read bytes match the actual number of bytes read.

Please let me know if this direction looks right.

PR #8684, opened 2026-03-17
fix(otelhttp): report correct read bytes in client instrumentation

Xio-Shark:
## Summary

Fix incorrect read-bytes reporting in `otelhttp` client instrumentation.

## Changes

- Add regression coverage for body-read timing
- Record read bytes only after the body reaches a final state
- Avoid double-recording across the EOF and Close paths

## Motivation

Issue #5087 describes this as a logical race, not a pure thread-safety bug.

## Testing

```bash
go test ./instrumentation/net/http/otelhttp/...
```


Xio-Shark:
Friendly follow-up: EasyCLA already reported success on March 17, 2026, but this PR still has the `blocked: CLA` label. @flc1125 could a maintainer please remove that label so the PR can proceed to normal review? Thanks!

Xio-Shark:
Pushed follow-up commit `890ab4f34` to address the unresolved EOF/Close review concern.

What changed:
- `wrappedBody.Read()` no longer calls `closeBody()` on `io.EOF`
- metrics/span finalization still happens on EOF via `recordMetricsOnce()`
- added a regression test to verify an `io.ReadWriteCloser` remains writable after EOF until the caller explicitly closes it

Local validation:
- `go test . -run 'TestWrappedBodyReadEOFThenCloseRecordsOnce|TestWrappedBodyReadEOFKeepsReadWriteCloserOpenUntilClose|TestTransportProtocolSwitch' -count=1`
- `go test .`

Both passed in `instrumentation/net/http/otelhttp`.

The `blocked: CLA` label is still present even though EasyCLA already reported success on March 17, 2026, so after this code-level review concern is out of the way the remaining blocker should still be maintainer-side label cleanup.

docs/agent-runs/2026-03-17-otelhttp-read-bytes.md
flc1125, line 1: This file is not necessary, please remove it.

Additionally, please read first: https://github.com/open-telemetry/community/blob/main/policies/genai.md#frequently-asked-questions---contributors

flc1125 (changes requested):

Xio-Shark:
Hi @flc1125, I've pushed the follow-up on top of the rebased branch.

Changes included:
- kept upgraded bodies open after EOF
- serialized metrics callback access to avoid the callback/assignment race around recordMetrics
- preserved error metrics on transport failures

Validation:
- in instrumentation/net/http/otelhttp:
  - go build ./...
  - go test . -run 'TestWrappedBodyReadEOFKeepsReadWriteCloserOpenUntilClose|TestTransportMetricsReportFinalRequestBodySize|TestTransportErrorTypeMetricAttribute' -count=1

Xio-Shark:
Hi @flc1125, thanks for the review. I've addressed all three points:

1. **Agent-run doc removed** — The rebased branch no longer contains `docs/agent-runs/2026-03-17-otelhttp-read-bytes.md`.

2. **P1 EOF/Close behavior** — Fixed in `eec1f7d19`: `wrappedBody.Read()` no longer calls `closeBody()` on `io.EOF`. A regression test verifies that upgraded bodies (`io.ReadWriteCloser`) remain writable after EOF until the caller explicitly closes them.

3. **P1 recordMetrics callback race** — Fixed in `2aa72ff1f`: the callback publication path is now serialized with a mutex so the transport write goroutine and `RoundTrip` cannot race on the captured function variable.

4. **Error metrics preservation** — Added in `e91b7421d`: ensures body-read errors are still reflected in `http.client.request.body.size` even when EOF does not trigger close.

I've also read the [OpenTelemetry GenAI policy](https://github.com/open-telemetry/community/blob/main/policies/genai.md#frequently-asked-questions---contributors) you linked.

Could you please take another look when you have a moment? The `blocked: CLA` label is still attached even though EasyCLA reported success on March 17—would appreciate it if a maintainer could clear that as well. Thanks!

Xio-Shark:
Hi @flc1125, friendly reminder that the EasyCLA check was already passed on 2026-03-17 ("The committers listed above are authorized under a signed CLA."). Could you please help remove the "blocked: CLA" label so this PR can move forward to review? Thank you!

Xio-Shark:
Follow-up update: I rebased the branch onto the latest main and pushed `7d6655703709`. I also resolved the previously open review threads after verifying the fixes are present in the current diff: upgraded bodies are no longer closed on EOF, the `recordMetrics` callback publication is synchronized, and the generated agent-run doc has been removed. Local validation passed for the targeted otelhttp tests, golangci-lint, and `git diff --check`. Could you please re-review when you have a chance? Thanks.

Xio-Shark:
Hi @flc1125, friendly ping on this PR. I've addressed all review feedback in the latest rebase (commit 7d665570):

- Upgraded bodies are no longer closed on EOF (P1 concern resolved)
- The `recordMetrics` callback publication is synchronized via `sync.Once`
- All review threads have been resolved

The PR is rebased onto the latest main and ready for re-review. Could you help take another look when you have time? Thank you!

Xio-Shark:
@flc1125 @dmathieu Ready for re-review when you have a moment.

All prior review feedback is addressed in `7d665570`:
- upgraded / duplex bodies are no longer closed on EOF
- `recordMetrics` callback publication is synchronized
- the agent-run doc has been removed

Branch is rebased onto main; local otelhttp tests passed previously. Thanks for another look!

pellared:
Please fix the PR so that it does not report problems with CLA. Please make sure all commits are only authored by yourself (and e.g. not by @tengyanxi)

Moreover, please translate the PR description to English

Xio-Shark:
Thanks @pellared — addressed:

1. **CLA** — Rewrote the changelog commit so all commits on this branch are authored solely by me (`Teng Yanxi` / `151488904+Xio-Shark@users.noreply.github.com`). Removed the Cursor co-author / `tengyanxi@meituan.com` authorship. EasyCLA is now passing.
2. **PR description** — Translated to English.

Could you please remove the `blocked: CLA` label when convenient? Thanks!

instrumentation/net/http/otelhttp/transport.go
```diff
@@ -82,6 +83,47 @@ func defaultTransportFormatter(_ string, r *http.Request) string {
...
+//
+// Close is the authoritative signal, but for known ContentLength values the
+// transport may finish reading slightly before Close runs. Treat that as
+// complete so response finalization can record metrics without racing the
+// Close callback.
+func (w *requestBodyTracker) finished(contentLength int64) bool {
+	if w.Closed() {
+		return true
+	}
+	return contentLength >= 0 && w.BytesRead() >= contentLength
```
MrAlias: `ContentLength == 0` with a non-nil client body means unknown length, not an empty body. A custom streaming body created with `http.NewRequest` commonly takes this path, so an early response makes `finished` return true immediately, records the partial `BytesRead`, and prevents the later `Close` callback from correcting it.

Could we require a positive content length here and add an unknown-length variant of `TestTransportMetricsReportFinalRequestBodySize`?

```suggestion
return contentLength > 0 && w.BytesRead() >= contentLength
```

instrumentation/net/http/otelhttp/transport.go
```diff
@@ -147,38 +218,60 @@ func (t *Transport) RoundTrip(r *http.Request) (*http.Response, error) {
...
-		},
-		t.semconv.MetricOptions(semconv.MetricAttributes{
-			Req:                  r,
-			StatusCode:           statusCode,
-			Err:                  err,
-			AdditionalAttributes: append(labeler.Get(), t.metricAttributesFromRequest(r)...),
-		}),
-	)
+	lastBodyMu.Lock()
+	recordMetrics = fn
```
MrAlias: Requests with no body, and bodies already closed before this callback is published, do not invoke `recordMetrics` here. They now depend on response EOF or `Close`, so a long-lived 101 or CONNECT response can defer `http.client.request.duration` and body-size metrics for the lifetime of the connection.

Could we invoke the callback immediately after publication when `lastTrackedBody` is nil or already finished, while leaving the close callback responsible for an upload still in flight?

instrumentation/net/http/otelhttp/transport.go
```diff
@@ -147,38 +218,60 @@ func (t *Transport) RoundTrip(r *http.Request) (*http.Response, error) {
...
+	lastBodyMu.Unlock()
 
 	if err != nil {
+		callRecordMetrics()
```
MrAlias: `RoundTripper` may close the request body asynchronously after returning, including on errors. Calling `recordMetrics` here wins the `sync.Once` with whatever has been read at return time; the later close callback cannot publish the final size.

Could we route errors through the same body-finalization gate and cover an erroring transport that finishes reading and closes the body after `RoundTrip` returns?

MrAlias (changes requested):
The new finalization gate is not aligned with the `net/http` body lifecycle, so it can either publish a partial request size or defer request metrics for the lifetime of a long-lived response. This should block merging.

Xio-Shark:
Hi @MrAlias, thanks for the review — all three points are addressed in 42c9560:
Applied your suggestion: finished() now requires contentLength > 0, so ContentLength == 0 with a non-nil body is treated as unknown length. Added an unknown-length variant of TestTransportMetricsReportFinalRequestBodySize.
The metrics callback now fires immediately after publication when the request body is nil or already finished; only an in-flight upload defers to the body's Close callback. Added a test asserting metrics are recorded while a long-lived response body is still open.
Errors now go through the same body-finalization gate instead of recording unconditionally. Added a test covering an erroring transport that drains and closes the body after RoundTrip returns.
go build, go vet, and go test ./... -race all pass locally; each new test fails against the previous revision.Could a maintainer also revisit the blocked: CLA label? All commits are now authored solely under my account and EasyCLA reports success. Thanks!

MrAlias (commented):
The finalization gate still treats a declared content length as a close signal. That can publish an incomplete request-body metric.

instrumentation/net/http/otelhttp/transport.go
```diff
@@ -82,6 +83,49 @@ func defaultTransportFormatter(_ string, r *http.Request) string {
...
+// transport may finish reading slightly before Close runs. Treat that as
+// complete so response finalization can record metrics without racing the
+// Close callback. A non-positive contentLength means the length is unknown
+// (e.g. ContentLength == 0 with a non-nil streaming body), so only Close is
+// authoritative in that case.
+func (w *requestBodyTracker) finished(contentLength int64) bool {
+	if w.Closed() {
+		return true
+	}
+	return contentLength > 0 && w.BytesRead() >= contentLength
```
MrAlias: When a body yields more bytes than its declared `ContentLength`, a `RoundTripper` can read through the declared length, return the response, then drain and close `Request.Body` asynchronously. This condition lets the immediate or response-finalization path win `recordMetricsOnce` with the prefix, so the close callback cannot publish the actual read size.

Could we make `Closed` the only completion condition for a non-nil body and add a regression for that lifecycle?

Xio-Shark: Addressed in 6300ee2: Closed is now the only completion condition for a non-nil body (the content-length heuristic is removed entirely), and TestTransportMetricsReportActualSizeOverDeclaredContentLength covers the over-declared-length lifecycle you described — it fails with the declared-length prefix (4096 vs 16384) without the fix.

instrumentation/net/http/otelhttp/transport_test.go
```diff
@@ -932,10 +1087,333 @@ func TestTransportMetrics(t *testing.T) {
...
+	require.NoError(t, err)
+	require.NoError(t, resp.Body.Close())
+
+	select {
+	case <-base.done:
+	case <-time.After(5 * time.Second):
+		t.Fatal("request body was not closed")
+	}
+
+	var rm metricdata.ResourceMetrics
```
dashpole, line 1519: Consider using `require.Eventually` for metric collection here since `base.done` closing can race with the `onClose` callback completing metric recording.

instrumentation/net/http/otelhttp/transport.go
```diff
@@ -147,28 +205,52 @@ func (t *Transport) RoundTrip(r *http.Request) (*http.Response, error) {
...
+		StatusCode:           statusCode,
+		Err:                  err,
+		AdditionalAttributes: append(labeler.Get(), t.metricAttributesFromRequest(r)...),
+	})
+
+	// Delay metric recording until the response body is finalized. The
+	// transport can continue reading the request body after RoundTrip returns.
+	var recordMetricsOnce sync.Once
+	fn := func() {
+		var requestSize int64
```
dashpole: Consider evaluating `requestSize` inside `recordMetricsOnce.Do` to ensure that the metric records the latest byte count if `callRecordMetrics` is triggered concurrently while the body is being read.

instrumentation/net/http/otelhttp/transport.go
```diff
@@ -82,6 +83,34 @@ func defaultTransportFormatter(_ string, r *http.Request) string {
...
+
+func newRequestBodyTracker(body io.ReadCloser, onClose func()) *requestBodyTracker {
+	return &requestBodyTracker{
+		BodyWrapper: request.NewBodyWrapper(body, func(int64) {}),
+		onClose:     onClose,
+	}
+}
+
+func (w *requestBodyTracker) Close() error {
+	err := w.BodyWrapper.Close()
```
MrAlias: `onClose` runs only after the underlying request body's `Close` returns. A `RoundTripper` may close the request body asynchronously after `RoundTrip` returns; if that `Close` blocks after the body has been drained, `Closed` never becomes true. Response EOF or `Close` then also declines to record, so both request metrics can remain absent indefinitely.

Could we mark the tracker closed and trigger finalization before delegating to the potentially blocking `Close`, mirroring `wrappedBody.Close`, and add a regression test with a blocking request-body close?

instrumentation/net/http/otelhttp/transport.go
```diff
@@ -117,22 +146,51 @@ func (t *Transport) RoundTrip(r *http.Request) (*http.Response, error) {
 
 	r = r.Clone(ctx) // According to RoundTripper spec, we shouldn't modify the origin request.
 
-	var lastBW *request.BodyWrapper // Records the last body wrapper. Can be nil.
+	// lastBodyMu protects both lastTrackedBody and recordMetrics from
+	// concurrent access. The request body tracker callback may fire on a
+	// different goroutine (e.g. during transport retries) while the main
+	// goroutine is still assigning recordMetrics after RoundTrip returns.
+	var (
```
MrAlias: The [contrib guide](https://github.com/open-telemetry/opentelemetry-go-contrib/blob/main/CONTRIBUTING.md#contributing-code) adopts the `opentelemetry-go` contributing requirements. Those require [`benchstat` for changes to performance-critical functionality](https://github.com/open-telemetry/opentelemetry-go/blob/main/CONTRIBUTING.md#tests), call for enabled instrumentation to [reduce allocation and computation overhead](https://github.com/open-telemetry/opentelemetry-go/blob/main/CONTRIBUTING.md#performance), and require benchmarks to report `allocs/op`, `B/op`, and `ns/op` for enabled/disabled scenarios.

This new per-request state and its closure graph are paid on every `RoundTrip`, including `nil` and `http.NoBody`; an isolated bodyless benchmark regresses from 24 allocs / 2,965 B to 32 allocs / 3,237 B per request.

Could we refactor this so the feature adds zero allocations over the current implementation for both bodyless and non-empty/streaming bodies? A single tracker/state object can replace the existing `BodyWrapper` allocation, with methods and explicit state replacing the escaping callback graph. Please include `benchstat` for those body shapes and enabled/disabled instrumentation in the PR description.

MrAlias (commented):
The request-body lifecycle still has a finalization path that can suppress client metrics indefinitely, and the implementation adds per-request allocations on a hot instrumentation path. Both should be addressed before merge.

Xio-Shark:
Thanks @MrAlias @dashpole — all four points addressed in e1940fa:
trackedRequestBody.Close finalizes telemetry before delegating to the potentially blocking underlying Close; regression test added.
Request size is now evaluated inside the once-guarded record path.
Metric assertions use require.Eventually.
Replaced the closure graph with a single requestTracker (embeds the first body wrapper; bodyless requests allocate nothing). benchstat vs main, allocs/op: NoBody 85→83, KnownLength 101→99, Streaming 99→97. Full tables in the PR description.

Xio-Shark:
Hi @MrAlias and @dashpole, just a gentle ping — the latest commits on this branch address all the feedback raised so far, and the branch is green (CI + EasyCLA passing).
  Specifically:
 - e1940fa (perf): replaced the per-request closure graph with a single requestTracker, so enabled instrumentation now allocates less than the untracked implementation for
    bodyless, known-length, and streaming bodies. BenchmarkTransportRoundTrip was extended to cover these body shapes with instrumentation both enabled and disabled.
  - e1940fa also finalizes telemetry before delegating to the potentially blocking Close, evaluates request size at record time (inside the once-guarded path), and regression
    tests use require.Eventually.
  - 6300ee2 / 71593e6 / 42c9560: made Close the only completion signal and aligned the finalization gate with the net/http body lifecycle, so a declared ContentLength is no
    longer treated as a close signal.
  - 640acf4 (+ 8f25385): added coverage for the request tracker's GetBody / filter pass-through paths.
 
  Could you take another look when you have a moment? Happy to adjust anything further.

MrAlias (approved):

Xio-Shark:
Thanks @MrAlias for the approval.

I merged the latest `main` into this branch and resolved the conflict in `instrumentation/net/http/otelhttp/transport.go`: kept the request-tracker finalization path, and incorporated `Resp: res` from main so client metrics still pick up the negotiated protocol version. The #8684 changelog entry is back under Unreleased after the 1.45.0 release.

@flc1125 could you re-review or dismiss the April changes-requested review? The agent-run doc was removed and the earlier EOF/callback issues were addressed.

@dmathieu @sonalgaud12 this is ready for a CODEOWNERS look when you have a moment.

sonalgaud12 (approved):

instrumentation/net/http/otelhttp/transport.go
```diff
@@ -82,6 +82,142 @@ func defaultTransportFormatter(_ string, r *http.Request) string {
...
+// requestTracker tracks the request body lifecycle of a single RoundTrip and
+// records the client metrics exactly once, after the request body reaches a
+// final state.
+//
+// A single tracker is allocated per request with a body, replacing the
+// body-wrapper allocation of the untracked implementation: the first request
+// body is wrapped by the embedded first field, and the lifecycle state is
+// held in explicit fields with methods instead of escaping closures. Only
+// GetBody retries allocate an additional trackedRequestBody.
+type requestTracker struct {
```
dashpole, line 94: This seems more complicated than necessary.  Can't we just store the things we want to record, and record the metric when both the round trip is done and body is closed?  I think that should still avoid allocations @MrAlias was initially concerned about. E.g.

```golang
type requestTracker struct {
	mu         sync.Mutex
	read       atomic.Int64
	bodyClosed bool
	rtDone     bool

	// Fields set when RoundTrip returns (zero closure allocations)
	ctx      context.Context
	duration time.Duration
	opts     semconv.MetricOpts
	t        *Transport
}

func (rt *requestTracker) onRoundTripDone(ctx context.Context, d time.Duration, opts semconv.MetricOpts) {
	rt.mu.Lock()
	rt.ctx, rt.duration, rt.opts = ctx, d, opts
	rt.rtDone = true
	if rt.bodyClosed {
		rt.record()
	}
	rt.mu.Unlock()
}

func (rt *requestTracker) onBodyClose() {
	rt.mu.Lock()
	rt.bodyClosed = true
	if rt.rtDone {
		rt.record()
	}
	rt.mu.Unlock()
}

func (rt *requestTracker) record() {
	rt.t.semconv.RecordMetrics(
		rt.ctx,
		semconv.MetricData{RequestSize: rt.read.Load(), RequestDuration: rt.duration},
		rt.opts,
	)
}
```

Xio-Shark: Hi @dashpole, thanks for the suggestion!

The current implementation actually follows this exact mental model (gating metric recording on both RoundTrip completion and body closure). The reason `requestTracker` has the extra structure is to address a few specific edge cases and constraints identified during earlier reviews:

1. **Bodyless / `nil` requests**: For requests without a body (e.g. standard `GET`), `onBodyClose` is never invoked. The check `(body == nil || body.closed.Load())` allows `publish` to record metrics immediately without waiting for a close event that will never fire.
2. **`GetBody` retries**: When the transport retries a request via `GetBody`, a new `io.ReadCloser` is generated. Storing `first` and updating `rt.body` dynamically ensures we track the active body attempt's bytes and closure correctly.
3. **Zero-allocation hot path**: Embedding the initial `trackedRequestBody` directly in `requestTracker.first` satisfies @MrAlias's earlier constraint to eliminate per-request heap allocations.
4. **Preventing stalls on blocking `Close`**: Telemetry is finalized right before delegating to the underlying body's `Close()`, mirroring `wrappedBody.Close` behavior so a blocking underlying close doesn't delay client metrics.

If you feel this can still be streamlined further while preserving these edge cases, I'm more than happy to adjust! Otherwise, if this context clarifies why the structure is needed, please let me know.

dashpole: Thanks. I had missed the GetBody retry requirement, and it makes sense now.

dashpole (approved):

CHANGELOG.md
```diff
@@ -16,10 +16,6 @@ This project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0.htm
 
 - Deprecate `go.opentelemetry.io/contrib/samplers/probability/consistent`. (#9633)
 
-### Fixed
-
-- Format span attributes in `go.opentelemetry.io/contrib/zpages` using `attribute.Value.String` instead of the deprecated `attribute.Value.Emit`, following the OpenTelemetry AnyValue representation for non-OTLP protocols. (#9453)
```
dmathieu, line 21: Why is this removed?

CHANGELOG.md
```diff
@@ -29,6 +25,7 @@ This project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0.htm
 
 - Format span attributes in `go.opentelemetry.io/contrib/zpages` using `attribute.Value.String` instead of the deprecated `attribute.Value.Emit`, following the OpenTelemetry AnyValue representation for non-OTLP protocols. (#9453)
 - Set `error.type` on the span and on the request-duration, request-body-size, and response-body-size metrics when a client disconnects mid-request in `go.opentelemetry.io/contrib/instrumentation/github.com/gin-gonic/gin/otelgin`, using the request context's cancellation error as the classification source when the handler has not already recorded an error via `c.Error`. Previously a disconnect was recorded with span status `Error` and no `error.type`, indistinguishable from a genuine server fault. (#9394)
+- Fix `http.client.request.body.size` recording for streaming request bodies in `go.opentelemetry.io/contrib/instrumentation/net/http/otelhttp`. (#8684)
```
dmathieu, line 28: This shouldn't be in the removed section.

CHANGELOG.md
Xio-Shark: Merged latest `main` to align with #9640; section structure is restored and no existing entries are removed.

CHANGELOG.md
Xio-Shark: Moved under `### Fixed`. The diff against `main` is now cleanly this single addition.

dmathieu (approved):

PR #8719, opened 2026-03-23
otelzap: set error field as record.SetErr 

iblancasa:
Fixes #8718

`go test -run='^$' -bench='BenchmarkSemanticConvWrite$' -benchmem -count=20`

```text
goos: darwin
goarch: arm64
pkg: go.opentelemetry.io/contrib/bridges/otelzap
cpu: Apple M4 Pro
                                                       │ enabled-before.txt │ enabled-after.txt │
                                                       │                                       sec/op                                       │                          sec/op                            vs base                │
ErrorFieldEnabledCompare/ErrorField-14                                                                                          69.54n ± 1%                                                  60.85n ± 2%   -12.49% (p=0.000 n=20)
ErrorFieldEnabledCompare/NamedErrorField-14                                                                                     70.91n ± 2%                                                  74.30n ± 1%    +4.78% (p=0.000 n=20)
ErrorFieldEnabledCompare/AnyErrorField-14                                                                                       69.06n ± 0%                                                  59.18n ± 1%   -14.31% (p=0.000 n=20)
ErrorFieldEnabledCompare/ErrorFieldWithStack-14                                                                                 75.84n ± 1%                                                  65.56n ± 0%   -13.56% (p=0.000 n=20)
ErrorFieldEnabledCompare/TenFieldsWithError-14                                                                                  317.7n ± 1%                                                  313.9n ± 3%        ~ (p=0.087 n=20)
ErrorFieldEnabledCompare/TenFieldsWithNamedError-14                                                                             320.4n ± 1%                                                  327.1n ± 5%    +2.11% (p=0.000 n=20)
ErrorFieldEnabledCompare/TenFieldsWithErrorAndStack-14                                                                          334.2n ± 2%                                                  324.1n ± 1%    -3.02% (p=0.000 n=20)
geomean                                                                                                                         136.4n                                                       128.7n         -5.67%

                                                       │ enabled-before.txt │ enabled-after.txt │
                                                       │                                        B/op                                        │                           B/op                             vs base                │
ErrorFieldEnabledCompare/ErrorField-14                                                                                           112.0 ± 0%                                                  112.0 ± 0%       ~ (p=1.000 n=20) ¹
ErrorFieldEnabledCompare/NamedErrorField-14                                                                                      112.0 ± 0%                                                  112.0 ± 0%       ~ (p=1.000 n=20) ¹
ErrorFieldEnabledCompare/AnyErrorField-14                                                                                        112.0 ± 0%                                                  112.0 ± 0%       ~ (p=1.000 n=20) ¹
ErrorFieldEnabledCompare/ErrorFieldWithStack-14                                                                                  112.0 ± 0%                                                  112.0 ± 0%       ~ (p=1.000 n=20) ¹
ErrorFieldEnabledCompare/TenFieldsWithError-14                                                                                 1.042Ki ± 0%                                                1.010Ki ± 0%    -3.07% (p=0.000 n=20)
ErrorFieldEnabledCompare/TenFieldsWithNamedError-14                                                                            1.018Ki ± 0%                                                1.018Ki ± 0%       ~ (p=1.000 n=20) ¹
ErrorFieldEnabledCompare/TenFieldsWithErrorAndStack-14                                                                         1.064Ki ± 0%                                                1.018Ki ± 0%    -4.40% (p=0.000 n=20)
geomean                                                                                                                          293.2                                                        290.0         -1.08%
¹ all samples are equal

                                                       │ enabled-before.txt │ enabled-after.txt │
                                                       │                                     allocs/op                                      │                         allocs/op                           vs base               │
ErrorFieldEnabledCompare/ErrorField-14                                                                                           3.000 ± 0%                                                   3.000 ± 0%       ~ (p=1.000 n=20) ¹
ErrorFieldEnabledCompare/NamedErrorField-14                                                                                      3.000 ± 0%                                                   3.000 ± 0%       ~ (p=1.000 n=20) ¹
ErrorFieldEnabledCompare/AnyErrorField-14                                                                                        3.000 ± 0%                                                   3.000 ± 0%       ~ (p=1.000 n=20) ¹
ErrorFieldEnabledCompare/ErrorFieldWithStack-14                                                                                  3.000 ± 0%                                                   3.000 ± 0%       ~ (p=1.000 n=20) ¹
ErrorFieldEnabledCompare/TenFieldsWithError-14                                                                                   12.00 ± 0%                                                   12.00 ± 0%       ~ (p=1.000 n=20) ¹
ErrorFieldEnabledCompare/TenFieldsWithNamedError-14                                                                              12.00 ± 0%                                                   12.00 ± 0%       ~ (p=1.000 n=20) ¹
ErrorFieldEnabledCompare/TenFieldsWithErrorAndStack-14                                                                           12.00 ± 0%                                                   12.00 ± 0%       ~ (p=1.000 n=20) ¹
geomean                                                                                                                          5.434                                                        5.434        +0.00%
¹ all samples are equal
```


dmathieu:
Could you provide benchmarks? Using reflect is quite slow, and this is a pretty hot path since it executes for every log entry.

iblancasa:
> Could you provide benchmarks? Using reflect is quite slow, and this is a pretty hot path since it executes for every log entry.

Sure. Will do.

iblancasa:
To avoid user issues, I decided to make this thing optional (disabled by default).

dmathieu:
Could you run the benchmarks before and after your change with benchstat (https://pkg.go.dev/golang.org/x/perf/cmd/benchstat) and add the results to the PR description?

iblancasa:
> Could you run the benchmarks before and after your change with benchstat (https://pkg.go.dev/golang.org/x/perf/cmd/benchstat) and add the results to the PR description?

Done

khushijain21 (approved):
LGTM

bridges/otelzap/core.go
```diff
@@ -114,13 +126,23 @@ func WithLoggerProvider(provider log.LoggerProvider) Option {
...
 }
 
+// WithExceptionSemanticConventions returns an [Option] that configures a [Core]
+// to map zap error fields to exception semantic convention attributes.
+func WithExceptionSemanticConventions() Option {
+	return optFunc(func(c config) config {
+		c.exceptionSemConv = true
+		return c
+	})
+}
```
pellared: I am not sure if we need such option. Do you suspect somebody would not like to use it? Note that we are already doing something similar for `context.Context`.

iblancasa: I understand this can have some performance impact (check the PR description). So, just to avoid impacting users... I created it as an option.

pellared: Ack. I am going to need more time to analyze it.

dmathieu:
Looking at your PR, I suspect what's slow is getting the exception type. 
Rather than an option that fully removes the exception attribute (which everyone will indeed likely want to enable), we make the option only enable/disable the exception type, and always emit the exception message?

iblancasa:
> Looking at your PR, I suspect what's slow is getting the exception type. Rather than an option that fully removes the exception attribute (which everyone will indeed likely want to enable), we make the option only enable/disable the exception type, and always emit the exception message?

Let me measure how it would be that approach. I think it makes sense.

bridges/otelzap/core.go
```diff
@@ -239,19 +272,57 @@ func (o *Core) Write(ent zapcore.Entry, fields []zapcore.Field) error {
...
 			ctx = ctxFld
 			continue
 		}
+		if o.exceptionSemConv && field.Type == zapcore.ErrorType {
+			err := field.Interface.(error)
+			if !hasException {
+				enc.root.attrs = slices.Grow(enc.root.attrs, 2)
+				enc.root.attrs = append(enc.root.attrs,
+					log.String(exceptionMessageKey, err.Error()),
+					exceptionType(err),
```
pellared: How about returning an `error` and calling `r.SetErr`, instead of using a `bool` and setting attributes? This would align more closely with the `context.Context` pattern.

The original intent behind `r.SetErr` was to avoid requiring bridges to set attributes themselves. It would also give SDK users more control over whether to emit things like `error.type` (even though that’s not currently supported).

bridges/otelzap/core.go
pellared: Even though it hits the performance, I would personally prefer to emit the `error.type` by default as I guess that for most users the speed penalty is worth the additional attribute.

CHANGELOG.md
```diff
@@ -12,6 +12,7 @@ This project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0.htm
 
 - Configuration file can now be set via `OTEL_CONFIG_FILE` in `go.opentelemetry.io/contrib/otelconf`. (#8639)
 - Added support for `service` resource detector in `go.opentelemetry.io/contrib/otelconf`. (#8674)
+- Add `WithExceptionSemanticConventions` option to `go.opentelemetry.io/contrib/bridges/otelzap` to emit `exception.type` from zap error fields; `exception.message` is emitted by default when error fields are present. (#8719)
```
pellared: This leaks SDK implementation details. Some API implementations may handle the errors differently.

iblancasa: I see. Was not sure about mentioning the details or not. Will change. Thanks!

pellared:
What are the benchmarks for `BenchmarkCoreWrite` if the option is set? I guess the impact would be close to none.

bridges/otelzap/core.go
pellared: Note also that this only impacts the logs that have errors. This is also why I am leaning towards not having this option unless there is a big demand for it. It feels YAGNI and I prefer decreasing the API surface.

iblancasa:
> What are the benchmarks for `BenchmarkCoreWrite` if the option is set? I guess the impact would be close to none.

```
goos: darwin
goarch: arm64
pkg: go.opentelemetry.io/contrib/bridges/otelzap
cpu: Apple M4 Pro
                                 │ corewrite-disabled.txt │ corewrite-enabled.txt │
                                 │         sec/op         │         sec/op         vs base              │
CoreWriteOption/10_fields-14               293.0n               293.9n       ~ (p=0.645 n=8)
CoreWriteOption/20_fields-14               603.1n               603.1n       ~ (p=0.878 n=8)
CoreWriteOption/Namespace-14               467.5n               464.6n       ~ (p=0.442 n=8)
CoreWriteOption/With10_fields-14           57.39n               57.18n       ~ (p=0.938 n=8)
CoreWriteOption/With20_fields-14           159.1n               160.9n       ~ (p=0.279 n=8)
CoreWriteOption/WithNamespace-14           2.169n               1.995n  -8.02% (p=0.002 n=8)
geomean                                     108.5n               107.1n       -1.30%

                                 │ corewrite-disabled.txt │ corewrite-enabled.txt │
                                 │          B/op          │          B/op          vs base                │
CoreWriteOption/10_fields-14              978.0                978.0       ~ (p=1.000 n=8)
CoreWriteOption/20_fields-14            2.129Ki              2.129Ki      ~ (p=1.000 n=8)
CoreWriteOption/Namespace-14            1.719Ki              1.719Ki      ~ (p=1.000 n=8)
CoreWriteOption/With10_fields-14          208.0                208.0       ~ (p=1.000 n=8)
CoreWriteOption/With20_fields-14          640.0                640.0       ~ (p=1.000 n=8)
CoreWriteOption/WithNamespace-14          0.000                0.000       ~ (p=1.000 n=8)

                                 │ corewrite-disabled.txt │ corewrite-enabled.txt │
                                 │       allocs/op        │       allocs/op        vs base                │
CoreWriteOption/10_fields-14              12.00                12.00       ~ (p=1.000 n=8)
CoreWriteOption/20_fields-14              20.00                20.00       ~ (p=1.000 n=8)
CoreWriteOption/Namespace-14              15.00                15.00       ~ (p=1.000 n=8)
CoreWriteOption/With10_fields-14          1.000                1.000       ~ (p=1.000 n=8)
CoreWriteOption/With20_fields-14          1.000                1.000       ~ (p=1.000 n=8)
CoreWriteOption/WithNamespace-14          0.000                0.000       ~ (p=1.000 n=8)
```

pellared (changes requested):
I do not think we need a new option: https://github.com/open-telemetry/opentelemetry-go-contrib/pull/8719#discussion_r3015382892

iblancasa:
> I do not think we need a new option: [#8719 (comment)](https://github.com/open-telemetry/opentelemetry-go-contrib/pull/8719#discussion_r3015382892)

Ok, I can revert it. WDYT @dmathieu ?

bridges/otelzap/core.go
```diff
@@ -239,19 +265,40 @@ func (o *Core) Write(ent zapcore.Entry, fields []zapcore.Field) error {
...
 			ctx = ctxFld
 			continue
 		}
+		if field.Type == zapcore.ErrorType {
+			if errField == nil {
+				errField = field.Interface.(error)
+			}
+			if field.Key == "error" {
+				continue
+			}
```
pellared: The current implementation looks buggy. A "random" named error can override an error with `error` name.

I think this should be

```suggestion
		if field.Type == zapcore.ErrorType && field.Key == "error" {
			errField = field.Interface.(error)
			continue
		}
```

1. I think that we should only convert `error` fields produced by https://pkg.go.dev/go.uber.org/zap#Error
2. It would be good to add unit tests that make sure that other named errors are serialized as before. We should e.g. add a unit test where this is only a named error (e.g. `db.error`).

iblancasa: Totally makes sense. thanks!

CHANGELOG.md
```diff
@@ -8,6 +8,10 @@ This project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0.htm
 
 ## [Unreleased]
 
+### Added
+
+- Emit zap error fields using OpenTelemetry exception semantic conventions by default in `go.opentelemetry.io/contrib/bridges/otelzap`. (#8719)
```
pellared: Can you move this to **Changed** and modify the entry to something similar to

> - Set error field as `record.SetErr` instead of a plain attribute in `go.opentelemetry.io/contrib/bridges/otellogrus`. (#8776)

E.g.

```suggestion
- Set the "error" field (e.g. created via `zap.Error`) as `record.SetErr` instead of a plain attribute in `go.opentelemetry.io/contrib/bridges/otelzap`. (#8719)
```

iblancasa: Oh, sure

pellared (approved):

dmathieu (approved):

bridges/otelzap/bench_test.go
```diff
@@ -107,3 +108,62 @@ func BenchmarkCoreWrite(b *testing.B) {
...
+		{name: "TenFieldsWithErrorAndStack", entry: errorEntry, fields: append(append([]zapcore.Field{}, tenFields...), zap.Error(testErr))},
+	}
+
+	for i := range benchmarks {
+		bm := &benchmarks[i]
+		b.Run(bm.name, func(b *testing.B) {
+			zc := NewCore(loggerName)
+			b.ReportAllocs()
+			b.ResetTimer()
+			for range b.N {
```
flc1125:
```suggestion
			for b.Loop {
```

see: https://go.dev/blog/testing-b-loop

flc1125 (approved):

PR #8782, opened 2026-04-08
Auto-ping codeowners on issues

dmathieu:
This takes the [ping code owners on issues](https://github.com/open-telemetry/opentelemetry-collector/blob/main/.github/workflows/ping-codeowners-issues.yml) action from the collector and applies it to this repository, so code owners pinging can happen automatically.


dmathieu:
After replacing the `gh comment create` with an `echo gh comment create`:

> open-telemetry/opentelemetry-go-contrib›  git:(issues-ping-codeowners) 1M COMPONENT='instrumentation: otelgrpc' ISSUE=8781 ./tools/ping_codeowners_issues.sh
gh issue comment 8781 --body Pinging code owners for instrumentation: otelgrpc: @dashpole.

dmathieu:
@open-telemetry/go-approvers ping for review?

dashpole (approved):

.github/workflows/ping-codeowners-issues.yml
```diff
@@ -0,0 +1,22 @@
...
+  ping-owners:
+    permissions:
+      issues: write
+    runs-on: ubuntu-24.04
+    if: ${{ github.repository_owner == 'open-telemetry' }}
+    steps:
+      - uses: actions/checkout@de0fac2e4500dabe0009e67214ff5f5447ce83dd # v6.0.2
+
+      - name: Run ping-codeowners-issues.sh
+        run: ./.github/workflows/scripts/ping-codeowners-issues.sh
```
MrAlias: This step calls `./.github/workflows/scripts/ping-codeowners-issues.sh`, but the script added in this patch is `tools/ping_codeowners_issues.sh`. As written, the workflow will fail with “No such file or directory” and never post owner pings.

```suggestion
        run: ./tools/ping_codeowners_issues.sh
```

Is the workflow able to run scripts outside of `.github`?

tools/ping_codeowners_issues.sh
```diff
@@ -0,0 +1,33 @@
...
+    exit 0
+fi
+
+CUR_DIRECTORY=$(dirname "$0")
+
+# Labels are formatted as "<type>: <component>" (e.g. "instrumentation: otelgrpc").
+# Extract the component name (the part after ': ') and resolve the full path
+# from CODEOWNERS, since get-codeowners.sh matches by path prefix.
+COMPONENT_NAME=$(echo "${COMPONENT}" | sed 's/^[^:]*:[[:space:]]*//')
+COMPONENT_PATH=$(grep -E "/${COMPONENT_NAME}[/[:space:]]" CODEOWNERS | awk '{ print $1 }' | sed 's|/$||' | head -1 || true)
```
MrAlias: `COMPONENT_NAME` is interpolated directly into `grep -E`, so label text is being treated as a regular expression. That can cause the script to match the wrong `CODEOWNERS` entry for labels containing regex metacharacters. Please escape the value or switch to fixed-string matching.

MrAlias: This reads `CODEOWNERS` via a repo-relative path. That works with the current workflow layout, but it is brittle and depends on the job running from the repository root. It would be safer to resolve `CODEOWNERS` from the script directory instead.

MrAlias (dismissed):
The script path needs to be updated.

flc1125 (approved):

PR #8784, opened 2026-04-08
split Handler and Transport options for type safety

### LINKED ISSUE #8107: Audit otelhttp API
We're starting to look into stabilizing the [go.opentelemetry.io/contrib/instrumentation/net/http/otelhttp](https://pkg.go.dev/go.opentelemetry.io/contrib/instrumentation/net/http/otelhttp) package.

For that, we need to audit the public API of the package, not for missing things (as those could be added later) but for things that should be changed before we can't do breaking changes anymore.

For reference, this is the API as it is with the latest release:

```
func ContextWithLabeler(parent context.Context, l *Labeler) context.Context
func ContextWithStartTime(parent context.Context, start time.Time) context.Context
func Get(ctx context.Context, targetURL string) (resp *http.Response, err error)
func Head(ctx context.Context, targetURL string) (resp *http.Response, err error)
func NewHandler(handler http.Handler, operation string, opts ...Option) http.Handler
func NewMiddleware(operation string, opts ...Option) func(http.Handler) http.Handler
func Post(ctx context.Context, targetURL, contentType string, body io.Reader) (resp *http.Response, err error)
func PostForm(ctx context.Context, targetURL string, data url.Values) (resp *http.Response, err error)
func StartTimeFromContext(ctx context.Context) time.Time
func Version() string
func WithRouteTag(route string, h http.Handler) http.Handler
type Filter
type Labeler
    func LabelerFromContext(ctx context.Context) (*Labeler, bool)
    func (l *Labeler) Add(ls ...attribute.KeyValue)
    func (l *Labeler) Get() []attribute.KeyValue

type Option
    func WithClientTrace(f func(context.Context) *httptrace.ClientTrace) Option
    func WithFilter(f Filter) Option
    func WithMessageEvents(events ...event) Option
    func WithMeterProvider(provider metric.MeterProvider) Option
    func WithMetricAttributesFn(metricAttributesFn func(r *http.Request) []attribute.KeyValue) Option
    func WithPropagators(ps propagation.TextMapPropagator) Option
    func WithPublicEndpoint() Option
    func WithPublicEndpointFn(fn func(*http.Request) bool) Option
    func WithServerName(server string) Option
    func WithSpanNameFormatter(f func(operation string, r *http.Request) string) Option
    func WithSpanOptions(opts ...trace.SpanStartOption) Option
    func WithTracerProvider(provider trace.TracerProvider) Option

type Transport
    func NewTransport(base http.RoundTripper, opts ...Option) *Transport
    func (t *Transport) RoundTrip(r *http.Request) (*http.Response, error)
```

This issue is open for any comment opinions before that stabilization.

## Critical Issues (Must Fix Before Stabilization)

- [ ] #8130
- [ ] [**Stabilization of error recording semantic conventions is a blocker**](https://github.com/open-telemetry/opentelemetry-go-contrib/issues/8107#issuecomment-3480832460) - Need to wait for stabilization of [error recording spec](https://github.com/open-telemetry/semantic-conventions/blob/main/docs/general/recording-errors.md#recording-errors-on-spans) before proceeding.
- [ ] [**Decide on span events vs log events and `WithMessageEvents()` implications**](https://github.com/open-telemetry/opentelemetry-go-contrib/issues/8107#issuecomment-3473157668) - Need to determine if span events should migrate to log events prior to stabilization, possibly behind a feature flag. This impacts the `WithMessageEvents()` option.
- [ ] [**Consider renaming `NewMiddleware`**](https://github.com/open-telemetry/opentelemetry-go-contrib/issues/8107#issuecomment-3481823580)
- [ ] [**Consider renaming `Transport`**](https://github.com/open-telemetry/opentelemetry-go-contrib/issues/8107#issuecomment-3481829841)
- [x] #8131
- [x] #8132

## Important Issues (Should Fix)

- [x] [**Deprecate and remove `WithRouteTag` option**](https://github.com/open-telemetry/opentelemetry-go-contrib/issues/8107#issuecomment-3476669317) - This option can be replaced with `WithMetricAttributesFn` for controlling `http.route` in metrics.
  - https://github.com/open-telemetry/opentelemetry-go-contrib/pull/8117
- [ ] [**Rethink `Labeler` type design and API integration**](https://github.com/open-telemetry/opentelemetry-go-contrib/issues/8107#issuecomment-3481788277) - The `Labeler` type is for metrics but isn't clearly connected to the rest of the API. Its purpose and representation need to be reconsidered.
- [ ] [**Consider splitting `Option` types for Handler and Transport**](https://github.com/open-telemetry/opentelemetry-go-contrib/issues/8107#issuecomment-3481794034) - Handler and Transport currently share the same option type, but options like `WithClientTrace` and `WithSpanNameFormatter` only apply to one. Should split if needed.
- [ ] [**Reconsider `ContextWithStartTime` / `StartTimeFromContext` as public API**](https://github.com/open-telemetry/opentelemetry-go-contrib/issues/8107#issuecomment-3481799512) - These functions are strange to have as part of (just) HTTP instrumentation and may not belong in the public API.
- [ ] [**Address issues #726 and #4501 before stabilization**](https://github.com/open-telemetry/opentelemetry-go-contrib/issues/8107#issuecomment-3477965937) - These issues significantly alter output behavior and should be resolved before declaring otelhttp stable.
- [ ] [**Consider renaming `Labeler` to `MetricLabeler`**](https://github.com/open-telemetry/opentelemetry-go-contrib/issues/8107#issuecomment-3473157668) - Since the godoc states it only applies to metrics, the name could be more specific. Alternatively, consider adding attributes to spans as well.
- [ ] [**Consider renaming `WithFilter` to `WithSpanFilter`**](https://github.com/open-telemetry/opentelemetry-go-contrib/issues/8107#issuecomment-3473157668) - Since the filter only applies to spans, a more specific name would be clearer.
- [ ] [**Consider refactoring the `filters` package**](https://github.com/open-telemetry/opentelemetry-go-contrib/issues/8107#issuecomment-3481806104) - Move filters to options in `otelhttp` or as other declarations.
- [ ] [**Decide if `Filter` needs to accept other parameters**](https://github.com/open-telemetry/opentelemetry-go-contrib/issues/8107#issuecomment-3481816310) - Address extensibility
- [ ] [**Ensure future extensibility**](https://github.com/open-telemetry/opentelemetry-go-contrib/issues/8107#issuecomment-3481941867) - HTTP/2, HTTP/3, WebSockets
- [ ] [**Rethink `WithMetricAttributesFunc`**](https://github.com/open-telemetry/opentelemetry-go-contrib/issues/8107#issuecomment-3481968468) - cache or redesign

## Nice to Have

- [ ] [**Refactor or rename `common.go` file**](https://github.com/open-telemetry/opentelemetry-go-contrib/issues/8107#issuecomment-3481782759) - The file is overly generic and not representative of what it contains.
- [ ] [**Comprehensive package documentation**](https://github.com/open-telemetry/opentelemetry-go-contrib/issues/8107#issuecomment-3481946201)
- [ ] [**Public API benchmarks**](https://github.com/open-telemetry/opentelemetry-go-contrib/issues/8107#issuecomment-3481992860)

## Context Notes

- Issue #2782 about generalist vs specific instrumentation is **not** resolved by this stabilization effort - any changes would introduce breaking changes and the issue hasn't generated momentum. Can be considered for a future v2.

NesterovYehor:
This PR addresses the v1.0 stabilization audit (Fixes #8107) by strictly separating `HandlerOption` and `TransportOption` in `otelhttp`. 

Previously, `NewHandler` and `NewTransport` shared the same `Option` type. This allowed users to accidentally pass transport-specific options (like `WithClientTrace`) to a server handler without compiler errors, resulting in silent runtime failures. This update enforces type safety at the API boundary.

## Changes Made
- Split the internal configuration into `handlerConfig` and `transportConfig`.
- Introduced distinct `HandlerOption` and `TransportOption` interfaces.
- Added a shared `Option` interface for common configurations (e.g., `WithPropagators`).
- Updated the constructors `NewHandler` and `NewTransport` to accept their respective, restrictive option interfaces.


instrumentation/net/http/otelhttp/config.go
```diff
@@ -18,45 +18,103 @@ import (
...
-	Propagators       propagation.TextMapPropagator
-	SpanStartOptions  []trace.SpanStartOption
-	PublicEndpointFn  func(*http.Request) bool
-	ReadEvent         bool
-	WriteEvent        bool
-	Filters           []Filter
-	SpanNameFormatter func(string, *http.Request) string
-	ClientTrace       func(context.Context) *httptrace.ClientTrace
+// handlerConfig represents the configuration options available for the http.Handler.
+type handlerConfig struct {
```
dmathieu: How about adding a `sharedConfig` struct that contains all the shared things, and the two configs include?

instrumentation/net/http/otelhttp/config.go
```diff
@@ -126,10 +212,8 @@ func WithSpanOptions(opts ...trace.SpanStartOption) Option {
 // If no filters are provided then all requests are traced.
 // Filters will be invoked for each processed request, it is advised to make them
 // simple and fast.
-func WithFilter(f Filter) Option {
-	return optionFunc(func(c *config) {
-		c.Filters = append(c.Filters, f)
-	})
+func WithFilter(f Filter) HandlerOption {
```
flc1125: This seems to have omitted support for `transportConfig`

instrumentation/net/http/otelhttp/config_test.go
```diff
@@ -16,6 +18,22 @@ import (
 	"go.opentelemetry.io/contrib/instrumentation/net/http/otelhttp"
 )
 
+func TestNewHandlerAndTransportOptions(t *testing.T) {
```
flc1125: Currently, it's not quite clear what this test is intended to verify.

CHANGELOG.md
```diff
@@ -11,6 +11,7 @@ This project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0.htm
 ### Changed
 
 - Set error field as `record.SetErr` instead of a plain attribute in `go.opentelemetry.io/contrib/bridges/otellogrus`. (#8776)
+- Split `otelhttp.Option` into `HandlerOption` and `TransportOption` in `go.opentelemetry.io/contrib/instrumentation/net/http/otelhttp`. (#8784)
```
flc1125: Since `Option` has been changed, it is best to have a description that explains the breaking changes.

flc1125 (commented):
For other reviewers' reference: `Option` has been changed, and there is a breaking change that needs to be assessed for acceptability.

pellared:
> For other reviewers’ reference: `Option` has been changed, introducing a breaking change that needs to be evaluated.

I agree with @flc1125. This should be handled as a migration effort spread over 2–3 releases rather than a single-step change.

@NesterovYehor, could we keep `Option` for now and mark it as deprecated? The existing behavior would continue to support it, while explicit usage would trigger warnings (e.g., via linters or IDEs). This would give users time to gradually refactor before the breaking change is enforced.

We might explore approaches like type embedding or type aliasing here. Could you sketch out what such a transition might look like?

NesterovYehor:
@pellared,
Thanks for the guidance! I haven't done a soft migration like this before, so I really appreciate the feedback.

To ensure we don't break downstream we can:

1. Keep NewHandler and NewTransport accepting the legacy ...Option signature for this release to prevent slice-breakage.
2. Mark the legacy Option interface (and the legacy With... functions) as // Deprecated.
3. Introduce the strict HandlerOption and TransportOption interfaces for the new API.
4. Inside NewHandler/NewTransport, use an internal adapter to seamlessly map the legacy Option values into the new strict config structs.

This way, existing code will compile flawlessly, but the // Deprecated tags will trigger the IDE/linter warnings you mentioned. In a future release, we can update the constructors to enforce the strict types.

Does this migration plan look good to you?

flc1125:
@NesterovYehor It might be more helpful for the next steps if we can outline the key changes in the upcoming few versions.

NesterovYehor:
Phase 1: 
- Introduce the new strict HandlerOption and TransportOption interfaces.
- Keep NewHandler and NewTransport accepting the legacy ...Option signature for now to prevent immediate slice-breakage.
- Mark the legacy Option interface and the specific legacy With... functions as // Deprecated.
- Add an internal adapter inside the constructors to map legacy options to the new strict configs seamlessly.

Impact: 100% backward compatible. Existing code compiles, but users get linter/IDE warnings on deprecated types to start refactoring.

Phase 2: Strict Enforcement (in 2–3 releases as @pellared suggested)
- Update NewHandler and NewTransport signatures to strictly accept ...HandlerOption and ...TransportOption.
- Update the specific With... functions to return the strict types.

Impact: Users who haven't migrated will experience build breaks here, but they had a full release cycle of deprecation warnings to prepare.

Phase 3: Cleanup (Future Release)

- Remove the legacy Option interface completely.
- Remove the internal adapters and legacy bridging code.

Does this versioning strategy align with the project's standard deprecation cycle?

flc1125:
> * Mark the legacy Option interface and the specific legacy With... functions as // Deprecated.

Does this impose a greater intrusion on the user?

---

What about this? Would it have less impact on the user:

```go
type HandlerOption interface {
	Option
	
	applyHandlerConfig(*handlerConfig)
}

type TransportOption interface {
	Option
	
	applyTransportConfig(*transportConfig)
}
```

NesterovYehor:
I am not sure that it will cause less impact, I think in the end it will be the same result until we change params of NewHandler and NewTransport to new option interfaces. But this can be implemented cleaner which I like. The only thing is that new interfaces will have to implement the apply method of Option, but it is still less than writing an adapter type.

flc1125:
> `NewHandler` and `NewTransport` to new option interfaces.

I do not deny this point; however, for users, if we can make the split as imperceptible as possible, it would be a relatively gentle situation. For most users, `NewXX` and `WithXX` are the more commonly used ways to configure things.

instrumentation/net/http/otelhttp/config.go
```diff
@@ -127,9 +213,10 @@ func WithSpanOptions(opts ...trace.SpanStartOption) Option {
 // Filters will be invoked for each processed request, it is advised to make them
 // simple and fast.
 func WithFilter(f Filter) Option {
-	return optionFunc(func(c *config) {
-		c.Filters = append(c.Filters, f)
-	})
+	return sharedOption{
+		handlerFunc:   func(c *handlerConfig) { c.Filters = append(c.Filters, f) },
+		transportFunc: func(c *transportConfig) { c.Filters = append(c.Filters, f) },
+	}
```
sonalgaud12, line 219: Should WithFilter remain shared, or should we introduce WithHandlerFilter/WithTransportFilter to complete the type-safety story?

NesterovYehor:
Looking at config.go, the bridging logic makes things a bit heavy and verbose right now, but once we fully deprecate and remove the legacy Option interface in future updates, all this mapping code can be ripped out, which will shrink and simplify the package significantly.

I made NewHandler and NewTransport still accept ...Option to support existing users' code. I added an internal adapter using applyLegacy. This allows us to apply the new HandlerOption and TransportOption directly, while quietly mapping the old legacy options during initialization.

Because this conversion logic only runs once during the initial setup, there is absolutely zero performance impact on the hot request and response paths.

I'm totally open to suggestions if anyone has a better idea or sees a cleaner way to handle this transition!

pellared:
@NesterovYehor, have you pushed new commits?

instrumentation/net/http/otelhttp/config.go
pellared: I think shared option is fine.

CHANGELOG.md
```diff
@@ -18,6 +18,10 @@ This project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0.htm
 ### Changed
 
 - Set error field as `record.SetErr` instead of a plain attribute in `go.opentelemetry.io/contrib/bridges/otellogrus`. (#8776)
+- Introduces breaking changes by refactoring `go.opentelemetry.io/contrib/instrumentation/net/http/otelhttp` to enforce distinct API boundaries between server and client instrumentation. (#8784)
+  - The `Option` interface has been split into `HandlerOption` and `TransportOption` to provide compile-time safety for configuration options.
+  - Internal configuration structures were refactored into `sharedConfig`, `handlerConfig`, and `transportConfig` to isolate server and client-specific settings.
```
pellared: Internals are not important for the end users.

CHANGELOG.md
```diff
@@ -18,6 +18,10 @@ This project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0.htm
...
+  - `NewHandler` and `NewTransport` now strictly accept their respective `HandlerOption` or `TransportOption` types.
```
pellared: We could also call out how the `With*` function changed.

CHANGELOG.md
```diff
@@ -18,6 +18,10 @@ This project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0.htm
 ### Changed
 
 - Set error field as `record.SetErr` instead of a plain attribute in `go.opentelemetry.io/contrib/bridges/otellogrus`. (#8776)
+- Introduces breaking changes by refactoring `go.opentelemetry.io/contrib/instrumentation/net/http/otelhttp` to enforce distinct API boundaries between server and client instrumentation. (#8784)
+  - The `Option` interface has been split into `HandlerOption` and `TransportOption` to provide compile-time safety for configuration options.
```
pellared: What about `CombinedOption`?

CHANGELOG.md
```diff
@@ -18,6 +18,10 @@ This project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0.htm
 ### Changed
 
 - Set error field as `record.SetErr` instead of a plain attribute in `go.opentelemetry.io/contrib/bridges/otellogrus`. (#8776)
+- Introduces breaking changes by refactoring `go.opentelemetry.io/contrib/instrumentation/net/http/otelhttp` to enforce distinct API boundaries between server and client instrumentation. (#8784)
```
```suggestion
- Refactor `go.opentelemetry.io/contrib/instrumentation/net/http/otelhttp` options to enforce distinct API boundaries between server and client instrumentation. (#8784)
```

instrumentation/net/http/otelhttp/config.go
```diff
@@ -38,28 +69,157 @@ type config struct {
...
+// CombinedOption is a combination of HandlerOption and TransportOption.
+// It is used for options that configure fields shared between handler and transport configs.
+type CombinedOption interface {
+	HandlerOption
+	TransportOption
+}
+
+// Option defines the legacy interface for setting optional configuration properties.
+//
+// Deprecated: Use HandlerOption or TransportOption instead. This interface will be removed in a future release.
```
pellared:
```suggestion
// Deprecated: Use [HandlerOption] or [TransportOption] instead. This interface will be removed in a future release.
```

instrumentation/net/http/otelhttp/config.go
```diff
@@ -38,28 +69,157 @@ type config struct {
...
+}
+
+// TransportOption defines an option for configuring an HTTP transport.
+// It extends the Option interface with transport-specific application logic.
+type TransportOption interface {
+	Option
+	applyTransport(*transportConfig)
+}
+
+// CombinedOption is a combination of HandlerOption and TransportOption.
```
pellared:
```suggestion
// CombinedOption is a combination of [HandlerOption[ and [TransportOption].
```

instrumentation/net/http/otelhttp/config.go
```diff
@@ -38,28 +69,157 @@ type config struct {
...
+// It is used for options that configure fields shared between handler and transport configs.
```
pellared:
```suggestion
// It is implemented by options that are applicable for both [NewHandler] and [NewTransport].
```

instrumentation/net/http/otelhttp/config.go
```diff
@@ -38,28 +69,157 @@ type config struct {
 	MetricAttributesFn func(*http.Request) []attribute.KeyValue
 }
 
-// Option interface used for setting optional config properties.
+// HandlerOption defines an option for configuring an HTTP handler.
+// It extends the Option interface with handler-specific application logic.
```
pellared:
```suggestion
```

instrumentation/net/http/otelhttp/config.go
```diff
@@ -38,28 +69,157 @@ type config struct {
...
+type HandlerOption interface {
+	Option
+	applyHandler(*handlerConfig)
+}
+
+// TransportOption defines an option for configuring an HTTP transport.
+// It extends the Option interface with transport-specific application logic.
```
pellared:
```suggestion
```

CHANGELOG.md
pellared: This is not addressed

PR #8785, opened 2026-04-08
docs(otelhttptrace): add performance guidance for WithoutSubSpans

imorte:
Hello 👋

We operate a GraphQL service that fans out each incoming request to ~20-30 upstream services. With default sub-span behavior, each outbound HTTP request creates up to 7 additional `trace.Span` heap objects. At scale this caused continuously growing memory usage - resolved by switching to `WithoutSubSpans()`. A separate service with identical tracing code but only 2-3 outbound calls per request had no issues, confirming the problem is specific to fan-out volume

Bildschirmfoto 2026-04-08 um 17 47 09

The current `WithoutSubSpans()` GoDoc explains what it does but not why - this PR adds performance guidance to help users in similar scenarios find the option before hitting the problem


instrumentation/net/http/httptrace/otelhttptrace/clienttrace.go
```diff
@@ -136,6 +148,13 @@ type clientTracer struct {
 // added as attributes to spans, although several headers will be automatically
 // redacted: Authorization, WWW-Authenticate, Proxy-Authenticate,
 // Proxy-Authorization, Cookie, and Set-Cookie.
+//
+// Sub-span creation carries non-trivial overhead: each outbound request may
```
dmathieu: While providing better doc on the option itself is good, I'm not sure we should mention it here. Otherwise, the constructor's doc could mention every option we have.

flc1125: Agreed, I would prefer to investigate the cause in order to fix this memory leak issue.

imorte: > Otherwise, the constructor's doc could mention every option we have.

Fair enough. We could remove this from the `NewClientTrace()` function and keep it for `WithoutSubSpans()`. What do you think?

imorte: > Agreed, I would prefer to investigate the cause in order to fix this memory leak issue.

I did a pretty thorough audit of the code and couldn't find an actual memory leak. What I believe is happening is memory pressure from the sheer volume of span objects (up to 7 per outbound request x many upstreams), which under sustained load outpaces Go's GC and looks like a leak but isn't one

imorte: pushed fix

dmathieu (dismissed):

imorte:
rebased with main

dmathieu:
Well, your rebase brought back the original comment on ClientTrace.
Also, please don't force-push, even on PRs.

imorte:
> Well, your rebase brought back the original comment on ClientTrace. Also, please don't force-push, even on PRs.

Should be okay now, we only missing "Skip Changelo" label

dmathieu (approved):

imorte:
@sonalgaud12 hey! could you please approve and merge this one as well?

sonalgaud12 (approved):

imorte:
@dmathieu Hello! Could you merge this one as well? 🙏

pellared (approved):

PR #8797, opened 2026-04-10
otelhttptrace: fix header attributes lost when using sub-spans

imorte:
When subspans are enabled (default), `wroteHeaderField()` writes header attributes to `ct.root` which points to the already ended `http.getconn` subspan. `SetAttributes` on an ended span is a silent noop, so all `http.request.header.*` attributes are quietly dropped.

This works correctly with `WithoutSubSpans()` because `ct.root` is set to the parent span from context, which is still active.

The fix sets `ct.root` to the parent span from context in both modes


dmathieu:

imorte:
> Could you add a changelog entry?

done

dmathieu:
With this change, we set ct.root twice in tge same method, both with the span from context.
It seems like we could set it only once, at the start of the method?

imorte:
> With this change, we set ct.root twice in tge same method, both with the span from context. It seems like we could set it only once, at the start of the method?

fixed

dmathieu (approved):

imorte:
Hello, @dmathieu I don't have rights to merge this PR. Just want to make sure that is't not lost here

dmathieu:
We require 2 approvals to merge.

sonalgaud12 (approved):

PR #8834, opened 2026-04-16
Fix flaky jaeger remote test

pellared:
The TestRemotelyControlledSampler_ImmediatelyUpdateOnStartup test in the jaegerremote sampler was using a fixed 100ms sleep to wait for a background goroutine to update the sampler's state. This is a common source of flakiness, as the sleep might be too short on slow CI environments.

I have replaced the fixed sleep with `assert.Eventually`, which polls for the condition and returns as soon as it's met, up to a 1s timeout. This makes the test both faster (when the update is quick) and more stable (allowing more time if needed).

I also added proper read-locking when checking the sampler's state during polling to ensure thread safety against the background update goroutine. Finally, I corrected a minor typo in a nearby comment.


rsmihir389:
Hey — I've been seeing this exact pattern a lot. Flaky tests like this usually cost teams $30-40 per rerun in CI time + developer context-switching.

I built a tool that detects flaky tests and shows the dollar cost: [kleore.com](https://www.kleore.com). Here's what a report looks like on a real repo: [sample report](https://www.kleore.com/report/cmn0knika00x7dc29p23yyha6).

60-second install as a GitHub App, free tier shows the cost. Might help here.

PR #8835, opened 2026-04-16
test(otelconf): replace expiring TLS fixtures with runtime certs

MrAlias:
## Summary
This fixes the repo-wide CI failures caused by expired `otelconf` TLS test certificates.

The failing gRPC exporter tests in `otelconf`, `otelconf/x`, and `otelconf/v0.3.0` were using static PEM fixtures that expired on April 15, 2026, which caused `main` and unrelated PRs to start failing with x509 certificate validity errors.

This change replaces those expiring test fixtures with runtime-generated CA, server, and client certificates created under `t.TempDir()`.

## Changes
- add a shared `otelconf/internal/testtls` helper to generate fresh test-only TLS materials at runtime
- update the gRPC log, metric, and trace exporter TLS tests in `otelconf`
- update the same TLS tests in `otelconf/x`
- update the same TLS tests in `otelconf/v0.3.0`

## Why this approach
Refreshing the checked-in PEM files would only defer the outage until the next expiry date. Generating certificates during the test run removes the calendar dependency while preserving the TLS and mTLS coverage those tests are meant to exercise.

## Verification
- `go test ./... -run "^Test_otlpGRPC(Log|Metric|Trace)Exporter$"` from `otelconf/`
- reran the same focused exporter suite with `GOARCH=386`

Both focused runs passed after the change.

### DIFF
--- a/otelconf/internal/testtls/testtls.go
+++ b/otelconf/internal/testtls/testtls.go
@@ -2,7 +2,7 @@
 // SPDX-License-Identifier: Apache-2.0
 
 // Package testtls provides runtime-generated TLS materials for tests.
-package testtls
+package testtls // import "go.opentelemetry.io/contrib/otelconf/internal/testtls"
 
 import (
 	"crypto/rand"
--- a/otelconf/internal/testtls/testtls.go
+++ b/otelconf/internal/testtls/testtls.go
@@ -1,0 +1,149 @@
+// Copyright The OpenTelemetry Authors
+// SPDX-License-Identifier: Apache-2.0
+
+// Package testtls provides runtime-generated TLS materials for tests.
+package testtls
+
+import (
+	"crypto/rand"
+	"crypto/rsa"
+	"crypto/x509"
+	"crypto/x509/pkix"
+	"encoding/pem"
+	"math/big"
+	"net"
+	"os"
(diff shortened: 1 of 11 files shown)


dmathieu:
It seems `config_test` still references those certificates. How about using the helper for them as well?
https://github.com/open-telemetry/opentelemetry-go-contrib/blob/main/otelconf/config_test.go#L458-L460

Then we could remove the expired certs from the repo.

codeboten (approved):

MrAlias:
> It seems `config_test` still references those certificates. How about using the helper for them as well? https://github.com/open-telemetry/opentelemetry-go-contrib/blob/main/otelconf/config_test.go#L458-L460
> 
> Then we could remove the expired certs from the repo.

Those tests only read the CA cert into a a root pool, it does not perform the server/client handshake paths that are failing CI.

If you want, I can look into updating all other places that reference the `testdata/*.crt`. It's going to expand the PR scope a bit though.

dashpole (approved):

dashpole:
I'm ok with the current scope to get tests fixed.

pellared (approved):
Let's merge it and address the comment in a separate PR.

PR #8844, opened 2026-04-20
Use dynamic certs everywhere in otelconf and remove the static ones

dmathieu:
This is a follow-up to https://github.com/open-telemetry/opentelemetry-go-contrib/pull/8835, using the dynamic certificate everywhere in otelconf.


dashpole (approved):
I didn't review the diffs in config_test.go files very closely. I think it is all identical, but github has a hard time rendering the diff

MrAlias (approved):

Issue #8210, opened 2025-11-21
otelconf: add support for `log_level`


codeboten:
### Component

Config

### Problem Statement

As per suggestion in https://github.com/open-telemetry/opentelemetry-go-contrib/pull/8026#discussion_r2548363259

### Proposed Solution

Use the `LogLevel` field and call `SetLogger` with the appropriate level

CtrlSpice:
Hi! Could I get assigned to this please?

flc1125:
@CtrlSpice It's yours.

codeboten:
Thanks @CtrlSpice!

PR #8847, opened 2026-04-20
otelconf: Add log_level support

sonalgaud12:
#8210

Wires up the LogLevel config field to the OTel SDK internal diagnostics logger via a new Logger() method on SDK.


otelconf/config_yaml.go
```diff
@@ -75,8 +75,8 @@ func (j *OpenTelemetryConfiguration) UnmarshalYAML(node *yaml.Node) error {
 	if sh.Disabled != nil {
 		sh.Plain.Disabled = sh.Disabled
 	} else {
-		// Configure the log level of the internal logger used by the SDK.
-		// If omitted, info is used.
+		// Configure if the SDK is disabled or not.
+		// If omitted, false is used.
```
codeboten: This could be submitted as a separate PR, but since its just comments its probably ok to have here

otelconf/config_yaml.go
```diff
@@ -97,10 +97,6 @@ func (j *OpenTelemetryConfiguration) UnmarshalYAML(node *yaml.Node) error {
 
 	if sh.LogLevel != nil {
 		sh.Plain.LogLevel = sh.LogLevel
-	} else {
-		// Configure the log level of the internal logger used by the SDK.
-		// If omitted, info is used.
-		sh.Plain.LogLevel = ptr(SeverityNumberInfo)
```
codeboten: The configuration schema states that info should be used if omitted: https://github.com/open-telemetry/opentelemetry-configuration/blob/d5f421535e704562a5b26a2edd9cf40317af7899/opentelemetry_configuration.json#L20

codeboten (commented):
Thanks @sonalgaud12, apologies for the delay in reviewing. One question around the default when log_level is not present

CHANGELOG.md
```diff
@@ -98,6 +98,7 @@ The next release will require at least [Go 1.26].
 
 ### Added
 
+- Add `log_level` support to `otelconf` to configure the OTel SDK internal logger. (#8847)
```
dashpole: This entry seems to have landed under the already-released 1.44.0 section. Consider moving it under the [Unreleased] section at the top of the file, and using the full module path go.opentelemetry.io/contrib/otelconf.

otelconf/config_test.go
```diff
@@ -399,6 +399,20 @@ func TestNewSDK(t *testing.T) {
...
+			cfg: []ConfigurationOption{
+				WithContext(t.Context()),
+				WithOpenTelemetryConfiguration(OpenTelemetryConfiguration{
+					LogLevel: ptr(SeverityNumber("invalid_level")),
+				}),
+			},
+			wantErr:            newErrInvalid(fmt.Sprintf("unsupported log_level %q", "invalid_level")),
+			wantTracerProvider: tracenoop.NewTracerProvider(),
+			wantMeterProvider:  metricnoop.NewMeterProvider(),
+			wantLoggerProvider: lognoop.NewLoggerProvider(),
```
dashpole: TestNewSDK asserts the other provider getters on SDK, but does not check sdk.Logger(). Consider adding an assertion for sdk.Logger() across the test cases (such as checking sdk.Logger().GetSink() is non-nil for configured cases and nil for noop/disabled).

PR #8850, opened 2026-04-21
otelconf: emit AdditionalProperties as map[string]any for mapstructure ,remain

### LINKED ISSUE #8842: [Bug]: AdditionalProperties type causes mapstructure decode error with ",remain"
### Component

Config

Generated structs in `otelconf` (e.g., `SpanProcessor`, `LogRecordProcessor`, `SpanExporter`, etc.) define the `AdditionalProperties` field as `interface{}` with the `mapstructure:",remain"` tag. 
The `mapstructure` decoder explicitly rejects non-map types for the `,remain` tag and throws a hard decode error. This breaks configuration loading for any application using `mapstructure` (directly or via config loaders) and prevents forward-compatibility with custom/extension config keys.

### Expected behavior

The JSON Schema → Go generator should emit `map[string]any` for fields tagged with `,remain` to comply with `mapstructure` requirements. The generated field should look like:
```go
AdditionalProperties map[string]any `mapstructure:",remain" json:"-" yaml:"-"`
```

This would allow the decoder to correctly capture and preserve unspecified configuration keys without failing.

### Steps to Reproduce

Run this minimal reproduction:

```go
package main

import (
	"fmt"

	"github.com/go-viper/mapstructure/v2"
)

type SpanProcessor struct {
	Batch                interface{} `mapstructure:"batch,omitempty"`
	AdditionalProperties interface{} `mapstructure:",remain"`
}

func main() {
	var cfg SpanProcessor
	raw := map[string]any{
		"batch":     map[string]any{},
		"extra_key": "value",
	}

	err := mapstructure.Decode(raw, &cfg)
	if err != nil {
		fmt.Printf("Decode failed: %v\n", err)
	}
    fmt.Printf("extra_key:%s", cfg.AdditionalProperties["extra_key"])
```
* Actual result: panic: reflect: Key of non-map type interface {})
* Expected result: extra_key:value

### Operating System

mac os

### Device Architecture

ARM64

### Go Version

1.26

### Component Version

go.opentelemetry.io/contrib/otelconf/x master

SAY-5:
## Summary

Fixes #8842.

`go-jsonschema` generates every schema-open struct with:

```go
AdditionalProperties interface{} `mapstructure:",remain"`
```

`mapstructure/v2`'s `,remain` decoder reflects on the destination field via `reflect.MapOf` and hard-fails when that field is a bare `interface{}`:

```
panic: reflect: Key of non-map type interface {}
```

This breaks any consumer - including the OpenTelemetry Collector's config loader - that routes YAML/JSON through `mapstructure` into `SpanProcessor`, `LogRecordProcessor`, `SpanExporter`, etc. Forward-compat for unknown config keys goes with it.

## Fix

1. Add a sed rule to `otelconf/jsonschema_patch.sed` that rewrites the generated field to `map[string]any` and hides it from JSON/YAML round-trips:

```go
AdditionalProperties map[string]any `mapstructure:",remain" json:"-" yaml:"-"`
```

2. Apply the same rewrite to the two committed generated files (`otelconf/generated_config.go`, `otelconf/x/generated_config.go`) so they match what the patched generator produces.

The existing internal assignment site in `config_common.go` (`plain.AdditionalProperties = raw` where `raw` is a `map[string]any`) and the existing test fixture in `config_test.go` (`AdditionalProperties: map[string]any{...}`) are already compatible with the new field type, so no consumer code changes.

The `json:"-" yaml:"-"` tags prevent the helper bucket from being emitted on round-trip: otelconf serialises only the named schema keys, and the catch-all stays a decode-side convenience - matching the collector's current expectations.

## Test plan

- [x] `go build ./...` in `otelconf/`, `otelconf/x/`, `otelconf/v0.2.0/`, `otelconf/v0.3.0/`
- [x] `go test ./...` across the otelconf module tree (no regressions)
- [ ] Regenerating from upstream schema (maintainers, locally): `make genjsonschema` produces an unchanged diff against this PR's files, confirming the sed rule is the source of truth


dashpole:
Can you add a test?  It would help me understand what is broken, and prevent regressions.

CHANGELOG.md
```diff
@@ -23,6 +23,7 @@ The next release will require at least [Go 1.26].
 - Reject OTLP exporter headers with an empty `name` in `go.opentelemetry.io/contrib/otelconf`, `go.opentelemetry.io/contrib/otelconf/x`, and `go.opentelemetry.io/contrib/otelconf/v0.3.0`, instead of forwarding invalid header names to OTLP exporters. (#9102)
 - `go.opentelemetry.io/contrib/detectors/aws/lambda` no longer returns an error when run outside of an AWS Lambda environment, matching the no-op behavior of other resource detectors. (#9464)
 - Convert Prometheus untyped metrics to OTLP Gauges in `go.opentelemetry.io/contrib/bridges/prometheus`. (#9099)
+- Type `AdditionalProperties` as `map[string]any` in the generated config structs of `go.opentelemetry.io/contrib/otelconf` and `go.opentelemetry.io/contrib/otelconf/x` so `mapstructure`'s `,remain` decoder no longer fails on a bare `interface{}`. (#8842)
```
dashpole: Please update this to reference the PR number `(#8850)` instead of the issue number `(#8842)`:
```suggestion
- Type `AdditionalProperties` as `map[string]any` in the generated config structs of `go.opentelemetry.io/contrib/otelconf` and `go.opentelemetry.io/contrib/otelconf/x` so `mapstructure`'s `,remain` decoder no longer fails on a bare `interface{}`. (#8850)
```

PR #8865, opened 2026-04-24
otelhttp: Add test scenarios for the 'no filters' case

Leshana:
No actual changes to behavior, simply an additional test scenario for the sake of clarity and completeness.

I don't think the function names Any/All/None are actually ambiguous or require further documentation, but mixing up how they apply to an empty set of criteria is an easy mistake to make.  Adding a test scenario spells it out explicitly to anyone who wonders and makes sure it doesn't change later by accident.


instrumentation/net/http/otelhttp/filters/filters_test.go
```diff
@@ -94,6 +106,12 @@ func TestNone(t *testing.T) {
 			req:    &http.Request{URL: &url.URL{Path: "/foo", Host: "bar.baz:8080"}},
 			exp:    false,
 		},
+		{
```
sonalgaud12, line 109: One small suggestion: It would be helpful to add an inline comment explaining the expected semantics.

Otherwise, LGTM.

dmathieu (approved):

dashpole (approved):

MrAlias (approved):

PR #8868, opened 2026-04-25
fix(instrumentation): http fallback method to _OTHER

### LINKED ISSUE #8866: http instrumentation should fall back to method _OTHER
We currently fall back to GET: https://github.com/open-telemetry/opentelemetry-go-contrib/blob/c17e09fe67f429ee048e6c9cf46e3c32e0758e0b/instrumentation/net/http/otelhttp/internal/semconv/client.go#L168-L175

https://github.com/open-telemetry/opentelemetry-go-contrib/blob/c17e09fe67f429ee048e6c9cf46e3c32e0758e0b/instrumentation/net/http/otelhttp/internal/semconv/server.go#L272-L285

We should instead fall back to `_OTHER` according to [http semconv](https://github.com/open-telemetry/semantic-conventions/blob/0e301ab9d72afbd962425bd85de1d33611b92c16/docs/http/http-spans.md?plain=1#L183-L184):

> If the HTTP request method is not known to instrumentation, it MUST set the `http.request.method` attribute to `_OTHER`.

This applies to all http instrumentation libraries, and requires updating the templates for those attributes.

Rahul-Baradol:
Fixes #8866

This pull request standardizes the handling of unknown or empty HTTP methods across multiple OpenTelemetry instrumentation packages. Instead of defaulting to "GET" for unknown or empty HTTP methods, the code now uses the "_OTHER" value as specified by the OpenTelemetry semantic conventions. Corresponding unit tests have also been updated to reflect this behavior.

**Standardization of HTTP method attribute values:**

* Updated the `method` helper functions in client and server semantic conventions for all instrumentations (`otelrestful`, `otelgin`, `otelmux`, `otelecho`, `otelhttptrace`, and `otelhttp`) to return `semconv.HTTPRequestMethodOther` ("_OTHER") instead of `semconv.HTTPRequestMethodGet` ("GET") when the HTTP method is unknown or empty. 

**Test updates to reflect new behavior:**

* Updated unit tests in all affected packages to expect "_OTHER" for unknown or empty HTTP methods instead of "GET". This ensures tests accurately verify the new standardized behavior. 

These changes improve compliance with OpenTelemetry semantic conventions and make HTTP method reporting more accurate and explicit across all supported frameworks.


dmathieu:

dashpole (approved):
lgtm after changelog added

MrAlias (commented):
This looks correct to me for the behavior change in `#8866`, but I think there’s still one worthwhile test gap before merge.

The PR updates the unknown/empty-method fallback in every generated `client.go`, while the new assertions added here are server-side only. `internal/shared/semconv/client_test.go.tmpl` still doesn’t exercise the new client fallback path, which also matches the remaining Codecov gap on this PR.

I’d suggest adding two client-path assertions in the shared client test template so the coverage applies across all generated instrumentations:
- `req.Method == ""` should produce `http.request.method = "_OTHER"` with no `http.request.method_original`
- an unknown mixed-case method like `"Unknown"` should produce `http.request.method = "_OTHER"` and `http.request.method_original = "Unknown"`

With that and the changelog entry already mentioned on the PR, I don’t have other concerns.

pellared:
@Rahul-Baradol, can you address the comments?

Rahul-Baradol:
Yes, will do it shortly

CHANGELOG.md
```diff
@@ -18,6 +18,7 @@ This project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0.htm
 - Set error field as `record.SetErr` instead of a plain attribute in `go.opentelemetry.io/contrib/bridges/otellogrus`. (#8776)
 - Set the "error" field (e.g. created via `zap.Error`) as `record.SetErr` instead of a plain attribute in `go.opentelemetry.io/contrib/bridges/otelzap`. (#8719)
 - Set emitted errors in `go.opentelemetry.io/contrib/bridges/otellogr` as record errors (`Record.SetErr`) instead of `exception.message` attributes. (#8775)
+- Unknown or empty HTTP methods now report "_OTHER" instead of "GET" across all HTTP instrumentations to align with OpenTelemetry semantic conventions. Updated related unit tests accordingly.
```
```suggestion
- Unknown or empty HTTP methods now report "_OTHER" instead of "GET" across all HTTP instrumentations to align with OpenTelemetry semantic conventions. (#8868)
```

MrAlias (approved):

pellared (approved):

PR #8869, opened 2026-04-25
feat(instrumentation): http server span supports url query

### LINKED ISSUE #8867: http server spans should include `url.query`
It doesn't look like url.query is supported by http trace instrumentation.

From [semconv](https://github.com/open-telemetry/semantic-conventions/blob/0e301ab9d72afbd962425bd85de1d33611b92c16/docs/http/http-spans.md?plain=1#L479):

> | [`url.query`](/docs/registry/attributes/url.md) | Stable | `Conditionally Required` If and only if one was received/sent. | string | The [URI query](https://www.rfc-editor.org/rfc/rfc3986#section-3.4) component [10] | `q=OpenTelemetry` |

I would expect this to be added next to url.path: https://github.com/open-telemetry/opentelemetry-go-contrib/blob/c17e09fe67f429ee048e6c9cf46e3c32e0758e0b/instrumentation/net/http/otelhttp/internal/semconv/server.go#L201-L203

This will require updating the template for http semconv.

Rahul-Baradol:
Fixes #8867 

This pull request enhances HTTP server instrumentation across multiple packages by adding support for capturing the `url.query` attribute in trace data, while also redacting sensitive query parameters. The changes ensure consistency in both production code and tests, and improve security by preventing sensitive data from being logged.

Key changes include:

**Security and Trace Attribute Improvements:**

* Updated the `RequestTraceAttrs` function in several HTTP server instrumentation packages (including `go-restful`, `gin-gonic`, `gorilla/mux`, `labstack/echo`, and standard library HTTP) to add the `url.query` attribute to trace spans. Sensitive query parameters such as `AWSAccessKeyId`, `Signature`, `sig`, and `X-Goog-Signature` are now redacted before being recorded. 

**Testing Consistency:**

* Modified test files in all affected packages to expect the new `url.query` attribute, ensuring test coverage for the updated trace data. 

These changes provide better visibility into HTTP requests while protecting sensitive information in query parameters.

### DIFF
--- a/internal/shared/semconv/server.go.tmpl
+++ b/internal/shared/semconv/server.go.tmpl
@@ -201,6 +201,24 @@ func (n HTTPServer) RequestTraceAttrs(server string, req *http.Request, opts Req
 	if req.URL != nil && req.URL.Path != "" {
 		attrs = append(attrs, semconv.URLPath(req.URL.Path))
 	}
+	
+	if req.URL != nil && req.URL.Query() != nil {
+		query := req.URL.Query()
+		sensitiveKeys := map[string]struct{}{
+			"AWSAccessKeyId":   {},
+			"Signature":        {},
+			"sig":              {},
+			"X-Goog-Signature": {},
+		}
+
+		for key := range query {
+			if _, isSensitive := sensitiveKeys[key]; isSensitive {
+				query.Set(key, "REDACTED")
+			}
+		}
+
+		attrs = append(attrs, semconv.URLQuery(query.Encode()))
+	}
 
 	if protoName != "" && protoName != "http" {
 		attrs = append(attrs, semconv.NetworkProtocolName(protoName))
--- a/instrumentation/github.com/emicklei/go-restful/otelrestful/internal/semconv/httpconvtest_test.go
+++ b/instrumentation/github.com/emicklei/go-restful/otelrestful/internal/semconv/httpconvtest_test.go
@@ -40,6 +40,7 @@ func TestNewTraceRequest(t *testing.T) {
 			attribute.String("client.address", req.clientIP),
 			attribute.String("network.protocol.version", "1.1"),
 			attribute.String("url.path", "/"),
+			attribute.String("url.query", ""),
 		}
 	}
 	testTraceRequest(t, serv, want)
--- a/instrumentation/github.com/emicklei/go-restful/otelrestful/internal/semconv/server.go
+++ b/instrumentation/github.com/emicklei/go-restful/otelrestful/internal/semconv/server.go
@@ -201,6 +201,24 @@ func (n HTTPServer) RequestTraceAttrs(server string, req *http.Request, opts Req
 	if req.URL != nil && req.URL.Path != "" {
 		attrs = append(attrs, semconv.URLPath(req.URL.Path))
 	}
+	
+	if req.URL != nil && req.URL.Query() != nil {
+		query := req.URL.Query()
+		sensitiveKeys := map[string]struct{}{
+			"AWSAccessKeyId":   {},
+			"Signature":        {},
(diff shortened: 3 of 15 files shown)


internal/shared/semconv/server.go.tmpl
```diff
@@ -201,6 +201,24 @@ func (n HTTPServer) RequestTraceAttrs(server string, req *http.Request, opts Req
 	if req.URL != nil && req.URL.Path != "" {
 		attrs = append(attrs, semconv.URLPath(req.URL.Path))
 	}
+	
+	if req.URL != nil && req.URL.Query() != nil {
+		query := req.URL.Query()
+		sensitiveKeys := map[string]struct{}{
+			"AWSAccessKeyId":   {},
```
sonalgaud12, line 208: I am not sure if we should hard-code this in instrumentation

internal/shared/semconv/server.go.tmpl
```diff
@@ -201,6 +201,24 @@ func (n HTTPServer) RequestTraceAttrs(server string, req *http.Request, opts Req
 	if req.URL != nil && req.URL.Path != "" {
 		attrs = append(attrs, semconv.URLPath(req.URL.Path))
 	}
+	
+	if req.URL != nil && req.URL.Query() != nil {
```
sonalgaud12, line 205: Will need to test this

internal/shared/semconv/server.go.tmpl
dmathieu: We definitely shouldn't have proprietary keys like this hardcoded in here. There could be an option to set this manually.

Rahul-Baradol: Sure, will see what I can do

dmathieu: IMHO, splitting that into 2 PRs (where the first one doesn't sanitize anything) would be better.

Rahul-Baradol: Got it!

dashpole: FYI these keys come from the semantic conventions directly, but are marked as in-development: https://opentelemetry.io/docs/specs/semconv/http/http-spans/

> Query string values for the following keys SHOULD be redacted by default and replaced by the value REDACTED:

> [AWSAccessKeyId](https://docs.aws.amazon.com/AmazonS3/latest/userguide/RESTAuthentication.html#RESTAuthenticationQueryStringAuth)
[Signature](https://docs.aws.amazon.com/AmazonS3/latest/userguide/RESTAuthentication.html#RESTAuthenticationQueryStringAuth)
[sig](https://learn.microsoft.com/azure/storage/common/storage-sas-overview#sas-token)
[X-Goog-Signature](https://cloud.google.com/storage/docs/access-control/signed-urls)
This list is subject to change over time.

> Matching of query parameter keys against the sensitive list SHOULD be case-sensitive.

> Instrumentation MAY provide a way to override this list via declarative configuration. If so, it SHOULD use the sensitive_query_parameters property (an array of case-sensitive strings with minimum items 0) under .instrumentation/development.general.sanitization.url. This list is a full override of the default sensitive query parameter keys, it is not a list of keys in addition to the defaults.

> When a query string value is redacted, the query string key SHOULD still be preserved, e.g. q=OpenTelemetry&sig=REDACTED.

dashpole: Seems potentially risky to include the query without any sanitization?

Rahul-Baradol: We can take the list from ENVIRONMENT during runtime, like 
`SENSITIVE_QUERY_KEYS="AWSAccessKeyId,Signature,sig,X-Goog-Signature"`

What do you think ?

PR #8871, opened 2026-04-27
Fix otelhttp span names to match the semantic conventions

### LINKED ISSUE #726: Spans created by HTTP instrumentation package does not conform to guidelines for span names

dmathieu:
Closes #726


dashpole (approved):
🚀

instrumentation/net/http/otelhttp/handler.go
```diff
@@ -35,8 +35,11 @@ type middleware struct {
 	semconv semconv.HTTPServer
 }
 
-func defaultHandlerFormatter(operation string, _ *http.Request) string {
-	return operation
+func defaultHandlerFormatter(_ string, r *http.Request) string {
+	if r.Pattern != "" {
+		return r.Method + " " + r.Pattern
```
MrAlias: This uses `r.Pattern` verbatim, but `Request.Pattern` is not always just the route path. For `http.ServeMux` it can include the registered method or host prefix, so this can produce names like `GET GET /foo/{id}` or `GET example.com/foo/{id}`.

I think the fix should reuse the same normalization we already apply when deriving `http.route`, so the span name is built from `{method} {route}` and not from the raw mux registration string.

instrumentation/net/http/otelhttp/handler.go
```diff
@@ -35,8 +35,11 @@ type middleware struct {
...
+	}
+	return r.Method
```
MrAlias: The fallback here still uses the raw request method, which does not match the HTTP span-name rules for nonstandard methods. In those cases the method token in the span name should be `HTTP`, not the original/custom verb.

Since `otelhttp` already has method canonicalization logic for `http.request.method`, I’d prefer reusing that same mapping here so the span name and emitted attributes stay consistent.

MrAlias (dismissed):
This is close, but I think the new default formatter still misses a couple of cases needed to actually match the HTTP span-name rules.

The main issue is that it uses `r.Pattern` verbatim, while the rest of `otelhttp` already treats that value as needing normalization before it becomes `http.route`. There is also still a gap for nonstandard methods, where the span name should fall back to `HTTP` instead of the raw method string.

I’d prefer to fix those in this PR so the new default naming is aligned with the existing route/method normalization logic.

dmathieu:
I have moved the span name computation into the semconv internal package, so we can reuse the `http.route` computation.

pellared (approved):

dashpole (approved):

sonalgaud12 (approved):

PR #8881, opened 2026-04-28
xrayconfig: improve mock collector synchronization and test assertions

pellared:
Improve the stability of AWS Lambda X-Ray configuration tests by replacing arbitrary delays with proper synchronization and polling assertions.


dmathieu (approved):

pellared:
Merging this with only one approval. I find this pretty sure that it will only improve the stability of tests.

Issue #8914, opened 2026-05-04
[Feature]: Extend otelconf resource detection capabilities


iblancasa:
### Component

AutoExporter

### Problem Statement

`otelconf/x` currently supports a small built-in set of resource detectors through `NewSDK(...)`:
- `container`
- `host`
- `process`
- `service`

### Proposed Solution

What I would like to do is to extend that support so `NewSDK(...)` can also run the resource detectors that already exist under `go-contrib/detectors`, including:

- `aws.ec2`
- `aws.ecs`
- `aws.eks`
- `aws.lambda`
- `azure.vm`
- `gcp`

iblancasa:
I will send a PR contributing one of the detectors so we can have a "model" to discuss.

paulojmdias:
Should we include all the detectors we will move there? I moved Hetzner already for example

iblancasa:
> Should we include all the detectors we will move there? I moved Hetzner already for example

Yes. Would like to get first some feedback on https://github.com/open-telemetry/opentelemetry-go-contrib/pull/8915 before continuing adding anything else

paulojmdias:
> > Should we include all the detectors we will move there? I moved Hetzner already for example
> 
> Yes. Would like to get first some feedback on [#8915](https://github.com/open-telemetry/opentelemetry-go-contrib/pull/8915) before continuing adding anything else

Thanks! Let me know if you need any help with it. Maybe on the ones I will move, I can include these changes after you get the feedback needed 👍

PR #8915, opened 2026-05-04
Add support for aws.ecs resource detector in otelconf/x

iblancasa:
Updates #8914


iblancasa:
cc @paulojmdias

paulojmdias (approved):
LGTM, thank you!

dashpole (approved):

iblancasa:
Had many conflicts and it was easy for me to recreate the branch and reapply the changes there.
Is there anything else needed in this PR?

Makefile
```diff
@@ -346,6 +346,8 @@ genjsonschema: genjsonschema-cleanup $(GOJSONSCHEMA)
 	@echo Modify jsonschema generated files.
 	sed -f ./otelconf/jsonschema_patch.sed ${GENERATED_EXPERIMENTAL_CONFIG} > ${GENERATED_EXPERIMENTAL_CONFIG}.tmp
 	cp ${GENERATED_EXPERIMENTAL_CONFIG}.tmp ${GENERATED_EXPERIMENTAL_CONFIG}
+	$(GO) fmt ${GENERATED_EXPERIMENTAL_CONFIG}
+	perl -0pi -e 's/type ExperimentalAWSECSResourceDetector map\[string\]interface\{\}\n(type ExperimentalServiceResourceDetector)/type ExperimentalAWSECSResourceDetector map[string]interface{}\n\n$$1/' ${GENERATED_EXPERIMENTAL_CONFIG}
```
paulojmdias: Is this needed ?

iblancasa: Same as here https://github.com/open-telemetry/opentelemetry-go-contrib/pull/9137/changes#r3414925575
If you can provide any better idea about how to implement this, I can go with it  😅

paulojmdias: I have no better idea, just asking 😅

paulojmdias (approved):

Makefile
iblancasa: Yep yep. I know is not... obvious. I spent some hours until making it work with this detector. After that, it was easy to apply to the others. But... I would be happy if we don't do this.

paulojmdias: Maybe an issue to revisit it later ?

otelconf/go.mod
```diff
@@ -6,6 +6,7 @@ require (
 	github.com/prometheus/client_golang v1.23.2
 	github.com/prometheus/otlptranslator v1.0.0
 	github.com/stretchr/testify v1.11.1
+	go.opentelemetry.io/contrib/detectors/aws/ecs v1.44.0
```
MrAlias, line 9: This needs a `replace` statement to use the local module. Run `make crosslink`.

Makefile
iblancasa: I agree. @dashpole what do you think?

dashpole: Sounds good to me.

otelconf/go.mod
MrAlias: The the resolution to this not get pushed, or maybe stepped on during a merge? I'm not seeing a `replace`.

iblancasa: Ooops. I ran the command and forgot to push.

otelconf/x/resource_test.go
```diff
@@ -147,6 +147,7 @@ func TestResourceOptsWithDetectors(t *testing.T) {
 		{
 			name: "all-detectors",
 			detectors: []ExperimentalResourceDetector{
+				{AWSECS: ExperimentalAWSECSResourceDetector{}},
```
MrAlias, line 162: Non-blocking: adding the ECS detector to this input executes the new branch, but the test only asserts attributes produced by the host, OS, process, and service detectors. Outside ECS, the ECS detector returns no resource, so this test still passes if the `d.AWSECS` branch in `resourceOpts` is removed. That leaves the core integration unprotected: `aws.ecs` could parse successfully while silently doing nothing.

I think we should add a focused test asserting that `resourceOpts` returns one option for an ECS-only input. This should not block this PR; we can follow up if it is not addressed before merge.

MrAlias (approved):

PR #8918, opened 2026-05-04
feat: integrate weaver live-check for semconv compliance validation (#8066)

### LINKED ISSUE #8066: Run weaver live checking on the telemetry instrumentations are emitting
We should check that all telemetry instrumentations are emitting is backed by semantic conventions.
weaver allows doing that with its [registry live-check](https://github.com/open-telemetry/weaver/blob/main/crates/weaver_live_check/README.md) tool.

The idea would be:

* Run weaver to start its OTLP endpoint.
* Have an E2E test that runs the instrumentation configured with the OTLP exporter sending the data to weaver.
* Create the JSON report.

We likely have many things that are not semconv valid (see https://github.com/open-telemetry/opentelemetry-go-contrib/issues/8048), so we should be able to warn at first, and actually fail CI only once we have fixed everything for an instrumentation.

gyanranjanpanda:
## Weaver Registry Live-Check Integration

This PR integrates the `weaver` registry live-check tool to automate validation of OpenTelemetry semantic conventions within our E2E telemetry generators.

### Objective
We have introduced a localized E2E testing framework to execute OpenTelemetry instrumentation (`otelhttp`) and emit traces/metrics to a local OTLP collector. The OTLP collector is the OpenTelemetry Weaver CLI's `live-check` engine, which dynamically fetches the latest semantic conventions from the central GitHub repository and verifies our instrumentation telemetry for compliance.

### Implementation Details
- **Architecture**: A new module at `internal/e2e/weaver` follows Clean Architecture:
  - `domain`: Interfaces and models.
  - `application`: Use case runner that handles the execution lifecycle.
  - `infrastructure`: Concrete `otelhttp` generator and OTLP gRPC initialization without experimental log bindings.
  - `interfaces`: Entrypoint for the E2E simulation logic.
- **Orchestration**: `tools/weaver-live-check.sh` is introduced to automatically download `weaver v0.23.0`, spin it up in the background with a healthy waiting mechanism, and trigger the E2E generation.
- **CI/CD Integration**: The `.github/workflows/ci.yml` pipeline now runs the `weaver-live-check` job locally to provide warn-only compliance verification, generating JSON compliance reports uploaded as GitHub artifacts.

### Testing
- Fully linter-compliant (`make golangci-lint/internal/e2e/weaver`).
- Successfully passes `make vanity-import-check`.
- Weaver E2E simulation verified via `./tools/weaver-live-check.sh`.

Closes #8066


gyanranjanpanda:
Could @dmathieu  @jensholdgaard review this waiting for your feedback

jensholdgaard:
I would like to provide my feedback, i think it would be worth looking into using otel weaver as a test container, rather then orchestration through the github actions, here is another example where it is done https://github.com/open-telemetry/opentelemetry-collector-contrib/issues/44905, perhaps we can use that as inspiration.

pellared:
> I would like to provide my feedback, i think it would be worth looking into using otel weaver as a test container, 

I totally agree. But could we please use https://github.com/ory/dockertest for it? We use it already e.g in https://github.com/open-telemetry/opentelemetry-ebpf-instrumentation

mikeblum:
pulled this down to test out the weaver output and the linter was complaining:

`make lint` flagged a few small nits:

```diff
@@ -31,8 +31,10 @@
-const weaverImage = "otel/weaver"
-const weaverTag = "v0.23.0"
+const (
+	weaverImage = "otel/weaver"
+	weaverTag   = "v0.23.0"
+)
```
> `gofumpt`: consecutive top-level consts should be grouped

```diff
@@ -80,7 +80,8 @@
-			conn, dialErr := net.DialTimeout("tcp", otlpEndpoint, 2*time.Second)
+			dialer := &net.Dialer{Timeout: 2 * time.Second}
+			conn, dialErr := dialer.DialContext(t.Context(), "tcp", otlpEndpoint)
```
> `noctx`: `net.DialTimeout` must not be called; use `(*net.Dialer).DialContext` with `(*net.Dialer).Timeout`

```diff
@@ -87,7 +90,7 @@
-	ctx := context.Background()
+	ctx := t.Context()
```
> `usetesting`: `context.Background()` can be replaced with `t.Context()` in `TestWeaverLiveCheck`

Happy to open a small PR against your branch if that's easier to apply.

internal/e2e/weaver/weaver_test.go
```diff
@@ -0,0 +1,283 @@
...
+	"go.opentelemetry.io/otel/exporters/otlp/otlptrace/otlptracegrpc"
+	"go.opentelemetry.io/otel/propagation"
+	sdkmetric "go.opentelemetry.io/otel/sdk/metric"
+	sdktrace "go.opentelemetry.io/otel/sdk/trace"
+
+	"go.opentelemetry.io/contrib/instrumentation/net/http/otelhttp"
+)
+
+const weaverImage = "otel/weaver"
+const weaverTag = "v0.23.0"
```
mikeblum: Should we make the weaver version/tag injectable via an env var / cli flag / both? seems like weaver is on a monthly release cadence: https://github.com/open-telemetry/weaver/releases but not sure what our posture is on taking the latest version

mikeblum:
can't seem to find `./tools/weaver-live-check.sh` or `.github/workflows/ci.yml` tweaks. just need to be pushed up?

internal/e2e/weaver/weaver_test.go
pellared: We could use Dependabot to bump this constant value using https://docs.renovatebot.com/modules/manager/regex/

gyanranjanpanda:
Thanks, addressed. I added a Renovate `customManagers` regex entry for `internal/e2e/weaver/weaver_test.go` so Renovate can detect `weaverImage` as the Docker dependency and `weaverTag` as the current value. I also verified the regex extracts `otel/weaver` and `v0.23.0` correctly.

om7057:
Went through the current diff rather than the PR description, since the description still talks about a `tools/weaver-live-check.sh` script, a "Clean Architecture" domain/application/infrastructure layout, and a CI workflow job, none of which are in the actual diff anymore. Worth updating the description so reviewers aren't evaluating a stale plan.

A few things on the current implementation:

1. `internal/e2e/weaver` has its own `go.mod`, and `ALL_COVERAGE_MOD_DIRS` in the Makefile auto-discovers every `go.mod` via `find`. That means `weaver_test.go` runs inside `make test-race` and `make test-coverage`, both of which run unconditionally on `ubuntu-latest` for every PR in the repo, not just weaver-related changes (I checked `.github/workflows/ci.yml:89-129`). Neither job passes `-short`, so `testing.Short()` doesn't skip it there, and both are required, blocking checks (`test-coverage` even sets `fail_ci_if_error: true`). So every PR now pulls `otel/weaver:v0.23.0` from Docker Hub and spins up a container as part of two required checks. If that pull is slow, rate-limited, or the runner's Docker daemon has a bad day, unrelated PRs go red. `compatibility-test` is fine since it runs `make test-short`, which does honor `testing.Short()`. Given the issue explicitly wants this to be warn-only until the repo is semconv-clean, I'd expect this test to live behind its own opt-in CI job (or a build tag / env var gate) rather than riding along in the default test-race/test-coverage matrix.

2. The test never inspects the live-check output. `logReports` just does `t.Logf` on the raw JSON and the test passes as long as weaver produced any file. It doesn't parse the report or fail (even non-fatally, e.g. via `t.Log` of a violation count) on semconv violations. Per the issue, the intent was "warn at first, fail once fixed." Right now there's no signal at all, so this establishes plumbing but not the actual compliance check the issue is asking for.

3. The `mikeblum` comment about making the weaver version injectable via env var/flag doesn't look resolved. The renovate.json custom manager (addressing pellared's comment) handles auto-bumping the pinned constant, but that's a different concern than runtime overridability, worth confirming with mikeblum whether renovate coverage is enough for them or they still want an override knob.

4. Scope is otelhttp-only, which is a reasonable first slice given the phased rollout in the issue, but the PR body should say so explicitly so reviewers aren't expecting broader coverage yet.

Nothing here is a design objection to the dockertest-based approach itself, that's a good simplification of the earlier "Clean Architecture" version and self-contained.

jensholdgaard:
Building on @om7057's point 2 (the report is logged but never inspected) — I run `weaver registry live-check` as a CI gate in a couple of projects ([an OTLP log backend](https://github.com/jensholdgaard/ourios/blob/main/.github/workflows/ci.yml) and a [recent port of the same gate to another repo](https://github.com/ruvnet/RuView/pull/1494)), and a few failure modes we hit in production may be useful while this is still taking shape:

1. **The zero-telemetry false pass.** If the OTLP export silently fails (mis-dialed endpoint, exporter dropped before delivery), weaver still writes a well-formed report — with `statistics.total_entities: 0` — and this test passes, because its only guard is "a JSON file exists". That's worse than no test: it green-lights a broken pipeline. Gating on `total_entities > 0` is cheap and catches it.

2. **Expected-signal presence.** One step further: assert that the specific signals this exercise must produce (the `http.server.*` / `http.client.*` metric and span names) actually appear in the report. Otherwise a refactor of the generator can silently stop exercising `otelhttp` and the check keeps passing on whatever residue still flows.

3. **"Warn now, fail later" works better as a ratchet than a flip.** Instead of a single global warn→fail transition once the repo is semconv-clean, parse the `PolicyFinding` entries and maintain a small exempt-list (by finding id / attribute name) of the *known* dirt — then fail on everything else immediately. New violations get caught from day one, and the exempt-list shrinks as cleanups land, entry by entry, instead of the gate staying toothless until some future all-clean day that keeps receding.

4. **Pin the compliance target.** Live-checking against the registry's moving head means the warn stream mixes real call-site drift with upstream semconv evolution, and the eventual fail-flip can never be held stable. `otelhttp` declares the schema version it targets — checking against *that* registry tag (bumped deliberately, e.g. by renovate like the weaver image) keeps the report's findings attributable to this repo's code.

5. **The Docker Hub pull in required checks has a registry-free alternative.** Weaver isn't published on GHCR today (only `otel/weaver` on Docker Hub), so the usual switch-to-ghcr fix isn't available — but the gate doesn't need a container at all: weaver ships single-binary GitHub release assets, and downloading the platform asset with a pinned sha256 removes both the Docker Hub rate-limit exposure and the docker-daemon dependency `dockertest` adds; it also makes the test runnable on machines without Docker. Longer-term, weaver publishing to GHCR alongside Docker Hub (as the collector releases already do) would make the container path robust too — that'd be an upstream ask on open-telemetry/weaver.

On session termination: relying on `--inactivity-timeout 30` plus the SIGHUP fallback works, but weaver's `--admin-port` exposes a `/stop` endpoint — ending the session deterministically right after the SDK shutdown flush is faster and avoids the idle-timer race entirely.

A concrete sketch of 1–3 on top of the existing `downloadReports` output (the `PolicyFinding`/`level`/`context.attribute_name` shapes are from real v0.23.0 reports):

```go
func gateReport(t *testing.T, raw string) {
	var stats struct {
		Statistics struct {
			TotalEntities int `json:"total_entities"`
		} `json:"statistics"`
	}
	if err := json.Unmarshal([]byte(raw), &stats); err != nil {
		t.Fatalf("parse live-check report: %v", err)
	}
	// A broken exporter still yields a valid report — with zero ingested
	// entities. That must fail, not pass. (1)
	if stats.Statistics.TotalEntities == 0 {
		t.Fatal("weaver ingested no telemetry — the OTLP export never reached it")
	}

	// The signals this exercise exists to produce must be observed. (2)
	for _, want := range []string{"http.server.request.duration", "http.client.request.duration"} {
		if !strings.Contains(raw, want) {
			t.Errorf("expected signal %q not observed in the report", want)
		}
	}

	// Known-dirty findings, removed one by one as cleanups land; anything
	// not on the list fails immediately. (3)
	exempt := map[string]bool{
		// "attribute.or.finding.id": true,
	}
	var doc any
	if err := json.Unmarshal([]byte(raw), &doc); err != nil {
		t.Fatalf("parse live-check report: %v", err)
	}
	var violations []string
	var walk func(v any)
	walk = func(v any) {
		switch node := v.(type) {
		case map[string]any:
			if node["type"] == "PolicyFinding" && node["level"] == "violation" {
				id := fmt.Sprint(node["id"])
				name := ""
				if ctx, ok := node["context"].(map[string]any); ok {
					name = fmt.Sprint(ctx["attribute_name"])
				}
				if !exempt[id] && !exempt[name] {
					violations = append(violations, fmt.Sprintf("%s (%s): %v", id, name, node["message"]))
				}
			}
			for _, child := range node {
				walk(child)
			}
		case []any:
			for _, child := range node {
				walk(child)
			}
		}
	}
	walk(doc)
	if len(violations) > 0 {
		t.Errorf("%d non-exempt semconv violation(s):\n%s", len(violations), strings.Join(violations, "\n"))
	}
}
```

None of this blocks the plumbing landing first — 1 and 2 are a few lines, and 3–5 can follow once the shape settles.

---
*Disclosure: this comment was drafted with AI assistance (Claude). The practices and failure modes are from my own CI gates linked above, and the report-shape claims are verified against real weaver v0.23.0 output — but the wording is largely the model's, so weigh it accordingly.*

PR #8924, opened 2026-05-06
otelhttp: separate client Labeler context key to prevent server attribute leak

### LINKED ISSUE #8920: [Bug]: http_client metrics have the wrong`http.route` label.
### Component

Instrumentation: otelhttp

`http_client_request_duration` histogram has `http_route` from the upstream server route which is confusing (and wrong). This attribute should not be available on the `http_client` metrics: https://opentelemetry.io/docs/specs/semconv/http/http-metrics/#http-client

### Expected behavior

`http_client` metrics don't have `http_route`.

### Steps to Reproduce

```go
// otelhttp Labeler leak repro.
//
// Run:   go run .
// Then:  curl http://localhost:8080/upstream
//        curl http://localhost:8080/metrics | grep http_route
//
// Expected: http.route is a server-side attribute. It should appear on
//   http_server_request_duration_seconds (server-side) but NOT on
//   http_client_request_duration_seconds (client-side).
//
// Actual: http_route="/upstream" appears on both, because the otelhttp
// Labeler is stored under a single context key shared by the server
// middleware (otelhttp.NewMiddleware) and the client transport
// (otelhttp.NewTransport). When the handler propagates the inbound
// request's context into an outbound request, the client transport reads
// the labeler the server middleware populated.

package main

import (
        "io"
        "log"
        "net/http"

        "github.com/prometheus/client_golang/prometheus"
        "github.com/prometheus/client_golang/prometheus/promhttp"
        "go.opentelemetry.io/otel"
        otelprom "go.opentelemetry.io/otel/exporters/prometheus"
        "go.opentelemetry.io/otel/sdk/metric"
)

func main() {
        reg := prometheus.NewRegistry()
        exporter, err := otelprom.New(otelprom.WithRegisterer(reg), otelprom.WithoutScopeInfo())
        if err != nil {
                log.Fatal(err)
        }
        otel.SetMeterProvider(metric.NewMeterProvider(metric.WithReader(exporter)))

        client := &http.Client{Transport: otelhttp.NewTransport(http.DefaultTransport)}

        upstream := http.HandlerFunc(func(w http.ResponseWriter, r *http.Request) {
                labeler, _ := otelhttp.LabelerFromContext(r.Context())
                labeler.Add(attribute.String("http.route", "/upstream"))

                req, _ := http.NewRequestWithContext(r.Context(), http.MethodGet, "https://google.com/downstream", nil)
                resp, err := client.Do(req)
                if err != nil {
                        http.Error(w, err.Error(), http.StatusBadGateway)
                        return
                }
                _, _ = io.Copy(io.Discard, resp.Body)
                _ = resp.Body.Close()
                w.WriteHeader(http.StatusOK)
        })

        mux := http.NewServeMux()
        mux.Handle("/upstream", otelhttp.NewMiddleware("upstream")(upstream))
        mux.Handle("/metrics", promhttp.HandlerFor(reg, promhttp.HandlerOpts{}))

        log.Println("listening on :8080 — try: curl http://localhost:8080/upstream")
        log.Fatal(http.ListenAndServe(":8080", mux))
}
```

### Operating System

NA

### Device Architecture

ARM64

### Go Version

1.26

### Component Version

otelhttp 0.68

wucm667:
Fixes #8920

## Problem

The `otelhttp` server middleware and client transport share the same context key (`labelerContextKey`) for the `Labeler`. When a server handler propagates its inbound request context into an outbound HTTP client request (a common pattern with `http.NewRequestWithContext(r.Context(), ...)`), the client transport reads the server's Labeler and incorrectly includes server-side attributes (e.g., `http.route`) in client-side metrics.

Per the [OpenTelemetry HTTP semantic conventions](https://opentelemetry.io/docs/specs/semconv/http/http-metrics/#http-client), `http.route` is a server-side attribute and should not appear on `http_client_*` metrics.

## Solution

Add a separate `clientLabelerContextKey` with `ContextWithClientLabeler` and `ClientLabelerFromContext` functions. The client transport now uses these instead of the server-side `LabelerFromContext` / `ContextWithLabeler`, preventing cross-contamination of attributes.

The existing `ContextWithLabeler` / `LabelerFromContext` API remains unchanged for server-side use. Users who need to add custom attributes to client-side metrics can use `ContextWithClientLabeler` / `ClientLabelerFromContext` or the existing `WithMetricAttributesFn` option.

## Changes

- `labeler.go`: Add `clientLabelerContextKey`, `ContextWithClientLabeler`, `ClientLabelerFromContext`
- `transport.go`: Use `ClientLabelerFromContext` / `ContextWithClientLabeler` instead of `LabelerFromContext` / `ContextWithLabeler`
- `transport_test.go`: Update `TestCustomAttributesHandling` to use `ContextWithClientLabeler`


instrumentation/net/http/otelhttp/labeler.go
```diff
@@ -56,3 +56,27 @@ func LabelerFromContext(ctx context.Context) (*Labeler, bool) {
...
+// for use with client-side HTTP instrumentation (otelhttp.Transport).
+// Attributes added to this labeler will be attached to client-side metrics.
+func ContextWithClientLabeler(parent context.Context, l *Labeler) context.Context {
+	return context.WithValue(parent, clientLabelerContextKey, l)
+}
+
+// ClientLabelerFromContext retrieves a Labeler instance from the provided context
+// for use with client-side HTTP instrumentation. If no Labeler was found, a new,
+// empty Labeler is returned and the second return value is false.
+func ClientLabelerFromContext(ctx context.Context) (*Labeler, bool) {
```
sonalgaud12, line 76: Can you add test?

sonalgaud12 (dismissed):

instrumentation/net/http/otelhttp/labeler.go
wucm667: Done! Added labeler_test.go with 3 test functions covering ContextWithClientLabeler, ClientLabelerFromContext, and a context isolation test that verifies server and client labelers don't interfere with each other. All tests pass locally.

sonalgaud12 (commented):
LGTM, Need review from maintainers.

CHANGELOG.md
```diff
@@ -69,6 +69,7 @@ This project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0.htm
 
 ### Fixed
 
+- Fix server-side `http.route` attribute leaking into client-side metrics in `go.opentelemetry.io/contrib/instrumentation/net/http/otelhttp` when a server handler propagates its request context into an outbound HTTP client request. (#8924)
```
dashpole, line 72: I would probably consider this a breaking change, albeit one that is one that many users probably want.

wucm667: That makes sense. This changes the emitted client metric attributes for users who were relying on the leaked server-side `http.route`.

Would you prefer this to be documented under `Changed` in the Unreleased section instead of `Fixed`, with wording that calls out the behavior change?

dashpole: Yes

wucm667: Implemented in b0deb30179719a192dd92dd8a3ef87863a10722e. The changelog now documents this under Unreleased Changed as a breaking migration: client metrics no longer read the server Labeler; users needing client attributes should use ContextWithClientLabeler / ClientLabelerFromContext or WithMetricAttributesFn. Server Labeler APIs are unchanged.

dmathieu (approved):

dashpole (approved):

Issue #8940, opened 2026-05-08
[detectors] Consolidate all resource detectors from opentelemetry-collector-contrib/processor/resourcedetectionprocessor


paulojmdias:
## Problem Statement

With [open-telemetry/opentelemetry-collector#14311](https://github.com/open-telemetry/opentelemetry-collector/issues/14311) landing, the collector's `service::telemetry` gains support for resource detectors from the `opentelemetry-go` SDK. Meanwhile, `processor/resourcedetectionprocessor` in `opentelemetry-collector-contrib` maintains its own parallel set of detector implementations (`system`, `env`, `ec2`, `gcp`, `azure`, `k8snode`, and many others).

This leaves two independent detection paths in the same binary:
- **SDK detectors** — used by `service::telemetry` (and, by extension, the attributes reported by the OpAMP extension via `include_resource_attributes`).
- **resourcedetectionprocessor detectors** — used by user telemetry pipelines.

They can — and likely will — drift: different attribute keys, different detection semantics, different failure modes, different priorities. Users will see inconsistent resource attributes on their collector's self-telemetry vs. the telemetry flowing through their pipelines, even with similar-looking configuration.

## Proposed Solution

Move all detectors to live in this repo (`opentelemetry-go-contrib/detectors/`), so that `processor/resourcedetectionprocessor` becomes a thin adapter delegating to these canonical Go SDK implementations. This eliminates the drift and gives users a single, consistent source of truth for resource detection.

For detectors that already exist in both repos, the work is to audit differences and reconcile them. For detectors that only exist in collector-contrib, the work is to port them here.

## Prior Art

- Original issue and discussion: open-telemetry/opentelemetry-collector-contrib#48159
- Go SIG approved this approach, with the requirement that each detector has a designated codeowner.

### Group A — Exist in both repos: Audit & reconcile

These detectors have implementations in both `go-contrib/detectors/` and `collector-contrib/processor/resourcedetectionprocessor/internal/`. The work is to diff attribute keys, values, semantics, and failure modes; fix any gaps in go-contrib; and ensure collector-contrib can safely delegate here.

- [ ] #8941
- [ ] #8943
- [ ] #8944
- [ ] #8945
- [ ] #8946
- [x] #8947
- [ ] #8948

### Group B — Only in collector-contrib: Contribute to go-contrib

[...]

PR #8979, opened 2026-05-11
detectors/hetzner: add Hetzner Cloud resource detector

### LINKED ISSUE #8962: [detectors/hetzner] Contribute hetzner detector from resourcedetectionprocessor
## Problem Statement

`collector-contrib/processor/resourcedetectionprocessor/internal/hetzner` implements a resource detector for Hetzner Cloud servers. There is currently no equivalent in `go-contrib/detectors/`, which means Hetzner detection is only available to collector-contrib pipelines and not to any Go SDK user.

## Proposed Solution

Add a new `detectors/hetzner` package implementing the Hetzner detector as a `resource.Detector`, ported from the collector-contrib implementation, so that `collector-contrib/processor/resourcedetectionprocessor` can delegate to it and any other Go SDK user can adopt it.

## Prior Art

- Collector-contrib implementation to port from: [`processor/resourcedetectionprocessor/internal/hetzner`](https://github.com/open-telemetry/opentelemetry-collector-contrib/tree/main/processor/resourcedetectionprocessor/internal/hetzner)

## Additional Context

- Queries the Hetzner Cloud Instance Metadata Service.
- Emits: `cloud.provider` (hetzner_cloud), `cloud.platform` (hetzner_cloud_hcloud), `cloud.region`, `host.id`, `host.name`.

paulojmdias:
Ports the Hetzner Cloud resource detector from `processor/resourcedetectionprocessor/internal/hetzner` in
`opentelemetry-collector-contrib` to a new standalone Go module `go.opentelemetry.io/contrib/detectors/hetzner`.

Detected attributes:
- `cloud.provider`
- `cloud.platform`
- `cloud.region`
- `cloud.availability_zone`
- `host.id`
- `host.name`

Returns `resource.Empty(), nil` when not running on a Hetzner Cloud server. Returns a partial resource and `resource.ErrPartialResource` if the platform is detected but individual metadata endpoints fail.

Also registers the detector in `detectors/autodetect` as `IDHetzner` (`"hetzner"`).

Fixes #8962.
Relates #8940.


paulojmdias:
\cc @dashpole @iblancasa 

If any of you want to be a codeowner with me here, please let me know, and I will update the codeowners list.

detectors/hetzner/hetzner.go
```diff
@@ -0,0 +1,94 @@
...
+}
+
+// ResourceDetector collects resource information of Hetzner Cloud servers.
+type ResourceDetector struct {
+	client *hcloudmeta.Client
+}
+
+// NewResourceDetector returns a [resource.Detector] that detects resource
+// attributes on Hetzner Cloud servers.
+func NewResourceDetector() *ResourceDetector {
```
dashpole: Consider making this accept options (see https://github.com/open-telemetry/opentelemetry-go/blob/main/CONTRIBUTING.md#configuration) in-case we want to support configuration in the future.

I could see this supporting a WithAttributeFilter(attribute.Filter) option, if you wanted to make this easier to use with the collector.

detectors/hetzner/README.md
```diff
@@ -0,0 +1,82 @@
...
+[Hetzner Cloud](https://www.hetzner.com/cloud) servers by querying the Hetzner
+Cloud Instance Metadata Service.
+
+## Installation
+
+```bash
+go get go.opentelemetry.io/contrib/detectors/hetzner
+```
+
+## Usage
```
dashpole: Consider using an example_test.go so that we can make sure the sample compiles and runs properly.

detectors/hetzner/README.md
```diff
@@ -0,0 +1,82 @@
+# Hetzner Cloud Resource Detector for Go
```
dashpole, line 1: We generally prefer using godocs, rather than READMEs.  See https://github.com/open-telemetry/opentelemetry-go/blob/main/CONTRIBUTING.md#documentation

detectors/hetzner/version.go
```diff
@@ -0,0 +1,17 @@
+// Copyright The OpenTelemetry Authors
+// SPDX-License-Identifier: Apache-2.0
+
+package hetzner // import "go.opentelemetry.io/contrib/detectors/hetzner"
+
+// Version is the current release version of the Hetzner resource detector.
+func Version() string {
+	return "1.43.0"
```
dashpole: Should this be marked stable?  Are the resource attributes it detects all stable?

dashpole:
I think this needs to be added to https://github.com/open-telemetry/opentelemetry-go-contrib/blob/main/versions.yaml

detectors/hetzner/version.go
paulojmdias: All the detected resource attributes are marked as `Development` in the semantic conventions. So I think we shouldn't mark this as stable. 👍

paulojmdias:
@dashpole thanks for the review. I think I addressed everything properly, PTAL when you have a chance 🙏

detectors/hetzner/hetzner.go
```diff
@@ -0,0 +1,123 @@
...
+	}
+	return &ResourceDetector{client: newHcloudClient(), cfg: cfg}
+}
+
+// Detect detects resource attributes of the Hetzner Cloud server the process
+// is running on. It returns an empty resource and no error when not running on
+// a Hetzner Cloud server. If the process is running on a Hetzner Cloud server
+// but some attributes cannot be retrieved, a partial resource is returned
+// together with [resource.ErrPartialResource].
+func (d *ResourceDetector) Detect(_ context.Context) (*resource.Resource, error) {
```
MrAlias, line 67: Non-blocking note for follow-up: this detector currently inherits the `hcloud` metadata client's request behavior, so `Detect` does not fully honor caller cancellation/timeouts. That seems acceptable for this PR since this is a straightforward port, but it may be worth revisiting later if we want stricter context propagation in autodetection.

MrAlias (approved):
Looks good to me. The earlier feedback around options, docs/examples, and release/versioning wiring appears to be addressed, and the new detector is integrated consistently with the existing detector modules.

detectors/hetzner/hetzner.go
paulojmdias: Should we create an issue to revisit later? And possibly improve upstream on `hcloud` package.

dashpole: Yes, please create an issue here, and upstream.

detectors/hetzner/example_test.go
```diff
@@ -0,0 +1,22 @@
...
+
+import (
+	"context"
+	"fmt"
+
+	"go.opentelemetry.io/contrib/detectors/hetzner"
+)
+
+func ExampleNewResourceDetector() {
+	detector := hetzner.NewResourceDetector()
```
dashpole: I believe the example can use dependencies (e.g. the trace SDK) without actually adding dependencies to the go module -- if you wanted to make this example more realistic by using it when creating an SDK.

dashpole (approved):

detectors/hetzner/hetzner.go
paulojmdias: Issues created:
- https://github.com/open-telemetry/opentelemetry-go-contrib/issues/8985
- https://github.com/hetznercloud/hcloud-go/issues/851

detectors/hetzner/example_test.go
paulojmdias: Thank you, I think I made it more realistic now 👍

paulojmdias:
Can this be merged? I would like ensure everything is correct before moving with the other detectors 🙏

PR #8990, opened 2026-05-13
otelconf: add default attributes to configuration-provided resource

peschinskiy:
Problem discussion at Slack: https://cloud-native.slack.com/archives/C0476L7UJT1/p1777290327086879

This PR fixes the issue of telemetry SDK and service name attributes not being added when a config resource is provided, while being present in the default resource. The behavior of the Go SDK is now aligned with other SDKs.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -10,7 +10,7 @@ This project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0.htm
 
 ### Added
 
-- The resource created by `go.opentelemetry.io/contrib/otelconf` now includes [default SDK attributes](https://pkg.go.dev/go.opentelemetry.io/otel/sdk/resource#Default). (#8990)
+- The resource created by `go.opentelemetry.io/contrib/otelconf` now includes [default SDK attributes](https://pkg.go.dev/go.opentelemetry.io/otel/sdk/resource#Default) and, when no schema URL is provided, falls back to the SDK's default schema URL. (#8990)
 - Add `azurecontainerapps` resource detector for Azure Container Apps. (#8939)
 
 ### Changed
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -10,10 +10,8 @@ This project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0.htm
 
 ### Added
 
+- The resource created by `go.opentelemetry.io/contrib/otelconf` now includes [default SDK attributes](https://pkg.go.dev/go.opentelemetry.io/otel/sdk/resource#Default). (#8990)
 - Add `azurecontainerapps` resource detector for Azure Container Apps. (#8939)
-### Changed
-
-- The configuration-provided resource now includes default SDK attributes. (#8990)
 
 ### Changed
 
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -15,7 +15,11 @@ This project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0.htm
 
 ### Changed
 
+<<<<<<< HEAD
 - The `Version()` function in `go.opentelemetry.io/contrib/instrumentation/github.com/gorilla/mux/otelmux` has been replaced by `const Version`. (#9076)
+=======
+- The configuration-provided resource now includes default SDK attributes. (#8990)
+>>>>>>> 052b058b8 (fix changelog)
 
 <!-- Released section -->
 <!-- Don't change this section unless doing release -->
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -35,7 +35,7 @@ This project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0.htm
 - Set the "error" field (e.g. created via `zap.Error`) as `record.SetErr` instead of a plain attribute in `go.opentelemetry.io/contrib/bridges/otelzap`. (#8719)
 - Set fields implementing `error` interface from `slog` records as `record.SetErr` instead of plain attributes in `go.opentelemetry.io/contrib/bridges/otelslog`. (#8774)
 - Set emitted errors in `go.opentelemetry.io/contrib/bridges/otellogr` as record errors (`Record.SetErr`) instead of `exception.message` attributes. (#8775)
-- The configuration-provided resource now includes default SDK attributes. ()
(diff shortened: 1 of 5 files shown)


peschinskiy:
@jack-berg @pellared Could you take a look at this PR when you have time? Thank you in advance.

otelconf/resource_test.go
```diff
@@ -14,6 +14,13 @@ import (
 )
 
 func TestNewResource(t *testing.T) {
+	withDefaultAttributes := func(attrs ...attribute.KeyValue) []attribute.KeyValue {
+		// Default attributes come first so they're overridden by config-provided attributes
+		return append(
+			resource.Default().Attributes(),
+			attrs...,
+		)
+	}
```
pellared: Can we have more granular assertion and not use the same code as in the implementation? There is no assertion that e.g. checks if the OpenTelemetry SDK attributes are added.

pellared:
CC @codeboten

otelconf/resource_test.go
peschinskiy: Made an explicit list of attributes expected in the resource.

peschinskiy:
Updated the tests. The checks failure seems flaky and unrelated to the changes.

codeboten (commented):
Thanks for addressing this, the code will need to be added in otelconf/x as well

peschinskiy:
Thanks for the review! Could you help me get this merged when you have a minute? @pellared @codeboten

bboreham:
The `govulncheck` failure would be cleared by updating to latest main branch, I think.  It was updated in #9015.
```
  Module: golang.org/x/net
    Found in: golang.org/x/net@v0.54.0
    Fixed in: golang.org/x/net@v0.55.0
```

codeboten (approved):

CHANGELOG.md
```diff
@@ -8,6 +8,10 @@ This project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0.htm
 
 ## [Unreleased]
 
+### Changed
+
+- The configuration-provided resource now includes default SDK attributes. (#8990)
```
```suggestion
### Added

- The resource created by `go.opentelemetry.io/contrib/otelconf` now includes [default SDK attributes](https://pkg.go.dev/go.opentelemetry.io/otel/sdk/resource#Default). (#8990)
```
```

File: otelconf/resource_test.go [hunk 3]
```diff
@@ -9,32 +9,32 @@ import (
...
 func TestNewResource(t *testing.T) {
 	tests := []struct {
-		name         string
-		config       *Resource
-		wantResource *resource.Resource
-		wantErrT     error
+		name          string
+		config        *Resource
+		wantSchemaURL string
+		wantAttrs     []attribute.KeyValue
```
pellared, line 21: Why are we not asserting against `wantResource`? Why do we now have `wantSchemaURL` and `wantAttrs` instead?

peschinskiy: To avoid duplicating the code from the `newResource` implementation when constructing expected resources. Presence of the telemetry SDK attributes is now a general expectation for every constructed resource and is therefore described in the assertion loop, not in every individual test case.

otelconf/resource_test.go
```diff
@@ -70,16 +66,28 @@ func TestNewResource(t *testing.T) {
...
-			assert.Equal(t, tt.wantResource, got)
+			require.ErrorIs(t, err, tt.wantErrT)
+
+			assert.Equal(t, tt.wantSchemaURL, got.SchemaURL())
+			assert.True(t, got.Set().HasValue(semconv.TelemetrySDKNameKey))
+			assert.True(t, got.Set().HasValue(semconv.TelemetrySDKLanguageKey))
+			assert.True(t, got.Set().HasValue(semconv.TelemetrySDKVersionKey))
+			for _, want := range tt.wantAttrs {
+				got, ok := got.Set().Value(want.Key)
+				if assert.True(t, ok) {
```
pellared: The failure messages of the `assert.True` do not say what went wrong.
From https://go.dev/wiki/CodeReviewComments#useful-test-failures

> Tests should fail with helpful messages saying what was wrong, with what inputs, what was actually got, and what was expected.

pellared (approved):

otelconf/x/resource.go
```diff
@@ -46,6 +46,7 @@ func newResource(r *Resource) (*resource.Resource, error) {
 		schema = *r.SchemaUrl
 	}
 	opts := []resource.Option{
+		resource.WithAttributes(resource.Default().Attributes()...),
```
dashpole, line 49: Is it not possible to use the detectors themselves, rather than detecting attributes and then taking the attributes?  Won't this lose the schemaURL for the default attributes?

peschinskiy: Omitting SchemaURL is intentional in this case. Otherwise, it would not be possible to provide a custom schema in the config resource because of merge conflict, which would be a breaking change. AFAIK, other SDKs set the default TelemetrySDK attributes without a schema when a config resource is present.

dashpole: If no schema url is set, I would expect to see the schema url from the default resource

peschinskiy: Valid point, thank you. Updated the PR.

dashpole (approved):

otelconf/x/resource.go
pellared: The applied changes seem to be against https://github.com/open-telemetry/opentelemetry-configuration/blob/f0604619b0aee2c5944138ab5bcf53cbf1e533d1/schema-docs.md?plain=1#L3151

> If omitted or null, no schema URL is used.


@open-telemetry/configuration-maintainers, is this intentional?

pellared (changes requested):
https://github.com/open-telemetry/opentelemetry-go-contrib/pull/8990#discussion_r3395212224

otelconf/x/resource.go
pellared: SIG meeting: revert the changes and create an issue in https://github.com/open-telemetry/opentelemetry-configuration and track it

peschinskiy: https://github.com/open-telemetry/opentelemetry-configuration/issues/654

pellared (approved):

pellared:
@peschinskiy , thank you!

PR #8996, opened 2026-05-13
[Prototype] bridges/otelslog: set log_bridge.name and log_bridge.version Instrumentation Scope attributes

cijothomas:
Draft prototype for the in-progress semantic convention proposed in open-telemetry/semantic-conventions#3716 (issue: open-telemetry/semantic-conventions#1550).

Sets `log_bridge.name` and `log_bridge.version` Instrumentation Scope attributes on every Logger created by `otelslog`, identifying the bridge package distinct from the producer (slog logger name) reported as the scope name.

Same trivial change applies to `otelzap`/`otellogrus`/`otellogr` — kept minimal here to drive the semconv discussion.

Related: open-telemetry/opentelemetry-specification#5089, open-telemetry/opentelemetry-rust#3515

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -13,7 +13,7 @@ This project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0.htm
 - Add `error.type` attribute to `http.client.request.duration` for transport failures in `otelhttp`. (#8801)
 - Add examples for prometheus compatibility document. (#8716)
 - Add `Resource` method to `SDK` in `go.opentelemetry.io/contrib/otelconf/x` to expose the resolved SDK resource from declarative configuration. (#8912)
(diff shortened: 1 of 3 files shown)


pellared (commented):
I approve this prototype. I still lean towards having `log.bridge.name` of instead something like `instrumentation.library.name` as it allows more flexibility (e.g. if one want to create log instrumentation using `slog`) and I find it is a more precise naming.

cijothomas:
@pellared @MrAlias 
This has no perf hit from what I can tell. Obtaining the logger is a separate act than the actual log emission itself, so this should be fine from perf standpoint? 
(I know Jack mentioned Java would likely hit perf issue, so trying to see if that is specific problem, not language wide.)

PR #8999, opened 2026-05-16
detectors/hetzner: respect context in Detect

### LINKED ISSUE #8985: [Bug]: detectors/hetzner: Detect does not honor context cancellation or timeouts
### Component

Config

(*ResourceDetector).Detect accepts a context.Context but ignores it entirely. The underlying hcloud-go metadata client hardcodes context.Background() inside its internal get method for every HTTP request, so any deadline or cancellation set by the caller is never forwarded.

hcloud-go's metadata.Client.get creates its own context.Background() instead of accepting one from the caller. Until that is fixed upstream https://github.com/hetznercloud/hcloud-go/issues/851, Detect cannot fully propagate cancellation.

### Expected behavior

Detect should respect the caller's context.Context. If the context is cancelled or its deadline exceeded before or during a metadata request, Detect should return promptly with the context error rather than blocking until the hcloud-go client's own 5-second timeout expires.

### Steps to Reproduce

1. Create a context with a short deadline:
      ctx, cancel := context.WithTimeout(context.Background(), 1*time.Millisecond)
   defer cancel()
2. Call Detect with it:
      res, err := hetzner.NewResourceDetector().Detect(ctx)
3. Observe that the call does not return early despite the context being expired — it blocks until the hcloud-go client's own timeout fires.

### Operating System

N/A

### Device Architecture

ARM64

### Go Version

1.25

### Component Version

detectors/hetzner v0.15.0

c-tonneslan:
Closes #8985.

The hcloud-go metadata client hardcodes \`context.Background()\` inside its internal HTTP \`get\` and ignores any context the caller passes in, so \`Detect\` with a tight deadline or a cancelled context blocked until hcloud-go's own ~5s internal timeout fired.

Until hetznercloud/hcloud-go#851 lands the proper upstream fix, run the existing detection synchronously in a goroutine and have \`Detect\` select on \`ctx.Done()\`. When the context expires before the metadata client returns, \`Detect\` surfaces \`ctx.Err()\` immediately. The in-flight goroutine is detached — it still completes in the background but no longer blocks the caller. When the context is live, the code path is unchanged.

Regression test points the detector at a metadata handler that blocks forever, cancels the context, and asserts \`Detect\` returns \`context.Canceled\` promptly. Existing tests still pass.

CHANGELOG entry added under \`[Unreleased]\`.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -8,6 +8,10 @@ This project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0.htm
 
 ## [Unreleased]
 
+### Fixed
+
+- `Detect` in `go.opentelemetry.io/contrib/detectors/hetzner` now returns `ctx.Err()` when the context is cancelled or its deadline expires before the metadata client can answer, instead of silently returning an empty resource. (#8999)
+
 <!-- Released section -->
 <!-- Don't change this section unless doing release -->
 
--- a/detectors/hetzner/hetzner.go
+++ b/detectors/hetzner/hetzner.go
@@ -63,9 +63,16 @@ func NewResourceDetector(opts ...Option) *ResourceDetector {
 // is running on. It returns an empty resource and no error when not running on
 // a Hetzner Cloud server. If the process is running on a Hetzner Cloud server
 // but some attributes cannot be retrieved, a partial resource is returned
-// together with [resource.ErrPartialResource].
+// together with [resource.ErrPartialResource]. If ctx is cancelled or its
+// deadline expires before detection completes, Detect returns ctx.Err().
 func (d *ResourceDetector) Detect(ctx context.Context) (*resource.Resource, error) {
 	if !d.client.IsHcloudServerWithContext(ctx) {
+		// IsHcloudServerWithContext returns a bool, so a false answer can mean
+		// either "not a Hetzner server" or "ctx was already done before we
+		// could find out". Peek at ctx to distinguish the two.
+		if err := ctx.Err(); err != nil {
+			return nil, err
+		}
 		return resource.Empty(), nil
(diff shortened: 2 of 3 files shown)


dashpole:
I think we probably just want to wait until the upstream changes. @paulojmdias thoughts?

paulojmdias:
> I think we probably just want to wait until the upstream changes. @paulojmdias thoughts?

Agree, I don't think we need to create any workaround here.

@c-tonneslan, feel free to suggest any upstream PR if you are open to.

c-tonneslan:
Took a stab at the upstream fix: hetznercloud/hcloud-go#852. Once that lands and gets released we can drop the goroutine wrapper here and just forward the ctx straight through.

paulojmdias:
@c-tonneslan are you planning to update the code here since the upstream fix was already released? 🙏

c-tonneslan:
Pushed. Bumped hcloud-go to v2.41.2, dropped the goroutine wrapper, and `Detect` now just forwards ctx into `IsHcloudServerWithContext` and the four `*WithContext` getters. `IsHcloudServerWithContext` only returns a bool, so there's a small `ctx.Err()` check after a false to distinguish 'not a Hetzner server' from 'we got cancelled before we could find out'. Added a deadline-exceeded test alongside the existing cancellation one.

paulojmdias:
@c-tonneslan please fix the conflicts

c-tonneslan:
Rebased onto main and force-pushed. #9026 already swapped the detector to the WithContext methods, so the only thing left here is the ctx.Err() peek after IsHcloudServerWithContext returns false (without it, a cancelled ctx silently turns into an empty resource and a nil error), plus the two regression tests and a CHANGELOG line

PR #9001, opened 2026-05-18
Add docker resource detector

### LINKED ISSUE #8951: [detectors/docker] Contribute docker detector from resourcedetectionprocessor
## Problem Statement

`collector-contrib/processor/resourcedetectionprocessor/internal/docker` implements a resource detector that collects container-level resource attributes from the Docker engine. There is currently no equivalent in `go-contrib/detectors/`, which means Docker-based detection is only available to collector-contrib pipelines and not to any Go SDK user.

## Proposed Solution

Add a new `detectors/docker` package implementing the docker detector as a `resource.Detector`, ported from the collector-contrib implementation, so that `collector-contrib/processor/resourcedetectionprocessor` can delegate to it and any other Go SDK user can adopt it.

## Prior Art

- Collector-contrib implementation to port from: [`processor/resourcedetectionprocessor/internal/docker`](https://github.com/open-telemetry/opentelemetry-collector-contrib/tree/main/processor/resourcedetectionprocessor/internal/docker)

## Additional Context

- Uses the Docker client to read the OS type from the Docker engine info endpoint.
- Emits `os.type` based on the Docker engine's reported OS.

ps-mir:
Port docker detector from collector-contrib as resource.Detector
Fixes: #8951 
Relates: #8940 

### Changes
1) The detector depends on a metaprovider 
`https://github.com/open-telemetry/opentelemetry-collector-contrib/tree/main/internal/metadataproviders/docker`

which is added to `provider.go`

2) The provider also calls a `internal` package function
`https://github.com/open-telemetry/opentelemetry-collector-contrib/blob/main/internal/metadataproviders/internal/internal.go`

It's imported as module under detector, not added in autodetect, but added in `versions.yaml`, since it needs to ship together with docker.

It cannot be put inside Docker because System Detector also uses it.

### Testing
Unit tests for happy and error path

Manual Integration Test (Not added to code)

```go

//go:build integration

package docker

import (
	"context"
	"testing"

	"github.com/stretchr/testify/require"
	semconv "go.opentelemetry.io/otel/semconv/v1.40.0"
)

func TestDockerResourceDetector_Integration(t *testing.T) {
	res, err := NewResourceDetector().Detect(context.Background())
	require.NoError(t, err)
	require.NotNil(t, res)

	if res.Len() == 0 {
		t.Log("detector returned empty resource, not running inside a Docker container")
		return
	}
	t.Logf("detector returned %d attribute(s)", res.Len())

	attrs := res.Set()

	validOSTypes := map[string]string{
		semconv.OSTypeLinux.Value.AsString():        "OSTypeLinux",
		semconv.OSTypeWindows.Value.AsString():      "OSTypeWindows",
		semconv.OSTypeDarwin.Value.AsString():       "OSTypeDarwin",
		semconv.OSTypeFreeBSD.Value.AsString():      "OSTypeFreeBSD",
		semconv.OSTypeNetBSD.Value.AsString():       "OSTypeNetBSD",
		semconv.OSTypeOpenBSD.Value.AsString():      "OSTypeOpenBSD",
		semconv.OSTypeDragonflyBSD.Value.AsString(): "OSTypeDragonflyBSD",
		semconv.OSTypeHPUX.Value.AsString():         "OSTypeHPUX",
		semconv.OSTypeAIX.Value.AsString():          "OSTypeAIX",
		semconv.OSTypeSolaris.Value.AsString():      "OSTypeSolaris",
		semconv.OSTypeZOS.Value.AsString():          "OSTypeZOS",
	}

	osType, exists := attrs.Value(semconv.OSTypeKey)
	assert.True(t, exists, "os.type attribute should be present")
	constName, valid := validOSTypes[osType.AsString()]
	assert.True(t, valid, "os.type %q is not a known semconv constant", osType.AsString())
	t.Logf("os.type: %q → semconv.%s", osType.AsString(), constName)

	hostname, exists := attrs.Value(semconv.HostNameKey)
	assert.True(t, exists, "host.name attribute should be present")
	assert.NotEmpty(t, hostname.AsString())
	t.Logf("host.name: %s", hostname.AsString())

	containerName, exists := attrs.Value(semconv.ContainerNameKey)
	assert.True(t, exists, "container.name attribute should be present")
	assert.NotEmpty(t, containerName.AsString())
	t.Logf("container.name: %s", containerName.AsString())

	containerImageID, exists := attrs.Value(semconv.ContainerImageIDKey)
	assert.True(t, exists, "container.image.id attribute should be present")
	assert.NotEmpty(t, containerImageID.AsString())
	t.Logf("container.image.id: %s", containerImageID.AsString())

	containerImage, exists := attrs.Value(semconv.ContainerImageNameKey)
	if exists {
		assert.NotEmpty(t, containerImage.AsString())
		assert.NotContains(t, containerImage.AsString(), ":", "container.image.name should not contain a tag")
		t.Logf("container.image.name: %s", containerImage.AsString())
	} else {
		t.Log("container.image.name: not present (container was referenced by bare image ID)")
	}

	containerImageTags, exists := attrs.Value(semconv.ContainerImageTagsKey)
	if exists {
		t.Logf("container.image.tags: %v", containerImageTags.AsStringSlice())
	} else {
		t.Log("container.image.tags: not present (image reference had no tag)")
	}
}

```

```bash
docker run --rm     -v /var/run/docker.sock:/var/run/docker.sock     -v $(pwd):/workspace     -v $(go env GOPATH)/pkg/mod:/go/pkg/mod     -w /workspace/detectors/docker     golang:1.25     go test -tags integration -v .
```
Output
```
=== RUN   TestDockerResourceDetector_Integration
    detector_integration_test.go:42: os.type: "linux" → semconv.OSTypeLinux
    detector_integration_test.go:47: host.name: coder-gwarps-otel-go-root
    detector_integration_test.go:52: container.name: /crazy_lederberg
    detector_integration_test.go:57: container.image.name: sha256:315d6d304226d0180a011ead57af82c9fe16f84fcf0eaa17d11034a53f63fbc4
--- PASS: TestDockerResourceDetector_Integration (0.03s)
PASS
ok  	go.opentelemetry.io/contrib/detectors/docker	0.034s
```

### Issues (Fixed)

Remapped OS type from internal.go (z_os) doesn't map to any existing semconv OsType constant (zos).
```go
func TestInternal(t *testing.T) {
	assert.Equal(t, semconv.OSTypeDragonflyBSD.Value.AsString(), internal.GOOSToOSType("dragonfly"))
	//The following two are failing
	assert.Equal(t, semconv.OSTypeZOS.Value.AsString(), internal.GOOSToOSType("zos"))
	assert.Equal(t, semconv.OSTypeZOS.Value.AsString(), internal.GOOSToOSType("z_os"))
}
```

Update: remap zos to z_os for Ostype has been removed.


ps-mir:
@iblancasa @paulojmdias PR is ready for review.

ps-mir:
@iblancasa @paulojmdias 
As [discussed](https://github.com/open-telemetry/opentelemetry-go-contrib/issues/8951#issuecomment-4481138763) with @dashpole, the Docker detector is complementary to the SDK-based container detector and cannot be merged due to portability differences.

The PR has some code duplication for OsType check (internal.go) with go-sdk, which is acceptable.

detectors/docker/detector.go
```diff
@@ -0,0 +1,52 @@
...
+	containerInfo, err := dockerProvider.ContainerInfo(ctx)
+	if err != nil {
+		return resource.Empty(), fmt.Errorf("failed getting container info: %w", err)
+	}
+
+	return resource.NewWithAttributes(
+		semconv.SchemaURL,
+		semconv.HostName(hostname),
+		semconv.OSTypeKey.String(osType),
+		semconv.ContainerName(containerInfo.Name),
```
thompson-tomo:
```suggestion
		semconv.ContainerId(containerInfo.Id),
		semconv.ContainerName(containerInfo.Name),
```

detectors/docker/detector.go
```diff
@@ -0,0 +1,52 @@
...
+		semconv.ContainerImageName(containerInfo.Image),
```
thompson-tomo: Is there a way to also get image.id as per semconv

thompson-tomo:
As per semconv https://opentelemetry.io/docs/specs/semconv/registry/entities/container/ shouldn't we be adding container.id & container.image.id

ps-mir:
> As per semconv https://opentelemetry.io/docs/specs/semconv/registry/entities/container/ shouldn't we be adding container.id & container.image.id

`container.id` is covered by sdk detector (WithContainer), so we can skip that. 
`container.image.id` can be added, but it doesn't port cleanly to k8s env. I would wait for a use case or a user to ask for this attribute before adding it. giving priority to the most used attributes.

@dashpole Any guidance on adding spec attributes to a resource?

dashpole:
Lets focus on parity with the collector implementation, and open follow-up issues for these questions to keep things simple.

detectors/docker/detector.go
```diff
@@ -0,0 +1,78 @@
...
+	createProvider func(...client.Opt) (provider, error)
+}
+
+// Detect returns a [resource.Resource] containing Docker host and container
+// attributes for the running container. It returns an empty resource and no
+// error when the Docker daemon is unreachable (not a Docker environment). If
+// the daemon is reachable but some attributes cannot be retrieved, a partial
+// resource is returned together with [resource.ErrPartialResource].
+func (r *resourceDetector) Detect(ctx context.Context) (*resource.Resource, error) {
+	// A provider failure means the Docker daemon is unreachable, which is
```
iblancasa: I think this fallback does not match the real runtime behavior. `newProvider()` only constructs the Docker client; it does not verify daemon reachability. Missing socket / refused connection / bad `DOCKER_HOST` errors happen later in `Info()` / `ContainerInspect()`, and those paths currently return `resource.ErrPartialResource`, not `resource.Empty(), nil`

detectors/docker/provider.go
```diff
@@ -0,0 +1,66 @@
...
+	OSType(context.Context) (string, error)
+	ContainerInfo(context.Context) (container.InspectResponse, error)
+}
+
+type dockerProviderImpl struct {
+	dockerClient *client.Client
+}
+
+func (d *dockerProviderImpl) Hostname(ctx context.Context) (string, error) {
+	sysInfoResult, err := d.dockerClient.Info(ctx, client.InfoOptions{})
```
iblancasa: Can we call `Info()` once and reuse it for both attributes?

detectors/autodetect/autodetect.go
```diff
@@ -154,6 +159,8 @@ var (
 		IDProcessRuntimeDescription: optFactory(resource.WithProcessRuntimeDescription()),
 
 		IDContainer: optFactory(resource.WithContainer()),
+
+		IDDocker: docker.NewResourceDetector,
```
iblancasa: Adding this to default `autodetect` seems too risky. Reachable Docker daemon != process is running in Docker. Today this detector can run on ordinary hosts with Docker installed, then fail `ContainerInspect(os.Hostname())` or return daemon-level metadata for a non-container process...

Maybe reading `/proc/self/cgroup`?

ps-mir:
@iblancasa ContainerInfo API is now included in detection since os, hostname work on host with Docker installed. This will make autodetect more accurate. Duplicate calls are removed.

Test script output

Inside docker container
```bash
=== RUN   TestDockerResourceDetector_Integration
    detector_integration_test.go:26: detector returned 4 attribute(s)
    detector_integration_test.go:48: os.type: "linux" → semconv.OSTypeLinux
    detector_integration_test.go:53: host.name: coder-gwarps-otel-go-root
    detector_integration_test.go:58: container.name: /relaxed_hawking
    detector_integration_test.go:63: container.image.name: golang:1.25
```

Outside container in host (docker installed)
```bash
=== RUN   TestDockerResourceDetector_Integration
    detector_integration_test.go:23: detector returned an empty resource, not running inside a Docker container
```

iblancasa (approved):

paulojmdias (approved):

detectors/docker/detector.go
```diff
@@ -0,0 +1,69 @@
...
+
+	if len(errs) > 0 {
+		return res, fmt.Errorf("%w: %w", resource.ErrPartialResource, errors.Join(errs...))
+	}
+	return res, nil
+}
+
+// NewResourceDetector returns a [resource.Detector] that detects resource
+// attributes for Docker containers using the local Docker daemon.
+func NewResourceDetector() resource.Detector {
```
dashpole: prefer returning concrete types, rather than interfaces (e.g. *ResourceDetector)

dashpole: Please also make this accept options, including WithAttributeFilter for consistency with other detectors.

versions.yaml
```diff
@@ -78,6 +78,8 @@ module-sets:
       - go.opentelemetry.io/contrib/detectors/autodetect
       - go.opentelemetry.io/contrib/detectors/azure/azurevm
       - go.opentelemetry.io/contrib/detectors/hetzner
+      - go.opentelemetry.io/contrib/detectors/docker
+      - go.opentelemetry.io/contrib/detectors/internal
```
dashpole: Please don't add this as a new module.  We don't allow internal cross-module dependencies.  Instead, please move this to detectors/docker/internal

detectors/docker/doc.go
```diff
@@ -0,0 +1,25 @@
+// Copyright The OpenTelemetry Authors
+// SPDX-License-Identifier: Apache-2.0
+
+/*
+Package docker provides a [resource.Detector] which supports detecting
+attributes for processes running inside Docker containers.
+
+The detector queries the Docker daemon via its API and requires the daemon
+socket to be accessible to the process (e.g. /var/run/docker.sock mounted
+into the container). If the socket is unavailable or any daemon call fails,
```
iblancasa: I think implementation returns no error for provider and container inspect failures, and partial error only for Info. Is that right? Can you fix?

detectors/docker/detector.go
```diff
@@ -0,0 +1,108 @@
...
+	if err != nil {
+		return resource.Empty(), nil
+	}
+
+	var (
+		attrs []attribute.KeyValue
+		errs  []error
+	)
+
+	attrs = append(attrs, semconv.ContainerName(containerInfo.Name), semconv.ContainerImageName(containerInfo.Config.Image))
```
iblancasa: Can `containerInfo.Config.Image` if `containerInfo.Config` is `nil`?

detectors/docker/detector.go
```diff
@@ -0,0 +1,108 @@
...
+func (r *ResourceDetector) Detect(ctx context.Context) (*resource.Resource, error) {
+	// A provider failure means the Docker daemon is unreachable, which is
+	// indistinguishable from not running in a Docker environment. Return an
+	// empty resource with no error so autodetect can continue with other detectors.
+	dockerProvider, err := r.createProvider()
+	if err != nil {
+		return resource.Empty(), nil
+	}
+
+	containerInfo, err := dockerProvider.ContainerInfo(ctx)
```
iblancasa, line 80: Maybe we should at least log something? We need to distinguish between "I'm not running on a Docker container" and "there was a problem detecting if I'm running on a Docker container"

iblancasa (approved):

ps-mir:
The provider interface is purely an adapter over the Docker client SDK. It exists to allow the detector to be unit-tested with a mock. Unsure of the value provider tests will add.

dashpole (approved):

dashpole:
@open-telemetry/go-approvers This needs another approval if anyone has time to review

ps-mir:
To ensure the use of the parent context, the `NewResourceDetector` only reads config. It doesn't do any work that could be cancelled or time-bounded.

`WithContext` plumbing is not needed for `NewResourceDetector`

@MrAlias @pellared

detectors/docker/detector.go
```diff
@@ -0,0 +1,117 @@
...
+		slog.Warn("docker container info", slog.Any("err", err))
+		return resource.Empty(), nil
+	}
+
+	var (
+		attrs []attribute.KeyValue
+		errs  []error
+	)
+
+	attrs = append(attrs, semconv.ContainerName(containerInfo.Name))
```
MrAlias, line 94: `Config.Image` is a Docker reference such as golang:1.25, but semantic conventions require `container.image.name` to contain only the image name; the tag belongs in `container.image.tags`.

detectors/docker/detector.go
```diff
@@ -0,0 +1,117 @@
...
+	if err != nil {
+		slog.Warn("docker client config", slog.Any("err", err))
+		return resource.Empty(), nil
+	}
+
+	// ContainerInfo is the primary detection mechanism
+	containerInfo, err := dockerProvider.ContainerInfo(ctx)
+	if err != nil {
+		slog.Warn("docker container info", slog.Any("err", err))
+		return resource.Empty(), nil
```
MrAlias: Returning errors is preferred to logging.


```suggestion
		return resource.Empty(), err
```

Or if you would like to keep the scope:


```suggestion
		fmt.Errorf("docker container info: %w" err)
		return resource.Empty(), err
```

detectors/docker/provider.go
```diff
@@ -0,0 +1,70 @@
...
+	var image *string
+	if result.Container.Config != nil {
+		image = &result.Container.Config.Image
+	}
+	return containerInfo{Name: result.Container.Name, Image: image}, nil
+}
+
+func newProvider(opts ...client.Opt) (provider, error) {
+	opts = append([]client.Opt{client.FromEnv}, opts...)
+	cli, err := client.New(opts...)
```
MrAlias, line 132: These clients are never closed. Lifecycle cleanup is needed here.

CHANGELOG.md
```diff
@@ -8,6 +8,8 @@ This project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0.htm
 
 ## [Unreleased]
 
+- Add `go.opentelemetry.io/contrib/detectors/docker`, a resource detector for docker, from `processor/resourcedetectionprocessor/internal/docker`. (#8951)
+
 ### Added
```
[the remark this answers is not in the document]
MrAlias: This is not resolved.

detectors/docker/detector.go
```diff
@@ -0,0 +1,120 @@
...
+	var cfg config
+	for _, opt := range opts {
+		opt.apply(&cfg)
+	}
+	return &ResourceDetector{createProvider: newProvider, cfg: cfg}
+}
+
+// Detect returns a [resource.Resource] containing Docker host and container
+// attributes for the running container. It returns an empty resource and no
+// error when the Docker daemon is unreachable (not a Docker environment). If
```
MrAlias: This is still not how detection behaves. `client.New` only constructs the client; it does not contact the daemon. A missing socket or refused connection fails in `ContainerInfo` below, which returns a non-nil error, so the documented empty-resource-without-error behavior does not occur when the daemon is unreachable.

Please either implement that fallback for daemon connection failures or update this comment, the package documentation, and the README to describe the actual error behavior.

detectors/docker/provider.go
```diff
@@ -0,0 +1,105 @@
...
+// registry host:port, e.g. "localhost:5000/name").
+func splitImageRef(ref string) (name, tag string) {
+	if i := strings.Index(ref, "@"); i != -1 {
+		ref = ref[:i] // strip the digest before looking for a tag
+	}
+	rest := ref
+	if slash := strings.LastIndex(ref, "/"); slash != -1 {
+		rest = ref[slash+1:]
+	}
+	if i := strings.LastIndex(rest, ":"); i != -1 {
```
MrAlias, line 40: `Config.Image` preserves the image reference supplied when the container is created. If the container was started by a local image ID such as `docker run sha256:<digest>`, this branch reports `container.image.name="sha256"` and puts the digest in `container.image.tags`. That is not a name/tag pair.

Please detect bare image IDs before splitting on `:` and avoid emitting them as `container.image.name` or `container.image.tags`. The inspect response's top-level `Image` field is the actual image ID and maps to `container.image.id` if we want to expose it.

detectors/docker/provider.go
```diff
@@ -0,0 +1,105 @@
...
+		tags  []string
+	)
+	if result.Container.Config != nil && result.Container.Config.Image != "" {
+		name, tag := splitImageRef(result.Container.Config.Image)
+		image = &name
+		if tag != "" {
+			tags = []string{tag}
+		}
+	}
+	return containerInfo{Name: result.Container.Name, Image: image, Tags: tags}, nil
```
MrAlias: `ContainerInspect` returns `Name` with a leading slash; the integration output in this PR shows values such as `/relaxed_hawking`. Emitting it unchanged makes `container.name` differ from the runtime name users see. The collector-contrib Docker receiver normalizes this with `strings.TrimPrefix`.

```suggestion
	return containerInfo{
		Name:  strings.TrimPrefix(result.Container.Name, "/"),
		Image: image,
		Tags:  tags,
	}, nil
```

MrAlias (commented):
There are three remaining correctness issues to address.

detectors/docker/provider.go
```diff
@@ -0,0 +1,117 @@
...
+	// they are equal, the operator referenced the container by bare digest
+	// (e.g. "docker run sha256:<id>"), so there is no name/tag to report.
+	if result.Container.Config != nil && result.Container.Config.Image != "" &&
+		result.Container.Config.Image != result.Container.Image {
+		name, tag := splitImageRef(result.Container.Config.Image)
+		imageName = &name
+		if tag != "" {
+			tags = []string{tag}
+		}
+	}
```
MrAlias, line 115: This still lets some bare image IDs through as image names. Docker accepts image ID prefixes and IDs without the `sha256:` prefix; in those cases `Config.Image` will not equal the resolved `Container.Image`, so this branch calls `splitImageRef` and emits an ID or `sha256` as `container.image.name`.

Please classify bare image IDs and ID prefixes before splitting the reference. Name/tag should only be emitted for real image references; the resolved image ID is already reported via `container.image.id`.

```suggestion
	var (
		imageName *string
		tags      []string
	)
	// Config.Image is the reference the operator supplied when creating the
	// container; Container.Image is always the resolved image digest. Bare image
	// ID references may be the full digest, the hex ID without the algorithm
	// prefix, or a unique prefix, so suppress name/tag for those forms.
	if result.Container.Config != nil && result.Container.Config.Image != "" {
		ref := result.Container.Config.Image
		id := strings.TrimPrefix(result.Container.Image, "sha256:")
		refID := strings.TrimPrefix(strings.ToLower(ref), "sha256:")
		isBareImageID := refID != "" && len(refID) >= 12 && len(refID) <= len(id) && strings.HasPrefix(id, refID)
		for _, c := range refID {
			if !(('0' <= c && c <= '9') || ('a' <= c && c <= 'f')) {
				isBareImageID = false
				break
			}
		}
		if !isBareImageID {
			name, tag := splitImageRef(ref)
			imageName = &name
			if tag != "" {
				tags = []string{tag}
			}
		}
	}
```

MrAlias (changes requested):
There is one remaining image-reference issue to resolve in the current patch. The existing thread about the Docker changelog entry being outside `### Added` is also still unresolved.

ps-mir:
@MrAlias Resolved the build issues and merge conflict. Would appreciate one more look.

detectors/docker/provider.go
```diff
@@ -0,0 +1,129 @@
...
+	// container ID by default, which ContainerInspect can resolve. This breaks
+	// when a custom hostname is set via --hostname or an orchestrator (e.g.
+	// Kubernetes pod.spec.hostname), in which case ContainerInspect returns
+	// "no such container" (a NotFound error) and Detect treats this the same
+	// as not running in a Docker container: an empty resource, no error.
+	hostname, err := os.Hostname()
+	if err != nil {
+		return containerInfo{}, err
+	}
+	result, err := d.dockerClient.ContainerInspect(ctx, hostname, client.ContainerInspectOptions{})
```
MrAlias, line 75: Using the process hostname as the container ID makes a Docker container started with `--hostname` (or an orchestrator-assigned hostname) look like a non-Docker process. `ContainerInspect` then returns 404 and `Detect` silently returns an empty resource.

Please resolve the container ID independently of the hostname—for example, with the cgroup parsing used by the SDK's container detector—and add a custom-hostname regression test.

MrAlias (commented):
Custom hostnames are not detected.

detectors/docker/internal/internal.go
```diff
@@ -0,0 +1,30 @@
...
+		return "arm32"
+	case "ppc64le":
+		return "ppc64"
+	case "386":
+		return "x86"
+	}
+
+	// Other cases either match the spec or are not well-known (so we use a custom value)
+	return goarch
+}
```
dashpole: I don't think this helper function is used?

detectors/docker/detector.go
```diff
@@ -0,0 +1,131 @@
...
+		filtered := attrs[:0]
+		for _, kv := range attrs {
+			if r.cfg.filter(kv) {
+				filtered = append(filtered, kv)
+			}
+		}
+		attrs = filtered
+	}
+
+	res := resource.NewWithAttributes(semconv.SchemaURL, attrs...)
```
MrAlias, line 125: This resource uses semconv v1.43.0, while the pinned `go.opentelemetry.io/otel/sdk` revision's built-in host and container detectors still emit v1.42.0. After a successful Docker detection, composing `IDDocker` with `IDHost` fails with `resource.ErrSchemaURLConflict`; I reproduced the 1.42.0/1.43.0 merge failure. `IDContainer` has the same failure when it finds an ID, even though this PR relies on that detector for `container.id`.

Please update the OpenTelemetry SDK dependency to the current `main` revision so all of these detectors use semconv v1.43.0; this should not be resolved by downgrading the Docker detector. Please also add a composition regression test for `IDDocker` with the built-in detectors.

MrAlias (commented):
The OpenTelemetry SDK dependency needs to be updated to the current `main` revision so successful Docker detection composes with the built-in resource detectors on the latest semconv schema. The custom-hostname behavior is covered in the existing thread.

CHANGELOG.md
```diff
@@ -28,6 +28,7 @@ This project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0.htm
...
 - Add `NewResourceDetector` along with the `WithAttributeFilter` and `WithTagKeyFilter` options in `go.opentelemetry.io/contrib/detectors/azure/azurevm`. `WithAttributeFilter` restricts the returned resource to the attributes the filter accepts. `WithTagKeyFilter` opts in to `azure.tag.<name>` attributes for the VM tags whose keys satisfy the provided predicate; no VM tags are emitted without it. (#9162)
 - Add `go.opentelemetry.io/contrib/detectors/vultr` — a new resource detector for Vultr Cloud Compute instances, ported from `processor/resourcedetectionprocessor/internal/vultr` in `opentelemetry-collector-contrib`. Detects `cloud.provider`, `cloud.platform`, `cloud.region`, `host.id`, and `host.name`. (#8995)
+- Add `go.opentelemetry.io/contrib/detectors/docker`, a resource detector for Docker, ported from `processor/resourcedetectionprocessor/internal/docker` in opentelemetry-collector-contrib. (#9001)
```
MrAlias: The v1.45.0/v0.17.0 release has already been cut, so this entry now changes the locked released section instead of documenting an unreleased change. This is why `protect-released-changelog` is failing. Please move it under `## [Unreleased]` in a new `### Added` section.

detectors/docker/version.go
```diff
@@ -0,0 +1,7 @@
+// Copyright The OpenTelemetry Authors
+// SPDX-License-Identifier: Apache-2.0
+
+package docker
+
+// Version is the current release version of the docker resource detector.
+const Version = "0.16.0"
```
MrAlias: This module missed the v1.45.0/v0.17.0 release synchronization. `Version` still reports `0.16.0`, while `versions.yaml` places Docker in the v0.17.0 detector set and autodetect requires `docker v0.17.0`; `detectors/docker/go.mod` also retains the pre-release OpenTelemetry revisions that the release replaced with v1.45.0. Please apply the same post-release version and dependency update to this module.

MrAlias (commented):
The schema-composition issue is resolved. The branch still needs the post-release metadata and changelog refresh; the custom-hostname limitation is already covered in the existing thread.

detectors/docker/provider.go
```diff
@@ -0,0 +1,129 @@
...
+		for _, c := range refID {
+			isHexDigit := ('0' <= c && c <= '9') || ('a' <= c && c <= 'f')
+			if !isHexDigit {
+				isBareImageID = false
+				break
+			}
+		}
+		if !isBareImageID {
+			name, tag := splitImageRef(ref)
+			imageName = &name
```
dashpole, line 110: If `name` is empty (such as for a digest reference like `@sha256:...`), `imageName` becomes a non-nil pointer to `""`, causing `Detect` to emit `container.image.name=""`. Consider only setting `imageName` when `name != ""`:
```suggestion
			if name != "" {
				imageName = &name
			}
```

detectors/docker/provider.go
```diff
@@ -0,0 +1,129 @@
...
+	)
+	// Config.Image is the reference the operator supplied when creating the
+	// container; Container.Image is always the resolved image digest. Bare image
+	// ID references may be the full digest, the hex ID without the algorithm
+	// prefix, or a unique prefix, so suppress name/tag for those forms.
+	if result.Container.Config != nil && result.Container.Config.Image != "" {
+		ref := result.Container.Config.Image
+		id := strings.TrimPrefix(result.Container.Image, "sha256:")
+		refID := strings.TrimPrefix(strings.ToLower(ref), "sha256:")
+		isBareImageID := refID != "" && len(refID) >= 12 && len(refID) <= len(id) && strings.HasPrefix(id, refID)
```
dashpole: If a container is started with a short `sha256:` prefix (e.g. `sha256:1a2b3c4d`), `len(refID) >= 12` evaluates to false, causing `splitImageRef` to split it into `name = "sha256"` and `tag = "1a2b3c4d"`. If `ref` explicitly starts with `sha256:`, consider treating it as a bare ID without requiring length >= 12.

detectors/docker/provider_test.go
```diff
@@ -0,0 +1,230 @@
...
+const testImageID = "sha256:e3b0c44298fc1c149afbf4c8996fb92427ae41e4649b934ca495991b7852b855"
+
+func TestDockerProviderImpl_ContainerInfo(t *testing.T) {
+	hostname, err := os.Hostname()
+	require.NoError(t, err)
+
+	p := newTestProvider(t, func(req *http.Request) (*http.Response, error) {
+		assert.Equal(t, "/v"+client.MaxAPIVersion+"/containers/"+hostname+"/json", req.URL.Path)
+		return jsonResponse(t, http.StatusOK, container.InspectResponse{
+			Name:   "my-container",
```
dashpole: Docker Engine returns container names with a leading slash (e.g. `"/my-container"`). Using `"/my-container"` in test fixtures would verify that `TrimPrefix` correctly strips the slash from genuine daemon responses.
```suggestion
			Name:   "/my-container",
```

detectors/docker/provider.go
MrAlias: This remains unresolved at 5152d826. `ContainerInfo` still passes `os.Hostname()` to `ContainerInspect`, so a container whose assigned hostname does not match its container ID or name still produces a not-found response that `Detect` converts to an empty resource. Please resolve the container ID independently of the hostname and add the requested regression test.

detectors/docker/provider.go
ps-mir: Pushback. `name is mandatory` is enforced at dockerd level. So it gets applied to all type of callers.

Example

```bash
$ curl --unix-socket /var/run/docker.sock -d '{"Image":"@sha256:28bd5fe8..."}' .../containers/create
  {"message":"invalid reference format"}

$ curl --unix-socket /var/run/docker.sock -d '{"Image":":latest"}' .../containers/create
  {"message":"invalid reference format"}
```

detectors/docker/provider.go
ps-mir: [77baad3](https://github.com/open-telemetry/opentelemetry-go-contrib/pull/9001/commits/77baad33e75ec534ed78522b67bb6515abd904da)

detectors/docker/provider_test.go
ps-mir: [7210875](https://github.com/open-telemetry/opentelemetry-go-contrib/pull/9001/commits/7210875bc446e3d0f8222a26e2a1e24bdc6b1bf6)

ps-mir:
#9549 is created to followup container ID check improvements. This PR can be merged as it covers the scope of porting logic from collector-contrib.

@MrAlias @pellared

MrAlias (approved):

MrAlias:
@dashpole PTAL

dashpole (approved):

pellared:
I moved to the next milestone as I am not sure if we want to merge this without https://github.com/open-telemetry/opentelemetry-go-contrib/issues/9549 in the same release. @MrAlias, @dashpole feel free to merge the PR if you think it can land in the upcoming release.

dashpole:
https://github.com/open-telemetry/opentelemetry-go-contrib/issues/9549 doesn't seem like a blocker to me.

ps-mir:
CI is green.

PR #9002, opened 2026-05-18
chore: re-enable revive.context-as-argument rule

### LINKED ISSUE #2877: Reenable context-as-argument linter.
context-as-arrugment linter was disabled because it can't be used with viper v1.13.0.  This should track [golangci-lint](https://github.com/golangci/golangci-lint/issues/3280) and be reenabled when it is compatible.

alexandear:
Enable `revive.context-as-argument` rule, since `allow-types-before` is no longer an issue.

Remove the deprecated `skip-package-name-checks` and `skip-package-name-collision-with-go-std` options from the `var-naming` rule.

Fixes #2877

P.S. I'm a Revive maintainer.

### DIFF
--- a/.golangci.yml
+++ b/.golangci.yml
@@ -109,11 +109,9 @@ linters:
         - name: blank-imports
         - name: bool-literal-in-expr
         - name: constant-logical-expr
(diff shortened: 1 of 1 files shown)


dashpole (approved):
Thanks @alexandear!

pellared (approved):

PR #9006, opened 2026-05-21
bump OTel Go

pellared:


zpages/tracez.go
```diff
@@ -232,7 +232,7 @@ func spanRows(s sdktrace.ReadOnlySpan) []spanRow {
 		sort.Sort(a)
 		var s []string
 		for i := range a {
-			s = append(s, fmt.Sprintf("%s=%v", a[i].Key, a[i].Value.Emit()))
+			s = append(s, fmt.Sprintf("%s=%v", a[i].Key, a[i].Value.Emit())) //nolint:staticcheck // Use deprecated method for formatting backward compatibility.
```
pellared, line 235: @dashpole, do we want to change this to follow non-OTLP string formatting per spec?

Reference: https://github.com/open-telemetry/opentelemetry-specification/blob/main/specification/common/README.md#anyvalue-representation-for-non-otlp-protocols

If so then I think I can do a separate PR to make it cleaner.

dashpole: I think it is a good idea to change this, and I agree we should do it in a follow-up

dashpole (approved):

zpages/tracez.go
pellared: I created https://github.com/open-telemetry/opentelemetry-go-contrib/issues/9008

XSAM (approved):

Issue #3185, opened 2023-01-26
Bug: http/2.0 requests tagged as http.flavor 1.1 (instrumentation/net/http/otelhttp)


ldemailly:
Making an http2.0 connection using net/http default transport (negotiating h2 over tls through ForceAttemptHTTP2:     true) wrapped by otelhttp.NewTransport generates a trace/span with the wrong flavor: 1.1 instead of 2

<img width="684" alt="Screen Shot 2023-01-26 at 11 03 29 AM" src="https://user-images.githubusercontent.com/3664595/214926286-2b9474f0-796c-4c4f-b344-d4e3216c98ed.png">

```
HTTP GET
Tags
http.flavor	1.1
http.host	 debug.fortio.org
http.method	GET
http.scheme	http
http.status_code	200
http.url	https://debug.fortio.org
internal.span.format	proto
otel.library.name	go.opentelemetry.io/contrib/instrumentation/net/http/otelhttp
otel.library.version	semver:0.37.0
span.kind	client
Process 
telemetry.sdk.language	go
telemetry.sdk.name	opentelemetry
telemetry.sdk.version	1.11.2
```

code: https://github.com/fortio/fortiotel/blob/main/simple/otel_httptrace_simple.go

output
```
2023/01/26 10:53:31 OTEL export pipeline setup successfully - making a single request to -url https://debug.fortio.org
2023/01/26 10:53:32 Response for "https://debug.fortio.org": 200 OK
2023/01/26 10:53:32 Φορτίο version 1.40.0 h1:jSDO/jGcyC/qTpMZZ84EZbn9BQawsWM9/RMQ9s6Cn3w= go1.19.5 arm64 linux (in fortio.org/proxy 1.6.1) echo debug server up for 13h0m11.1s - request from 216.194.105.121:28096 https TLS_AES_128_GCM_SHA256

GET / HTTP/2.0

headers:

Host: debug.fortio.org
Accept-Encoding: gzip
Traceparent: 00-fa939e35d51ce76cb7064f77867ee33c-d67f1ef81823588e-01
User-Agent: Go-http-client/2.0

body:



2023/01/26 10:53:32 OTEL export pipeline shut down successfully
```
(as you can see it did make an h2 call, the debug endpoint echoes back everything and there is no proxy in between the client and that server)

Sovietaced:
I think this issue may be obsolete since protocol name/version are no longer recorded attributes.

dmathieu:
[...]

PR #9010, opened 2026-05-22
fix(otelhttp): record network.protocol.version from the response

### LINKED ISSUE #9007: [Bug]: otelhttp records `network.protocol.version` from request struct, ignoring ALPN-negotiated wire protocol
### Component

Instrumentation: otelhttp

`go.opentelemetry.io/contrib/instrumentation/net/http/otelhttp` records `network.protocol.version` (and `network.protocol.name`) from `req.Proto` rather than from the response's actual wire protocol. When a request created by `http.NewRequest` (which always stamps `Proto = "HTTP/1.1"`) is upgraded to HTTP/2 via ALPN by the underlying transport, both the client span and the request-duration / request-body-size metrics still report `network.protocol.version=1.1`, masking the real wire protocol from anyone using these signals to identify HTTP/1 vs HTTP/2 traffic.

This is the same issue as reported by #3185 ("http/2.0 requests tagged as http.flavor 1.1"). That issue was closed in 2023 because semconv 1.20 dropped `http.flavor` from the recorded set. The attribute was later re-added under the new name `network.protocol.version` during the migration through semconv 1.24 -> 1.39 -> 1.40, sourced again from `req.Proto`, so the underlying defect returned, just renamed.

The current code paths are in the shared template `internal/shared/semconv/client.go.tmpl` (regenerated into all six HTTP consumer packages):
- `HTTPClient.RequestTraceAttrs` reads `req.Proto`.
- `HTTPClient.MetricAttributes` reads `req.Proto`.
- `HTTPClient.ResponseTraceAttrs` does not currently emit network protocol attributes at all, so nothing later overwrites the stale request-side values on the span.

### Expected behavior

Per the [OpenTelemetry HTTP semconv](https://opentelemetry.io/docs/specs/semconv/http/http-spans/) for `network.protocol.version`:

> If protocol version is subject to negotiation (for example using ALPN), this attribute SHOULD be set to the negotiated version.

For an HTTP/1.1-shaped `*http.Request` whose round trip is served by HTTP/2, both the client span and the corresponding metric data points should carry `network.protocol.version=2.0` (the value reflected in `resp.Proto`), not `1.1`.

### Steps to Reproduce
```go
package repro

import (
	"io"
	"net/http"
	"net/http/httptest"
	"testing"

)

func TestProtocolVersionBug(t *testing.T) {
	sr := tracetest.NewSpanRecorder()
	tp := sdktrace.NewTracerProvider(sdktrace.WithSpanProcessor(sr))

	ts := httptest.NewUnstartedServer(http.HandlerFunc(func(w http.ResponseWriter, _ *http.Request) {
		w.WriteHeader(http.StatusOK)
	}))
	ts.EnableHTTP2 = true

	ts.StartTLS()
	defer ts.Close()

	c := &http.Client{
		Transport: otelhttp.NewTransport(ts.Client().Transport,
			otelhttp.WithTracerProvider(tp),
		),
	}

	req, err := http.NewRequestWithContext(t.Context(), http.MethodGet, ts.URL, http.NoBody)
	if err != nil {
		t.Fatal(err)
	}

	res, err := c.Do(req)
	if err != nil {
		t.Fatal(err)
	}

	_, _ = io.Copy(io.Discard, res.Body)
	_ = res.Body.Close()

	if res.Proto != "HTTP/2.0" {
		t.Fatalf("test setup did not negotiate HTTP/2; got %q", res.Proto)
	}

	for _, kv := range sr.Ended()[0].Attributes() {
		if kv.Key == "network.protocol.version" {
			if got := kv.Value.AsString(); got != "2.0" {
				t.Errorf("network.protocol.version = %q, want %q (request was served over %s)", got, "2.0", res.Proto)
			}

			return
		}
	}

	t.Error("span has no network.protocol.version attribute")
}
```
### Operating System

macOS 15

### Device Architecture

ARM64

### Go Version

1.26

### Component Version

`go.opentelemetry.io/contrib/instrumentation/net/http/otelhttp` v0.68.0 (also reproduces on `main` at `ba143fc95`)

danielnovograd:
# otelhttp: record `network.protocol.version` from the response, not the request

Fixes #9007
Relates #3185

## Summary

`otelhttp` currently records `network.protocol.version` from `req.Proto`, which `http.NewRequest` always stamps as `"HTTP/1.1"`. When the underlying transport negotiates HTTP/2 via ALPN, the client span and request metrics still report `1.1`. Per the OTel HTTP semconv spec, `network.protocol.version` SHOULD reflect the negotiated wire protocol when negotiation occurs.

This PR sources the attribute from `resp.Proto` for both trace attributes and metric attributes on the client side. On error (no response), it falls back to `req.Proto` as before.

This is the same defect as #3185, which was closed when semconv 1.20 dropped `http.flavor`. The attribute returned under the new name `network.protocol.version` during the migration through 1.24 -> 1.39 -> 1.40, again sourced from `req.Proto`.

## Changes

- `internal/shared/semconv/client.go.tmpl`
  - `ResponseTraceAttrs` now emits `network.protocol.{name,version}` from `resp.Proto`.
  - `MetricAttributes` signature changes to `(req, resp, additional)` semantics by consolidating the previously-separate `statusCode int` parameter into the existing `*http.Response`. When `resp != nil`, `resp.Proto` and `resp.StatusCode` are used; otherwise `req.Proto` is used and the `http.response.status_code` attribute is omitted.
  - `MetricOptions` interprets `Resp == nil` as "no response, this was a transport error" and only then appends `error.type`.
- `internal/shared/semconv/server.go.tmpl`
  - `MetricAttributes.StatusCode` moved up to `ServerMetricData`. Server-side handlers don't have a `*http.Response` to thread through; status code is recorded by the server wrapper directly.
- `instrumentation/net/http/otelhttp/transport.go`
  - On `RoundTrip` error, `Resp` is left nil so the metric falls back to `req.Proto` and skips the status-code attribute. On success, `Resp` is set to the actual response.
- All consumers regenerated. `otelhttp/handler.go`, `otelmux/mux.go`, `otelgin/gin.go`, and `otelecho/echo.go` updated to pass `StatusCode` at the `ServerMetricData` level. `otelhttptrace` and `otelrestful` don't construct `ServerMetricData` (the former is client-only; the latter records traces but not server metrics), so only their regenerated `internal/semconv` files changed.

## Test plan

- [x] New end-to-end test `TestTransportProtocolVersionHTTP2` in `instrumentation/net/http/otelhttp/transport_test.go` verifies that a request served over HTTP/2 (via `httptest.Server.EnableHTTP2`) records `network.protocol.version=2.0` on the client span.
- [x] New unit test cases in `TestHTTPClient_MetricAttributes` (template, regenerated to all 6 consumers):
  - `nil response falls back to request proto and omits status code` — locks in the no-response branch.
  - `response proto overrides request proto` — locks in the actual fix at unit level (HTTP/1.1 request, HTTP/2.0 response -> `version=2.0`).
  - `http/3 response proto` — documents that the version parser is generic over the `Proto` string, so wrapping a third-party HTTP/3 `http.RoundTripper` (e.g. quic-go's `http3.Transport`, which sets `resp.Proto = "HTTP/3.0"`) records `network.protocol.version=3.0` with no extra code.
- [x] Existing `TestTransportErrorTypeMetricAttribute` extended to assert `http.response.status_code` is absent when the round trip fails.
- [x] `go test ./...` passes in all six affected modules: `otelhttp`, `otelhttptrace`, `otelmux`, `otelgin`, `otelecho`, `otelrestful`.

## Notes

- No public-API change in the consumer packages — `MetricAttributes` is in `internal/semconv`, so its signature change does not affect users.
- The CHANGELOG entry is filed under "Fixed".


MrAlias:
Closing as we are not able to consider changes for merge unless the CLA has been signed. Please re-open if you are able to sign the CLA in the future.

danielnovograd:
Hey @MrAlias I'm unable to re-open. The CLA is signed, would you mind re-opening?

PR #9063, opened 2026-05-29
fix(runtime): use per-bucket count for histogram sum

### LINKED ISSUE #9055: runtime: Runtime histogram sum miscalculated in new producer
### Description

The runtime histogram producer computes histogram sums with cumulative counts instead of per-bucket counts, inflating exported sums. 

`convertRuntimeHistogram` accumulates a cumulative `count` and uses it in the sum (`sum += bounds[i-1] * float64(count)`), which multiplies lower bounds by the running total rather than the per-bucket count (producer.go:78-106). This overestimates sums rather than underestimates them.
Runtime histogram semantics show that `Counts` are per-bucket weights (`Count[n]` is weight for range `[bucket[n], bucket[n+1])`) in Go's `runtime/metrics` definition (GOROOT/src/runtime/metrics/histogram.go:8-12). Therefore, the sum should use `c`, not cumulative `count`.

### Steps To Reproduce

[22-8e2d5e5abf6481918cd93c4f6a1537cb-runtime-histogram-sum-miscalculated-in-new-producer.zip](https://github.com/user-attachments/files/28385614/22-8e2d5e5abf6481918cd93c4f6a1537cb-runtime-histogram-sum-miscalculated-in-new-producer.zip)

A PoC unit test (`instrumentation/runtime/poc_sum_test.go`) demonstrates the issue with buckets [0,1,2,3] and counts [2,3,4], where the correct lower-bound sum is 11 but the current code would compute 50 (inflated). 

1. Run the `histogram_sum_poc` test from the artifact zip.
2. Use bucket bounds and counts where per-bucket lower-bound sum is known.
3. Observe that the current calculation uses the running total and produces a larger incorrect sum.

### Expected behavior

Histogram sum calculation should multiply each bucket lower bound by that bucket count, not by the cumulative count.

dashpole:
The runtime histogram sum calculation used cumulative counts instead of per-bucket counts, leading to inflated sums.

Fixes #9055


CHANGELOG.md
```diff
@@ -8,6 +8,10 @@ This project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0.htm
 
 ## [Unreleased]
 
+### Fixed
+
+- Fix incorrect (overestimated) sum calculation for runtime histograms in `go.opentelemetry.io/contrib/instrumentation/runtime`. (#9055)
```
```suggestion
- Fix incorrect (overestimated) sum calculation for runtime histograms in `go.opentelemetry.io/contrib/instrumentation/runtime`. (#9063)
```

instrumentation/runtime/producer.go
```diff
@@ -106,7 +106,7 @@ func convertRuntimeHistogram(runtimeHist *metrics.Float64Histogram, ts time.Time
 		// This computed sum is an underestimate, since it assumes each
 		// observation happens at the bucket's lower bound.
 		if i > 0 && count != 0 {
```
pellared: I think this can be

```suggestion
		if i > 0 && c != 0 {
```

pellared (approved):

ps-mir (approved):
Reviewed carefully .. in the duplicate.

CHANGELOG.md
```diff
@@ -8,6 +8,10 @@ This project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0.htm
 
 ## [Unreleased]
 
+### Fixed
```
MrAlias: This is a duplicate `### Fixed` section. Please move this entry to the existing section.

MrAlias (approved):

PR #9064, opened 2026-05-29
Add AGENTS.md for opentelemetry-go-contrib

dashpole:
This is the same as the one in core, with https://github.com/open-telemetry/opentelemetry-go/pull/8393 added


pellared (approved):

pellared:
Can you also add https://github.com/open-telemetry/opentelemetry-go/blob/main/CLAUDE.md ?

PR #9068, opened 2026-05-30
fix(otelzap): avoid panic on malformed error fields

### LINKED ISSUE #9043: otelzap error-field handling can panic on nil error
### Description

`bridges/otelzap` can panic while converting zap error fields whose key is `error` but whose `Interface` value is nil or not an `error`. Current-tree validation showed `bridges/otelzap/core.go` still performs an unguarded `field.Interface.(error)` assertion for `zapcore.ErrorType` fields named `error`.

### Steps To Reproduce

[04-317e4de46f508191b04cd9fff3217ab5-otelzap-error-field-handling-can-panic-on-nil-error.zip](https://github.com/user-attachments/files/28384481/04-317e4de46f508191b04cd9fff3217ab5-otelzap-error-field-handling-can-panic-on-nil-error.zip)

1. Use `go.opentelemetry.io/contrib/bridges/otelzap`.
2. Construct or log a zap field equivalent to `zapcore.Field{Key: "error", Type: zapcore.ErrorType, Interface: nil}`.
3. Pass the field through the otelzap conversion path, or run the PoC test from the artifact zip.
4. Observe a panic at the unguarded `field.Interface.(error)` assertion.

### Expected behavior

otelzap should handle nil or unexpected `ErrorType` field values without panicking, either by ignoring the malformed error field, preserving it as a normal attribute, or recording a safe error value.

omer-cengel:
Fixes https://github.com/open-telemetry/opentelemetry-go-contrib/issues/9043

## Summary

Fixes a panic in `bridges/otelzap` when handling malformed zap error fields.

`convertField` previously performed an unguarded type assertion for fields with:

* `Type == zapcore.ErrorType`
* `Key == "error"`

This could panic if `field.Interface` was `nil` or not an `error`.

The conversion now safely checks the type assertion and ignores malformed error fields while preserving the existing behavior for valid `zap.Error(err)` fields.

## Changes

* Guarded the `field.Interface.(error)` type assertion in `convertField`
* Added tests covering:

  * `nil` error field values
  * invalid non-`error` field values

## Testing

Verified that:

* the new tests fail on the previous implementation
* all tests pass with this change
* valid `zap.Error(err)` behavior remains unchanged

### DIFF
--- a/bridges/otelzap/core_test.go
+++ b/bridges/otelzap/core_test.go
@@ -642,3 +642,41 @@ func TestConvertLevel(t *testing.T) {
 		}
 	}
 }
+
+func TestCoreMalformedErrorFieldDoesNotPanic(t *testing.T) {
+	tests := []struct {
+		name       string
+		fieldValue any
+	}{
+		{
+			name:       "NilError",
+			fieldValue: nil,
+		},
+		{
+			name:       "InvalidErrorType",
+			fieldValue: "not-an-error",
+		},
+	}
+
+	for _, tt := range tests {
+		t.Run(tt.name, func(t *testing.T) {
+			p := newCaptureProvider()
+			logger := zap.New(NewCore(loggerName, WithLoggerProvider(p)))
+
+			ce := logger.Check(zap.ErrorLevel, testMessage)
+			require.NotNil(t, ce)
+
+			require.NotPanics(t, func() {
+				ce.Write(zapcore.Field{
+					Key:       "error",
+					Type:      zapcore.ErrorType,
+					Interface: tt.fieldValue,
+				})
+			})
+
+			r := p.logger.lastRecord(t)
+			require.NoError(t, r.err)
+			require.Empty(t, r.attrs)
+		})
+	}
+}
--- a/bridges/otelzap/core.go
+++ b/bridges/otelzap/core.go
@@ -275,7 +275,9 @@ func convertField(fields []zapcore.Field) (context.Context, []log.KeyValue, erro
 			continue
 		}
 		if field.Type == zapcore.ErrorType && field.Key == "error" {
-			errField = field.Interface.(error)
+			if err, ok := field.Interface.(error); ok && err != nil {
+				errField = err
+			}
 			continue
 		}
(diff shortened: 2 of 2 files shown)


dashpole (approved):

dashpole:
Doesn't require a changelog IMO

pellared (approved):

bridges/otelzap/core_test.go
```diff
@@ -642,3 +642,41 @@ func TestConvertLevel(t *testing.T) {
 		}
 	}
 }
+
+func TestCoreMalformedErrorFieldDoesNotPanic(t *testing.T) {
```
pellared, line 646:
```suggestion
func TestCoreMalformedErrorField(t *testing.T) {
```

bridges/otelzap/core_test.go
```diff
@@ -642,3 +642,41 @@ func TestConvertLevel(t *testing.T) {
...
+			ce := logger.Check(zap.ErrorLevel, testMessage)
+			require.NotNil(t, ce)
+
+			require.NotPanics(t, func() {
+				ce.Write(zapcore.Field{
+					Key:       "error",
+					Type:      zapcore.ErrorType,
+					Interface: tt.fieldValue,
+				})
+			})
```
pellared, line 675: I do not think we need `require.NotPanics` given we also have other assertions

```suggestion
			ce.Write(zapcore.Field{
				Key:       "error",
				Type:      zapcore.ErrorType,
				Interface: tt.fieldValue,
			})
		})
```

MrAlias (approved):

bridges/otelzap/core_test.go
MrAlias: Won't this test panic if run on `main`, hiding the other assertions?

pellared: > Won't this test panic if run on main, hiding the other assertions?

It would.  If you do not like this than please resolve my comment so that the PR does not apply my suggestion.

khushijain21 (approved):

PR #9074, opened 2026-06-01
Add support for azure.vm resource detector in otelconf/x

iblancasa:


CHANGELOG.md
```diff
@@ -10,6 +10,7 @@ This project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0.htm
 
 ### Added
 
+- Add support for the `azure.vm` resource detector in `go.opentelemetry.io/contrib/otelconf/x`. (#8914)
```
dashpole: i don't know if we have rules for this, but dots in the resource detector name seems... odd? Are these always one word?  (e.g. azurecontainerapps)?

iblancasa: I agree. But used this https://github.com/open-telemetry/opentelemetry-go-contrib/blob/e45db6388afccda386073d11479477728b2bf7df/detectors/autodetect/autodetect.go#L51

I can change but tried to attach to what is already there.

dashpole (approved):

MrAlias (approved):

otelconf/x/resource_test.go
```diff
@@ -148,6 +148,7 @@ func TestResourceOptsWithDetectors(t *testing.T) {
 			name: "all-detectors",
 			detectors: []ExperimentalResourceDetector{
 				{AWSEKS: ExperimentalAWSEKSResourceDetector{}},
+				{AzureVM: ExperimentalAzureVMResourceDetector{}},
```
MrAlias: Non-blocking: adding the Azure VM detector to this input performs a real request to `169.254.169.254`, but the test only asserts attributes produced by the host, OS, process, and service detectors. Outside Azure, the detector returns an empty resource, so this test still passes if the `d.AzureVM` branch in `resourceOpts` is removed. It also leaves the test dependent on link-local network behavior.

Could we keep this hermetic with a focused test asserting that `resourceOpts` returns one option for an Azure-only input, without applying that option through `newResource`? This should not block this PR; we can follow up if it is not addressed before merge.

MrAlias (approved):
Looks good 👍

otelconf/x/resource_test.go
iblancasa: I see. Makes sense. I will send a new change.

MrAlias:
@iblancasa this look ready to merge after the conflicts are resolved. Thanks for the contribution 😄

PR #9080, opened 2026-06-02
fix(autoprop): return no-op propagator instead of nil for empty input

### LINKED ISSUE #9057: autoprop: TextMapPropagator returns nil for empty input
### Description

`TextMapPropagator(names ...string)` documents that empty input returns a no-operation propagator. Current code returns nil when no propagators are selected, so callers that trust the documented contract can panic when calling `Inject`, `Extract`, or `Fields`.

`TextMapPropagator` explicitly states that if no names are provided, a no-op propagator is returned (registry.go:120-127). The implementation instead returns `nil` when `len(props)==0` (registry.go:160-162). A caller that trusts the documented behavior and calls `Inject`/`Extract` on the return value will hit a nil interface method call panic.

### Steps To Reproduce

[24-5aa2ad1652108191a90f69e0ab65f95f-textmappropagator-returns-nil-for-empty-input.zip](https://github.com/user-attachments/files/28385840/24-5aa2ad1652108191a90f69e0ab65f95f-textmappropagator-returns-nil-for-empty-input.zip)

1. Call `autoprop.TextMapPropagator()` with no names.
2. Confirm it returns `nil` instead of a no-op propagator.
3. Call `Inject`, `Extract`, or `Fields` on the returned value.
4. Observe a nil interface method call panic.

### Expected behavior

Empty input should return a non-nil no-op propagator, equivalent to `propagation.NewCompositeTextMapPropagator()`, so method calls are safe no-ops.

toller892:
## Problem

`TextMapPropagator(names ...string)` [documents](https://github.com/open-telemetry/opentelemetry-go-contrib/blob/main/propagators/autoprop/registry.go#L116-L117) that empty input returns a no-operation propagator:

> If "none" is included in the arguments, or no names are provided, the returned TextMapPropagator will be a no-operation implementation.

The implementation returns `nil` when `len(props)==0` ([registry.go:152](https://github.com/open-telemetry/opentelemetry-go-contrib/blob/main/propagators/autoprop/registry.go#L152)), so callers that trust the documented contract panic when calling `Inject`, `Extract`, or `Fields` on the return value.

## Fix

Change `return nil, err` to `return propagation.NewCompositeTextMapPropagator(), err` in the `case 0` branch, matching the behavior already used for the `"none"` name on line 133.

Also add `TestTextMapPropagatorEmptyNoError` to verify the fix.

Fixes #9057

### DIFF
--- a/propagators/autoprop/propagator_test.go
+++ b/propagators/autoprop/propagator_test.go
@@ -50,6 +50,19 @@ func TestNewTextMapPropagatorSingleNoOverhead(t *testing.T) {
 	assert.Same(t, p, NewTextMapPropagator(p))
 }
 
+func TestTextMapPropagatorEmptyNoError(t *testing.T) {
+	// Empty input should return a no-op propagator, not nil.
+	// See https://github.com/open-telemetry/opentelemetry-go-contrib/issues/9057
+	p, err := TextMapPropagator()
+	assert.NoError(t, err)
+	assert.NotNil(t, p, "empty TextMapPropagator should return a non-nil no-op propagator")
+
+	// Verify it is safe to call methods on the returned propagator.
+	ctx := p.Inject(context.Background(), nil)
+	assert.NotNil(t, ctx)
+	assert.Empty(t, p.Fields())
+}
+
 func TestNewTextMapPropagatorMultiEnvNone(t *testing.T) {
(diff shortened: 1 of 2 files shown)


MrAlias:
@toller892 please sign the CLA so these changes can be reviewed.

PR #9081, opened 2026-06-02
fix(otelmux): copy MultipartForm back for temp file cleanup

### LINKED ISSUE #9070: otelmux: MultipartForm temp files leak when middleware substitutes request context
### Component

instrumentation: otelmux

otelmux.Middleware passes r.WithContext(ctx) to the downstream handler instead of the original request. If the handler calls ParseMultipartForm, the MultipartForm is stored on the context copy, not the original request. net/http's finishRequest only cleans up MultipartForm on the exact request it handed to ServeHTTP, so the temp files are never reaped and accumulate on disk.

### Expected behavior

After ServeHTTP returns, the original request's MultipartForm should be non-nil so that net/http can clean up any temporary files created during multipart parsing.

### Steps to Reproduce

1. Set up a gorilla/mux router with otelmux.Middleware
2. Add a handler that calls r.ParseMultipartForm()
3. Send a multipart POST request
4. After ServeHTTP returns, check r.MultipartForm — it is nil
5. Temp files from the multipart upload are never cleaned up

### Operating System

windows

### Device Architecture

x86_64

### Go Version

1.26.3

### Component Version

latest

toller892:
## Problem

`otelmux.Middleware` passes `r.WithContext(ctx)` to the downstream handler instead of the original request. If the handler calls `ParseMultipartForm`, the `MultipartForm` is stored on the context copy, not the original request. `net/http`'s `finishRequest` only cleans up `MultipartForm` on the exact request it handed to `ServeHTTP`, so temp files are never reaped and accumulate on disk.

Fixes #9070

## Fix

After `ServeHTTP` returns, copy `MultipartForm` from the context-derived request back to the original request so `net/http` can clean up the temp files.

**Changed file**: `instrumentation/github.com/gorilla/mux/otelmux/mux.go` (6-line change)

## Testing

Added `TestMultipartFormCleanup` — sends a multipart POST through the middleware and verifies temp files don't leak.

`go test ./...` passes (all existing tests + new test).

### DIFF
--- a/instrumentation/github.com/gorilla/mux/otelmux/multipartform_test.go
+++ b/instrumentation/github.com/gorilla/mux/otelmux/multipartform_test.go
@@ -1,0 +1,171 @@
+// Copyright The OpenTelemetry Authors
+// SPDX-License-Identifier: Apache-2.0
+
+package otelmux_test
+
+import (
+	"bytes"
+	"mime/multipart"
+	"net/http"
+	"net/http/httptest"
+	"os"
+	"strings"
+	"testing"
+
+	"github.com/gorilla/mux"
+	"github.com/stretchr/testify/assert"
+	"github.com/stretchr/testify/require"
+	sdktrace "go.opentelemetry.io/otel/sdk/trace"
+	"go.opentelemetry.io/otel/sdk/trace/tracetest"
+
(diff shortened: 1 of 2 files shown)


MrAlias:
@toller892 blocked on signing the CLA. Please sign the CLA and we can reopen.

PR #9082, opened 2026-06-02
fix(examples/namedtracer): add baggagecopy span processor

Icarus9913:
The example creates baggage with keys like `ex.com/foo` and `ex.com/bar`, but those values are stored in context and are not automatically exported as span attributes. Adding the baggage copy span processor makes the example behavior easier to understand and aligns the output with user expectations.


Icarus9913:
cc @pellared

examples/namedtracer/main.go
```diff
@@ -38,6 +39,7 @@ func initTracer() error {
 	tp = sdktrace.NewTracerProvider(
 		sdktrace.WithSampler(sdktrace.AlwaysSample()),
 		sdktrace.WithSpanProcessor(bsp),
+		sdktrace.WithSpanProcessor(baggagecopy.NewSpanProcessor(baggagecopy.AllowAllMembers)),
```
ps-mir: This is an example of baggage copying in single in-process shared context. Would be even better to show with a propagation as baggage are useful on process boundaries.              
                                                                                                                                                                                                          
Service A (ex.com/foo in baggage) -> HTTP call (bagagge W3C headers) -> Service B (`Parent.Remote: true`) recovers and adds it into span.

Icarus9913:
Looking into the CI failure.

Icarus9913:
cc @dashpole

Icarus9913:
Hey @dashpole, could you please help re-run the CI again, the lint stuff is a little bit annoying. I've run the golang-lint in my local environment, now it should be fine.
Thank you very much :slightly_smiling_face:

Icarus9913:
@dashpole Thanks for you help, now the CI pass

PR #9088, opened 2026-06-03
Add instrumentation for labstack echo v5

### LINKED ISSUE #9003: Upgrade otelecho from using echo/v4 to using echo/v5
Echo v5 has been released with several breaking changes, some requiring changes here for compatability. The release itself can be found [here](https://github.com/labstack/echo/releases/tag/v5.0.0) with a link to a detailed document of changes.

From what I can gather from looking at otelecho the necessary changes should be the following:
```
find . -type f -name "*.go" -exec sed -i 's/ echo.Context/ *echo.Context/g' {} +
find . -type f -name "*.go" -exec sed -i 's/echo\/v4/echo\/v5/g' {} +
```

shonigbaum:
Fixes #9003

## Changes

Adds OpenTelemetry instrumentation for [Echo v5](https://github.com/labstack/echo) at `instrumentation/github.com/labstack/echo/v5/otelecho`.

Echo v5 introduced breaking changes from v4:
- `Context` changed from an interface to a concrete struct (`*echo.Context`)
- `c.Response()` now returns `http.ResponseWriter` instead of `*echo.Response`
- `c.Error()` was removed; error handling is done via `c.Echo().HTTPErrorHandler()`
- `SetParamNames`/`SetParamValues` replaced by `SetPathValues(echo.PathValues{...})`

This adds a separate instrumentation package for v5 alongside the existing v4 package.

## Testing

- Unit tests passing
- Example with docker-compose verified

## Questions

- We discovered that there's a new library directly from Labstack regarding OpenTelemetry: https://github.com/labstack/echo-opentelemetry - Will this be the new way to go?
- Is the path okay like this? Or should I change it to `instrumentation/github.com/labstack/echo-v5/otelecho` to reduce one level.


pellared:
@flc1125, PTAL as code owner

pellared (changes requested):
Why not using https://github.com/labstack/echo-opentelemetry? Can you comment in the issue? If there are pboelsm then probably it would be better addressing them in https://github.com/labstack/echo-opentelemetry

shonigbaum:
> Why not using https://github.com/labstack/echo-opentelemetry? Can you comment in the issue? If there are pboelsm then probably it would be better addressing them in https://github.com/labstack/echo-opentelemetry

Hey, was this question for me? As I wrote in the description under "Questions", I also discovered this library and asked, if it's the new way to go. Seems so?

I will close this PR then.

PR #9097, opened 2026-06-05
fix(otelconf): surface Prometheus metrics server errors in v0.2.0

### LINKED ISSUE #9056: otelconf: Prometheus metrics server errors are silently ignored
### Description

Prometheus metrics server errors can be silently ignored.

The goroutine now calls otel.Handle only when errors.Is(err, http.ErrServerClosed) is true (config/metric.go:264-266). Compared to the previous revision: the condition used to be `err != http.ErrServerClosed` (git show 9f2491a1c0c6aac2774d1184c2924dc705d3a43d^:config/metric.go:264-266), so the check is inverted. http.ErrServerClosed represents normal shutdown; thus the new logic reports clean shutdowns as "unexpected" and suppresses real unexpected Serve errors (bind/accept/runtime), matching the reported regression. (4) 

### Steps To Reproduce

[23-65a6d958fbf4819196da4f467c2152d4-prometheus-metrics-server-errors-are-silently-ignored.zip](https://github.com/user-attachments/files/28385793/23-65a6d958fbf4819196da4f467c2152d4-prometheus-metrics-server-errors-are-silently-ignored.zip)

1. Run the Prometheus error-handler PoC from the artifact zip.
2. Force the metrics server startup or serve path to return an error.
3. Observe that the error is not surfaced to the caller/operator.

### Expected behavior

Metrics server errors should be returned or logged clearly so operators can detect failed metrics exposure.

arpitjain099:
Fixes #9056

### Problem

The Prometheus metrics HTTP server goroutine in `otelconf/v0.2.0/metric.go` has an inverted error check. It reports an error only when the returned error *is* `http.ErrServerClosed`:

```go
go func() {
    if err := server.Serve(lis); err != nil && errors.Is(err, http.ErrServerClosed) {
        otel.Handle(fmt.Errorf("the Prometheus HTTP server exited unexpectedly: %w", err))
    }
}()
```

`http.ErrServerClosed` is the sentinel that `Serve` returns on a clean `Shutdown`/`Close`. With this condition:

- a normal shutdown is reported as "the Prometheus HTTP server exited unexpectedly", and
- a genuine `Serve` error (bind/accept/runtime) does not match the sentinel, so it is silently swallowed.

That is the exact regression described in the issue: real metrics-server failures are invisible to operators.

The git-blame in the issue points at the introducing change on the pre-rename path (`config/metric.go`, before the directory was renamed to `otelconf/v0.2.0/`), where the original correct form `err != http.ErrServerClosed` was inverted. The two sibling variants in this repo, `otelconf/v0.3.0/metric.go` and `otelconf/x/metric.go`, already use the correct form, which confirms `v0.2.0` is the one left inverted.

### Fix

Flip the condition so only non-clean-shutdown errors are reported, leaving the error-channel/logging behavior otherwise unchanged:

```go
if err := server.Serve(lis); err != nil && !errors.Is(err, http.ErrServerClosed) {
    otel.Handle(fmt.Errorf("the Prometheus HTTP server exited unexpectedly: %w", err))
}
```

Before / after, same scenarios:

- clean shutdown (`http.ErrServerClosed`): before -> reported as unexpected error; after -> not reported.
- real serve error (e.g. listener closed under the server): before -> swallowed; after -> surfaced via `otel.Handle`.

### Tests

Added `TestPrometheusServerErrorHandling` to `otelconf/v0.2.0/metric_test.go`, using `otel.SetErrorHandler` to capture what the goroutine reports:

- `clean shutdown is not reported as an error`: builds the reader through `prometheusReader`, calls `Shutdown`, and asserts no error is handled. This subtest fails against the unfixed source and passes after the fix.
- `real serve error is reported`: reproduces the same server goroutine, closes the listener out from under `Serve` to produce a real (non-`ErrServerClosed`) error, and asserts it is surfaced.

Verified the test is a real regression guard by reverting only the source change: the clean-shutdown subtest fails, and passes again once the fix is restored. `go test ./...` is green for the whole `otelconf` module and `go build ./...` succeeds.


dashpole (approved):

arpitjain099:
Fixed the lint failure: the `noctx` linter flagged `net.Listen` in the new test, so I switched it to `(*net.ListenConfig).Listen` with a context. `go test` passes locally and noctx is clean. (The codecov/patch gate is coverage-only.)

iblancasa (approved):

arpitjain099:
Rebased onto main. The conflicts were both additive: a CHANGELOG collision in the Unreleased Fixed list, and the `net` import in `metric_test.go`. Nothing about the fix itself changed, so the two approvals should still apply.

`go build ./...` and `go test ./...` are clean in `otelconf/v0.2.0`.

CHANGELOG.md
```diff
@@ -17,6 +17,7 @@ This project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0.htm
 - Report `ot-baggage-*` extraction errors from `go.opentelemetry.io/contrib/propagators/ot` to `otel.Handle` instead of silently discarding them, while still attaching the successfully parsed baggage members to the context. (#9395)
 - Set `error.type` on the `rpc.client.call.duration` and `rpc.server.call.duration` metrics in `go.opentelemetry.io/contrib/instrumentation/google.golang.org/grpc/otelgrpc` when the RPC fails with a non-OK status, per the RPC semantic conventions. (#9429)
 - Reject OTLP exporter headers with an empty `name` in `go.opentelemetry.io/contrib/otelconf`, `go.opentelemetry.io/contrib/otelconf/x`, and `go.opentelemetry.io/contrib/otelconf/v0.3.0`, instead of forwarding invalid header names to OTLP exporters. (#9102)
+- Report genuine Prometheus metrics HTTP server errors instead of swallowing them in `go.opentelemetry.io/contrib/otelconf/v0.2.0`. The error check was inverted, so a clean shutdown (`http.ErrServerClosed`) was reported as unexpected while real `Serve` errors were ignored. (#9056)
```
```suggestion
- Report genuine Prometheus metrics HTTP server errors instead of swallowing them in `go.opentelemetry.io/contrib/otelconf/v0.2.0`. The error check was inverted, so a clean shutdown (`http.ErrServerClosed`) was reported as unexpected while real `Serve` errors were ignored. (#9097)
```

arpitjain099: Thanks, and thanks for merging main in as well. You are right on the reference: the three entries above it cite PR numbers (#9395, #9429, #9102) while #9056 is the issue, so #9097 is the consistent one.

I pulled your commit down and re-ran the module after the main merge, `go build ./...` and `go test ./v0.2.0/...` are both clean. CI is still running on your commit at the moment.

Nothing outstanding from my side.

otelconf/v0.2.0/metric_test.go
```diff
@@ -1414,3 +1416,86 @@ func TestPrometheusReaderHostParsing(t *testing.T) {
...
+		}))
+		t.Cleanup(func() { otel.SetErrorHandler(prev) })
+
+		var lc net.ListenConfig
+		lis, err := lc.Listen(t.Context(), "tcp", "127.0.0.1:0")
+		require.NoError(t, err)
+
+		server := http.Server{Handler: http.NewServeMux()}
+		go func() {
+			if err := server.Serve(lis); err != nil && !errors.Is(err, http.ErrServerClosed) {
```
pellared: This subtest recreates the server goroutine and copies the production predicate, so it does not exercise the positive-error branch in `prometheusReader`. This is also why the optional `codecov/patch` check remains red.

The clean-shutdown subtest catches the exact historical inversion. To cover both invariants, consider extracting the error classification/reporting into a small production helper and testing it with `http.ErrServerClosed` and a genuine error.

otelconf/v0.2.0/metric_test.go
```diff
@@ -1414,3 +1416,86 @@ func TestPrometheusReaderHostParsing(t *testing.T) {
...
+	t.Run("clean shutdown is not reported as an error", func(t *testing.T) {
+		handled := make(chan error, 1)
+		prev := otel.GetErrorHandler()
+		otel.SetErrorHandler(otel.ErrorHandlerFunc(func(err error) {
+			select {
+			case handled <- err:
+			default:
+			}
+		}))
+		t.Cleanup(func() { otel.SetErrorHandler(prev) })
```
pellared: On the first `otel.SetErrorHandler` call, `prev` is OpenTelemetry’s initial delegating handler. Restoring that object does not restore its original behavior because it remains permanently delegated to this temporary channel handler. Later `otel.Handle` calls in the test process can therefore be consumed or dropped by the completed subtest.

The current remaining tests do not appear affected. An injected reporting callback or shared helper would avoid mutating the process-global handler.

pellared (commented):
The production fix looks correct. It restores the intended `http.ErrServerClosed` handling, matches the `v0.3.0` and `x` implementations, and directly addresses #9056.

I left two comments about test coverage and isolation.

otelconf/v0.2.0/metric_test.go
arpitjain099: You are right, and this is the more important of the two. That subtest asserted on a copy of the predicate and never entered `prometheusReader`, so it would have passed just as happily with the production code still inverted. That is the whole point of the test, and it was not doing it.

Done as you suggested, in b2aac113. The classification is now a production helper:

```go
func handleServeErr(err error) {
	if err == nil || errors.Is(err, http.ErrServerClosed) {
		return
	}

	reportErr(fmt.Errorf("the Prometheus HTTP server exited unexpectedly: %w", err))
}
```

and the goroutine is `go func() { handleServeErr(server.Serve(lis)) }()`. The table covers nil, the sentinel, a wrapped sentinel, a genuine error and a wrapped genuine error.

I checked it actually bites rather than assuming it does. Restoring the pre-fix inverted predicate and re-running:

```
--- FAIL: TestHandleServeErr/clean_shutdown_sentinel_is_not_reported
--- FAIL: TestHandleServeErr/wrapped_clean_shutdown_sentinel_is_not_reported
--- FAIL: TestHandleServeErr/a_genuine_serve_error_is_reported
--- FAIL: TestHandleServeErr/a_wrapped_genuine_serve_error_is_reported
--- FAIL: TestPrometheusServerErrorHandling
```

I kept the end-to-end clean-shutdown test as well, since it is the one that covers the goroutine wiring rather than just the predicate.

One consistency note: `v0.3.0/metric.go:375` and `x/metric.go:446` still have the inline form. I left them alone to keep this PR to the changelog entry, but happy to follow up with the same extraction there if you want them to match.

otelconf/v0.2.0/metric_test.go
arpitjain099: Good catch, and thanks for spelling out the mechanism, I had not thought about the delegating handler not being restorable.

Fixed in the same commit by going the injected-callback route. Production reports through a package variable:

```go
var reportErr = otel.Handle
```

so the tests swap `reportErr` and never touch `otel.SetErrorHandler` at all. Both `SetErrorHandler` calls are gone from the file, and with them the `otel` import in the test. The swap lives in one `swapReportErr(t)` helper with a `t.Cleanup` restore, so a `prev` that is genuinely restorable.

These tests do not call `t.Parallel`, so the shared variable is safe as written. `go test -race ./v0.2.0/...` is clean.

PR #9099, opened 2026-06-05
Prometheus Bridge: Support untyped metrics

### LINKED ISSUE #9098: Prometheus Bridge: Support UNTYPED metrics
While working on https://github.com/prometheus/prometheus-opentelemetry-collector/pull/16, which adapts the [mysqld_exporter](https://github.com/prometheus/mysqld_exporter) into an OpenTelemetry Collector receiver, I realized that the exporter has a bunch of UNTYPED metrics.

We're using this bridge to transform Prometheus metrics into OTLP, and realized the error comes from
https://github.com/open-telemetry/opentelemetry-go-contrib/blob/3282a0ef867eb72df691104b5583807afefa0b14/bridges/prometheus/producer.go#L103-L106

The current spec says [Untyped metrics become OTLP Gauges](https://github.com/open-telemetry/opentelemetry-specification/blob/main/specification/compatibility/prometheus_and_openmetrics.md#unknown-typed). So I think we can align with that :)

ArthurSens:
Fixes #9098


dashpole (approved):

ArthurSens:
Hello hello, anything needed from my side here? :)

dashpole:
(just a second review). @open-telemetry/go-approvers

ArthurSens:
Friendly bump! Maybe @pellared could take a look?

ArthurSens:
Hello hello, anything I could do to get this merged? 👼

bridges/prometheus/producer.go
```diff
@@ -138,6 +140,24 @@ func convertGauge(metrics []*dto.Metric, now time.Time) metricdata.Gauge[float64
...
+	otelGauge := metricdata.Gauge[float64]{
+		DataPoints: make([]metricdata.DataPoint[float64], len(metrics)),
+	}
+	for i, m := range metrics {
+		dp := metricdata.DataPoint[float64]{
+			Attributes: convertLabels(m.GetLabel()),
+			Time:       now,
+			Value:      m.GetUntyped().GetValue(),
+		}
+		if m.GetTimestampMs() != 0 {
```
pellared, line 153: `TimestampMs` is optional, so this conflates an absent timestamp with an explicitly supplied Unix-epoch timestamp (`0`). In the latter case, the timestamp is incorrectly replaced by `now`. The [OpenTelemetry compatibility specification](https://github.com/open-telemetry/opentelemetry-specification/blob/main/specification/compatibility/prometheus_and_openmetrics.md#timestamps) requires a present Prometheus sample timestamp to be converted to the OTLP data point timestamp. Could we check `m.TimestampMs != nil` instead and add an epoch regression test?

pellared (commented):
One comment

bridges/prometheus/producer.go
pellared: I see that this is not specific to the newly added untyped converter: the existing gauge, counter, exponential histogram, summary, and histogram converters use the same presence check. I opened #9539 to track a consistent fix and regression coverage across all converters, so it can be addressed separately from this PR.

pellared (approved):

PR #9102, opened 2026-06-06
otelconf: reject OTLP exporter headers with empty name

### LINKED ISSUE #9053: otelconf: Headers config now accepts empty header names
### Description

Header configuration can accept an empty header name and pass it to OTLP exporters. The historical path `config/v0.3.0/config.go` is missing in this checkout, so current-tree confirmation was not possible here.

`createHeadersConfig` inserts headers with non-nil values without validating `kv.Name` (config/v0.3.0/config.go:183-201). `NameStringValuePair` defines `Name` as a plain string (defaults to "" if omitted) and `Value` as *string, so a YAML header entry missing `name` can yield `Name==""` with a non-nil value (config/v0.3.0/generated_config.go:248-253). The resulting map is passed directly into OTLP exporters via `otlptracegrpc.WithHeaders`/`otlptracehttp.WithHeaders` (config/v0.3.0/trace.go:128-180), so empty header keys propagate to exporter clients. A standard-library demo run in this container (`go run /workspace/poc_http_invalid_header.go`) prints `roundtrip error: net/http: invalid header field name ""`, confirming Go's HTTP stack rejects empty header names; thus a misconfigured header entry would break export.

### Steps To Reproduce

[20-1b8acc8a26ac8191abaf141f478c624d-headers-config-now-accepts-empty-header-names.zip](https://github.com/user-attachments/files/28385482/20-1b8acc8a26ac8191abaf141f478c624d-headers-config-now-accepts-empty-header-names.zip)

1. Create a header entry with an omitted or empty `name` and a non-empty value.
2. Load the config and pass the resulting headers to an OTLP HTTP or gRPC exporter.
3. Observe exporter/client errors such as invalid header field name.

### Expected behavior

Header names should be validated and empty names should be rejected before exporter construction.

viacheslav-danilin:
Fixes #9053.

`createHeadersConfig` in `otelconf` previously inserted header entries with an empty `Name` into the map passed to OTLP exporters via `otlptracegrpc.WithHeaders` / `otlptracehttp.WithHeaders` (and the equivalent metric/log exporters). Go's HTTP stack rejects empty header field names, so a misconfigured header entry would break export at runtime.

This change validates header entries up-front: if `Value != nil` and `Name == ""`, `createHeadersConfig` returns an error before any exporter is constructed. Existing merge/precedence semantics between `headers` and `headers_list` are preserved.

The fix is applied to all current header-config paths:

- `go.opentelemetry.io/contrib/otelconf` (stable)
- `go.opentelemetry.io/contrib/otelconf/x` (experimental)
- `go.opentelemetry.io/contrib/otelconf/v0.3.0` (historical)


pellared:
This misses a changelog entry. Please make sure that the entry follows the existing conventions.

pellared:
@codeboten, PTAL

viacheslav-danilin:
> This misses a changelog entry. Please make sure that the entry follows the existing conventions.

Thanks, added the changelog entry.

iblancasa (approved):
I guess a good extra test can be empty name but nil value passes

codeboten (approved):

MrAlias (approved):

Issue #8953, opened 2026-05-08
[detectors/k8s_api] Contribute k8snode detector from resourcedetectionprocessor


paulojmdias:
## Problem Statement

`collector-contrib/processor/resourcedetectionprocessor/internal/k8s_api` implements a resource detector that collects Kubernetes node-level resource attributes by querying the Kubernetes API. There is currently no equivalent in `go-contrib/detectors/`, which means k8s node detection is only available to collector-contrib pipelines and not to any Go SDK user.

## Proposed Solution

Add a new `detectors/k8s_api` package implementing the k8s_api detector as a `resource.Detector`, ported from the collector-contrib implementation, so that `collector-contrib/processor/resourcedetectionprocessor` can delegate to it and any other Go SDK user can adopt it.

## Prior Art

- Collector-contrib implementation to port from: [`processor/resourcedetectionprocessor/internal/k8s_api`](https://github.com/open-telemetry/opentelemetry-collector-contrib/tree/main/processor/resourcedetectionprocessor/internal/k8s_api)
- Tracking issue: open-telemetry/opentelemetry-collector-contrib#48159

## Additional Context

What the collector-contrib detector does:

- Uses the in-cluster Kubernetes client to fetch the Node object by node name (from the `K8S_NODE_NAME` environment variable).
- Emits: `k8s.node.name`, `k8s.node.uid`.

ps-mir:
Working on this one.

ps-mir:
@dashpole @paulojmdias @iblancasa Name and scope for this detector will change to k8s_api after collector-contrib [PR #48597](https://github.com/open-telemetry/opentelemetry-collector-contrib/pull/48597)

paulojmdias:
Done in https://github.com/open-telemetry/opentelemetry-go-contrib/pull/9108

PR #9108, opened 2026-06-08
[detector/k8sapi] Add resource detector

ps-mir:
The scope of this detector changed slightly due to migration of k8snode -> k8s_api in collector-contrib (See `Also View`). 

- k8s.node.name
- k8s.node.uid
- k8s.cluster.uid (additional attribute that needs RBAC permissions to kube-system namespace)

## Detection
K8s API can extract partial attributes depending on env setup
* Node attributes if env variable `K8S_NODE_NAME` is set, and required node RBAC is present
   * k8s.node.name
   * k8s.node.uid
* Cluster attributes from kube-system namespace, if required RBAC is present
  * k8s.cluster.uid

There has been [discussion](https://github.com/open-telemetry/opentelemetry-collector-contrib/pull/48597#discussion_r3333547908) about decoupling detection of node and cluster attribute, but existing detection seems ok as Working k8 client config is also a requirement for `cluster.uid`.

Node attributes can only be extracted from inside **Pod** unless env variable `K8S_NODE_NAME` is explicitly set.

k8s_api has been changed to `k8sapi`

### Fixes
* #8953 

### Related to
* #8940 

### Also View
* open-telemetry/opentelemetry-collector-contrib#48597


detectors/k8sapi/detector.go
```diff
@@ -0,0 +1,143 @@
...
+// The node name is read from the K8S_NODE_NAME environment variable by
+// default. Use [WithNodeEnvVar] to customise the variable name. The variable
+// is typically populated via the Kubernetes downward API:
+//
+//	env:
+//	  - name: K8S_NODE_NAME
+//	    valueFrom:
+//	      fieldRef:
+//	        fieldPath: spec.nodeName
+func NewResourceDetector(opts ...Option) resource.Detector {
```
dashpole: I probably missed this on previous detector PRs, but we should ideally return concrete types (e.g. *ResourceDetector), rather than interfaces.

detectors/k8sapi/doc.go
```diff
@@ -0,0 +1,34 @@
...
+typically via the Kubernetes downward API:
+
+	env:
+	  - name: K8S_NODE_NAME
+	    valueFrom:
+	      fieldRef:
+	        fieldPath: spec.nodeName
+
+The cluster UID is derived from the kube-system namespace UID and requires
+the following RBAC:
```
dashpole, line 31: Please document the RBAC required for getting the node UID as well.

detectors/k8sapi/detector.go
```diff
@@ -0,0 +1,143 @@
...
+
+	ns, err := client.CoreV1().Namespaces().Get(ctx, "kube-system", metav1.GetOptions{})
+	if err != nil {
+		res := resource.NewWithAttributes(semconv.SchemaURL, attrs...)
+		return res, fmt.Errorf("%w: %w", resource.ErrPartialResource, err)
+	} else if uid := string(ns.UID); uid != "" {
+		attrs = append(attrs, semconv.K8SClusterUID(uid))
+	}
+
+	if rd.cfg.filter != nil {
```
dashpole, line 115: Ideally we wouldn't attempt to get resource attributes that are not part of the filter, right?

E.g. if I filter out k8s.node.uid, I would not expect the detector to fail if I don't have permissions to get that one.

detectors/k8sapi/detector.go
```diff
@@ -0,0 +1,143 @@
...
+// Detect returns a [*resource.Resource] describing the Kubernetes node. If the
+// node-name environment variable is not set or the process is not running
+// inside a cluster, an empty resource and no error are returned.
+func (rd *resourceDetector) Detect(ctx context.Context) (*resource.Resource, error) {
+	nodeName := os.Getenv(rd.cfg.nodeEnvVar)
+	if nodeName == "" {
+		return resource.Empty(), nil
+	}
+
+	client := rd.cfg.kubeClient
```
dashpole, line 77: Would it be better to create the client in NewDetector and return the error there? How do other detectors handle this? I do like that the way you've done it simplifies the initialization in the example.

detectors/k8sapi/detector.go
```diff
@@ -0,0 +1,143 @@
...
+		var clientErr error
+		client, clientErr = kubernetes.NewForConfig(conf)
+		if clientErr != nil {
+			return nil, fmt.Errorf("k8sapi detector: failed to create Kubernetes client: %w", clientErr)
+		}
+	}
+
+	node, err := client.CoreV1().Nodes().Get(ctx, nodeName, metav1.GetOptions{})
+	if err != nil {
+		return nil, fmt.Errorf("k8sapi detector: failed to get node %q: %w", nodeName, err)
```
dashpole: We should support partial resources here as well (for all cases where we could get some, but not all resource attributes.

dashpole:
cc @paulojmdias @iblancasa

detectors/k8sapi/detector.go
ps-mir: Approach seen so far,
1) NewResourceDetector() cannot return an error as per autodetect's usage:
https://github.com/open-telemetry/opentelemetry-go-contrib/blob/becb15cb091ef0f6c61874c72d18dce0f0656b4f/detectors/autodetect/autodetect.go#L130-L133

This means any client-related errors cannot be processed in NewResourceDetector; you can only do basic wiring.

2) Detect() has to absorb any issues and return EmptyResource in case of any non-recoverable client error. So that autodetect logic can move on to other detectors.

Best error visibility in detectors is log-only, and it's not being used at the moment.

detectors/k8sapi/detector.go
ps-mir: Achievable even when filters are applied primitively at end. It may come with partial error as it will still fire API for all attributes, which may look odd with filtered attributes.

But the requested attribute should appear in the response. Applying proactively would be much smarter.

ps-mir:
@dashpole Among other corrections, using filters early to skip API calls if attribute filtered out. Ex; no failure due to RBAC if clusterUID is not requested in the first place.

CHANGELOG.md
```diff
@@ -24,6 +24,7 @@ This project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0.htm
 - Add support for `cardinality_limits` in `PeriodicMetricReader` in `otelconf`. (#8885)
 - Add `Resource` method to `SDK` in `go.opentelemetry.io/contrib/otelconf/x` to expose the resolved SDK resource from declarative configuration. (#8913)
 - Add `go.opentelemetry.io/contrib/detectors/hetzner`, a new resource detector for Hetzner Cloud servers, ported from `github.com/open-telemetry/opentelemetry-collector-contrib/processor/resourcedetectionprocessor/internal/hetzner`. Detects `cloud.provider`, `cloud.platform`, `cloud.region`, `cloud.availability_zone`, `host.id`, and `host.name`. (#8979)
+- Add `go.opentelemetry.io/contrib/detectors/k8sapi`, a new resource detector that queries the Kubernetes API. Detects `k8s.node.name` and `k8s.node.uid` when `K8S_NODE_NAME` is set via the downward API, and `k8s.cluster.uid` derived from the kube-system namespace UID (works on any Kubernetes distribution). (#8953)
```
iblancasa: You need to move this under `Unreleased`

detectors/k8sapi/detector.go
```diff
@@ -0,0 +1,160 @@
...
+}
+
+// Compile-time interface assertion.
+var _ resource.Detector = (*ResourceDetector)(nil)
+
+// Detect returns a [*resource.Resource] describing the Kubernetes node. If the
+// node-name environment variable is not set or the process is not running
+// inside a cluster, an empty resource and no error are returned.
+func (rd *ResourceDetector) Detect(ctx context.Context) (*resource.Resource, error) {
+	nodeName := os.Getenv(rd.cfg.nodeEnvVar)
```
iblancasa: This early return also prevents `k8s.cluster.uid` detection when `K8S_NODE_NAME` is unset, even though cluster UID only needs the kube-system namespace lookup. If cluster UID is meant to be independently detectable, this should only gate node attribute detection.

detectors/k8sapi/detector.go
```diff
@@ -0,0 +1,160 @@
...
+			return nil, fmt.Errorf("k8sapi detector: failed to create Kubernetes client: %w", clientErr)
+		}
+	}
+
+	var (
+		attrs []attribute.KeyValue
+		errs  []error
+	)
+
+	needsNode := rd.cfg.filter == nil ||
```
iblancasa: This checks the filter with an empty node name before we fetch the real node. That can give the wrong answer.

Example: if my filter says “only keep node name `my-node`”, this code asks it about an empty node name first, gets `false`, and then never fetches the real `my-node`.

Maybe we should only use the filter after we have the real values, or only do this early skip for filters that are definitely key-only. What do you think?

ps-mir: @dashpole This is a valid point. Filters use a predicate defined over a k-v pair; using such a filter with empty values is not reliable.

Using filters for preflight optimization needs some discussion. For now will revert to post-fetch filtering, which is what all detectors follow.

detectors/k8sapi/detector.go
ps-mir: Closing this as filter's implementation is switched back to result projection at end.

detectors/k8sapi/go.mod
```diff
@@ -0,0 +1,56 @@
+module go.opentelemetry.io/contrib/detectors/k8sapi
+
+go 1.26.0
+
+require (
+	go.opentelemetry.io/otel v1.44.0
+	go.opentelemetry.io/otel/sdk v1.43.0
+	k8s.io/apimachinery v0.36.1
+	k8s.io/client-go v0.36.1
+)
```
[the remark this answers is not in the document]
ps-mir: module bumped as part of the semconv v1.42.0 upgrade

detectors/k8sapi/detector.go
```diff
@@ -0,0 +1,143 @@
...
+		semconv.K8SNodeUID(string(node.UID)),
+	}
+
+	ns, err := client.CoreV1().Namespaces().Get(ctx, "kube-system", metav1.GetOptions{})
+	if err != nil {
+		res := resource.NewWithAttributes(semconv.SchemaURL, attrs...)
+		return res, fmt.Errorf("%w: %w", resource.ErrPartialResource, err)
+	} else if uid := string(ns.UID); uid != "" {
+		attrs = append(attrs, semconv.K8SClusterUID(uid))
+	}
```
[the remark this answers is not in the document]
ps-mir: Implementation changed. The current code accumulates errors into errs and always applies the filter at the end before returning the result

detectors/k8sapi/doc.go
```diff
@@ -0,0 +1,34 @@
...
+
+According to semantic conventions for [k8s] attributes,
+each of the following attributes is added if it is available:
+
+  - k8s.node.name
+  - k8s.node.uid
+  - k8s.cluster.uid
+
+Node attributes require the K8S_NODE_NAME environment variable to be set,
+typically via the Kubernetes downward API:
```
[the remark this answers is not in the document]
ps-mir: Behavior corrected. k8s.cluster.uid is now documented separately with its own RBAC requirements.

ps-mir:
Minor changes to improve testability of the code. Result from Integration tests

### 1) No cluster. Standalone run.
```bash
=== RUN   TestDetect_Integration
    detector_integration_test.go:39: KUBECONFIG not set, relying on in-cluster auto-detection
2026/07/01 20:51:29 WARN k8sapi detector: not running in a Kubernetes cluster err="unable to load in-cluster configuration, KUBERNETES_SERVICE_HOST and KUBERNETES_SERVICE_PORT must be defined"
    detector_integration_test.go:47: detector returned empty resource: not running in a Kubernetes cluster
--- PASS: TestDetect_Integration (0.00s)
PASS
ok      go.opentelemetry.io/contrib/detectors/k8sapi    0.014s
```

### 2) Running outside k8 cluster, connected via WithKubeClient
```bash
=== RUN   TestDetect_Integration
    detector_integration_test.go:32: KUBECONFIG set, building an external client from /tmp/kind-kubeconfig
    detector_integration_test.go:50: detector returned 3 attribute(s)
    detector_integration_test.go:57: k8s.cluster.uid: 98c76c5b-1ef4-4e78-924f-e3c7e82cb103
    detector_integration_test.go:65: k8s.node.name: k8sapi-test-control-plane
    detector_integration_test.go:70: k8s.node.uid: 4d25f337-0911-4d6e-901f-b75a0faca613
--- PASS: TestDetect_Integration (0.05s)
```

> It shows node detection as well, since `K8S_NODE_NAME` was also set during testing. Should be absent ideally in production scenario.

### 3) Inside Pod (Container)
```bash
  === RUN   TestDetect_Integration                                                                                                                                                                        
      detector_integration_test.go:39: KUBECONFIG not set, relying on in-cluster auto-detection                                                                                                           
      detector_integration_test.go:50: detector returned 3 attribute(s)                                                                                                                                   
      detector_integration_test.go:57: k8s.cluster.uid: 98c76c5b-1ef4-4e78-924f-e3c7e82cb103                                                                                                              
      detector_integration_test.go:65: k8s.node.name: k8sapi-test-control-plane                                                                                                                           
      detector_integration_test.go:70: k8s.node.uid: 4d25f337-0911-4d6e-901f-b75a0faca613                                                                                                                 
  --- PASS: TestDetect_Integration (0.03s)                                                                                                                                                                
  PASS   
  ```

File: detectors/k8sapi/detector.go [hunk 12]
```diff
@@ -0,0 +1,159 @@
...
+// inside a cluster, an empty resource and no error are returned.
+func (rd *ResourceDetector) Detect(ctx context.Context) (*resource.Resource, error) {
+	client := rd.cfg.kubeClient
+	if client == nil {
+		conf, err := rd.inClusterConfig()
+		if err != nil {
+			if errors.Is(err, rest.ErrNotInCluster) {
+				slog.Warn("k8sapi detector: not running in a Kubernetes cluster", "err", err)
+				return resource.Empty(), nil
+			}
```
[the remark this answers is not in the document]
ps-mir: The error was being suppressed, as its not considered as partial resource error, hence we don't return it to avoid breaking the autodetect sequence.

Logging is acceptable.

detectors/k8sapi/detector.go
```diff
@@ -0,0 +1,159 @@
...
+			attrs = append(attrs, semconv.K8SNodeName(node.Name), semconv.K8SNodeUID(string(node.UID)))
+		}
+	}
+
+	ns, err := client.CoreV1().Namespaces().Get(ctx, "kube-system", metav1.GetOptions{})
+	if err != nil {
+		errs = append(errs, fmt.Errorf("kube-system namespace: %w", err))
+	} else if uid := string(ns.UID); uid != "" {
+		attrs = append(attrs, semconv.K8SClusterUID(uid))
+	}
```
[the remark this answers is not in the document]
ps-mir: Description fixed.

detectors/k8sapi/detector.go
```diff
@@ -0,0 +1,159 @@
...
+	createProvider  func(*rest.Config) (kubernetes.Interface, error)
+	inClusterConfig func() (*rest.Config, error)
+}
+
+// Compile-time interface assertion.
+var _ resource.Detector = (*ResourceDetector)(nil)
+
+// Detect returns a [*resource.Resource] describing the Kubernetes node. If the
+// node-name environment variable is not set or the process is not running
+// inside a cluster, an empty resource and no error are returned.
```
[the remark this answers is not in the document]
ps-mir: Comment updated.

detectors/k8sapi/detector.go
```diff
@@ -0,0 +1,159 @@
...
+	f(c)
+}
+
+// WithNodeEnvVar sets the environment variable name from which the Kubernetes
+// node name is read. Defaults to "K8S_NODE_NAME".
+func WithNodeEnvVar(name string) Option {
+	return optionFunc(func(c *config) { c.nodeEnvVar = name })
+}
+
+// WithKubeClient sets the Kubernetes client used to query the node. If not
```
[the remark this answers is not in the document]
ps-mir: Updated.

detectors/k8sapi/detector.go
```diff
@@ -0,0 +1,159 @@
...
+
+func (f optionFunc) apply(c *config) {
+	f(c)
+}
+
+// WithNodeEnvVar sets the environment variable name from which the Kubernetes
+// node name is read. Defaults to "K8S_NODE_NAME".
+func WithNodeEnvVar(name string) Option {
+	return optionFunc(func(c *config) { c.nodeEnvVar = name })
+}
```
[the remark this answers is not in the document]
ps-mir: Possible but very low odds of overriding env. Skipping.

CHANGELOG.md
```diff
@@ -10,6 +10,7 @@ This project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0.htm
 
 ### Added
 
+- Add `go.opentelemetry.io/contrib/detectors/k8sapi`, a new resource detector that queries the Kubernetes API. Detects `k8s.node.name` and `k8s.node.uid` when `K8S_NODE_NAME` is set via the downward API, and `k8s.cluster.uid` derived from the kube-system namespace UID (works on any Kubernetes distribution). (#8953)
```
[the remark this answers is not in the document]
ps-mir: Changed to PR number.

detectors/k8sapi/go.mod
```diff
@@ -0,0 +1,56 @@
...
+	github.com/mailru/easyjson v0.7.7 // indirect
+	github.com/modern-go/concurrent v0.0.0-20180306012644-bacd9c7ef1dd // indirect
+	github.com/modern-go/reflect2 v1.0.3-0.20250322232337-35a7c28c31ee // indirect
+	github.com/munnerz/goautoneg v0.0.0-20191010083416-a7dc8b61c822 // indirect
+	github.com/pmezard/go-difflib v1.0.1-0.20181226105442-5d4384ee4fb2 // indirect
+	github.com/stretchr/testify v1.11.1
+	github.com/x448/float16 v0.8.4 // indirect
+	go.opentelemetry.io/auto/sdk v1.2.1 // indirect
+	go.opentelemetry.io/otel/metric v1.44.0 // indirect
+	go.opentelemetry.io/otel/trace v1.44.0 // indirect
```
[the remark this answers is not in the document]
ps-mir: Fixed.

detectors/k8sapi/go.mod
```diff
@@ -0,0 +1,56 @@
...
+	github.com/munnerz/goautoneg v0.0.0-20191010083416-a7dc8b61c822 // indirect
+	github.com/pmezard/go-difflib v1.0.1-0.20181226105442-5d4384ee4fb2 // indirect
+	github.com/stretchr/testify v1.11.1
+	github.com/x448/float16 v0.8.4 // indirect
+	go.opentelemetry.io/auto/sdk v1.2.1 // indirect
+	go.opentelemetry.io/otel/metric v1.44.1-0.20260623111333-65f30a1ab958 // indirect
+	go.opentelemetry.io/otel/trace v1.44.1-0.20260623111333-65f30a1ab958 // indirect
+	go.yaml.in/yaml/v2 v2.4.3 // indirect
+	go.yaml.in/yaml/v3 v3.0.4 // indirect
+	golang.org/x/net v0.49.0 // indirect
```
MrAlias: This needs to be upgraded to address the CI failure of the vulnerability checker.

detectors/k8sapi/doc.go
MrAlias, line 1: These docs look good. We need a minimal README file to [comply with policy](https://github.com/open-telemetry/opentelemetry-go/blob/main/CONTRIBUTING.md#readme-files) though.

detectors/k8sapi/detector_test.go
```diff
@@ -0,0 +1,230 @@
...
+		semconv.K8SNodeUID(string(nodeUID)),
+	)
+	assert.Equal(t, expected, res)
+}
+
+func TestDetectClusterUIDOnlyNoNodeEnv(t *testing.T) {
+	clusterUID := uuid.NewUUID()
+
+	client := k8sfake.NewClientset(newFakeNamespace(clusterUID))
+	// K8S_NODE_NAME intentionally not set — cluster UID is independently detectable
```
MrAlias: This inherits `K8S_NODE_NAME`. Make the assertion explicit.

```suggestion
	t.Setenv("K8S_NODE_NAME", "")
	clusterUID := uuid.NewUUID()

	client := k8sfake.NewClientset(newFakeNamespace(clusterUID))
```

ps-mir: Explicit condition added.

detectors/k8sapi/doc.go
ps-mir: Added. Expect lint failure as godoc link doesn't exist yet.

detectors/k8sapi/go.mod
ps-mir: upgraded to fixed version. I thought this will be taken care by renovate.

iblancasa (approved):

dashpole (approved):

paulojmdias (approved):

detectors/k8sapi/version.go
```diff
@@ -0,0 +1,17 @@
...
+	return "0.16.0"
+	// This string is updated by the pre_release.sh script during release
+}
+
+// SemVersion is the semantic version to be supplied to tracer/meter creation.
+//
+// Deprecated: Use [Version] instead.
+func SemVersion() string {
+	return Version()
+}
```
MrAlias: `SemVersion` was deprecated for compatibility with existing users. This is a new module, so there are no existing users to support and no reason to add an already-deprecated API. The current pattern is `const Version`, as used by `detectors/azure/azurecontainerapps`. Please update `version_test.go` to read `k8sapi.Version` directly.

```suggestion
// Version is the current release version of the Kubernetes API resource detector.
const Version = "0.16.0"
```

MrAlias (changes requested):
There is one public API issue to address before this is ready.

detectors/k8sapi/version.go
ps-mir: Accepted.

MrAlias (approved):

CHANGELOG.md
```diff
@@ -10,8 +10,9 @@ This project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0.htm
 
 ### Added
 
-- Add new `elasticbeanstalk` resource detector for AWS Elastic Beanstalk, ported from `processor/resourcedetectionprocessor/internal/aws/elasticbeanstalk` in opentelemetry-collector-contrib. (#8993)
```
MrAlias:
```suggestion
```

PR #9112, opened 2026-06-09
envcar: ignore non-normalized env names on Get and Keys

pellared:
Follows https://github.com/open-telemetry/opentelemetry-specification/pull/5144 and https://github.com/open-telemetry/opentelemetry-specification/pull/5163.

Also related to the follow-up caching/stability discussion in https://github.com/open-telemetry/opentelemetry-specification/pull/5166.

This updates `go.opentelemetry.io/contrib/propagators/envcar` so reads from the current process environment follow the Environment Variables as Context Propagation Carriers key-normalization model without pre-populating a normalized copy of the whole environment.

## Changes

- Update `Carrier.Get` to normalize the requested key and perform a direct lookup of the matching normalized environment variable.
- Update `Carrier.Keys` to return only environment variable names that are already normalized, so non-normalized names such as `traceparent` are ignored instead of being normalized into `TRACEPARENT`.
- Remove the carrier's cached environment snapshot so `Get` and `Keys` use the same live-read observation model and avoid extra state/enumeration overhead.
- Add `normalized` helper coverage and update empty-key normalization so `normalize("")` returns `_`, matching the documented valid environment-name conversion.
- Expand GoDoc documentation to describe envcar as process-boundary startup input: applications should extract from the environment once during initialization and keep using the extracted context instead of repeatedly treating environment variables as a live parent-context source.
- Update the child-process example so it injects context from a real SDK span.

## Rationale

The PR discussion raised two related concerns:

- A cache can make repeated `Get` calls stable if a process mutates its environment later.
- Avoiding a cache removes carrier state and avoids enumerating the whole environment for common extraction paths.

This PR keeps the Go implementation lightweight and documents the intended usage instead: extract environment context once near application startup, then pass/use that extracted context. It also keeps `Get` and `Keys` consistent by making both read directly from the current environment rather than mixing a live `Get` with snapshot-style `Keys`.


propagators/envcar/carrier_example_test.go
```diff
@@ -34,10 +34,6 @@ func ExampleCarrier_extractFromParent() {
 	fmt.Printf("Trace ID: %s\n", spanCtx.TraceID())
 	fmt.Printf("Span ID: %s\n", spanCtx.SpanID())
 	fmt.Printf("Sampled: %t\n", spanCtx.IsSampled())
-	// Output:
-	// Trace ID: 0102030405060708090a0b0c0d0e0f10
-	// Span ID: 0102030405060708
-	// Sampled: true
 }
```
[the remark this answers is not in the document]
pellared: The tests can break the host if these env vars have been already used

propagators/envcar/carrier_example_test.go
```diff
@@ -77,5 +73,4 @@ func ExampleCarrier_childProcess() {
 		return
 	}
 	fmt.Print(string(out))
-	// Output: 00-0102030405060708090a0b0c0d0e0f10-0102030405060708-01
 }
```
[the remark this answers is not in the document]
pellared: The tests can break the host if these env vars have been already used

propagators/envcar/carrier_example_test.go
```diff
@@ -34,10 +34,6 @@ func ExampleCarrier_extractFromParent() {
 	fmt.Printf("Trace ID: %s\n", spanCtx.TraceID())
 	fmt.Printf("Span ID: %s\n", spanCtx.SpanID())
 	fmt.Printf("Sampled: %t\n", spanCtx.IsSampled())
-	// Output:
```
Joibel, line 37: Is this removal deliberate? This is the test outcome (it asserts that the Printf results are what we expect), otherwise this isn't much of a test.

propagators/envcar/carrier_example_test.go
```diff
@@ -77,5 +73,4 @@ func ExampleCarrier_childProcess() {
 		return
 	}
 	fmt.Print(string(out))
-	// Output: 00-0102030405060708090a0b0c0d0e0f10-0102030405060708-01
```
Joibel, line 80: Same here.

propagators/envcar/carrier_example_test.go
pellared: Removing the output changes that this example is no longer executed while running `go test`. The reason is that it was affecting other unit tests as it is setting the env vars for the whole process. I decided that having the example as tests hurt the test isolation and it does not bring much value.

propagators/envcar/carrier_example_test.go
pellared: https://github.com/open-telemetry/opentelemetry-go-contrib/pull/9112#discussion_r3381391050

propagators/envcar/carrier_example_test.go
Joibel: Can we either rename this to remove the _test,  or preferably add a defer(os.Setenv("TRACEPARENT","")) assuming that fixes the problem? I liked having examples that also proved they work.

propagators/envcar/carrier_example_test.go
pellared: TBH, I am not sure why I removed this one :wink:

pellared: https://github.com/open-telemetry/opentelemetry-go-contrib/pull/9112/commits/d74740b6cd3bf716d76c47ac01634efd3ea96a9d

propagators/envcar/carrier_example_test.go
pellared: https://github.com/open-telemetry/opentelemetry-go-contrib/pull/9112/commits/d74740b6cd3bf716d76c47ac01634efd3ea96a9d

Joibel (approved):
Looks good, thank you

kamphaus (approved):

dashpole (approved):

propagators/envcar/carrier.go
```diff
@@ -35,24 +38,28 @@ type Carrier struct {
...
-// Get returns the value associated with the normalized passed key.
+// Get returns the value associated with the normalized key.
 // The first call to [Carrier.Get] or [Carrier.Keys] for a
-// given Carrier will read and store the values from the
-// environment and all future reads will be from that store.
+// given Carrier will read and store the values from environment variables
+// whose names are already normalized, and all future reads will be from that
+// store.
 func (c *Carrier) Get(key string) string {
 	c.fetch()
```
MrAlias: This still enumerates the whole process environment on the first `Get` for a carrier. The spec change this PR tracks says `Get` should normalize the requested key and use that normalized key name to read from the carrier; the linked issue specifically calls out avoiding the old full-environment loop when env propagation is not configured.

Can we make `Get` do a direct lookup of `normalize(key)` instead, and leave enumeration to `Keys`?

MrAlias (commented):
I left a few comments. One is a thread that looks resolved in the PR UI, but the current code still has the same issue.

propagators/envcar/normalize.go
```diff
@@ -46,3 +46,26 @@ func normalize(s string) string {
 	}
 	return string(b)
 }
+
+// normalized reports whether s is already a normalized environment variable
+// name.
+func normalized(s string) bool {
+	if s == "" {
+		return true
```
MrAlias: The merged spec defines normalized names as matching `^[A-Z_][A-Z0-9_]*$`, so an empty name is not normalized. Returning true here means an inherited entry like `=X=bad` can make `Keys` return `""` and make `Get("")` read it.

Can we return false for `""` and add that case to the test table?

propagators/envcar/carrier_test.go
```diff
@@ -147,13 +152,15 @@ func TestInjectTraceContextEnvCarrier(t *testing.T) {
 }
 
 func TestCarrierKeys(t *testing.T) {
-	t.Setenv("traceparent", "value")
+	t.Setenv("TRACEPARENT", "value")
+	t.Setenv("envcar_non_normalized_key", "ignored")
```
MrAlias, line 155: This looks resolved in the PR UI, but the test is still not hermetic. If the outer environment already contains `ENVCAR_NON_NORMALIZED_KEY`, `Carrier.Keys()` will include it and this assertion can fail even though the implementation is correct.

Can we explicitly clear or restore the normalized variant before calling `Keys`? The same issue applies to `ENVCAR_GET_NON_NORMALIZED_KEY` in `TestCarrierGetIgnoresNonNormalizedEnvNames`.

propagators/envcar/normalize.go
pellared: Good catch.

I think the more important thing is that we should also normalize an empty string to `_` as an empty key value is not POSIX compliant.

Does it make sense?

propagators/envcar/carrier_test.go
pellared: I resolved because I thought that this is almost impossible to have such keys, but lets play safe :wink:

propagators/envcar/carrier.go
pellared: I support your proposal. 

Addressed in https://github.com/open-telemetry/opentelemetry-go-contrib/pull/9112/commits/ca0cacfb9daae79c96c67fa46358d516d320e97b and https://github.com/open-telemetry/opentelemetry-go-contrib/pull/9112/commits/fd398ac66899bbde4052dfe9dd16bef4f651f787

Some spec wording that can motivate the original decision from https://github.com/open-telemetry/opentelemetry-specification/blob/main/specification/context/env-carriers.md#operational-guidance

> Once set for a process, environment variables SHOULD be treated as immutable within that process:
> Applications SHOULD read context-related environment variables during initialization.


I also had a discussion with @dashpole that probably the "Operational Guidance" should use non-normative wording as it should indicate how the carrier should be used. Maybe some normative wording could be use to call out the languages should document guidance how the carriers should be used.

I tried to "implement" the guidance here https://github.com/open-telemetry/opentelemetry-go-contrib/pull/9112/commits/5b1606fabd376ac0b55b82c225d749205a9144ac

propagators/envcar/normalize.go
pellared: Spec PR: https://github.com/open-telemetry/opentelemetry-specification/pull/5163

propagators/envcar/carrier_test.go
pellared: Addressed https://github.com/open-telemetry/opentelemetry-go-contrib/pull/9112/commits/f57490e4040338cb5ade486b040be6a183dda02b

propagators/envcar/carrier_test.go
```diff
@@ -147,13 +151,16 @@ func TestInjectTraceContextEnvCarrier(t *testing.T) {
...
+	unsetenv(t, "ENVCAR_NON_NORMALIZED_KEY")
```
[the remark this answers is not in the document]
pellared: https://github.com/open-telemetry/opentelemetry-go-contrib/pull/9112/commits/c48452538ea12a3122702e67072f9f19f4f5f40f

pellared:
@Joibel, @dashpole, @MrAlias, can you please take another look?

propagators/envcar/carrier_test.go
```diff
@@ -147,13 +149,16 @@ func TestInjectTraceContextEnvCarrier(t *testing.T) {
...
+	t.Setenv("envcar-non-normalized-key", "ignored")
 
 	c := envcar.Carrier{}
 	keys := c.Keys()
 
 	assert.Contains(t, keys, "TRACEPARENT")
-	assert.NotContains(t, keys, "traceparent")
+	assert.NotContains(t, keys, "envcar_non_normalized_key")
+	assert.NotContains(t, keys, "ENVCAR_NON_NORMALIZED_KEY")
 }
```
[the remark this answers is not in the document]
pellared: https://github.com/open-telemetry/opentelemetry-go-contrib/pull/9112/commits/591c0c3a004b6ccdce6ef9b14694ee70e16bb758

propagators/envcar/carrier.go
```diff
@@ -25,37 +27,38 @@ import (
...
 	c.once.Do(func() {
 		environ := os.Environ()
-		c.values = make(map[string]string, len(environ))
+		c.keys = make(map[string]struct{}, len(environ))
 		for _, kv := range environ {
 			kvPair := strings.SplitN(kv, "=", 2)
-			key := normalize(kvPair[0])
-			c.values[key] = kvPair[1]
+			key := kvPair[0]
+			if !normalized(key) {
```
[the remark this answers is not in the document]
pellared: https://github.com/open-telemetry/opentelemetry-go-contrib/pull/9112/commits/060e0d3763ec0b474e16b33316eac2de1d3ee355

propagators/envcar/carrier.go
```diff
@@ -25,37 +27,37 @@ import (
...
-// Get returns the value associated with the normalized passed key.
-// The first call to [Carrier.Get] or [Carrier.Keys] for a
-// given Carrier will read and store the values from the
-// environment and all future reads will be from that store.
-func (c *Carrier) Get(key string) string {
-	c.fetch()
-	return c.values[normalize(key)]
+// Get returns the value associated with the normalized key.
+// It reads the normalized key directly from the current process environment.
+func (*Carrier) Get(key string) string {
```
Joibel, line 48: Why the change back from store on first call? The Carrier was not supposed to (potentially) return different values for the same key, see the original PR for discussions on this.

pellared: Please check this discussion https://github.com/open-telemetry/opentelemetry-go-contrib/pull/9112#discussion_r3432179522

This is also a related PR https://github.com/open-telemetry/opentelemetry-specification/pull/5166

pellared: Long story short the appropriate usage is to use this carrier once near application startup.

Caching and enumeration causes unnecessary overhead that we want to avoid.

pellared: I unresolved the mentioned thread so that you can navigate to it

Joibel: Does that mean that Keys and Get can respond differently (if used contra to that startup expectation). Shouldn't we drop the store behaviour for Keys too? If the expectation is basically only one call to Keys ever then the store is pretty pointless.

kamphaus: I get wanting to avoid enumerating all environment variables for efficiency, however `Get` now reads directly from the env vars.

The mutability of env vars is a concern of mine and I worry that reading directly from env vars without a cache could lead to unpredictable runtime bugs (eg. when an instrumented process emits several spans using the env var context as parent context and the vars were changed in between). Even instantiating the carrier once during startup would not protect against this.

The more I think about it, the more I want to add normative language to the spec like: `Get MUST NOT return different values when called with the same argument` implying the use of a cache.

pellared: @Joibel, @kamphaus, thanks a lot for your questions/concerns. These are exactly the discussions that we need to have before going RC.

> Does that mean that Keys and Get can respond differently (if used contra to that startup expectation). Shouldn't we drop the store behaviour for Keys too? If the expectation is basically only one call to Keys ever then the store is pretty pointless.

@Joibel, I agree that `Keys` and `Get` should have the same observation model. If the expected usage is extracting once near startup, then `Keys` will typically be called at most once for that extraction path (e.g. when using https://pkg.go.dev/go.opentelemetry.io/contrib/propagators/ot which is the only known propagator that uses `Keys` method).

Removing the `Keys` cache also seems consistent with the current intended usage, and I think it would avoid the confusing case where `Get` is live but `Keys` is a snapshot. It also looks fine for your vendored carrier usage here: https://github.com/argoproj/argo-workflows/blob/4cb7eb4e89299595ba8fddd3b7f9d95872c3b39d/workflow/executor/tracing/tracing.go#L26.

Addressed https://github.com/open-telemetry/opentelemetry-go-contrib/pull/9112/commits/aeaccb08719bfc01986cf8b781855b2981756528

Just to double-check. Would you see any benefit from caching `Get` and `Keys`, or would it just be unnecessary state/overhead?

> Even instantiating the carrier once during startup would not protect against this.

@kamphaus, agreed that construction alone would not protect against later environment mutation. What I meant is that applications should extract from the environment once during startup and then use the extracted context, rather than repeatedly using the process environment as a live parent-context source.

Addressed https://github.com/open-telemetry/opentelemetry-go-contrib/pull/9112/commits/ef3ca092ca04a0c708c93506ae624c708c18b9bd

> The more I think about it, the more I want to add normative language to the spec like: `Get MUST NOT return different values when called with the same argument` implying the use of a cache.

@kamphaus, I do not think we have evidence yet that this is something which needs to be normative.

For OTel Go, C++ and Rust, we usually care very much about any overhead, I think it is fine if we can expect users to not call `Get` more than once for the same argument.

At the same time, I think that for some languages it may be fine to cache. But maybe even for these languages it should be opt-in as it cause overhead that users may not want.

At this point of time, I lean towards the non-normative proposal from https://github.com/open-telemetry/opentelemetry-specification/pull/5166

_I do not take this lightly and I plan to further discuss this with maintainers of other language implementations._

@Joibel, WDYT?

kamphaus: > I do not take this lightly and I plan to also discuss this with other language maintainers.

I agree that this needs to be discussed with other language maintainers and thank you for taking this seriously.

Might I suggest in the meantime to document in the godoc that applications using envcar/Carrier should cache the context based on it?

kamphaus (approved):

propagators/envcar/carrier.go
pellared: > Might I suggest in the meantime to document in the godoc that applications using envcar/Carrier should cache the context based on it?

I did this before I read your comment :wink: 

PTAL https://github.com/open-telemetry/opentelemetry-go-contrib/commit/ef3ca092ca04a0c708c93506ae624c708c18b9bd

kamphaus: > PTAL https://github.com/open-telemetry/opentelemetry-go-contrib/commit/ef3ca092ca04a0c708c93506ae624c708c18b9bd

And I reviewed it before seeing this.

propagators/envcar/go.mod
```diff
@@ -5,15 +5,20 @@ go 1.25.0
 require (
 	github.com/stretchr/testify v1.11.1
 	go.opentelemetry.io/otel v1.44.0
+	go.opentelemetry.io/otel/sdk v1.44.0
```
MrAlias: Non-blocking question: since `go.opentelemetry.io/otel/sdk` is only used by the executable example, this adds the SDK and its indirect requirements to the `envcar` module graph for documentation/test code.

Could we keep the example on the API-only `trace.ContextWithSpanContext` path, or is using a real SDK span important enough here to justify the added module dependency? This should not affect users' binaries, but it can affect downstream build lists, vendoring, and module metadata.

MrAlias (approved):
Thanks, this looks good overall. I left one non-blocking question about the example-only SDK dependency.

propagators/envcar/go.mod
kamphaus: The use of the SDK was introduced in 19c35a09ec2a97f9d7667b2042eecc03bb2f1608.
The test was passing before. Why was this changed?

pellared: I changed this to have a more realistic example. I can revert this.

pellared: Reverted https://github.com/open-telemetry/opentelemetry-go-contrib/pull/9112/commits/f15acc3594013c97fad75e9a0295fb2bde4252ed

PR #9126, opened 2026-06-10
otelconf: allow null OTLP endpoint to use defaults

### LINKED ISSUE #9052: otelconf: Endpoint null bypasses new OTLP exporter validation
### Description

Endpoint `null` can bypass new OTLP exporter endpoint validation in JSON/YAML config handling.

In otelconf/config_json.go, UnmarshalJSON only checks that the raw JSON key exists (`if sh.Endpoint == nil { return newErrRequired(...) }`) then unmarshals `sh.Endpoint` into `*string` (lines 370-386). If the JSON is `{"endpoint":null}`, `sh.Endpoint` is non-nil (RawMessage == "null"), so the required check is bypassed and `json.Unmarshal("null", &sh.Plain.Endpoint)` sets the pointer to nil without error. YAML has the same pattern: UnmarshalYAML only verifies the key exists (`hasYAMLMapKey`) and decodes into the struct (lines 186-199), so `endpoint: null` yields a nil pointer without error. The Endpoint field is a `*string` (generated_config.go lines 641-655), so nil is a valid decoded value. This confirms the required-field validation is bypassed when endpoint is explicitly set to null.

### Steps To Reproduce

[19-4526869d3c6c8191b6fd789c14d6fde4-endpoint-null-bypasses-new-otlp-exporter-validation.zip](https://github.com/user-attachments/files/28385397/19-4526869d3c6c8191b6fd789c14d6fde4-endpoint-null-bypasses-new-otlp-exporter-validation.zip)

1. Create an OTLP exporter config with `endpoint: null`.
2. Load it through the JSON or YAML config path.
3. Observe that validation is bypassed or produces behavior different from an omitted/invalid endpoint.

### Expected behavior

A null endpoint should be rejected or normalized consistently with the schema and endpoint validation rules.

viacheslav-danilin:
Closes #9052.

The required `endpoint` field for OTLP exporter configs only checked that the JSON/YAML key was present, so `{"endpoint":null}` and `endpoint: null` decoded to a nil `*string` and bypassed `newErrRequired`. Add a post-unmarshal nil check in the manual JSON and YAML unmarshalers for `OTLPHttpExporter`, `OTLPGrpcExporter`, `OTLPHttpMetricExporter`, and `OTLPGrpcMetricExporter` in `otelconf` and `otelconf/x`, returning the same `newErrRequired(j, "endpoint")` used for the missing-key case.

Add `null endpoint` cases to the existing `TestUnmarshalOTLP*` table tests covering both JSON and YAML inputs.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -17,6 +17,10 @@ This project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0.htm
 
 - The `Version()` function in `go.opentelemetry.io/contrib/instrumentation/github.com/gorilla/mux/otelmux` has been replaced by `const Version`. (#9076)
 
+### Fixed
+
+- Allow `endpoint` to be omitted or set to `null` for OTLP exporters in `go.opentelemetry.io/contrib/otelconf` and `go.opentelemetry.io/contrib/otelconf/x`; the SDK exporter defaults are used in this case, matching the OpenTelemetry configuration schema. (#9052)
+
 <!-- Released section -->
 <!-- Don't change this section unless doing release -->
 
--- a/otelconf/config_json.go
+++ b/otelconf/config_json.go
@@ -470,96 +470,56 @@ func (j *SpanLimits) UnmarshalJSON(value []byte) error {
 // UnmarshalJSON implements json.Unmarshaler.
 func (j *OTLPHttpMetricExporter) UnmarshalJSON(b []byte) error {
 	type Plain OTLPHttpMetricExporter
-	type shadow struct {
-		Plain
-		Endpoint json.RawMessage `json:"endpoint"`
-	}
-	var sh shadow
-	if err := json.Unmarshal(b, &sh); err != nil {
+	var plain Plain
+	if err := json.Unmarshal(b, &plain); err != nil {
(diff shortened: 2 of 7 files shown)


codeboten (commented):
Thanks for your PR @viacheslav-danilin! According to the configuration schema, a null endpoint should default to an appropriate endpoint, see https://github.com/open-telemetry/opentelemetry-configuration/blob/639833e2ca33855d7c24d93483ad58337f2557d3/opentelemetry_configuration.json#L1785

For grpc, the endpoint should default to localhost:4317 and for http, it should default to http://localhost:4318/v1/<signal>

viacheslav-danilin:
> Thanks for your PR @viacheslav-danilin! According to the configuration schema, a null endpoint should default to an appropriate endpoint, see https://github.com/open-telemetry/opentelemetry-configuration/blob/639833e2ca33855d7c24d93483ad58337f2557d3/opentelemetry_configuration.json#L1785
> 
> For grpc, the endpoint should default to localhost:4317 and for http, it should default to http://localhost:4318/v1/

Thanks, updated the PR to follow the schema behavior for omitted and null endpoints.

PR #9131, opened 2026-06-11
Implement resource detection attribute filtering in otelconf/x

### LINKED ISSUE #9130: Implement resource detection attribute filtering in otelconf/x
### Component

Config

### Problem Statement

`otelconf/x` exposes resource.detection/development.attributes.included and excluded in the generated config model, but resource creation does not currently apply those filters to detector-produced attributes.

### Proposed Solution

Resource detection should apply the existing `IncludeExclude` filtering behavior to attributes produced by resource detectors. Configured resource attributes should be built separately from detected attributes, so detection filters only affect detector output and do not remove explicitly configured `resource.attributes`.

### Alternatives

_No response_

### Prior Art

_No response_

### Additional Context

_No response_

iblancasa:
Fixes #9130


otelconf/x/resource.go
```diff
@@ -50,9 +50,46 @@ func newResource(r *Resource) (*resource.Resource, error) {
...
+	if r.DetectionDevelopment == nil {
+		return base, nil
+	}
+
+	detected, err := newDetectedResource(r.DetectionDevelopment)
+	if err != nil {
+		return nil, err
+	}
+
+	return resource.Merge(base, detected)
```
ps-mir: Does detection has higher precedence than base ?

otelconf/x/resource.go
```diff
@@ -50,9 +50,46 @@ func newResource(r *Resource) (*resource.Resource, error) {
...
 	}
 
-	return resource.New(context.Background(), opts...)
+	if r.DetectionDevelopment == nil {
+		return base, nil
+	}
+
+	detected, err := newDetectedResource(r.DetectionDevelopment)
+	if err != nil {
+		return nil, err
```
ps-mir: If are introducing new error path, is it ok to retain base (with ErrPartialResource) in case something goes wrong with detection ?

otelconf/x/resource.go
iblancasa: Mmm I think it makes sense to keep configured attributes with higher precedence. I will fix it

otelconf/x/resource.go
iblancasa: I will change this to only return `nil` when SDK returns no resource. If SDK returns a partial resource plus error, `attrs` still get filtered and merged with base, and `error` is preserved via `errors.Join`. Does it make sense?

otelconf/x/resource.go
```diff
@@ -51,9 +52,50 @@ func newResource(r *Resource) (*resource.Resource, error) {
...
 	}
 
-	return resource.New(context.Background(), opts...)
+	if r.DetectionDevelopment == nil {
+		return base, nil
+	}
+
+	detected, err := newDetectedResource(r.DetectionDevelopment)
```
MrAlias: This continues the existing behavior where resource detection uses `context.Background()` instead of the context from `NewSDK(WithContext(...))`. That seems worth fixing, but it is separate from the filtering behavior here and probably better tracked in a follow-up issue (I'll look into adding one).

That said, can you add the plumbing here to pass along the context to `newDetectedResource` which also uses a `context.Background`?

MrAlias: https://github.com/open-telemetry/opentelemetry-go-contrib/issues/9158

otelconf/x/resource_test.go
```diff
@@ -160,3 +168,123 @@ func TestResourceOptsWithDetectors(t *testing.T) {
...
+		},
+		{
+			name: "exclude-os",
+			attrs: &IncludeExclude{
+				Excluded: []string{string(semconv.OSTypeKey)},
+			},
+			wantHost:   true,
+			wantCustom: true,
+		},
+	}
```
ps-mir, line 212: I think it could be more useful to separate out Filter Mechanics with Scope on Custom Attributes. That way it will make it more clear that `include-host` and `exclude-os` could lead to same outcome, and also separate emphasis on Custom Attributes not being part of Filter scope.

otelconf/x/resource.go
ps-mir: Are we keeping base, if detection fails entirely (err with detected=nil). If not why ?

otelconf/x/resource_test.go
```diff
@@ -160,3 +168,138 @@ func TestResourceOptsWithDetectors(t *testing.T) {
...
+		Attributes: []AttributeNameValue{
+			{Name: "custom", Value: "value"},
+		},
+		DetectionDevelopment: &ExperimentalResourceDetection{
+			Detectors: []ExperimentalResourceDetector{
+				{Host: ExperimentalHostResourceDetector{}},
+			},
+			Attributes: &IncludeExclude{
+				Included: []string{string(semconv.HostNameKey)},
+			},
```
ps-mir, line 249: Nit: Can we add a exclude filter for custom value. Improves informative aspect of the test.

ps-mir (approved):
Added a Nit, rest seems ok.

MrAlias (approved):

codeboten (approved):

iblancasa:
Can we merge this?

dashpole (approved):

PR #9132, opened 2026-06-11
feat(detectors): add ability to inject logger to aws ec2 client

skl:
This PR adds the ability to inject a custom logger to the AWS EC2 resource detector’s AWS SDK client.

The intention is to allow consumers to communicate log levels to their current logging infrastructure, rather than relying on the default SDK logger.

## API Stability

In order to maintain public API stability, the existing constructor (`NewResourceDetector`) behaviour is kept the same and a new constructor (`NewResourceDetectorWithOptions`) is introduced. In particular:

- `NewResourceDetector` currently silently drops client creation errors, that specific behaviour is kept the same to prevent surprises for existing users.
- `NewResourceDetectorWithOptions` has a different signature: it accepts `opts` and also returns an error. This could become the new default constructor at the next major version if desired, for now they can coexist.

## Design Decisions

1. I tried to keep the public API type dependencies small, introducing `logging.Logger` from `aws/smithy-go` only. This has a very small and stable API surface. If this is not desirable, we could instead introduce an intermediate interface to be used by all AWS detectors.
2. I did not expose a generic `WithAWSConfigOptions` to avoid introducing a dependency on the entire AWS Configuration API surface.
3. To prevent the unit tests from executing AWS SDK code directly, `awsconfig.LoadDefaultConfig` is now set via a package-level var to allow for fake substitution.

### Example Usage

Here's an example of how a consumer could inject `slog`, and handle configuration errors:

```go
import (
	"context"
	"fmt"
	"log/slog"
	"os"

	"github.com/aws/smithy-go/logging"
	"go.opentelemetry.io/contrib/detectors/aws/ec2/v2"
)

slogger := slog.New(slog.NewJSONHandler(os.Stderr, nil))

// The SDK only emits Warn and Debug.
awsLogger := logging.LoggerFunc(func(c logging.Classification, format string, v ...any) {
	var level slog.Level
	switch c {
	case logging.Warn:
		level = slog.LevelWarn
	case logging.Debug:
		level = slog.LevelDebug
	default:
		level = slog.LevelInfo
	}
	slogger.Log(context.Background(), level, fmt.Sprintf(format, v...))
})

detector, err := ec2.NewResourceDetectorWithOptions(ec2.WithLogger(awsLogger))
if err != nil {
	// handle AWS SDK configuration error
}

res, err := resource.New(ctx, resource.WithDetectors(detector))
```


pellared:
@pyohannes, PTAL as a code owner

detectors/aws/ec2/v2/ec2.go
```diff
@@ -40,7 +41,43 @@ var _ resource.Detector = (*resourceDetector)(nil)
...
 func NewResourceDetector() resource.Detector {
-	return &resourceDetector{c: newClient()}
+	// Drop error to preserve stable API.
+	client, _ := newClient(config{})
+	return &resourceDetector{c: client}
+}
+
+// NewResourceDetectorWithOptions returns a resource detector that will detect
+// AWS EC2 resources. The options configure the AWS SDK client.
+func NewResourceDetectorWithOptions(opts ...Option) (resource.Detector, error) {
```
dashpole: I think this is the only detector that returns an error from the New function.  Others just defer client creation until Detect() and return the error there.  Should we switch to that pattern?

detectors/aws/ec2/v2/ec2.go
```diff
@@ -40,7 +41,43 @@ var _ resource.Detector = (*resourceDetector)(nil)
...
+	}
+	client, err := newClient(c)
+	if err != nil {
+		return nil, err
+	}
+	return &resourceDetector{c: client}, nil
+}
+
+// WithLogger passes the [logging.Logger] to the AWS SDK client.
+func WithLogger(logger logging.Logger) Option {
```
dashpole: Just to confirm: there isn't a bridge from LoggerProvider to the aws logger, right?  It would be nice if OTel components accepted LoggerProviders, but it isn't a hard requirement.  Should this be WithAWSLogger or something to indicate it isn't an otel logger?

detectors/aws/ec2/v2/ec2.go
skl: Done in 0b7176194

detectors/aws/ec2/v2/ec2.go
```diff
@@ -40,7 +41,43 @@ var _ resource.Detector = (*resourceDetector)(nil)
...
+	return optionFunc(func(c *config) {
+		c.logger = logger
+	})
+}
+
+// Option represents a AWS SDK client configuration option function.
```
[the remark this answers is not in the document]
skl: Done in 0b7176194

detectors/aws/ec2/v2/ec2.go
skl: Assuming you mean these [bridges](https://github.com/open-telemetry/opentelemetry-go-contrib/tree/main/bridges), looks like one does not exist for https://pkg.go.dev/github.com/aws/smithy-go/logging.

For now, I renamed `WithLogger` to `WithAWSLogger` in d6aff0445, can always also add `WithLoggerProvider` in future (assuming that we might provide both?)

Happy to take a look at creating the bridge here, or in a follow-up PR? Am I right in thinking that we would export both `WithAWSLogger` and `WithLoggerProvider`?

dashpole: Probably? I think its fine to just stick with WithAWSLogger for now (or smithy logger, or whatever makes the most sense).

ps-mir (approved):

iblancasa (approved):

paulojmdias (approved):

detectors/aws/ec2/v2/ec2.go
skl: I don't feel strongly either way on the naming, it seems Smithy is the canonical AWS logger for the AWS SDK, so either works for me.

detectors/aws/ec2/v2/ec2.go
```diff
@@ -48,19 +47,45 @@ var _ resource.Detector = (*resourceDetector)(nil)
...
+	logger logging.Logger
 }
 
 // Detect detects associated resources when running in AWS environment.
 func (detector *resourceDetector) Detect(ctx context.Context) (*resource.Resource, error) {
-	// Return nil if not able to establish valid client
-	client := detector.getClient()
-	if client == nil {
-		return nil, errClient
+	client, err := newClient(detector.cfg)
```
MrAlias: This changes the detector to create a new IMDS client on every `Detect` call. `imds.NewFromConfig` does not expose teardown, but it does build a client with its own HTTP transport and IMDS token provider, so repeated resource detection loses connection/token reuse and can leave idle transport resources around until they age out.

I think this should lazily initialize and cache the client on the detector, protected by a mutex and populated only after successful construction. Pass the current `Detect` context into AWS config loading while constructing the client, but store only the resulting client; later calls should reuse that client and still pass their own `ctx` to the IMDS operations.

MrAlias (commented):
One lifecycle issue to consider before this is ready.

detectors/aws/ec2/v2/ec2.go
```diff
@@ -92,13 +117,18 @@ func (detector *resourceDetector) Detect(ctx context.Context) (*resource.Resourc
...
 
-func newClient() client {
-	cfg, err := loadDefaultConfig(context.Background())
+func newClient(c config) (client, error) {
+	var optFns []func(*awsconfig.LoadOptions) error
+	if c.logger != nil {
+		optFns = append(optFns, awsconfig.WithLogger(c.logger))
+	}
+
+	cfg, err := loadDefaultConfig(context.Background(), optFns...)
```
dashpole: Lets plumb context to this call instead of using context.Background() to ensure cancellation works properly.

skl: Done in 61d81e4abd9908eab1cb839c1c65074e37fcc1a2

detectors/aws/ec2/v2/ec2.go
skl: Done in 61d81e4abd9908eab1cb839c1c65074e37fcc1a2

dashpole (approved):

pellared (approved):
The PR description is outdated, but approving and merging 😉

PR #9137, opened 2026-06-15
Add support for gcp resource detector in otelconf/x

iblancasa:


otelconf/jsonschema_patch.sed
```diff
@@ -1,4 +1,12 @@
 # go-jsonschema always generates patternProperties as
 # map[string]interface{}, for more specific types, they must
 # be replaced here
-s+type Headers.*+type Headers map[string]string+g
\ No newline at end of file
+s+type Headers.*+type Headers map[string]string+g
```
dashpole, line 4: Why is this patch necessary?

iblancasa: This is needed because the generated config types come from the upstream declarative config schema, which does not define things like `gcp`.

To be honest, I didn't find a better way to solve this situation. I also had to do some hack in the `Makefile`... which I don't feel very confortable with. I had to do this for all the similar PRs I created (those are similar but the detector is the thing that changes).

I'm totally open to change this. But not sure how.

dashpole (approved):

iblancasa:
@open-telemetry/go-maintainers, please take a look

CHANGELOG.md
```diff
@@ -10,7 +10,8 @@ This project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0.htm
 
 ### Added
 
-- The resource created by `go.opentelemetry.io/contrib/otelconf` now includes [default SDK attributes](https://pkg.go.dev/go.opentelemetry.io/otel/sdk/resource#Default). (#8990)
+- The resource created by `go.opentelemetry.io/contrib/otelconf` now includes [default SDK attributes](https://pkg.go.dev/go.opentelemetry.io/otel/sdk/resource#Default). (#9137)
```
MrAlias: Should this have changed?


```suggestion
- The resource created by `go.opentelemetry.io/contrib/otelconf` now includes [default SDK attributes](https://pkg.go.dev/go.opentelemetry.io/otel/sdk/resource#Default). (#8990)
```

MrAlias (approved):

CHANGELOG.md
iblancasa: I guess I messed up during a merge.

iblancasa:
I did a bad rebase but guess I fixed it now.

otelconf/go.mod
```diff
@@ -6,7 +6,8 @@ require (
 	github.com/prometheus/client_golang v1.23.2
 	github.com/prometheus/otlptranslator v1.0.0
 	github.com/stretchr/testify v1.11.1
-	go.opentelemetry.io/contrib/detectors/aws/eks v1.44.0
+	go.opentelemetry.io/contrib/detectors/aws/eks v0.0.0-00010101000000-000000000000
```
MrAlias: This placeholder version is only valid while the local `replace` is in effect. Downstream consumers ignore `replace` directives in dependencies, so resolving `otelconf` would try to download `go.opentelemetry.io/contrib/detectors/aws/eks@v0.0.0-00010101000000-000000000000`, which does not exist.

Keep the released EKS version in the requirement; the local replace already selects the workspace module here.

```suggestion
	go.opentelemetry.io/contrib/detectors/aws/eks v1.44.0
```

MrAlias (commented):
This should not merge with a placeholder EKS dependency: it makes the published otelconf module unresolvable for consumers.

otelconf/go.mod
iblancasa: 🤦🏻 sorry for this. I checked the PR 3 times and didn't notice

MrAlias:
@iblancasa this looks ready to merge after the conflicts have been addressed. Thanks for the contribution 😄

PR #9138, opened 2026-06-15
Add support for aws.eks resource detector in otelconf/x

iblancasa:


otelconf/x/config_json.go
```diff
@@ -192,6 +207,7 @@ func (j *ExperimentalResourceDetector) UnmarshalJSON(b []byte) error {
 	type Plain ExperimentalResourceDetector
 	type shadow struct {
 		Plain
+		AWSEKS    json.RawMessage `json:"aws.eks"`
```
dashpole, line 210: just curious (apologies if I already asked this elsewhere). How did you decide on `aws_eks` vs `aws.eks` here?

iblancasa: I followed whatever `autodetect.go` had https://github.com/open-telemetry/opentelemetry-go-contrib/blob/abd6958a5b955e3a2bd8c11f968b221c7cace421/detectors/autodetect/autodetect.go#L39

dashpole: ah, got it.

dashpole (approved):

otelconf/x/resource.go
```diff
@@ -10,12 +10,17 @@ import (
 	"go.opentelemetry.io/otel/attribute"
 	"go.opentelemetry.io/otel/sdk/resource"
 
+	eksdetector "go.opentelemetry.io/contrib/detectors/aws/eks"
```
MrAlias, line 13: Non-blocking: this links the EKS detector's full Kubernetes clientset into every `otelconf/x` binary, even when `aws.eks` is not configured. A minimal binary that only calls `NewSDK()` grows from 27 MB to 75 MB and links 268 Kubernetes packages.

This should not block this PR, but I think we should follow up by changing `detectors/aws/eks` to use the typed CoreV1 client, or another lightweight client, instead of `kubernetes.NewForConfig`.

MrAlias (approved):

otelconf/x/resource.go
iblancasa: I see. I will create a follow up PR as soon as we merge this.

PR #9139, opened 2026-06-15
Add support for aws.ec2 resource detector in otelconf/x

iblancasa:


dashpole (approved):

otelconf/x/resource_test.go
```diff
@@ -125,6 +125,7 @@ func TestResourceOptsWithDetectors(t *testing.T) {
 		{
 			name: "all-detectors",
 			detectors: []ExperimentalResourceDetector{
+				{AWSEC2: ExperimentalAWSEC2ResourceDetector{}},
```
MrAlias, line 165: Should this stub out IMDS and not probe the real EC2 client?

iblancasa:
Is there anything else to do here?

otelconf/go.mod
```diff
@@ -32,6 +33,20 @@ require (
 )
 
 require (
+	github.com/aws/aws-sdk-go-v2 v1.42.0 // indirect
```
MrAlias: The local replacement at the bottom of this file now resolves to an EC2 module that requires newer AWS SDK versions than this block records. Both `lint` and `govulncheck` fail with `updates to go.mod needed`, so this cannot pass the required checks.

Please rebase onto `main`, run `make crosslink` and `go mod tidy`, and commit the resulting module changes.

otelconf/x/resource_test.go
```diff
@@ -152,6 +155,11 @@ func TestResourceOptsWithDetectors(t *testing.T) {
 	}
 	for _, tt := range tests {
 		t.Run(tt.name, func(t *testing.T) {
+			for _, detector := range tt.detectors {
+				if detector.AWSEC2 != nil {
+					stubIMDS(t)
```
MrAlias, line 194: Non-blocking: this stubs IMDS, but the test still only asserts attributes produced by the host, OS, process, and service detectors. Removing the `d.AWSEC2` branch still leaves this test green.

Could we assert an EC2 attribute such as `cloud.provider`, `cloud.platform`, or `host.id` so the integration is protected?

MrAlias (commented):
The EC2 integration is reasonable, but the stale branch and broken module graph need to be resolved before this is ready to merge.

MrAlias (approved):

PR #9150, opened 2026-06-16
instrumentation/otelgrpc: add rpc.client.attempt.started and rpc.server.call.started metrics

om7057:
#### Description

Add the `rpc.client.attempt.started` and `rpc.server.call.started` counters to `otelgrpc`.

Both counters are emitted from `stats.Begin` and provide visibility into the start of client call attempts and server call handling, respectively.

The client-side counter enables observation of retry behavior, which is not visible from the existing per-call duration metric alone. Both counters carry the same method/service attributes used by the duration metrics. They do not include `rpc.response.status_code`, since that information is only available at `stats.End`.

As with the existing duration instruments, a noop fallback is used if instrument creation fails.

#### Related issues

* Part of #2840
* Requested from open-telemetry/opentelemetry-collector#15169

cc @iblancasa

### DIFF
--- a/instrumentation/google.golang.org/grpc/otelgrpc/grpc_stats_handler_test.go
+++ b/instrumentation/google.golang.org/grpc/otelgrpc/grpc_stats_handler_test.go
@@ -282,6 +282,22 @@ func checkClientMetrics(t *testing.T, reader metric.Reader, addr, stabilityOptIn
 	switch stabilityOptIn {
 	case "rpc/old":
 		expectedMetrics = append(expectedMetrics, metricdata.Metrics{
+			Name:        "rpc.client.attempt.started",
+			Description: "The number of client call attempts started.",
+			Unit:        "{attempt}",
+			Data: metricdata.Sum[int64]{
+				Temporality: metricdata.CumulativeTemporality,
+				IsMonotonic: true,
+				DataPoints: []metricdata.DataPoint[int64]{
+					{Attributes: attribute.NewSet(attribute.String("rpc.system", "grpc"), attribute.String("rpc.service", "grpc.testing.TestService"), attribute.String("rpc.method", "EmptyCall"), testMetricAttr)},
+					{Attributes: attribute.NewSet(attribute.String("rpc.system", "grpc"), attribute.String("rpc.service", "grpc.testing.TestService"), attribute.String("rpc.method", "UnaryCall"), testMetricAttr)},
+					{Attributes: attribute.NewSet(attribute.String("rpc.system", "grpc"), attribute.String("rpc.service", "grpc.testing.TestService"), attribute.String("rpc.method", "StreamingInputCall"), testMetricAttr)},
+					{Attributes: attribute.NewSet(attribute.String("rpc.system", "grpc"), attribute.String("rpc.service", "grpc.testing.TestService"), attribute.String("rpc.method", "StreamingOutputCall"), testMetricAttr)},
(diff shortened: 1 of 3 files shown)


dashpole:
I thought we were going to do one PR per metric?

om7057:
> I thought we were going to do one PR per metric?

Oh, right, apologies for the confusion. I initially grouped the two started counters together, thinking they shared the same code path, but I understand the value of keeping each metric strictly separate - it makes the review, discussion, and potential revert of each metric independent. I've updated the PR to cover only rpc.client.attempt.started and will open separate PRs for the remaining metrics.

ps-mir:
@dashpole @iblancasa 

Any reason for moving forward with metrics which are not part of semconv yet ? Not sure what can be said about stability of these metrics for end users.

om7057:
> @dashpole @iblancasa
> 
> Any reason for moving forward with metrics which are not part of semconv yet ? Not sure what can be said about stability of these metrics for end users.

That's a valid concern. These metrics aren't formally in semconv yet, the mapping work is tracked in open-telemetry/semantic-conventions#3229. The motivation here is that these metrics are currently not recoverable at all from otelgrpc, and the gap is documented. That said, if the preferred path is to wait for semconv to stabilize the names first before implementing them here, I'm happy to hold off. @dashpole @iblancasa - what's your preference on this?

dashpole:
They DO need to be in semconv, but they don't need to be stable for inclusion (this module is still not 1.0).

om7057:
> They DO need to be in semconv, but they don't need to be stable for inclusion (this module is still not 1.0).

Thanks for clarifying. Looking at the current semconv docs, rpc.client.attempt.started doesn't appear to be defined there yet; the gRPC page only references the general rpc.client.call.duration metric. Is the right next step to open a PR in the semantic-conventions repo to add these metrics first, and then come back here once they're accepted even as experimental?

dashpole:
Is there an issue in the semconv repo tracking the new metrics?  That should be the first step. There is an RPC semantic conventions group/SIG that you will want to get opinions from before opening a PR to the semantic conventions repo.

om7057:
I checked, and there doesn't appear to be an existing issue in the semconv repo tracking these specific metrics. The referenced https://github.com/open-telemetry/semantic-conventions/pull/3229 was a compatibility mapping document (already merged), it documented the gap, but didn't propose adding the metrics to semconv. I'll open an issue in the semantic-conventions repo to propose adding these metrics and bring it to the RPC SIG for feedback. I'll keep this PR on hold until there's consensus there.
Will it do @dashpole ?

dashpole:
Sounds reasonable to me. I moved this to draft while you work on updating the conventions.

om7057:
Closing this PR based on update from the semantic-conventions maintainers.

After opening [open-telemetry/semantic-conventions#3804](https://github.com/open-telemetry/semantic-conventions/issues/3804) to check if these metrics should be in semconv, @lmolkova (semconv maintainer) confirmed that **semconv will not add these metrics**. 

The RPC semantic conventions focus on logical-layer (per-call) metrics, not transport/attempt-level metrics like retry attempts and compressed message sizes. The recommendation is to use gRPC's native instrumentation (`google.golang.org/grpc/stats/opentelemetry`) directly.

Since these metrics won't be part of semconv, adding them to `otelgrpc` (which follows semconv) would create non-standard metrics under the `rpc.*` namespace, which isn't appropriate.

I'll return to the original collector approach (PR #15446) in: open-telemetry/opentelemetry-collector#15169

Thanks @dashpole and @iblancasa for the guidance throughout this process!

PR #9160, opened 2026-06-17
Honor WithContext for otelconf resources

### LINKED ISSUE #9158: Honor `WithContext` during `otelconf/x` resource detection
### Component

Config

### Problem Statement

`otelconf/x.NewSDK` accepts `WithContext(ctx)` and stores it in the internal configuration options. Most SDK construction paths use that context when creating exporters, processors, and metric readers.

Resource construction is an exception. `otelconf/x/resource.go` calls `resource.New(context.Background(), ...)`, including when `resource.detection/development` enables SDK resource detectors. This means resource detection ignores the context supplied with `NewSDK(WithContext(...))`.

The current built-in detectors mostly do local work and may not make this very visible today, but `resource.New` accepts a context for detector execution and the rest of `otelconf/x` already treats `WithContext` as the SDK construction context. Resource detection should follow the same behavior.

### Expected Behavior

When callers configure `otelconf/x.NewSDK(WithContext(ctx), ...)`, resource construction and resource detector execution should use that context.

### Actual Behavior

`otelconf/x` uses `context.Background()` while creating the resource, so caller cancellation/deadline is not propagated to resource detection.

### Proposed Solution

Thread the SDK construction context into resource creation:

- Change `newResource` in `otelconf/x` to accept a `context.Context`.
- Call it from `NewSDK` with `o.ctx`.
- Ensure any helper that builds detected resources also uses that context when calling `resource.New`.

It may also be worth updating stable `otelconf/resource.go` to accept the SDK construction context for consistency, although stable `otelconf` does not currently expose `resource.detection/development`.

### Additional Context

This came up while reviewing #9131. That PR makes resource detection handling more visible, but the context behavior predates that change and is separate from attribute filtering.

MrAlias:
## Summary

This updates `go.opentelemetry.io/contrib/otelconf` and `go.opentelemetry.io/contrib/otelconf/x` so SDK resource construction uses the context configured with `WithContext`.

Previously, resource construction called `resource.New(context.Background(), ...)`, which meant caller cancellation and deadlines were not propagated to resource detector execution even though other SDK construction paths already used the configured context.

The change threads the SDK construction context into resource creation, covers both stable and experimental otelconf packages for consistency, and adds regression coverage around the context passed to resource builders.

Fixes #9158.

## Testing

- `go test ./...` from `otelconf`
- `make precommit`

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -19,7 +19,7 @@ This project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0.htm
 
 ### Fixed
 
-- Honor the context configured with `WithContext` when constructing resources in `go.opentelemetry.io/contrib/otelconf` and `go.opentelemetry.io/contrib/otelconf/x`. (#TBD)
+- Honor the context configured with `WithContext` when constructing resources in `go.opentelemetry.io/contrib/otelconf` and `go.opentelemetry.io/contrib/otelconf/x`. (#9160)
 
 <!-- Released section -->
 <!-- Don't change this section unless doing release -->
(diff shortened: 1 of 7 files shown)


dashpole (approved):

codeboten (approved):
Thanks @MrAlias

pellared (approved):

PR #9162, opened 2026-06-17
detectors/azure/azurevm: reconcile detected attributes with collector-contrib Azure detector

### LINKED ISSUE #8947: [detectors/azure/azurevm] Audit and reconcile differences with resourcedetectionprocessor azure detector
## Problem Statement

`go-contrib/detectors/azure/azurevm` and `collector-contrib/processor/resourcedetectionprocessor/internal/azure` both implement an Azure VM resource detector, but they have diverged. Two parallel implementations in the same binary can produce inconsistent resource attributes — one seen on the collector's self-telemetry, another on user pipeline telemetry.

## Proposed Solution

Audit all attribute keys, values, semantics, and failure modes across both implementations. Reconcile any gaps in this repo so that `go-contrib/detectors/azure/azurevm` becomes the canonical implementation, and `collector-contrib/processor/resourcedetectionprocessor` can delegate to it without loss of functionality.

## Prior Art

- Existing go-contrib detector: [`detectors/azure/azurevm`](https://github.com/open-telemetry/opentelemetry-go-contrib/tree/main/detectors/azure/azurevm)
- Collector-contrib detector: [`processor/resourcedetectionprocessor/internal/azure`](https://github.com/open-telemetry/opentelemetry-collector-contrib/tree/main/processor/resourcedetectionprocessor/internal/azure)

## Additional Context

Known differences to investigate:

- **Azure resource tags:** collector-contrib supports fetching Azure VM resource tags (regex-filtered via `tags` config) and emits them as `azure.tag.*` attributes. go-contrib has no equivalent.
- **Extra attribute `azure.vm.name`:** collector-contrib sets this in addition to `host.name` (which uses `osProfile.computerName` with a fallback to `name`). go-contrib sets `host.name` only from the VM name field.
- **`host.name` fallback logic:** collector-contrib uses `osProfile.computerName` first, falling back to `name`. go-contrib behaviour should be audited for alignment.
- **Attribute coverage:** both set `cloud.provider`, `cloud.platform`, `cloud.region`, `cloud.account.id`, `host.id`, `host.name`, `host.type`, `os.type`, `azure.vm.sku` — verify values and semconv versions are aligned.

paulojmdias:
### Description

Reconciles the `azurevm` detector with the collector-contrib Azure detector so it emits a superset of both.

Adds `cloud.account.id`, `cloud.availability_zone`, `azure.vm.name`, `azure.vm.size`, `azure.vm.scaleset.name`, and `azure.resourcegroup.name`; `host.name` now prefers `osProfile.computerName` and falls back to the VM name; new `WithTagKeyRegexps` option emits `azure.tag.<name>` for matching VM tags. `New()` stays backward compatible.

### Link to tracking issue

Fixes #8947

### Testing

Extended tests with full-payload attribute assertions, tag filtering (match/drop), computerName fallback, tags-off-by-default, plus existing error cases.

### Documentation

Updated `doc.go` and added a `CHANGELOG.md` entry.


paulojmdias:
\cc @iblancasa

detectors/azure/azurevm/vm.go
```diff
@@ -17,24 +18,69 @@ import (
...
+	Value string `json:"value"`
+}
+
+// Option configures a [ResourceDetector].
+type Option func(*ResourceDetector)
+
+// WithTagKeyRegexps configures the detector to emit an azure.tag.<name>
+// attribute for every VM tag whose key matches any of the provided regular
+// expressions. When no regexps are configured, VM tags are not emitted.
+func WithTagKeyRegexps(res ...*regexp.Regexp) Option {
```
dashpole: I wonder if we should have this take a filter function as input vs regular expressions? Or are there any other things you considered for this option?

iblancasa: I like that idea.

paulojmdias: Changed it to take a filter function instead of regexes. It's more flexible, and you can still get the regex behavior by passing one in. 👍 

Thanks!

iblancasa (approved):

dashpole (approved):
@pyohannes would love a review from you. This PR also proposes additional codeowners who have been maintaining the resource detector in the collector.  Please let me know (here or on slack) if you have concerns.

MrAlias:
Thanks for the contribution, and apologies for the delay here. Since this was opened, #9289 and #9290 established the option and constructor conventions used by the newer Azure resource detectors. Please rebase this PR onto `main` and align it with those conventions:

- Keep `New()` with its existing no-argument signature. Changing it to a variadic function breaks callers that use `azurevm.New` as a function value. Add `NewResourceDetector(opts ...Option)` for the configurable constructor.
- Use the `Option`/config pattern used by the other Azure detectors and add `WithAttributeFilter(attribute.Filter)`. The collector adapter will need this to preserve its `resource_attributes` configuration.
- Keep Azure tag collection opt-in through a separate tag-key filter. Tags should remain disabled unless explicitly requested.
- Use `semconv.AzureResourceGroupName` for `azure.resource_group.name`, which is now available following #9337. The collector currently emits the legacy `azure.resourcegroup.name`, so the PR should also explain how compatibility will be handled when the collector delegates to this detector.
- Add the proposed owners specifically to `detectors/azure/azurevm` rather than changing ownership for the entire `detectors/azure/` tree.
- Update the changelog entry to reference this PR (`#9162`) rather than the tracking issue.

Please request another review once those changes are in.

paulojmdias:
@MrAlias, I addressed everything. Please review when you get a chance 🙏

MrAlias (approved):

PR #9163, opened 2026-06-18
fix(autoprop): return no-op for empty TextMapPropagator input

### LINKED ISSUE #9057: autoprop: TextMapPropagator returns nil for empty input

arpitjain099:
Fixes #9057.

`autoprop.TextMapPropagator` returned a nil propagator when called with no names, even though its docstring says empty input should give back a no-op propagator. Calling `Inject`, `Extract`, or `Fields` on that nil interface panics.

Empty input now returns the no-op propagator with a nil error, which is the same thing `none` already does. Both paths reuse the package-level no-op the registry registers as `none` rather than building a fresh empty composite at each return site, so the two stay visibly equivalent.

Names that all fail to resolve still return `nil` along with the error. That distinction matters: `NewTextMapPropagator` only falls back to its default when the propagator is nil, so handing back a no-op there would turn a typo like `OTEL_PROPAGATORS=invalid` into silently disabled propagation instead of a fallback to TraceContext and Baggage.

On the tests, there's one for empty input returning the no-op with no error, and one for all-unknown input returning nil plus the error. `TestNewTextMapPropagatorInvalidEnvVal` now also asserts the fallback really produces the default propagator fields, since that's what catches the regression above. There's a changelog entry under Unreleased Fixed.

`go test -race ./...` passes in `propagators/autoprop`, and vet and gofmt are clean.


dashpole (approved):

propagators/autoprop/registry_test.go
```diff
@@ -78,3 +79,16 @@ func TestDuplicateRegisterTextMapPropagatorPanics(t *testing.T) {
...
+	p, err := TextMapPropagator()
+	require.NoError(t, err)
+	assert.Equal(t, noop, p)
+
+	carrier := propagation.MapCarrier{}
+	assert.NotPanics(t, func() {
+		p.Inject(context.Background(), carrier)
+		_ = p.Extract(context.Background(), carrier)
+		_ = p.Fields()
+	})
```
pellared: I am not sure what is the value of this

```suggestion
```

propagators/autoprop/registry.go
```diff
@@ -149,7 +149,7 @@ func TextMapPropagator(names ...string) (propagation.TextMapPropagator, error) {
 
 	switch len(props) {
 	case 0:
-		return nil, err
+		return propagation.NewCompositeTextMapPropagator(), err
```
MrAlias: Returning a no-op whenever `props` is empty also changes the all-unknown case. For example, with `OTEL_PROPAGATORS=invalid`, `TextMapPropagator` now returns a non-nil no-op along with the error. `NewTextMapPropagator` logs the error, sees the non-nil propagator, and returns it instead of falling back to the default TraceContext+Baggage propagator. That means a typo in the environment variable disables propagation.

Please keep returning `nil, err` when no valid propagators were selected due to unknown names, and only return the no-op for the actual empty-input or `none` cases.

MrAlias: The empty `propagation.NewCompositeTextMapPropagator()` is the right no-op implementation available from the OTel API; it is also what the OTel global propagator uses internally. Since this package already registers that no-op as `none`, please reuse the same package-level value here instead of constructing a new empty composite at each return site. That keeps the `none` and empty-input behavior visibly equivalent.

MrAlias (changes requested):
Thanks for fixing the nil return path for empty input. Before merge, please address the invalid-environment-value regression noted inline.

Because this is a user-visible bug fix, please also add an Unreleased `Fixed` changelog entry for `go.opentelemetry.io/contrib/propagators/autoprop`.

arpitjain099:
> Thanks for fixing the nil return path for empty input. Before merge, please address the invalid-environment-value regression noted inline.
> 
> Because this is a user-visible bug fix, please also add an Unreleased `Fixed` changelog entry for `go.opentelemetry.io/contrib/propagators/autoprop`.

@MrAlias noted, I will make the changes today

arpitjain099:
Thanks @MrAlias, good catch on the all-unknown regression. I'd flattened that case so OTEL_PROPAGATORS=invalid handed back a non-nil no-op and NewTextMapPropagator returned it, which would silently disable propagation on a typo.

Fixed it so the empty-input case (and `none`) returns the no-op with a nil error, while names that all fail to resolve return `nil, err` again, letting NewTextMapPropagator fall back to the default TraceContext + Baggage. I also reused the package-level no-op the registry already registers as `none` instead of building a fresh composite at each return site, so the two paths stay visibly equivalent.

Added an Unreleased Fixed changelog entry for the autoprop module (#9163).

@pellared you were right that the carrier no-panic block wasn't adding much (it was really exercising the upstream propagation package), so I dropped it. The empty-input test now just asserts the no-op is returned with no error, and I added a separate test for the all-unknown case returning nil + the error. I also strengthened TestNewTextMapPropagatorInvalidEnvVal to assert the fallback to the default propagators, which is what actually catches this regression. `go test ./propagators/autoprop/...` passes (go vet and gofmt clean).

MrAlias (approved):

PR #9197, opened 2026-06-26
processors/minsev: fix Severity text round trip for FATAL offsets

Xio-Shark:
`Severity.String` formats values above `SeverityFatal4` with the same zero-skipping offset convention used by the other severity groups. For example, `SeverityFatal4 + 2` is formatted as `FATAL+6`.

Before this change, parsing `FATAL+6` added the offset directly to `SeverityFatal1`, producing `SeverityFatal4 + 3` instead of the original value. That breaks the documented `MarshalText` / `UnmarshalText` round trip.

This change adjusts positive `FATAL` offsets above the named range during parsing and adds a round-trip test over the existing encoding cases.

Validation:

```bash
cd processors/minsev
go test ./...
```


MrAlias (approved):

ps-mir (approved):

dashpole (approved):

MrAlias:
@Xio-Shark please take a look at and address copilot review. I think this is ready to merge once that is complete (and updated with `main`).

Xio-Shark:
@MrAlias Thanks — Copilot comments are addressed and the branch is up to date with `main`. Ready for merge whenever you have a moment.

Issue #9189, opened 2026-06-24
[Feature]: otelgrpc: Allow setting OTel statuses based on gRPC status and path/payload


bendiknesbo:
### Component

Instrumentation: otelgrpc

### Problem Statement

The StatHandlers for the gRPC instrumentation use a small [hard-coded mapping](https://github.com/open-telemetry/opentelemetry-go-contrib/blob/main/instrumentation/google.golang.org/grpc/otelgrpc/interceptor.go#L77-L98) for the server-handler, and [always a `codes.Error`](https://github.com/open-telemetry/opentelemetry-go-contrib/blob/main/instrumentation/google.golang.org/grpc/otelgrpc/stats_handler.go#L304-L306) for the client-handler.

As far as I understand, we can call the `Span.SetStatus` manually inside each handler or our own interceptor, but this adds another place where we **have** to extract span-information and add custom handling. Having everything related to setting the span status in once place would reduce the amount of places we have to add custom instrumentation.

### Proposed Solution

Add an optional consumer-provided function that runs in `*config.handleRPC`, which allows the consumer to introspect some parts (like the gRPC method name and returned gRPC status-code) and then return an OTel `codes.Code`. This should still adhere to the [semantic conventions for gRPC](https://opentelemetry.io/docs/specs/semconv/rpc/grpc/), with emphasis on SHOULD instead of MUST:

Client: 
> - rpc.response.status_code: All status codes except OK **SHOULD** be considered errors.

Server:
>rpc.response.status_code: The following status codes **SHOULD** be considered errors:
> - UNKNOWN
> - DEADLINE_EXCEEDED
> - UNIMPLEMENTED
> - INTERNAL
> - UNAVAILABLE
> - DATA_LOSS

### Alternatives

Document how to set whether or not a span should be an error, and in which cases the instrumentation will override the status set by the consumer.

### Prior Art

Datadog's `dd-trace-go` has a map of gRPC status code which should not be treated as an error:
https://github.com/DataDog/dd-trace-go/blob/main/contrib/google.golang.org/grpc/grpc.go#L88-L90

DCchoudhury15:
@bendiknesbo can i work on this issue?

bendiknesbo:
[...]

PR #9205, opened 2026-06-28
instrumentation/otelgrpc: add WithSpanStatusFn for custom span status

DCchoudhury15:
## Description

Implements #9189.

The `otelgrpc` stats handlers currently use a fixed mapping from gRPC status codes to OpenTelemetry span status:

* **Server:** only `UNKNOWN`, `DEADLINE_EXCEEDED`, `UNIMPLEMENTED`, `INTERNAL`, `UNAVAILABLE`, and `DATA_LOSS` are marked as span errors.
* **Client:** every non-`OK` status is marked as an error.

This makes it difficult to customize span status for specific RPCs or status codes without adding a separate interceptor.

## Changes

This PR adds a `SpanStatusFn` type along with a `WithSpanStatusFn` option for both `NewServerHandler` and `NewClientHandler`.

When provided, the callback replaces the default status mapping. It receives:

* The full gRPC method name (for example, `/package.Service/Method`)
* The parsed `*status.Status` (`nil` when the RPC completed successfully)

The function returns the OpenTelemetry `codes.Code` and description that should be applied to the span.

This follows the gRPC semantic conventions, which describe span status classification as a **SHOULD**, allowing applications to override the default behaviour when it better matches their use case.

### Example

```go
// Treat NotFound as a non-error for lookup RPCs only.
otelgrpc.NewServerHandler(
    otelgrpc.WithSpanStatusFn(func(_ context.Context, fullMethod string, grpcStatus *status.Status) (codes.Code, string) {
        if grpcStatus != nil && grpcStatus.Code() == grpc_codes.NotFound {
            return otelcodes.Unset, ""
        }

        switch grpcStatus.Code() {
        case grpc_codes.Unknown, grpc_codes.DeadlineExceeded, grpc_codes.Unimplemented,
            grpc_codes.Internal, grpc_codes.Unavailable, grpc_codes.DataLoss:
            return otelcodes.Error, grpcStatus.Message()
        }

        return otelcodes.Unset, ""
    }),
)
```

## Testing

Added unit tests in `config_test.go` covering:

* Option wiring
* Nil callback guard
* Argument passthrough

Added integration tests in `grpc_stats_handler_test.go` covering:

* Server: overriding the default behaviour to treat `NotFound` as an error
* Server: suppressing `Internal` for a specific RPC method
* Client: treating `NotFound` as a non-error instead of the default error behaviour
* Default client and server behaviour when no custom callback is configured

Verified with:

```bash
go test -race ./...
```

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -10,6 +10,7 @@ This project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0.htm
 
 ### Added
 
+- Add `WithSpanStatusFn` option for `NewServerHandler` and `NewClientHandler` in `go.opentelemetry.io/contrib/instrumentation/google.golang.org/grpc/otelgrpc` to allow per-method override of the default gRPC → OTel span status mapping. (#9189)
 - The resource created by `go.opentelemetry.io/contrib/otelconf` now includes [default SDK attributes](https://pkg.go.dev/go.opentelemetry.io/otel/sdk/resource#Default). (#8990)
 - Add `azurecontainerapps` resource detector for Azure Container Apps. (#8939)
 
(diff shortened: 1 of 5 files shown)


dashpole:
Lets discuss on the issue first.

Issue #8984, opened 2026-05-12
[Feature]: Add Azure Container Apps resource detector


kathiehuang:
### Component

Detector: GCP

### Problem Statement

***Component should be "Detector: Azure" but it wasn't an option**.

The [Azure resourcedetectionprocessor](https://github.com/open-telemetry/opentelemetry-collector-contrib/tree/main/processor/resourcedetectionprocessor/internal/azure) doesn't support Azure Container Apps. It queries the Azure Instance Metadata Service, which doesn't work for a handful of Azure workloads.

There are detectors for Azure Container Apps in the other runtime SDKs, but not in Go. Currently, [there is a migration](https://github.com/open-telemetry/opentelemetry-go-contrib/issues/8940) where the `resourcedetectionprocessor` in `opentelemetry-collector-contrib` will begin to delegate to the detectors in `opentelemetry-go-contrib`. As a result, adding support in the Go SDK will also enable support in the `resourcedetectionprocessor`, making resource detection work consistently in Azure Container Apps for all runtimes without requiring application code changes.

### Proposed Solution

Azure Container Apps injects platform metadata as [built-in environment variables](https://learn.microsoft.com/en-us/azure/container-apps/environment-variables?tabs=portal#built-in-environment-variables) at runtime. This issue tracks adding a resource detector that reads those variables and emits standard OTel resource attributes.

## What it detects

The detector gates on `CONTAINER_APP_NAME`. If that variable is not set, it returns an empty resource. When set, it emits:

| Source                       | Resource Attribute                        |
| ---------------------------- | ----------------------------------------- |
| *(constant)*                 | `cloud.provider = "azure"`                |
| *(constant)*                 | `cloud.platform = "azure.container_apps"` |
| `CONTAINER_APP_NAME`         | `service.name`                            |
| `CONTAINER_APP_REPLICA_NAME` | `service.instance.id`                     |

Other resource attributes that could be added for other Azure workloads (standard Azure resource information):
`cloud.account.id`: Azure subscription ID
`azure.resourcegroup.name`: Azure resource group

### Prior Art

- [azurevm implementation](https://github.com/open-telemetry/opentelemetry-go-contrib/tree/main/detectors/azure/azurevm)
- [lambda implementation](https://github.com/open-telemetry/opentelemetry-go-contrib/tree/main/detectors/aws/lambda) (which also uses env vars for detecting information)
[...]

PR #8939, opened 2026-05-08
feat: add Azure Container Apps resource detector

kathiehuang:
Adds a new resource detector for Azure Container Apps.

Azure Container Apps [injects platform metadata as environment variables](https://learn.microsoft.com/en-us/azure/container-apps/environment-variables?tabs=portal#built-in-environment-variables). This detector reads those variables and emits standard OTel resource attributes:

| Environment Variable         | Resource Attribute | Example Value |
| ---------------------------- | --- | --- |
| *(constant)*                 | `cloud.provider` | azure
| *(constant)*                 |`cloud.platform` | azure.container_apps
| `CONTAINER_APP_NAME`         |`service.name` | my-app
| `CONTAINER_APP_REPLICA_NAME` |`service.instance.id` | my-app--abc123-0

Related PR in resource detection processor: [[processor/resourcedetection] Add Azure Container Apps resource detector](https://github.com/open-telemetry/opentelemetry-collector-contrib/pull/48209)

## Testing
Tested this as a local Go module. Deployed an Azure Container App and confirmed that the expected resource attributes are emitted in metric payload.

Tracking issue: https://github.com/open-telemetry/opentelemetry-go-contrib/issues/8984

PR #9208, opened 2026-06-29
fix(detectors/azurecontainerapps): use azure.container_app.instance.id for replica name instead of service.instance.id

kathiehuang:
Changes the resource attribute used for the Azure Container Apps replica name from `service.instance.id` to `azure.container_app.instance.id`.

The Go SDK's default resource detector auto-populates `service.instance.id` [with a random UUID](https://github.com/open-telemetry/opentelemetry-go/blob/18ef80827fc3965a41742417b6937c370f8d9817/sdk/resource/builtin.go#L108), which would conflict with the Azure Container App detector's replica name value for Go users. `azure.container_app.instance.id` avoids that collision, is more precise, and is consistent with the precedent set by the [Node.js SDK](https://github.com/open-telemetry/opentelemetry-js-contrib/tree/main/packages/resource-detector-azure#readme). A follow-up PR will be made in the semantic-conventions repo to make this a semantic convention.

This was raised during review of the collector-contrib adapter PR for this detector: [discussion 1](https://github.com/open-telemetry/opentelemetry-collector-contrib/pull/49342#discussion_r3492746358), [discussion 2](https://github.com/open-telemetry/opentelemetry-go-contrib/pull/8939#issuecomment-4835231076)

Link to tracking issue: https://github.com/open-telemetry/opentelemetry-go-contrib/issues/8984

Follow up to PR #8939

## Testing
Updated existing tests to use faas.instance. All tests pass.

## Documentation
Updated doc.go to reference faas.instance and link to the FaaS semantic conventions.

### DIFF
--- a/detectors/azure/azurecontainerapps/detector.go
+++ b/detectors/azure/azurecontainerapps/detector.go
@@ -71,10 +71,13 @@ func (d *ResourceDetector) Detect(context.Context) (*resource.Resource, error) {
 		semconv.ServiceName(appName),
 	}
 
+	// azure.container_app.instance.id is a platform-specific attribute for the replica name
+	// as there lacks an appropriate attribute in semantic conventions. This is planned
+	// to be added to semantic conventions.
 	var partialErr error
 	replicaName := os.Getenv(containerAppReplicaNameEnvVar)
 	if replicaName != "" {
-		attrs = append(attrs, semconv.FaaSInstance(replicaName))
+		attrs = append(attrs, attribute.String("azure.container_app.instance.id", replicaName))
 	} else {
 		partialErr = fmt.Errorf("%w: %s not set", resource.ErrPartialResource, containerAppReplicaNameEnvVar)
 	}
--- a/detectors/azure/azurecontainerapps/detector.go
+++ b/detectors/azure/azurecontainerapps/detector.go
@@ -74,7 +74,7 @@ func (d *ResourceDetector) Detect(context.Context) (*resource.Resource, error) {
 	var partialErr error
 	replicaName := os.Getenv(containerAppReplicaNameEnvVar)
 	if replicaName != "" {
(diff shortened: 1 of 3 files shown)


kathiehuang:
Should we add the "Skip Changelog" label?

dashpole:
Added, and added to the milestone. What are your thoughts on faas.name vs service.name?

kathiehuang:
@dashpole Azure Container Apps run as long-lived services rather than functions, and every runtime SDK uses `service.name`, so `faas.*` feels like a slight semantic mismatch. It also seems that the default `service.name` population [is designed to be overridden](https://github.com/open-telemetry/opentelemetry-go/blob/18ef80827fc3965a41742417b6937c370f8d9817/sdk/resource/builtin.go#L97) since it puts `unknown_service:...`. And it looks like for users using the resourcedetectionprocessor, [override defaults to true](https://github.com/open-telemetry/opentelemetry-collector-contrib/blob/277638be46c8653cc6574f2f9a6d2597bd5c12da/processor/resourcedetectionprocessor/config.go#L47).

It looks like Google Cloud Run uses `faas.name` too though. What's your preference?

kathiehuang:
cc @iblancasa @paulojmdias

kathiehuang:
@dashpole I raised this at the Go SIG meeting today and since Azure Container Apps aren't technically Function-as-a-Service, the consensus was to use a more precise platform-specific attribute (`azure.container_app.instance.id`, following the same pattern as the [Node.js SDK](https://github.com/open-telemetry/opentelemetry-js-contrib/tree/main/packages/resource-detector-azure#readme)) for now, with the intent to propose it as a semantic convention later.

dashpole:
sounds good.  Thanks for following up

dashpole (approved):

XSAM (approved):

paulojmdias (approved):

iblancasa (approved):

PR #9209, opened 2026-06-30
instrumentation/otelgrpc: add WithNonErrorCodes to suppress error span status for specific gRPC codes

bendiknesbo:
## Description

Implements #9189. Different alternative to https://github.com/open-telemetry/opentelemetry-go-contrib/pull/9205

The `otelgrpc` stats handlers currently use a fixed mapping from gRPC status codes to OpenTelemetry span status:

* **Server:** only `UNKNOWN`, `DEADLINE_EXCEEDED`, `UNIMPLEMENTED`, `INTERNAL`, `UNAVAILABLE`, and `DATA_LOSS` are marked as span errors.
* **Client:** every non-`OK` status is marked as an error.

This makes it difficult to customize span status for specific RPCs or status codes without adding a separate interceptor.

## Changes

This PR adds a `WithNonErrorCodes` option for both `NewServerHandler` and `NewClientHandler` that prevents the handler from setting an error span status for the provided gRPC codes. The codes apply handler-wide to every RPC. An existing span status set by the application is left unchanged.

This follows the gRPC semantic conventions, which describe span status classification as a **SHOULD**, allowing applications to override the default behaviour when it better matches their use case.

### Example

```go
// Treat NotFound as a non-error status-code.
otelgrpc.NewServerHandler(
   	otelgrpc.WithNonErrorCodes(codes.NotFound),
)
```

## Testing

Added unit tests in `config_test.go` covering:

* Option wiring
* Map ownership
* Empty values guard

Added integration tests in `grpc_stats_handler_test.go` covering:

* Server: overriding the default behaviour to treat `Internal` as Unset (Instead of Error)
* Server: does nothing to an application-set Error status
* Server: does nothing to an application-set Ok status
* Client: overriding the default behaviour to treat `NotFound` as Unset (Instead of Error)
* Default client and server behaviour when no custom callback is configured
* Client: does nothing to an application-set Error status
* Client: does nothing to an application-set Ok status

Verified with:

```bash
go test -race ./...
```


instrumentation/google.golang.org/grpc/otelgrpc/config.go
```diff
@@ -244,3 +246,14 @@ func WithMetricAttributesFn(fn func(ctx context.Context) []attribute.KeyValue) O
...
 	})
 }
+
+// WithNonErrorCodes sets a custom map of gRPC codes that should be treated as non-errors.
+// This map is used to determine the span status code for a gRPC call.
+// When set, this map is used in addition to the default status mapping defined by the OpenTelemetry gRPC semantic conventions.
+func WithNonErrorCodes(nonErrorCodes map[grpc_codes.Code]struct{}) Option {
+	return optionFunc(func(c *config) {
+		if len(nonErrorCodes) > 0 {
+			c.NonErrorCodes = nonErrorCodes
```
dashpole: Copy nonErrorCodes to prevent data races if the caller modifies the map after passing it.
```suggestion
		if len(nonErrorCodes) > 0 {
			c.NonErrorCodes = make(map[grpc_codes.Code]struct{}, len(nonErrorCodes))
			for k, v := range nonErrorCodes {
				c.NonErrorCodes[k] = v
			}
		}
```

MrAlias: The handler keeps this caller-owned map and reads it concurrently for every RPC. If the caller reuses or updates the map after constructing the handler, existing handler behavior changes and a concurrent update can panic with `concurrent map read and map write`.

Please clone the map while applying the option, or accept variadic codes and build an owned map. Add a test that mutations to the input do not affect the config.

instrumentation/google.golang.org/grpc/otelgrpc/stats_handler.go
```diff
@@ -384,6 +384,12 @@ func (*config) handleRPC(
 		}
 		if span.IsRecording() {
 			if s != nil {
+				if c.NonErrorCodes != nil {
+					if _, ok := c.NonErrorCodes[s.Code()]; ok {
+						// Override the status code to OK for non-error codes. This will take priority over any later span.SetStatus calls, including the default behavior below.
+						span.SetStatus(codes.Ok, "")
```
MrAlias: Setting `Ok` here is stronger than skipping the default mapping: OpenTelemetry defines `Ok` as final, so it overrides an `Error` already set by an interceptor or application code. The public option currently only says these codes are non-errors.

Please either document that this option deliberately replaces application-set errors and add that regression test, or leave the status unset when only the default gRPC classification should be suppressed.

instrumentation/google.golang.org/grpc/otelgrpc/grpc_stats_handler_test.go
```diff
@@ -809,3 +810,79 @@ func TestMetricsSemconvOptIn(t *testing.T) {
...
 	}
 }
+
+func TestNonErrorCodes(t *testing.T) {
+	end := func(err error) *stats.End {
+		now := time.Now()
+		return &stats.End{Error: err, BeginTime: now, EndTime: now}
+	}
+
+	t.Run("server: promotes NotFound to OK (overrides default unset)", func(t *testing.T) {
```
MrAlias: `NotFound` is already `Unset` under the server mapping, so this case only tests promotion from one non-error status to another. It does not protect the advertised server override.

Please add a server case for one of the default errors, such as `Internal` or `Unknown`, and assert that it becomes `Ok`.

CHANGELOG.md
```diff
@@ -10,6 +10,7 @@ This project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0.htm
 
 ### Added
 
+- Add `WithNonErrorCodes` option for `NewServerHandler` and `NewClientHandler` in `go.opentelemetry.io/contrib/instrumentation/google.golang.org/grpc/otelgrpc` to allow extra gRPC → OTel span status mapping. (#9189)
```
MrAlias: The changelog must reference this PR, not the issue.

```suggestion
- Add `WithNonErrorCodes` option for `NewServerHandler` and `NewClientHandler` in `go.opentelemetry.io/contrib/instrumentation/google.golang.org/grpc/otelgrpc` to allow extra gRPC → OTel span status mapping. (#9209)
```

MrAlias (changes requested):
The option is useful, but its configuration ownership and final status-precedence contract need to be addressed before this is safe to merge.

instrumentation/google.golang.org/grpc/otelgrpc/stats_handler.go
bendiknesbo: What is the wanted behaviour? To set `Ok`, or to leave it as `Unset`?

instrumentation/google.golang.org/grpc/otelgrpc/config.go
bendiknesbo: Great find. I changed the function to be a variadic function: https://github.com/open-telemetry/opentelemetry-go-contrib/pull/9209/changes/07e1caf1a6bf48e7b84c71b3e2dd3c30af3a1b32

instrumentation/google.golang.org/grpc/otelgrpc/grpc_stats_handler_test.go
bendiknesbo: I have added a case for `Internal` becoming `Ok` instead of `Error`, but I have left the `NotFound` as-is.
I'm open for suggestions for other test-names, but I believe the test-code as it stands today is valuable, so that we test both `Unset`->`Ok` and `Error`->`Ok`

instrumentation/google.golang.org/grpc/otelgrpc/config_test.go
```diff
@@ -60,3 +61,38 @@ func TestParseSemconvMode(t *testing.T) {
...
+		assert.Contains(t, c.NonErrorCodes, grpc_codes.NotFound)
+	})
+
+	t.Run("mutation of input-slice does not affect config", func(t *testing.T) {
+		slice := []grpc_codes.Code{grpc_codes.NotFound}
+		c := newConfig([]Option{WithNonErrorCodes(slice...)})
+		assert.Len(t, c.NonErrorCodes, 1)
+		assert.Contains(t, c.NonErrorCodes, grpc_codes.NotFound)
+
+		_ = append(slice, grpc_codes.Internal)
```
MrAlias: This append result is discarded, so the caller's slice never changes and the test cannot catch retaining it. Mutating the existing element exercises the ownership guarantee.

```suggestion
		slice[0] = grpc_codes.Internal
```

instrumentation/google.golang.org/grpc/otelgrpc/config.go
```diff
@@ -244,3 +246,17 @@ func WithMetricAttributesFn(fn func(ctx context.Context) []attribute.KeyValue) O
 		}
 	})
 }
+
+// WithNonErrorCodes sets a custom map of gRPC codes that should be treated as non-errors.
```
MrAlias: This option applies to every RPC processed by the handler; it cannot make `NotFound` a non-error for only the lookup RPC shown in the PR description. The description also still passes the removed map-shaped argument. Could we update both the GoDoc and PR example to match the variadic, handler-wide API?

MrAlias (commented):
The production ownership fix is sound. I would leave configured non-error statuses unset so an error already set by server application or interceptor code remains authoritative; the current `Ok` behavior still needs to change before this is ready.

instrumentation/google.golang.org/grpc/otelgrpc/stats_handler.go
bendiknesbo: I have rewritten this part, to treat the `NonErrorCodes` as `Unset` instead of `Ok`

instrumentation/google.golang.org/grpc/otelgrpc/config_test.go
bendiknesbo: Thanks, fixed now.

instrumentation/google.golang.org/grpc/otelgrpc/config.go
bendiknesbo: Updated the GoDoc comment and PR body

bendiknesbo:
Thanks for the very good feedback @MrAlias , I think I have addressed every point, and the PR should be ready to merge now, if you agree :)

instrumentation/google.golang.org/grpc/otelgrpc/stats_handler.go
dashpole: I think we want to not set it, rather than set it to `Unset`. We don't want to clear application-set errors. i.e. it should just suppress the setting of `Error` status, not actively change the status.

bendiknesbo: Setting `Unset` is the current behaviour for e.g. `NotFound` code today.

This is the code for the `SetStatus` function in the `Span` interface:
```go
	// SetStatus sets the status of the Span in the form of a code and a
	// description, provided the status hasn't already been set to a higher
	// value before (OK > Error > Unset). The description is only included in a
	// status when the code is for an error.
	SetStatus(code codes.Code, description string)
```

So setting `Unset` should be fine. I have already added a unit-test that tests that we don't end up clearing application-set errors (or `Ok`)

instrumentation/google.golang.org/grpc/otelgrpc/config.go
```diff
@@ -244,3 +246,17 @@ func WithMetricAttributesFn(fn func(ctx context.Context) []attribute.KeyValue) O
 		}
 	})
 }
+
+// WithNonErrorCodes accepts a set of gRPC codes that should be treated as non-errors.
+// These codes are used to determine the span status code for a gRPC call.
+// When set, these codes are used to prevent the SDK from treating them as errors.
```
MrAlias: The SDK does not classify gRPC status codes; this option only suppresses the stats handlers' default `Error` mapping, and `SetStatus(Unset, "")` leaves application-set statuses intact. The current GoDoc promises a stronger final classification and still does not say the codes apply handler-wide. Could we document that contract directly? The PR description should match it and also fix the extra `}` in the example and the stale client `Ok` test description.

```suggestion
// WithNonErrorCodes returns an Option that prevents [NewClientHandler] and
// [NewServerHandler] from setting an error span status for the provided gRPC
// codes. The codes apply to every RPC. An existing span status set by the
// application is left unchanged.
```

CHANGELOG.md
```diff
@@ -69,7 +70,7 @@ This project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0.htm
 - Fix header attributes lost when using sub-spans in `go.opentelemetry.io/contrib/instrumentation/net/http/httptrace/otelhttptrace`. (#8797)
 - Validate `encoding` configuration for OTLP HTTP exporters in `go.opentelemetry.io/contrib/otelconf`. (#8772)
 - Remove the custom body wrapper from the request's body after the request is processed to allow body type comparisons with the original type in `go.opentelemetry.io/contrib/instrumentation/net/http/otelhttp` and `go.opentelemetry.io/contrib/instrumentation/github.com/gorilla/mux/otelmux`. (#6914)
-- Unknown or empty HTTP methods now report "_OTHER" instead of "GET" across all HTTP instrumentations to align with OpenTelemetry semantic conventions. (#8868)
+- Unknown or empty HTTP methods now report "\_OTHER" instead of "GET" across all HTTP instrumentations to align with OpenTelemetry semantic conventions. (#8868)
```
MrAlias: This PR only needs the new Unreleased entry, but this commit also rewrites 18 lines in released sections below the `Don't change this section unless doing release` marker. Could we revert the unrelated historical changelog formatting and keep only the entry for #9209?

MrAlias (commented):
The status-precedence behavior now looks sound. The public contract and changelog scope still need cleanup before this is ready.

CHANGELOG.md
bendiknesbo: I'm terribly sorry, I accidentally save the changelog, which triggered Prettier to format it 🤦🏼  I have reverted the accidental changes to the formatting

instrumentation/google.golang.org/grpc/otelgrpc/config.go
bendiknesbo: I have accepted your suggestion, and done a rewording of the PR description and changelog-entry. I have fixed the syntax in the example. I have updated the stale Client `Ok` test description.

MrAlias (approved):

MrAlias (changes requested):
Blocking until we get resolution on https://github.com/open-telemetry/opentelemetry-go-contrib/issues/9378

ps-mir:
@MrAlias 

Based on SIG discussion and context from #9378, the PR is already based on approach 3, and will additionally need to include documentation on setting span status via stats.Handler.

Documentation Context: https://github.com/open-telemetry/opentelemetry-go-contrib/issues/9378#issuecomment-5218038831 by @dashpole

instrumentation/google.golang.org/grpc/otelgrpc/config.go
```diff
@@ -244,3 +246,18 @@ func WithMetricAttributesFn(fn func(ctx context.Context) []attribute.KeyValue) O
...
 }
+
+// WithNonErrorCodes returns an Option that prevents [NewClientHandler] and
+// [NewServerHandler] from setting an error span status for the provided gRPC
+// codes. The codes apply to every RPC. An existing span status set by the
+// application is left unchanged.
+func WithNonErrorCodes(nonErrorCodes ...grpc_codes.Code) Option {
+	return optionFunc(func(c *config) {
+		if len(nonErrorCodes) > 0 {
+			c.NonErrorCodes = make(map[grpc_codes.Code]struct{})
```
dashpole, line 257: I would expect WithNonErrorCodes(foo), WithNonErrorCodes(bar) to behave the same as `WithNonErrorCodes(foo, bar)`.  Should we initialize NonErrorCodes as an empty map in newConfig, rather than here?  That would also allow us to remove the nil check in stats_handler.go

PR #9268, opened 2026-07-10
Remove vanity import tooling

MrAlias:
Supersedes #9264.

This follows the same cleanup direction as open-telemetry/opentelemetry-go#8585.

## Summary

- Remove canonical import path comments from contrib packages and generated bridge output.
- Remove porto and the vanity-import check/fix targets from the Makefile and CI workflow.
- Drop porto from the tools module now that golangci-lint modernize is enabled for linting.
- Run `for d in $( find -type f -name 'go.mod' -exec dirname {} \; ); do (cd $d && go run golang.org/x/tools/go/analysis/passes/modernize/cmd/modernize@latest -fix ./...); done`

## Motivation

Canonical import path comments are legacy package annotations that conflict with modernize expectations in module-mode repositories. Keeping them also requires a separate porto-based check and generation rewrite path, which adds tooling that duplicates the linting workflow.

## Validation

- Ran golangci-lint v2.12.2 with only the modernize linter enabled across all Go modules.
- Ran make precommit.


dashpole (approved):

pellared (approved):

samplers/jaegerremote/sampler_remote.go
```diff
@@ -71,7 +71,7 @@ type samplerUpdater interface {
 type Sampler struct {
 	// These fields must be first in the struct because `sync/atomic` expects 64-bit alignment.
 	// Cf. https://github.com/jaegertracing/jaeger-client-go/issues/155, https://pkg.go.dev/sync/atomic#pkg-note-BUG
-	closed int64 // 0 - not closed, 1 - closed
+	closed atomic.Int64 // 0 - not closed, 1 - closed
```
pellared, line 74: This does more than just removes the vanity import but I like it

MrAlias: It was a part of the modernize fix that I figured was worth including.

PR #9272, opened 2026-07-11
docs: otelmongo: fix OTEL_SEMCONV_STABILITY_OPT_IN behaviour docs

### LINKED ISSUE #9207: [Docs Bug]: otelmongo discrepancy between documentation and component behaviour about semconv version
### Component

Instrumentation: otelmongo

As per https://github.com/open-telemetry/opentelemetry-go-contrib/blob/main/instrumentation/go.mongodb.org/mongo-driver/mongo/otelmongo/doc.go#L16-L21, when `OTEL_SEMCONV_STABILITY_OPT_IN` env var is unset, this package should emit `v1.21.0` semconv by default.
This is contradictory with the code at https://github.com/open-telemetry/opentelemetry-go-contrib/blob/main/instrumentation/go.mongodb.org/mongo-driver/mongo/otelmongo/internal/semconv/event_monitor.go#L90. 

Current behaviour is that 
OTEL_SEMCONV_STABILITY_OPT_IN = "database/dup" -> emit both v1.21.0 and v1.42.0 (latest)
else: emit semconv v1.42.0 (latest)

### Expected behavior

Either docs should be updated to reflect latest code behaviour or code needs to be updated as per the docs.
Any one should be fine.

Seems like this has been the case since long time so I'm not sure what appropriate change would be for past releases.

### Steps to Reproduce

Checkout out TestCommandStartedTraceAttrs. This doesn't require anything else for repro.

TestCommandStartedTraceAttrs
- no version and unsupported version expects v1260 
- database/dup and mixed categories expect both v1210, v1260

This probably means OTEL_SEMCONV_STABILITY_OPT_IN = "database/dup" is required for v1210 compatibilty.

### Operating System

MacOS

### Device Architecture

ARM64

### Go Version

1.26

### Component Version

otelmongo@v0.69.0

mahendrabishnoi2:
Fixes https://github.com/open-telemetry/opentelemetry-go-contrib/issues/9207

Updated docs to reflect current OTEL_SEMCONV_STABILITY_OPT_IN behaviour.

### DIFF
--- a/instrumentation/go.mongodb.org/mongo-driver/mongo/otelmongo/doc.go
+++ b/instrumentation/go.mongodb.org/mongo-driver/mongo/otelmongo/doc.go
@@ -14,11 +14,11 @@
 //   - https://github.com/DataDog/dd-trace-go/tree/v1.23.3/ddtrace/ext
 //
 // The "OTEL_SEMCONV_STABILITY_OPT_IN" environment variable can be used to opt
-// into the latest semantic conventions:
-//   - "database": emit the latest semantic conventions
-//   - "": emit v1.21.0 (default) semantic conventions
-//   - "database/dup": emit v1.21.0 (default) and the latest semantic
+// into emitting the previous (v1.21.0) semantic conventions alongside the
+// current stable ones:
+//   - "": (default) emit the latest stable semantic conventions
+//   - "database/dup": emit both v1.21.0 and the latest stable semantic
 //     conventions
 //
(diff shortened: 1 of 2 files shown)


mahendrabishnoi2:
Not sure if this requires a changelog entry. For now, I didn't add one.
Let me know if its required else please add `Skip Changelog` label.

prestonvasquez (approved):

pellared (approved):

PR #9273, opened 2026-07-11
propagators/b3: report `b3` header from Fields() for default encoding

### LINKED ISSUE #9058: b3: B3 Fields() no longer matches default single-header injection
### Description

B3 propagation has inconsistent header reporting for unspecified/default injection encoding. `Inject()` treats `B3Unspecified` as single-header B3 and writes `b3`, while `Fields()` reports only `x-b3-*` fields and omits `b3`. Middleware that relies on `Fields()` to whitelist, copy, preallocate, or clear propagation headers can drop the injected `b3` header or fail to clear stale `b3` values. 

Inject() treats `B3Unspecified` as single-header and sets the `b3` header (b3_propagator.go:107-123), while Fields() omits `b3` unless B3SingleHeader is explicitly set and for `B3Unspecified` only returns the multi-header keys (b3_propagator.go:175-181). This validates the reported functional regression: default injection writes `b3`, but Fields() does not list it, so header whitelists/clearers based on Fields() will omit or fail to clear the `b3` header.

### Steps To Reproduce

[25-9435b774ac188191b95c8983d0859a26-b3-fields-no-longer-matches-default-single-header-injection.zip](https://github.com/user-attachments/files/28385868/25-9435b774ac188191b95c8983d0859a26-b3-fields-no-longer-matches-default-single-header-injection.zip)

1. Create or use a B3 propagator whose `InjectEncoding` is `B3Unspecified`.
2. Call `Inject()` with a valid span context and text-map carrier.
3. Observe that `Inject()` sets the single `b3` header.
4. Call `Fields()` on the same propagator.
5. Observe that `Fields()` returns only multi-header names and omits `b3`.

### Expected behavior

`Fields()` should include every header that `Inject()` may write for the configured encoding. For `B3Unspecified`/default single-header injection, `Fields()` should include `b3`.

ATKasem:
Fixes #9058

## Problem

The B3 propagator's `Inject` and `Fields` disagree about which headers the default (`B3Unspecified`) encoding uses. `config` documents that `B3Unspecified` defaults to single-header injection:

```go
// If no encoding is specified (i.e. `B3Unspecified`) `B3SingleHeader` will be used as the default.
```

Accordingly, `Inject` writes the single `b3` header for `B3Unspecified`:

```go
if b3.cfg.InjectEncoding.supports(B3SingleHeader) || b3.cfg.InjectEncoding == B3Unspecified { // writes `b3`
...
if b3.cfg.InjectEncoding.supports(B3MultipleHeader) { // writes x-b3-*
```

But `Fields` attached the `|| == B3Unspecified` clause to the **multi**-header branch instead of the single-header one, so for `B3Unspecified` it reported `x-b3-*` and omitted `b3` — exactly the inverse of what `Inject` writes. Middleware that relies on `Fields()` to whitelist, copy, preallocate, or clear propagation headers would drop or fail to clear the `b3` header.

## Fix

Align `Fields`'s encoding conditions with `Inject`'s: report `b3` for `B3SingleHeader`/`B3Unspecified`, and `x-b3-*` for `B3MultipleHeader`.

## Testing

- Updated the "no encoding specified" case in the existing `Fields` table test (`TestB3Propagator_Fields`), which had encoded the old (inverted) behavior. The updated test fails against the previous `Fields` implementation and passes with this change.
- `go test ./...`, `go vet ./...`, and `gofmt` are clean for `propagators/b3`.


MrAlias (approved):

MrAlias:
Please address the CI failures.

ATKasem:
Thanks @MrAlias — done, and rebased onto main to clear the conflict.

The failures were test expectations that still assumed multi-header injection for the default `B3Unspecified` encoding. Updated in three places:

- `otelconf/propagator_test.go` — `valid b3`, `multiple propagators`, `multiple propagators via composite_list` (this is what CI reported)
- `otelconf/x/propagator_test.go` — same three cases
- `propagators/autoprop/example_test.go` — three `// Output:` blocks

CI stopped at `otelconf` before reaching `autoprop`, so those three weren't in the failure list yet but would have broken next. `valid b3multi` is unchanged — it explicitly selects multi-header encoding and still reports the four `x-b3-*` headers.

Verified green locally across every module that depends on `propagators/b3`: `otelconf`, `propagators/autoprop`, `propagators/b3`, `otelgin`, `otelecho`, `otelrestful`.

dashpole (approved):

PR #9274, opened 2026-07-13
fix(otelmongo): strip connection number from connection ID before deriving network.peer.address

AMDaban:
## Problem

`peerInfo` passes the driver's connection ID straight to `net.SplitHostPort`. The v2 driver formats that ID as `<address>[-<connection number>]` ([`topology.newConnection`](https://github.com/mongodb/mongo-go-driver/blob/master/x/mongo/driver/topology/connection.go)), for example `127.0.0.1:27017[-13]`.

`net.SplitHostPort` rejects the trailing `]` (`unexpected ']' in address`), so `peerInfo` falls into its fallback branch and uses the **entire connection ID** as the hostname:

```go
hostname, portStr, err := net.SplitHostPort(connectionID)
if err != nil {
	// If parsing fails, assume default MongoDB port and return the entire ConnectionID as hostname
	hostname = connectionID
	port = defaultMongoPort
	return hostname, port
}
```

The connection number is unique per pooled connection, so `network.peer.address` takes a new value for **every connection the pool opens**. On spans this is merely noisy, but `db.client.operation.duration` is recorded with the same attribute set, so the metric's cardinality grows without bound for the lifetime of the process.

## Impact

This took down monitoring for our service in production. A single pod reached:

- **5,305** distinct `network.peer.address` values (e.g. `10.0.0.1:27017[-1708]`)
- **54,570** `db_client_operation_duration_seconds` bucket series — ~96% of all series the process exported
- a **24.7 MB** `/metrics` payload

which exceeded the scrape size limit, so the scrape was rejected **outright** — and *every* metric from the process disappeared, not just the MongoDB ones. It came back on restart (fresh registry) and then grew back over the following hours.

## Fix

Strip the connection number before parsing, so the attribute becomes the MongoDB host and stays bounded by the size of the replica set.

The bracket is only trimmed when it opens **after the final colon**, which leaves bracketed IPv6 literals such as `[::1]` intact — the existing `IPv6 without port with square brackets` test case covers that and still passes.

## Tests

Four cases added to the existing `TestPeerInfo` table, covering the format the driver actually emits (which was previously untested):

| connection ID | host | port |
| --- | --- | --- |
| `127.0.0.1:27018[-13]` | `127.0.0.1` | 27018 |
| `[::1]:27019[-13]` | `::1` | 27019 |
| `example.com:27020[-13]` | `example.com` | 27020 |
| `example.com[-13]` | `example.com` | 27017 |

All pre-existing cases still pass.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -31,6 +31,7 @@ This project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0.htm
 - Honor the context configured with `WithContext` when constructing resources in `go.opentelemetry.io/contrib/otelconf` and `go.opentelemetry.io/contrib/otelconf/x`. (#9160)
 - Handle nil response bodies from custom `RoundTripper` implementations in `go.opentelemetry.io/contrib/instrumentation/net/http/otelhttp` without panicking. (#9184)
 - Fix incorrect (overestimated) sum calculation for runtime histograms in `go.opentelemetry.io/contrib/instrumentation/runtime`. (#9063)
+- Strip the connection number from the driver's connection ID before deriving `network.peer.address` in `go.opentelemetry.io/contrib/instrumentation/go.mongodb.org/mongo-driver/v2/mongo/otelmongo`. The address previously fell back to the full connection ID, which is unique per pooled connection and gave the recorded metrics unbounded cardinality. (#9274)
 
 <!-- Released section -->
 <!-- Don't change this section unless doing release -->
--- a/instrumentation/go.mongodb.org/mongo-driver/v2/mongo/otelmongo/mongo.go
+++ b/instrumentation/go.mongodb.org/mongo-driver/v2/mongo/otelmongo/mongo.go
@@ -8,6 +8,7 @@ import (
 	"errors"
 	"net"
 	"strconv"
+	"strings"
 	"sync"
 
 	"go.mongodb.org/mongo-driver/v2/bson"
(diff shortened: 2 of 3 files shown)


AMDaban:
I'm unfortunately not able to sign the CLA, so I can't take this PR through to merge.

Rather than just abandon it, I'd like to leave the analysis behind, since the underlying bug is real and still present on `main`.

### The issue

`peerInfo` passes the driver's connection ID straight to `net.SplitHostPort`. The v2 driver formats that ID as `<address>[-<connection number>]` (set in `topology.newConnection`), e.g. `10.0.0.1:27017[-13]`. `SplitHostPort` rejects the trailing `]`, so `peerInfo` falls into its fallback branch and uses the **entire connection ID** — connection number included — as the hostname.

Since the connection number is unique per pooled connection, `network.peer.address` takes a new value for every connection the pool opens. On spans that's merely noisy, but `db.client.operation.duration` is recorded with the same attribute set, so its cardinality grows without bound for the lifetime of the process.

The practical impact is worse than it first looks: in our case a single process accumulated ~5.3k distinct `network.peer.address` values and ~54.5k `db_client_operation_duration_seconds` series (~24.7 MB of exposition). That exceeded the collector's scrape size limit, so the scrape was rejected **outright** — and *every* metric from the process disappeared, not just the MongoDB ones. It returned on restart (fresh registry) and then grew back over the following hours, which made it look like an intermittent outage rather than a cardinality problem.

Worth noting: the existing `TestPeerInfo` cases only cover addresses *without* the connection-number suffix, so the format the driver actually emits isn't currently exercised by the tests.

### Suggested fix

Strip the connection number before parsing. The one subtlety is to trim the bracket only when it opens *after* the final colon, so that bracketed IPv6 literals such as `[::1]` aren't truncated.

Please feel free to implement this independently — there's no need to take anything from this PR. I'm happy to answer questions or provide more detail, and equally happy to close this and open a plain issue instead if that's easier for tracking.

AMDaban:
Closing this, since I can't sign the CLA and it therefore can't be merged. I've opened #9275 to track the bug itself, so it doesn't get lost. The diff here is available as a reference, but please feel free to implement the fix independently — there's no need to take anything from this PR.

PR #9277, opened 2026-07-13
bridges/logutil: fix recursive log conversion crash by enforcing max depth

### LINKED ISSUE #9044: otellogr: Recursive log conversion can crash on cyclic data
### Description

### Steps To Reproduce

### Expected behavior

tharunn0:
## Summary

`convertValue` in all four log bridges (`otellogr`, `otellogrus`, `otelslog`, and `otelzap`) recursively traversed maps, slices, arrays, pointers, and interfaces without a depth limit. Cyclic or excessively deep values could cause unbounded recursion, leading to stack exhaustion and a process crash.

## Changes

- Added `maxConvertDepth = 32`.
- Introduced `convertValueDepth(v any, depth int)`; `convertValue` now calls it with `depth = 0`.
- Incremented the depth on every recursive descent.
- Return `attribute.StringValue("truncated: max depth exceeded")` when the limit is exceeded.
- Added tests covering cyclic and over-deep values.
- Regenerated all four bridge implementations with `go generate`.

## Testing

- Passed all related test

Fixes: #9044


tharunn0:
I'd appreciate feedback on these:

Maximum depth of 32 and is it worth adding explicit cycle detection instead of relying solely on the depth limit ?

pellared:
1. Can you try adding just adding explicit cycle detection instead? Preferably in a separate PR so that we can compare the PRs. I think this would be my implementation preference.
2. For this PR I propose to have the default limit as 64 and make this configurable per https://github.com/open-telemetry/opentelemetry-specification/pull/5186

MrAlias:
@tharunn0 are you able to address [the feedback](https://github.com/open-telemetry/opentelemetry-go-contrib/pull/9277#issuecomment-4971765396) left by @pellared?

tharunn0:
@MrAlias yes, got a little busy, will be updating and raising the pr in a day or two

internal/shared/logutil/convert_test.go.tmpl
```diff
@@ -273,3 +273,15 @@ func TestConvertValueFloat32(t *testing.T) {
...
+func TestConvertValueMaxDepth(t *testing.T) {
+	m := map[string]any{}
+	m["self"] = m
+	assert.Equal(t, attribute.MAP, convertValue(m).Type())
+
+	var nested any = "leaf"
+	for i := 0; i < maxConvertDepth+5; i++ {
+		nested = []any{nested}
+	}
+	assert.Equal(t, attribute.SLICE, convertValue(nested).Type())
```
MrAlias, line 287: The single-edge cycle proves that conversion returns, but these assertions only inspect the root container types. The deeply nested slice would pass without any truncation, so this does not pin down the new boundary or replacement behavior. Could we assert the terminal value immediately before and after the limit, and add a branching-cycle case once cycle detection is implemented?

CHANGELOG.md
```diff
@@ -26,6 +26,7 @@ This project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0.htm
 
 ### Fixed
 
+- Fix recursive log value conversion in `go.opentelemetry.io/contrib/bridges/otellogr`, `go.opentelemetry.io/contrib/bridges/otellogrus`, `go.opentelemetry.io/contrib/bridges/otelslog`, and `go.opentelemetry.io/contrib/bridges/otelzap` to enforce a maximum recursion depth of 32, preventing stack exhaustion from cyclic or deeply-nested values.
```
MrAlias: The PR number is known now, so this entry needs the repository-required reference.

```suggestion
- Fix recursive log value conversion in `go.opentelemetry.io/contrib/bridges/otellogr`, `go.opentelemetry.io/contrib/bridges/otellogrus`, `go.opentelemetry.io/contrib/bridges/otelslog`, and `go.opentelemetry.io/contrib/bridges/otelzap` to enforce a maximum recursion depth of 32, preventing stack exhaustion from cyclic or deeply-nested values. (#9277)
```

MrAlias (commented):
The depth-only approach still leaves the branching-cycle resource-exhaustion case described in #9044 unresolved.

tharunn0:
A separate PR with cycle detection has raised as @pellared suggested. [bridges/logutil: fix recursive log conversion crash by adding cycle detection- #9398
](https://github.com/open-telemetry/opentelemetry-go-contrib/pull/9398)

internal/shared/logutil/convert.go.tmpl
```diff
@@ -16,8 +16,24 @@ import (
...
+
 // convertValue converts various types to attribute.Value.
 func convertValue(v any) attribute.Value {
+	return convertValueDepth(v, 0)
+}
+
+// convertValueDepth converts various types to attribute.Value with depth tracking
+// to prevent unbounded recursion on cyclic or deeply nested values.
+func convertValueDepth(v any, depth int) attribute.Value {
+	if depth > DefaultAttributeValueDepthLimit {
```
MrAlias, line 33: This limit is applied before inspecting the value, and `depth` is also incremented for pointers and interfaces. A chain of 65 pointers to a scalar therefore becomes `attribute.EMPTY`, even though the merged attribute-limit rules only increment through arrays/maps and only replace an over-limit array/map.

Could we count container depth independently of Go indirection and add boundary coverage for pointer/interface chains and scalar leaves?

internal/shared/logutil/convert_test.go.tmpl
```diff
@@ -274,3 +274,15 @@ func TestConvertValueFloat32(t *testing.T) {
...
 	assert.InDelta(t, value.AsFloat64(), want.AsFloat64(), 0.0001)
 }
+
+func TestConvertValueMaxDepth(t *testing.T) {
+	m := map[string]any{}
+	m["self"] = m
+	assert.Equal(t, attribute.MAP, convertValue(m).Type())
+
+	var nested any = "leaf"
+	for i := 0; i < DefaultAttributeValueDepthLimit+5; i++ {
```
MrAlias, line 284: The lint job is red on this generated loop: `intrange` requires integer range syntax. Could we fix the shared template and regenerate all four bridge tests?

```suggestion
	for range DefaultAttributeValueDepthLimit + 5 {
```

MrAlias (commented):
The latest update changes the conversion boundary inconsistently with the merged attribute-limit rules, and the branch is not passing lint.

pellared:
I created https://github.com/open-telemetry/opentelemetry-go-contrib/pull/9542 instead.

PR #9284, opened 2026-07-14
Change k8s client for aws eks resource detector

### LINKED ISSUE #9281: Change k8s client for aws eks resource detector
Non-blocking: this links the EKS detector's full Kubernetes clientset into every `otelconf/x` binary, even when `aws.eks` is not configured. A minimal binary that only calls `NewSDK()` grows from 27 MB to 75 MB and links 268 Kubernetes packages.

This should not block this PR, but I think we should follow up by changing `detectors/aws/eks` to use the typed CoreV1 client, or another lightweight client, instead of `kubernetes.NewForConfig`.

_Originally posted by @MrAlias in https://github.com/open-telemetry/opentelemetry-go-contrib/pull/9138#discussion_r3572556490_

iblancasa:
Fixes #9281


detectors/aws/eks/detector.go
```diff
@@ -16,7 +16,7 @@ import (
 	"go.opentelemetry.io/otel/sdk/resource"
 	semconv "go.opentelemetry.io/otel/semconv/v1.42.0"
 	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
-	"k8s.io/client-go/kubernetes"
+	corev1 "k8s.io/client-go/kubernetes/typed/core/v1"
```
MrAlias: #9281 suggested the typed CoreV1 client, so this is a reasonable interpretation of the requested change. The generated client is still much heavier than expected, though: for a minimal `otelconf/x` binary that only calls `NewSDK()`, this patch reduces the binary from 74.9 MB to 62.7 MB and the Kubernetes dependency count from 268 to 164.

Could we avoid the generated client and use `rest.InClusterConfig` with `rest.HTTPClientFor` to issue and decode only the ConfigMap requests? I prototyped that approach and the same binary dropped to 33.3 MB with 71 Kubernetes packages, much closer to the 27 MB baseline from #9281. This keeps the change scoped to the EKS detector. A focused `httptest` for a successful ConfigMap response and a non-2xx response should cover the new request path.

MrAlias (dismissed):
I like the direction here, but I think there is more we can do to reduce the binary-size impact for `otelconf/x` users.

detectors/aws/eks/go.mod
```diff
@@ -43,18 +31,16 @@ require (
 	go.opentelemetry.io/otel/metric v1.44.1-0.20260625150014-c84013202f01 // indirect
 	go.opentelemetry.io/otel/trace v1.44.1-0.20260625150014-c84013202f01 // indirect
 	go.yaml.in/yaml/v2 v2.4.4 // indirect
-	go.yaml.in/yaml/v3 v3.0.4 // indirect
-	golang.org/x/net v0.57.0 // indirect
+	golang.org/x/net v0.56.0 // indirect
```
MrAlias: Non-blocking: could we preserve `golang.org/x/net` at `v0.57.0`? This version change is unrelated to the Kubernetes dependency reduction, and retaining it avoids an unnecessary downgrade.

```suggestion
	golang.org/x/net v0.57.0 // indirect
```

MrAlias (approved):

dashpole (approved):

PR #9295, opened 2026-07-15
feat(samplers/jaegerremote): add OTEP 235 tracestate probability sampling support

yvrhdn:
Hi! This PR adds an opt-in `WithTraceStateSamplingEnabled()` option that implements the OpenTelemetry tracestate probability sampling specification (OTEP 235).

While it samples at the same rate, it changes _which_ traces are sampled so the option is disabled by default to be backwards compatibility. Rate-limiting strategies are not probabilistic and never set th.

### DIFF
--- a/samplers/jaegerremote/README.md
+++ b/samplers/jaegerremote/README.md
@@ -43,6 +43,26 @@ Notes:
   the backend for the sampling strategy for this service.
 * Both Jaeger Agent and OpenTelemetry Collector implement the Jaeger sampling service endpoint.
 
+## Consistent probability sampling (tracestate)
+
+By default, the probabilistic and per-operation sampling strategies make sampling decisions the same way they always have and do not touch the [W3C `tracestate`](https://www.w3.org/TR/trace-context/#tracestate-header) header.
+
+Passing `jaegerremote.WithTraceStateSamplingEnabled()` to `New` switches those strategies to the OpenTelemetry [tracestate probability sampling](https://opentelemetry.io/docs/specs/otel/trace/tracestate-probability-sampling/) specification (OTEP 235).
+The sampling decision is made by comparing a 56-bit randomness value (an explicit `rv` tracestate value if present, otherwise the trace ID's least significant 56 bits) against a rejection threshold derived from the configured sampling rate.
+On a positive decision, that threshold is published via the `th` sub-key of the `ot` tracestate entry.
+This lets downstream consumers, such as span-to-metrics extrapolation, compute a reliable adjusted count for sampled spans.
+
+```go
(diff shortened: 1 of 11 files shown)


samplers/jaegerremote/README.md
```diff
@@ -43,6 +43,26 @@ Notes:
   the backend for the sampling strategy for this service.
 * Both Jaeger Agent and OpenTelemetry Collector implement the Jaeger sampling service endpoint.
 
+## Consistent probability sampling (tracestate)
+
+By default, the probabilistic and per-operation sampling strategies make sampling decisions the same way they always have and do not touch the [W3C `tracestate`](https://www.w3.org/TR/trace-context/#tracestate-header) header.
+
+Passing `jaegerremote.WithTraceStateSamplingEnabled()` to `New` switches those strategies to the OpenTelemetry [tracestate probability sampling](https://opentelemetry.io/docs/specs/otel/trace/tracestate-probability-sampling/) specification (OTEP 235).
```
yurishkuro, line 50: The main point of "jaegerremote" is that the exact sampling policy comes from the backend via sampling config, to avoid having to redeploy hundreds of microservices with new sampling setups. Given that, it is actually appropriate to have the behavior changed via in-code option?

yvrhdn: sorry for the late reply!

that's fair, I was looking into making jaegerremote tracestate-aware and this seemed like the best approach to follow the spec with a limited change to the jaegerremote protocol.

in terms of impact:
- this change is opt-in, users don't need to switch. it is an option for users that want to adopt OTEP-235.
- if you use jaegerremote with parent based sampling it should be okay to gradually roll out the change across your fleet. once your gateways switch the downstream services will follow this decision.

what do you think of instead only setting the tracestate attributes but not changing the sampling logic?

yurishkuro: 1. The actual tracestate sampler does not belong here, why is it not a separate reusable sampler that people can wire in programmatically?
2. Once it's available elsewhere, it's ok to wire it in instead of the traditional remote sampler that `jaegerremote` invokes.

The remaining design choice is whether the switch from purely random sampler to tracestate-based sampler is controlled centrally through Jaeger's remote sampler protocol or individually by each SDK in each service. Going for protocol change would be much more work, so I would agree that as a shortcut you can simply make it a config option of jaegerremote that toggles random-vs-tracestate

dashpole: I believe that is https://github.com/open-telemetry/opentelemetry-go/pull/8123, which has stalled. I agree that we shouldn't be implementing sampler behavior in this component.

PR #9301, opened 2026-07-16
otelconf: support wildcard include/exclude filters

### LINKED ISSUE #9300: IncludeExclude filters treat wildcard patterns as literal attribute keys
### Component

Config

### Describe the issue you are facing

The declarative configuration contract defines `IncludeExclude.included` and `IncludeExclude.excluded` entries as case-sensitive wildcard patterns: `*` matches zero or more characters and `?` matches exactly one character. Exclusions are applied after inclusions.

The Go implementation in `newIncludeExcludeFilter` instead converts entries to `map[attribute.Key]struct{}` and performs exact-key lookups. As a result, wildcard patterns are treated as literal attribute keys.

For example, this configuration does not exclude `secret.token` or `process.command_args`:

```yaml
with_resource_constant_labels:
  excluded:
    - "secret.*"
    - "process.*"
```

The helper also returns an error when an identical literal appears in both lists. The configuration contract says exclusions take precedence, so overlapping include/exclude entries should be valid.

The affected helper is copied across:

- `otelconf/metric.go`
- `otelconf/v0.2.0/metric.go`
- `otelconf/v0.3.0/metric.go`
- `otelconf/x/metric.go`

It is used for view attribute filters and, depending on the schema package, resource detector and Prometheus resource-label filters.

Specification references:

- [Pattern-matching rules](https://github.com/open-telemetry/opentelemetry-configuration/blob/v0.3.0/README.md#properties-which-pattern-matching)
- [IncludeExclude type descriptions](https://github.com/open-telemetry/opentelemetry-configuration/blob/v0.3.0/schema/type_descriptions.yaml)

This is a configuration correctness issue, not an endpoint-wide redaction mechanism. In particular, Prometheus `target_info` is controlled separately from `WithResourceAsConstantLabels`; tests for resource constant labels should assert the labels on a named application metric rather than search the entire scrape body.

### Expected behavior

- `*` and `?` use the specified case-sensitive wildcard semantics.
- If `included` is empty, attributes are included unless an exclusion matches.
- If `included` is non-empty, an attribute must match at least one inclusion.
- Exclusions are evaluated after inclusions and always win.
- Overlapping include/exclude entries are accepted.

### Steps to Reproduce

Add focused tests next to `newIncludeExcludeFilter`:

```go
func TestIncludeExcludeWildcard(t *testing.T) {
    filter, err := newIncludeExcludeFilter(&IncludeExclude{
        Excluded: []string{"secret.*", "process.?ommand_args"},
    })
    require.NoError(t, err)

    assert.False(t, filter(attribute.String("secret.token", "value")))
    assert.False(t, filter(attribute.String("process.command_args", "value")))
    assert.True(t, filter(attribute.String("service.name", "svc")))
}

func TestIncludeExcludeOverlap(t *testing.T) {
    filter, err := newIncludeExcludeFilter(&IncludeExclude{
        Included: []string{"service.name"},
        Excluded: []string{"service.name"},
    })
    require.NoError(t, err)
    assert.False(t, filter(attribute.String("service.name", "svc")))
}
```

On current code, the wildcard assertions fail because the keys are compared literally, and the overlap test fails because constructing the filter returns an error.

### Operating System

Linux

### Device Architecture

x86_64

### Go Version

go1.26.4

### Component Version

`go.opentelemetry.io/contrib/otelconf` v0.24.0 and current main (`234cdab235e4`)

malikov73:
## Summary

- support case-sensitive `*` and `?` patterns in `IncludeExclude` filters
- apply exclusions after inclusions across stable, `x`, `v0.2.0`, and `v0.3.0`
- preserve the zero-allocation literal fast path and add regression coverage

## Details

`newIncludeExcludeFilter` stored configuration entries as literal attribute keys, so wildcard patterns never matched. It also rejected identical entries in the include and exclude lists even though the configuration contract gives exclusions precedence.

This change adds a shared internal matcher that separates literal and wildcard patterns. `*` matches zero or more characters, `?` matches exactly one character, matching is case-sensitive, and exclusions always win.

Fixes #9300.

## Testing

- `make precommit`
- `go test -race ./...` from `otelconf`
- `go test -run '^$' -bench '^BenchmarkIncludeExcludeFilter' -benchmem -count=10 .`

## Benchmarks

All cases report `0 B/op` and `0 allocs/op`.

| Case | Before | After | Delta |
| --- | ---: | ---: | ---: |
| Exact match | 5.667 ns/op | 5.732 ns/op | +1.16% |
| Exact no match | 7.203 ns/op | 6.971 ns/op | -3.21% |
| Wildcard match | — | 27.40 ns/op | — |
| Wildcard no match | — | 12.00 ns/op | — |
| Wildcard excluded | — | 75.00 ns/op | — |

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -27,7 +27,7 @@ This project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0.htm
 
 ### Fixed
 
-- Fix wildcard matching and exclusion precedence for include/exclude filters in `go.opentelemetry.io/contrib/otelconf`.
+- Fix wildcard matching and exclusion precedence for include/exclude filters in `go.opentelemetry.io/contrib/otelconf`. (#9301)
 - Fix Prometheus reader resource label filter configuration in `go.opentelemetry.io/contrib/otelconf/v0.2.0`. (#9045)
 - Apply `resource.detection/development.attributes.included` and `excluded` filtering to resource detector attributes in `go.opentelemetry.io/contrib/otelconf/x`. (#9131)
 - Honor the context configured with `WithContext` when constructing resources in `go.opentelemetry.io/contrib/otelconf` and `go.opentelemetry.io/contrib/otelconf/x`. (#9160)
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -27,6 +27,7 @@ This project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0.htm
 
 ### Fixed
 
+- Fix wildcard matching and exclusion precedence for include/exclude filters in `go.opentelemetry.io/contrib/otelconf`.
 - Fix Prometheus reader resource label filter configuration in `go.opentelemetry.io/contrib/otelconf/v0.2.0`. (#9045)
 - Apply `resource.detection/development.attributes.included` and `excluded` filtering to resource detector attributes in `go.opentelemetry.io/contrib/otelconf/x`. (#9131)
 - Honor the context configured with `WithContext` when constructing resources in `go.opentelemetry.io/contrib/otelconf` and `go.opentelemetry.io/contrib/otelconf/x`. (#9160)
--- a/otelconf/internal/wildcard/wildcard.go
+++ b/otelconf/internal/wildcard/wildcard.go
@@ -1,0 +1,146 @@
+// Copyright The OpenTelemetry Authors
+// SPDX-License-Identifier: Apache-2.0
+
+// Package wildcard matches strings against case-sensitive wildcard patterns.
+package wildcard // import "go.opentelemetry.io/contrib/otelconf/internal/wildcard"
+
+import (
+	"strings"
+	"unicode/utf8"
+)
+
+// Matcher matches strings against included and excluded wildcard patterns.
+// Patterns in each set are combined with a logical OR, and exclusions take
+// precedence over inclusions. An asterisk matches zero or more characters and
+// a question mark matches exactly one character.
+type Matcher struct {
+	included         patternSet
+	excluded         patternSet
+	includeAll       bool
+	excludeAll       bool
+	excludeNone      bool
+	includeExactOnly bool
+}
+
+type patternSet struct {
+	exact     map[string]struct{}
+	wildcards []string
+	matchAll  bool
+}
+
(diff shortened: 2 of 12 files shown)


CHANGELOG.md
```diff
@@ -27,6 +27,7 @@ This project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0.htm
 
 ### Fixed
 
+- Fix wildcard matching and exclusion precedence for include/exclude filters in `go.opentelemetry.io/contrib/otelconf`. (#9301)
```
iblancasa, line 30: There are modifications across some packages. Maybe those changes should be mentioned here?

otelconf/internal/wildcard/wildcard.go
```diff
@@ -0,0 +1,146 @@
...
+	exact     map[string]struct{}
+	wildcards []string
+	matchAll  bool
+}
+
+// NewMatcher returns a Matcher for the included and excluded patterns.
+func NewMatcher(included, excluded []string) Matcher {
+	includedSet := newPatternSet(included)
+	excludedSet := newPatternSet(excluded)
+	return Matcher{
```
iblancasa, line 35: I think returning this by value and using value receivers on `Match/match` means every `attribute.Filter` invocation copies the whole struct twice (once entering `Match` and again entering `match`). This filter runs per attribute per measurement recording, so it's a hot path

otelconf/internal/wildcard/wildcard.go
```diff
@@ -0,0 +1,146 @@
...
+				patternIndex++
+				continue
+			default:
+				_, patternSize := utf8.DecodeRuneInString(pattern[patternIndex:])
+				_, valueSize := utf8.DecodeRuneInString(value[valueIndex:])
+				if patternSize == valueSize &&
+					pattern[patternIndex:patternIndex+patternSize] == value[valueIndex:valueIndex+valueSize] {
+					patternIndex += patternSize
+					valueIndex += valueSize
+					continue
```
iblancasa, line 128: Maybe
```suggestion
				if pattern[patternIndex] == value[valueIndex] {
					patternIndex++
					valueIndex++
					continue
				}
```

dashpole:
LGTM after comments are addressed

pellared:
@malikov73, can you address the comments?

PR #9303, opened 2026-07-17
chore: docker housekeeping

thompson-tomo:
Explicitly pin docker version to improve info/context available to renovate.

The jaegar tracing image has been switched as existing image is eol jaegertracing/jaeger#6321


instrumentation/github.com/labstack/echo/otelecho/example/docker-compose.yml
```diff
@@ -3,7 +3,7 @@
 version: "3.7"
 services:
   echo-client:
-    image: golang:alpine@sha256:0178a641fbb4858c5f1b48e34bdaabe0350a330a1b1149aabd498d0699ff5fb2
+    image: golang:1.26.0-alpine@sha256:0178a641fbb4858c5f1b48e34bdaabe0350a330a1b1149aabd498d0699ff5fb2
```
MrAlias: This tag and digest do not identify the same manifest. Docker rejects the exact reference with `manifest verification failed`, so `docker compose up` cannot start `echo-client`.

The digest is for `golang:1.26.5-alpine`; keep the tag aligned with the pinned image.

```suggestion
    image: golang:1.26.5-alpine@sha256:0178a641fbb4858c5f1b48e34bdaabe0350a330a1b1149aabd498d0699ff5fb2
```

MrAlias (changes requested):
The Jaeger migration looks compatible, but the Go image reference must be corrected before this is ready.

thompson-tomo:
Note the failing test looks unrelated https://github.com/open-telemetry/opentelemetry-go-contrib/pull/9249

@MrAlias should we look to exclude the file resulting in the invalid link to be checked?

MrAlias (approved):

dashpole (approved):

sonalgaud12 (approved):

PR #9352, opened 2026-07-24
Fix/otelmongo peer address cardinality

### LINKED ISSUE #9275: otelmongo (mongo-driver/v2): network.peer.address includes the connection number, giving db.client.operation.duration unbounded cardinality
### Description

`peerInfo` passes the driver's connection ID straight to `net.SplitHostPort`. The v2 driver formats that ID as `<address>[-<connection number>]` (set in `topology.newConnection`), for example `10.0.0.1:27017[-13]`.

`net.SplitHostPort` rejects the trailing `]` (`unexpected ']' in address`), so `peerInfo` falls into its fallback branch and uses the **entire connection ID** — connection number included — as the hostname:

```go
hostname, portStr, err := net.SplitHostPort(connectionID)
if err != nil {
	// If parsing fails, assume default MongoDB port and return the entire ConnectionID as hostname
	hostname = connectionID
	port = defaultMongoPort
	return hostname, port
}
```

Because the connection number is unique per pooled connection, `network.peer.address` takes a new value for **every connection the pool opens**. On spans this is merely noisy, but `db.client.operation.duration` is recorded with the same attribute set, so the metric's cardinality grows without bound for the lifetime of the process.

### Impact

The practical consequence is worse than it first appears. In our case a single process accumulated:

- ~**5,300** distinct `network.peer.address` values (e.g. `10.0.0.1:27017[-1708]`)
- ~**54,500** `db_client_operation_duration_seconds` bucket series — roughly 96% of all series the process exported
- a ~**24.7 MB** `/metrics` payload

That exceeded the collector's scrape size limit, so the scrape was rejected **outright** — meaning *every* metric from the process disappeared, not just the MongoDB ones. Metrics returned on restart (fresh registry) and then grew back over the following hours, which made it present as an intermittent service outage rather than a cardinality problem. It took a while to trace back to here.

### Environment

- `go.opentelemetry.io/contrib/instrumentation/go.mongodb.org/mongo-driver/v2/mongo/otelmongo`
- `go.mongodb.org/mongo-driver/v2`
- Metrics exported via the OTel Prometheus exporter

Still reproducible on `main` as of today.

### Note on test coverage

The existing `TestPeerInfo` cases only cover addresses *without* the connection-number suffix, so the format the driver actually emits isn't currently exercised. That's likely why this went unnoticed.

### Suggested fix

Strip the connection number from the connection ID before parsing, so `network.peer.address` becomes the MongoDB host and stays bounded by the size of the replica set.

The one subtlety: trim the bracket only when it opens *after* the final colon, so bracketed IPv6 literals such as `[::1]` aren't truncated (the existing `IPv6 without port with square brackets` case covers this).

I opened #9274 with a fix and tests, but I'm not able to sign the CLA, so it can't be merged. Please feel free to implement this independently — there's no need to take anything from that PR. Happy to answer questions or provide more detail.

om7057:
Fixes #9275

With go.mongodb.org/mongo-driver/v2, pooled connection IDs are formatted as `<address>[-<connection number>]`, e.g. `10.0.0.1:27017[-13]`. peerInfo() passes that straight into net.SplitHostPort, which chokes on the trailing suffix and falls back to using the whole connection string as the hostname. End result: network.peer.address on db.client.operation.duration ends up unique per pooled connection instead of stable per host, so the metric's cardinality grows unbounded over the life of the process.

Fix strips the connection number suffix before parsing, so host/port come out right and network.peer.address stays stable.

Added tests covering IPv4, IPv6, and hostname forms of the connection ID.


om7057:
Hi @pellared, @MrAlias, @XSAM,

This addresses #9275.

**Summary**

* Strip the `[-<connection number>]` suffix from `ConnectionID` in `peerInfo()` for `go.mongodb.org/mongo-driver/v2/mongo/otelmongo`.
* Add `TestPeerInfo` coverage for IPv4, IPv6, and hostname connection IDs with pooled connection suffixes, while preserving bracketed IPv6 literals (for example, `[::1]`).
* Update `CHANGELOG.md` under **Unreleased → Fixed**.

I'd appreciate it if you could take a look when you have a chance. Thanks!

om7057:
Rebased this onto current main, it had picked up an unrelated commit from an old branch of mine that was causing the conflict. Should be clean now.

MrAlias (approved):

instrumentation/go.mongodb.org/mongo-driver/v2/mongo/otelmongo/mongo.go
```diff
@@ -212,6 +213,7 @@ func NewMonitor(opts ...Option) *event.CommandMonitor {
 // peerInfo will parse the hostname and port from the mongo connection ID.
 func peerInfo(connectionID string) (hostname string, port int) {
 	defaultMongoPort := 27017
+	connectionID = trimConnectionSuffix(connectionID)
```
prestonvasquez: [nit] `connectionID` will always be of the form %s[-%d], we can just strip:

```
host, _, _ := strings.Cut(connectionID, "[-")
hostname, portStr, err := net.SplitHostPort(host)
```

prestonvasquez (commented):
The peerInfo parser is pretty unfortunate, I've opened the following issue to remove it once the Go Driver has implemented better behavior: https://github.com/open-telemetry/opentelemetry-go-contrib/issues/9370

om7057:
> The peerInfo parser is pretty unfortunate, I've opened the following issue to remove it once the Go Driver has implemented better behavior: #9370

Applied, thanks. Also removed isDigits/the digit validation since it's no longer needed given the guaranteed format.

instrumentation/go.mongodb.org/mongo-driver/v2/mongo/otelmongo/mongo.go
om7057: @prestonvasquez Removed trimConnectionSuffix and isDigits entirely since the digit check isn't needed for a guaranteed format. Also updated the two malformed-suffix test cases since they were asserting the old defensive behavior.

om7057:
@prestonvasquez @dashpole I've addressed the requested changes, please review it once.

instrumentation/go.mongodb.org/mongo-driver/v2/mongo/otelmongo/mongo.go
```diff
@@ -210,12 +211,18 @@ func NewMonitor(opts ...Option) *event.CommandMonitor {
...
 // peerInfo will parse the hostname and port from the mongo connection ID.
+//
+// ConnectionID is always formatted by the driver as "<address>[-<connection
+// number>]", so the pooled connection number suffix can be stripped by
+// cutting at the first "[-".
+// See https://github.com/open-telemetry/opentelemetry-go-contrib/issues/9370.
 func peerInfo(connectionID string) (hostname string, port int) {
 	defaultMongoPort := 27017
-	hostname, portStr, err := net.SplitHostPort(connectionID)
+	host, _, _ := strings.Cut(connectionID, "[-")
```
dashpole, line 221: If the Unix socket path contains `[-` (e.g., `/tmp/mongodb[-test].sock[-1]`), `strings.Cut` will split at the first occurrence, resulting in an incorrect hostname. Using `strings.LastIndex` to cut at the last `[-` is safer:
```suggestion
	host := connectionID
	if idx := strings.LastIndex(connectionID, "[-"); idx != -1 {
		host = connectionID[:idx]
	}
```

prestonvasquez (approved):

dashpole (commented):
Should we fix this issue in the v1 instrumentation as well?

instrumentation/go.mongodb.org/mongo-driver/v2/mongo/otelmongo/mongo.go
om7057: Got it @dashpole , fixed in e9f77a9. Switched to strings.LastIndex so a Unix socket path containing [- won't get truncated early.

prestonvasquez: @dashpole A unix domain doesn't have a host or port, we should ignore these cases, which IIUC is caught by the existing error check. 

Suggest no changes, @om7057.

om7057: @prestonvasquez Understood, reverted strings.Cut.

om7057:
Yes, `go.mongodb.org/mongo-driver` v1.17.9 formats pooled connection IDs the same way (`fmt.Sprintf("%s[-%d]", addr, nextConnectionID())` in `x/mongo/driver/topology/connection.go`), so `peerInfo` in the v1 instrumentation's `internal/semconv/event_monitor.go` has the identical bug. Pushed a fix using the same `strings.Cut` approach we landed on here, added test cases mirroring the v2 ones (IPv4/IPv6/hostname with connection number suffix), and a matching CHANGELOG entry.

om7057:
One correction to my earlier comment: the v1 module has no metrics pipeline, so the bug there only affects the `network.peer.address` and `net.peer.name` span attributes rather than the metric cardinality growth described in #9275. The underlying `net.SplitHostPort` defect is the same, but the impact is narrower. I've updated the CHANGELOG entry to say that instead of reusing the v2 wording.

The v1 fix uses `strings.Cut` on the first `[-`, consistent with @prestonvasquez's call above that a Unix domain socket has no host or port and is already handled by the existing error check. No reason for v1 to diverge from v2 here.

dashpole (approved):

om7057:
Thank you so much @dashpole for approval, and @MrAlias for the merge!

PR #9359, opened 2026-07-26
xray: fix predictable trace/span IDs when crypto/rand seed read fails

### LINKED ISSUE #9048: xray: Predictable trace/span IDs in AWS X-Ray ID generator
### Description

The AWS X-Ray ID generator uses `math/rand` for trace and span IDs after seeding it from `crypto/rand.Reader`. Current-tree validation showed seed-read errors are ignored; if the seed read fails, the seed remains zero and generated IDs become deterministic. Even when seeding succeeds, `math/rand` is not intended for unpredictable identifiers.

### Steps To Reproduce

[14-ad458fee1a048191a08bca66901af3a9-predictable-trace-span-ids-in-aws-x-ray-id-generator.zip](https://github.com/user-attachments/files/28384898/14-ad458fee1a048191a08bca66901af3a9-predictable-trace-span-ids-in-aws-x-ray-id-generator.zip)

1. Use `go.opentelemetry.io/contrib/propagators/aws/xray`.
2. Simulate or force a failed `crypto/rand` seed read, or run the validation PoC from the artifact zip.
3. Generate trace and span IDs from the resulting generator.
4. Observe repeatable ID sequences when the seed remains zero.

### Expected behavior

ID generation should either use a cryptographically secure random source directly or fail safely when seeding fails, rather than continuing with deterministic `math/rand` output.

om7057:
Fixes #9048

the seed read from crypto/rand here just ignores its error:

```go
var rngSeed int64
_ = binary.Read(crand.Reader, binary.LittleEndian, &rngSeed)
gen.randSource = rand.New(rand.NewSource(rngSeed))
```

if that read fails for whatever reason the seed is just 0 and you get the same ids every time.

simplest fix is to drop the manual seeding altogether and use math/rand/v2's global funcs, which the runtime seeds itself. also means no more mutex since those are already concurrency safe. core sdk does the same thing for its default generator so figured it's fine to match that here.

didn't change the id format or anything public, existing tests still pass. added one more test that just hammers the generator from a bunch of goroutines under -race, mostly to make sure removing the mutex doesn't come back to bite someone later.


propagators/aws/xray/idgenerator.go
```diff
@@ -5,32 +5,30 @@ package xray // import "go.opentelemetry.io/contrib/propagators/aws/xray"
...
 
 var _ sdktrace.IDGenerator = &IDGenerator{}
 
 // NewSpanID returns a non-zero span ID from a randomly-chosen sequence.
 func (gen *IDGenerator) NewSpanID(context.Context, trace.TraceID) trace.SpanID {
-	gen.Lock()
-	defer gen.Unlock()
 	sid := trace.SpanID{}
-	_, _ = gen.randSource.Read(sid[:])
+	binary.NativeEndian.PutUint64(sid[:], rand.Uint64())
```
dashpole: If rand.Uint64 returns 0, this returns an invalid all-zero SpanID. Consider looping to ensure a non-zero ID:
```suggestion
	for {
		binary.NativeEndian.PutUint64(sid[:], rand.Uint64())
		if sid.IsValid() {
			break
		}
	}
```

propagators/aws/xray/idgenerator.go
```diff
@@ -40,26 +38,20 @@ func (gen *IDGenerator) NewSpanID(context.Context, trace.TraceID) trace.SpanID {
...
 	tid := trace.TraceID{}
 	currentTime := getCurrentTimeHex()
 	copy(tid[:4], currentTime)
-	_, _ = gen.randSource.Read(tid[4:])
+	binary.NativeEndian.PutUint64(tid[4:12], rand.Uint64())
+	binary.NativeEndian.PutUint32(tid[12:16], rand.Uint32())
 
 	sid := trace.SpanID{}
-	_, _ = gen.randSource.Read(sid[:])
+	binary.NativeEndian.PutUint64(sid[:], rand.Uint64())
```
dashpole: Same as above:
```suggestion
	for {
		binary.NativeEndian.PutUint64(sid[:], rand.Uint64())
		if sid.IsValid() {
			break
		}
	}
```

om7057:
@dashpole Thanks for the review. I've updated both NewSpanID and NewIDs to loop until we get a valid non-zero ID, which ensures we never return an all-zero span ID as the spec requires. All tests pass with the changes.

propagators/aws/xray/idgenerator.go
om7057: Addressed.

propagators/aws/xray/idgenerator.go
om7057: Addressed.

propagators/aws/xray/idgenerator.go
```diff
@@ -5,32 +5,35 @@ package xray // import "go.opentelemetry.io/contrib/propagators/aws/xray"
...
-type IDGenerator struct {
-	sync.Mutex
-	randSource *rand.Rand
-}
+//
+// math/rand/v2's top-level generator is used instead of a per-instance
+// math/rand.Rand seeded from crypto/rand: the former is safe for concurrent
+// use without a mutex and, unlike a locally seeded math/rand.Rand, has no
+// seed-read step that can silently fail and leave it deterministic.
+type IDGenerator struct{}
```
MrAlias: Embedding `sync.Mutex` exposes an exported `Mutex` field and promotes `Lock`, `Unlock`, and `TryLock` into `*IDGenerator`'s public method set. Replacing the struct with an empty type removes that API, so existing consumers can no longer compile; I confirmed `NewIDGenerator().Lock()` builds against the base and fails against this head. Since `go.opentelemetry.io/contrib/propagators/aws` is stable-v1, please preserve the existing public type shape here or defer this removal to a major version.

CHANGELOG.md
```diff
@@ -10,12 +10,40 @@ This project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0.htm
...
+- Fix `Severity.UnmarshalText` round trip for positive `FATAL` offsets above the named range in `go.opentelemetry.io/contrib/processors/minsev`. (#9197)
+- `TextMapPropagator` in `go.opentelemetry.io/contrib/propagators/autoprop` returns the no-op propagator for empty input, matching the behavior of `none`. An unknown `OTEL_PROPAGATORS` value still returns an error with a nil propagator so `NewTextMapPropagator` falls back to the default TraceContext and Baggage propagators instead of disabling propagation. (#9163)
+- Preserve error-valued attributes nested in a group as grouped attributes instead of silently dropping them in `go.opentelemetry.io/contrib/bridges/otelslog`. (#9238)
+- Fix a data race in `go.opentelemetry.io/contrib/bridges/otelslog` where concurrent `Handle` calls could corrupt each other's log attributes because `kvBuffer.KeyValues` returned a slice aliasing a shared buffer. (#9229)
+- Fix `go.opentelemetry.io/contrib/propagators/aws/xray` producing deterministic trace and span IDs when the seed read from `crypto/rand` silently failed, by switching to `math/rand/v2`'s securely auto-seeded top-level generator. (#9359)
```
MrAlias: Non-blocking: `securely auto-seeded` implies a cryptographic guarantee that `math/rand/v2` does not make. The important behavior is that the top-level generator avoids the per-instance seed-read failure mode.

```suggestion
- Fix `go.opentelemetry.io/contrib/propagators/aws/xray` producing deterministic trace and span IDs when the seed read from `crypto/rand` silently failed, by switching to `math/rand/v2`'s concurrency-safe top-level generator. (#9359)
```

MrAlias (changes requested):
The `math/rand/v2` change is reasonable and consistent with the relevant specifications. However, this patch removes public API from `IDGenerator`, breaking compatibility in a stable-v1 module. That API break needs to be resolved before merge.

propagators/aws/xray/idgenerator.go
om7057: @MrAlias I've added the sync.Mutex field back to preserve the public API shape even though we don't use it internally now. Existing code that calls NewIDGenerator().Lock() will continue to compile. Added a comment explaining it's kept for backward compatibility.

CHANGELOG.md
om7057: @MrAlias Changed it from "securely auto-seeded" to "concurrency-safe" to be more accurate about what math/rand/v2 actually guarantees. The important behavior is avoiding the per-instance seed-read failure mode, which the top-level generator handles correctly.

om7057:
> The `math/rand/v2` change is reasonable and consistent with the relevant specifications. However, this patch removes public API from `IDGenerator`, breaking compatibility in a stable-v1 module. That API break needs to be resolved before merge.

@MrAlias Thank you for the review.

I've preserved the sync.Mutex field in the struct to maintain the public API surface. The field is documented as kept for backward compatibility, so existing code that interacts with the Lock/Unlock methods will continue to compile. 
Could you please check them once?

om7057:
Pushed a follow-up cleaning up the lint fallout from preserving the mutex and adding the loop: `//nolint:gosec` on the `rand.Uint64`/`rand.Uint32` calls (matches this repo's existing convention for non-security-sensitive randomness), `//nolint:revive` on `NewSpanID`/`NewIDs` since the receiver is required by the `sdktrace.IDGenerator` interface but no longer read in the body, and a `TestIDGeneratorLockUnlockCompat` test that exercises `Lock`/`TryLock`/`Unlock` so a future refactor that drops the embedded mutex fails CI instead of silently breaking callers. `golangci-lint` and `go test -race` are clean on the package now.

propagators/aws/xray/idgenerator_test.go
```diff
@@ -88,3 +89,36 @@ func TestSpanIDIsNotNil(t *testing.T) {
...
+
+	assert.True(t, stillLocked, "TryLock should fail while the mutex is held")
+}
+
+// TestIDGeneratorConcurrentUse guards against reintroducing a shared,
+// unsynchronized random source: IDGenerator no longer holds a mutex, so this
+// only stays safe as long as ID generation goes through math/rand/v2's
+// top-level functions, which are documented as safe for concurrent use.
+// Run with -race to catch a regression.
+func TestIDGeneratorConcurrentUse(t *testing.T) {
```
MrAlias, line 140: Could we add a regression test for the seed-read failure this fixes? This concurrency test checks race safety, but the previous zero-seeded implementation would still pass it. Replacing `crypto/rand.Reader` with a failing reader and comparing IDs from two independently constructed generators fails on `main` and passes here.

MrAlias (approved):
The stable-v1 compatibility issue is resolved on the current head.

om7057:
@MrAlias, thank you for the approval.
Also, @dashpole, please verify the changes made and check if the PR is ready for merge.

propagators/aws/xray/idgenerator_test.go
om7057: Added TestIDGeneratorSeedReadFailure in https://github.com/open-telemetry/opentelemetry-go-contrib/commit/196812ed04071b9001445936232de2e75e4b7cdb. It swaps crypto/rand.Reader for a reader that always errors, constructs two generators independently, and asserts their IDs differ.

Verified both directions: against current main it fails with the two generators emitting byte-identical trace and span IDs (0x6a6abc68...), which is the zero-seed determinism this PR fixes. It passes on this branch, where ID generation no longer touches crypto/rand at all.

dashpole (approved):

om7057:
I appreciate the approval @dashpole, Also thanks for the merge @MrAlias!

PR #9361, opened 2026-07-26
fix(otelmux): copy MultipartForm back to original request for temp file cleanup

pujitha24:
Motivation

otelmux.Middleware passes the downstream handler a context-derived copy
of the request (r.WithContext(ctx)) instead of the original. If the
handler calls ParseMultipartForm, the resulting *multipart.Form -
which references any temp files created for large uploads - is stored
on that copy's MultipartForm field, not on the original request that
net/http's http.Server holds a reference to. After ServeHTTP returns,
net/http's response.finishRequest only removes temp files via
MultipartForm.RemoveAll() on the original request object, so the temp
files are never reaped and accumulate on disk.

Approach

Keep a reference to the context-derived request and, after the wrapped
handler returns, copy its MultipartForm back onto the original request
so net/http can find and clean up the temp files.

Scope note: when otelmux.Middleware is used directly (wrapping a
plain http.Handler, or via any router that doesn't itself re-derive
the request before invoking the middleware chain), this fully fixes
the leak. When used via router.Use(...) on a gorilla/mux.Router,
gorilla/mux's own routing step (requestWithRoute) performs an
additional, unrelated context copy before calling the middleware
chain, and that copy is never written back to the request object
gorilla/mux itself received. That remaining leak lives in gorilla/mux
and is outside this package's control; a maintainer noted intent to
file it upstream. This change does not claim to eliminate that case.

Validation

go test ./instrumentation/github.com/gorilla/mux/otelmux/... passes,
including new regression test TestMultipartFormCopiedToOriginalRequest,
which applies Middleware directly to a handler that calls
ParseMultipartForm and asserts MultipartForm is copied back onto the
original request. Reverting the fix causes this test to fail.

golangci-lint run (scoped to this package) reports 0 issues.

This change fixes the direct-wrapper path only. The `router.Use(...)` integration still leaks because gorilla/mux re-derives the request during routing (see gorilla/mux#777), so #9070 must stay open until that lands. Intentionally a non-closing reference.

Relates to #9070


MrAlias (commented):
This patch addresses only one request-copy boundary, so it is not ready to close #9070 or ship the current cleanup claim.

instrumentation/github.com/gorilla/mux/otelmux/mux.go
```diff
@@ -178,7 +178,15 @@ func (tw traceware) ServeHTTP(w http.ResponseWriter, r *http.Request) {
...
 
-	tw.handler.ServeHTTP(w, r.WithContext(ctx))
+	rCtx := r.WithContext(ctx)
+	tw.handler.ServeHTTP(w, rCtx)
+	// Copy MultipartForm back to the original request so net/http's
+	// finishRequest can find and remove any temp files ParseMultipartForm
+	// created on the context-derived request copy.
+	// See https://github.com/open-telemetry/opentelemetry-go-contrib/issues/9070
+	if rCtx.MultipartForm != nil {
+		r.MultipartForm = rCtx.MultipartForm
```
MrAlias, line 188: The package example installs this middleware with `router.Use(...)`. In that path, `r` is already gorilla/mux's context-derived copy, so assigning `rCtx.MultipartForm` here only updates the router-owned request; the request held by `net/http` remains nil and disk-backed uploads still leak.

Could we treat this as one half of the fix, link [gorilla/mux#777](https://github.com/gorilla/mux/pull/777), and keep #9070 open until the router copy-back is available? The changelog should state the direct-wrapper scope rather than claiming `net/http` cleanup for the documented integration.

pujitha24:
Thanks both for the review.

@Copilot: the copy-back is now deferred so it still runs if the wrapped handler panics after `ParseMultipartForm`, and the regression test now uses `maxMemory=0` and asserts the file part is disk-backed (`*os.File`) so it actually exercises the temp-file leak instead of an in-memory form. Added a panic-safety regression test as well. Also added the missing PR number to the changelog line.

@MrAlias: agreed — this only fixes the direct-wrapper case, not `router.Use(...)`, where gorilla/mux's own routing step makes its own uncopied-back request. I've updated the changelog and PR description to state that scope explicitly, link gorilla/mux#777, and stop claiming/closing #9070 (changed the description so it no longer auto-closes the issue on merge). Leaving #9070 open for the router-level gap as you suggested.

instrumentation/github.com/gorilla/mux/otelmux/mux.go
```diff
@@ -178,7 +178,18 @@ func (tw traceware) ServeHTTP(w http.ResponseWriter, r *http.Request) {
 		},
 	})
 
-	tw.handler.ServeHTTP(w, r.WithContext(ctx))
+	rCtx := r.WithContext(ctx)
+	// Copy MultipartForm back to the original request so net/http's
+	// finishRequest can find and remove any temp files ParseMultipartForm
+	// created on the context-derived request copy. Deferred so the copy-back
+	// still runs if the wrapped handler panics.
```
MrAlias: `net/http` does not call `finishRequest` after a handler panic. The panic unwinds from `serverHandler.ServeHTTP` to `conn.serve`'s recovery defer, bypassing the later `w.finishRequest()` call. This copies the form pointer back, but no code calls `MultipartForm.RemoveAll()`, so disk-backed uploads still leak.

Please either remove the panic-cleanup claim and regression test, or clean up the form in the middleware while preserving the panic.

instrumentation/github.com/gorilla/mux/otelmux/mux_test.go
```diff
@@ -48,6 +52,82 @@ func TestRequestBodyWrapper(t *testing.T) {
...
+	require.NoError(t, writer.Close())
+
+	r := httptest.NewRequestWithContext(t.Context(), http.MethodPost, "/upload", &body)
+	r.Header.Set("Content-Type", writer.FormDataContentType())
+	w := httptest.NewRecorder()
+
+	handler.ServeHTTP(w, r)
+
+	assert.Equal(t, http.StatusOK, w.Code)
+	require.NotNil(t, r.MultipartForm, "MultipartForm should be copied back to the original request so net/http can clean up its temp files")
```
MrAlias, line 96: These tests call the middleware directly, so `http.Server.finishRequest` never runs; forcing disk storage makes the tests themselves leak `multipart-*` files. Running them with an isolated `TMPDIR` left two files behind.

Please register cleanup in both tests after confirming the form is non-nil.

```suggestion
	require.NotNil(t, r.MultipartForm, "MultipartForm should be copied back to the original request so net/http can clean up its temp files")
	t.Cleanup(func() {
		require.NoError(t, r.MultipartForm.RemoveAll())
	})
```

MrAlias (changes requested):
The direct-wrapper fix handles the normal return path, but the patch still claims cleanup semantics beyond what it delivers. This should block merging, and the PR description's `Fixes #9070` must become a non-closing reference while `router.Use(...)` remains unresolved.

pujitha24:
Thanks for catching this — you're right that the deferred copy-back never actually helped in the panic case, since `net/http` never calls `finishRequest` (and thus `RemoveAll`) after a handler panic. I went with the option you offered of removing the overclaim rather than implementing panic-safe cleanup: reverted to a plain, non-deferred copy-back, deleted `TestMultipartFormCopiedToOriginalRequestOnPanic`, and updated the CHANGELOG to say the fix does not cover a handler panic.

I also added `t.Cleanup` to `TestMultipartFormCopiedToOriginalRequest` so it removes its own disk-backed temp files — confirmed with an isolated `TMPDIR` that no `multipart-*` files are left behind after the test run.

Happy to look into real panic-safe cleanup as a follow-up if that's preferred over documenting the gap.

pujitha24:
@MrAlias I noticed my earlier reply claimed the PR description had already been updated to stop auto-closing #9070 on merge, but that was inaccurate — the description still ended with `Fixes #9070`. I've corrected that now: it reads "Relates to #9070" with an explicit note that this only covers the direct-wrapper case, not `router.Use(...)` (linking gorilla/mux#777) or the handler-panic case, so #9070 stays open as you asked. No code changes were needed here, just the description fix. Apologies for the earlier inaccuracy.

instrumentation/github.com/gorilla/mux/otelmux/mux.go
pujitha24: You're right — this only covers the direct-wrapper boundary. I've scoped it accordingly: the CHANGELOG now states the direct-wrapper scope and explicitly calls out that it does not cover `router.Use(...)` (where gorilla/mux re-derives the request), linking gorilla/mux#777. I also changed the description's `Fixes #9070` to a non-closing `Relates to #9070` so #9070 stays open until the router copy-back lands upstream. Treating this as one half of the fix.

instrumentation/github.com/gorilla/mux/otelmux/mux.go
pujitha24: Good catch — confirmed: `net/http` unwinds a handler panic straight to `conn.serve`'s recovery defer and never reaches `w.finishRequest()`, so the copy-back can't help there and no `RemoveAll()` runs. I removed the panic-cleanup claim and its regression test rather than overstate what this delivers; the copy-back now runs only on the normal return path, which is the only case it can actually guarantee.

instrumentation/github.com/gorilla/mux/otelmux/mux_test.go
pujitha24: Applied your suggestion — the test now registers `t.Cleanup(func() { require.NoError(t, r.MultipartForm.RemoveAll()) })` after asserting the form is non-nil. I verified with an isolated `TMPDIR` that the package tests now leave 0 `multipart-*` files behind (they leaked before).

pujitha24:
Thanks for the thorough review, @MrAlias — I think this now matches what it actually delivers:

- **Scope**: narrowed to the direct-wrapper path. CHANGELOG spells out that it does **not** cover `router.Use(...)` (gorilla/mux re-derives the request) and links gorilla/mux#777.
- **Non-closing**: the description now says `Relates to #9070` instead of `Fixes #9070`, so #9070 stays open until the router copy-back lands.
- **Panic claim removed**: since `net/http` never calls `finishRequest` after a handler panic, I dropped the panic-cleanup claim and its test rather than overstate it.
- **Test cleanup**: added `t.Cleanup(... RemoveAll())`; verified with an isolated `TMPDIR` that the tests leave 0 `multipart-*` files behind.

Ready for another look whenever you have time.

MrAlias (approved):

instrumentation/github.com/gorilla/mux/otelmux/mux_test.go
```diff
@@ -48,6 +52,55 @@ func TestRequestBodyWrapper(t *testing.T) {
...
```
dashpole: nit: If `require.NotNil` fails here, `t.Cleanup` won't be registered, which leaves temp files on disk. Consider moving `t.Cleanup` before `require.NotNil` and add a nil check for r.MultipartForm.

dashpole (approved):

pujitha24:
Thanks @dashpole — good catch. I moved `t.Cleanup` ahead of `require.NotNil` in `TestMultipartFormCopiedToOriginalRequest` and added a nil check inside the cleanup closure, so cleanup is registered (and safely no-ops if `MultipartForm` is nil) before the assertion that could fail first. Confirmed `TestMultipartFormCopiedToOriginalRequest` and the full `otelmux` package test suite still pass.

mahendrabishnoi2:
@MrAlias As per [SIG meeting notes](https://github.com/open-telemetry/opentelemetry-go/issues/6648#issuecomment-3479584781) back in November last year, consensus was to remove the temp file using `MultipartForm.RemoveAll()` method. (This was for otelgin)

There is an open (stale) PR https://github.com/open-telemetry/opentelemetry-go-contrib/pull/6609 for otelgin, never got merged.
Given, most of http lib instumentation have this issue (ref: https://github.com/open-telemetry/opentelemetry-go-contrib/issues/9070#issuecomment-4944139785),  can we say this is the approach for fixing this problem? I guess it would be good to solve it similarly in other http lib instumentation as well. 

Any thoughts?

mahendrabishnoi2:
Fixes https://github.com/open-telemetry/opentelemetry-go-contrib/issues/9070

Yiling-J:
Have the team decided that “copy the cloned `Request.MultipartForm` value back to the original request before returning” is the expected approach for this class of problems?

I can update my stale PR https://github.com/open-telemetry/opentelemetry-go-contrib/pull/6609 to use the same approach. FYI, the Go team has some suggestions: https://github.com/golang/go/issues/74455#issuecomment-3099652232

@pujitha24 @mahendrabishnoi2

PR #9387, opened 2026-07-29
otelrestful: record error.type on server span

### LINKED ISSUE #9377: otelrestful: client-initiated request termination is recorded as a 500 with span status Error and no error.type
Split out from #9358 as requested in https://github.com/open-telemetry/opentelemetry-go-contrib/issues/9358#issuecomment-5108871781. Same defect class as the `otelhttp` report, confirmed independently against `otelrestful`. This package was not named in the original thread, but it is affected for the same reason, so raising it here to keep the set complete.

### Symptom

When a client disconnects mid-request, the handler observes the cancelled request context and typically returns an error, which surfaces as a 500. `otelrestful` then records `http.response.status_code=500` and sets the span status to `Error`, and sets no `error.type` at all, so a client disconnect is indistinguishable from a genuine server fault.

`OTelFilter` reads `resp.StatusCode()` after `chain.ProcessFilter` and never inspects `req.Request.Context().Err()`:

https://github.com/open-telemetry/opentelemetry-go-contrib/blob/5f30960897bf324a81b44247dec93e51f902cfd1/instrumentation/github.com/emicklei/go-restful/otelrestful/restful.go#L69-L75

`otelrestful` is the thinnest of the server instrumentations: it wraps neither the request body nor the response writer, so there is no observed read or write error to fall back on, and it does not record metrics at all (`semconv.NewHTTPServer(nil)`), so the metric side of the parent issue does not apply here. The request-context error is the only signal available, and it is not consulted.

### Reproduction

Confirmed on `main` at 5f30960 with a real HTTP/1.1 disconnect. The handler waits on `req.Request.Context().Done()` and then writes 500:

```go
container.Filter(otelrestful.OTelFilter("srv", otelrestful.WithTracerProvider(tp)))
ws.Route(ws.GET("/hello").To(func(req *restful.Request, resp *restful.Response) {
    <-req.Request.Context().Done()
    resp.WriteHeader(http.StatusInternalServerError)
    _, _ = resp.Write([]byte("cancelled"))
}))
```

```
span name   = "/hello"
span status = Error ""
```

### Suggested fix

Mirroring the triage in https://github.com/open-telemetry/opentelemetry-go-contrib/issues/9358#issuecomment-5095882060, reduced to what applies here:

1. Prefer the request-context error as the failure cause over the bare status code.
4. Add a real disconnect regression test so cancellation is detected before the filter returns.

Whether `otelrestful` should also start wrapping the body and response writer to observe read/write errors — and whether it should record metrics at all — feels like separate scope; happy to raise that separately if it is wanted.

### Note on shared code

`otelrestful/internal/semconv/server.go` is generated from `internal/shared/semconv/server.go.tmpl` and is byte-identical to the copy vendored into `otelhttp` (verified by `diff`). The same is true for `otelgin`, `otelmux` and `otelecho`. So whatever `error.type` classification lands in the shared template is picked up by all of them, but each middleware still needs its own wiring to detect and pass the failure cause, plus its own regression test.

Relevant requirements: [HTTP span conventions](https://opentelemetry.io/docs/specs/semconv/http/http-spans/#http-server-span) and the [dropped-connection example](https://opentelemetry.io/docs/specs/semconv/http/http-spans/#http-server-call-connection-dropped-before-response-body-was-sent).

dashpole:
Fix client-initiated request termination being recorded as a 500 status without an error.type attribute.

- Inspect request context error after chain.ProcessFilter returns
- Set span status to Error and record error.type attribute when request context error occurs
- Set error.type to the HTTP status code string for 5xx responses without a context error
- Add disconnect regression test

Fixes #9377


instrumentation/github.com/emicklei/go-restful/otelrestful/restful.go
```diff
@@ -69,7 +73,18 @@ func OTelFilter(service string, opts ...Option) restful.FilterFunction {
...
+		err := req.Request.Context().Err()
+		if err != nil {
+			span.SetStatus(codes.Error, err.Error())
+			span.SetAttributes(otelsemconv.ErrorType(err))
+		} else {
+			stCode, stMsg := semconvServer.Status(status)
+			span.SetStatus(stCode, stMsg)
+			if stCode == codes.Error {
+				span.SetAttributes(otelsemconv.ErrorTypeKey.String(strconv.Itoa(status)))
+			}
```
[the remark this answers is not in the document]
dashpole: addressed in 440f97d5f

instrumentation/github.com/emicklei/go-restful/otelrestful/restfultest_test.go
```diff
@@ -188,10 +191,77 @@ func TestSpanStatus(t *testing.T) {
...
+
+	ctx, cancel := context.WithCancel(t.Context())
+	req, err := http.NewRequestWithContext(ctx, http.MethodGet, ts.URL+"/hello", http.NoBody)
+	require.NoError(t, err)
+
+	client := ts.Client()
+	go func() {
+		time.Sleep(50 * time.Millisecond)
+		cancel()
+	}()
```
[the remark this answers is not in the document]
dashpole: addressed in e410f30b7

instrumentation/github.com/emicklei/go-restful/otelrestful/restfultest_test.go
```diff
@@ -188,10 +191,77 @@ func TestSpanStatus(t *testing.T) {
...
+	assert.Equal(t, codes.Error, span.Status().Code)
+	assert.Contains(t, span.Attributes(), attribute.Int("http.response.status_code", http.StatusInternalServerError))
+	hasErrorType := false
+	for _, attr := range span.Attributes() {
+		if attr.Key == "error.type" && attr.Value.AsString() != "" {
+			hasErrorType = true
+			break
+		}
+	}
+	assert.True(t, hasErrorType, "span should have non-empty error.type attribute")
```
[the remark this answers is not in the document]
dashpole: addressed in 76fbe6215

instrumentation/github.com/emicklei/go-restful/otelrestful/restful.go
```diff
@@ -69,7 +73,18 @@ func OTelFilter(service string, opts ...Option) restful.FilterFunction {
 		chain.ProcessFilter(req, resp)
 
 		status := resp.StatusCode()
-		span.SetStatus(semconvServer.Status(status))
+		err := req.Request.Context().Err()
```
pellared: Inspect the context captured before downstream filters.

`req.Request` remains mutable throughout the filter chain. A downstream filter can install a scoped child context and `defer cancel()`. That cancellation runs before control returns here, causing a successful 2xx request to be recorded with span status `Error` and `error.type`.

This regression test reproduces the problem on the current head:

```go
func TestDownstreamContextCancellationIgnored(t *testing.T) {
    sr := tracetest.NewSpanRecorder()
    provider := sdktrace.NewTracerProvider(sdktrace.WithSpanProcessor(sr))

    scopedContextFilter := func(
        req *restful.Request,
        resp *restful.Response,
        chain *restful.FilterChain,
    ) {
        ctx, cancel := context.WithCancel(req.Request.Context())
        defer cancel()
        req.Request = req.Request.WithContext(ctx)
        chain.ProcessFilter(req, resp)
    }

    ws := &restful.WebService{}
    ws.Route(ws.GET("/").To(func(
        _ *restful.Request,
        resp *restful.Response,
    ) {
        resp.WriteHeader(http.StatusOK)
    }))

    container := restful.NewContainer()
    container.Filter(otelrestful.OTelFilter(
        "test",
        otelrestful.WithTracerProvider(provider),
    ))
    container.Filter(scopedContextFilter)
    container.Add(ws)

    container.ServeHTTP(
        httptest.NewRecorder(),
        httptest.NewRequestWithContext(
            t.Context(),
            http.MethodGet,
            "/",
            http.NoBody,
        ),
    )

    require.Len(t, sr.Ended(), 1)
    span := sr.Ended()[0]
    assert.Equal(t, codes.Unset, span.Status().Code)
    for _, attr := range span.Attributes() {
        assert.NotEqual(t, otelsemconv.ErrorTypeKey, attr.Key)
    }
}
```

The test currently fails with status `Error` and an `error.type` attribute. Reading the pre-chain `ctx` already captured above instead `err := ctx.Err()` makes it pass while preserving genuine client-disconnect detection.

pellared (commented):
One observability correctness issue found.

instrumentation/github.com/emicklei/go-restful/otelrestful/restful.go
dashpole: addressed in 28aa36ce2

pellared (approved):

instrumentation/github.com/emicklei/go-restful/otelrestful/restful.go
```diff
@@ -69,7 +73,18 @@ func OTelFilter(service string, opts ...Option) restful.FilterFunction {
 		chain.ProcessFilter(req, resp)
 
 		status := resp.StatusCode()
-		span.SetStatus(semconvServer.Status(status))
+		err := ctx.Err()
```
dashpole: A successfully-served request is reported as an error when the client disconnects afterwards. Gate the context branch on "no response status made it out".

dashpole: Context branch pre-empts the status-code branch, so `error.type` is never the 5xx status when both apply. Check the status first and fall back to the context error only when no error status is available:.

dashpole: Diverges from every other contrib HTTP server instrumentation; logic belongs in the shared template. Put the 5xx → `error.type` logic in `internal/shared/semconv/server.go.tmpl`'s `ResponseTraceAttrs` (mirroring `client.go.tmpl:178`), regenerate, and let otelhttp/otelmux/otelecho/otelgin/otelrestful all pick it up in one PR.

instrumentation/github.com/emicklei/go-restful/otelrestful/restful.go
```diff
@@ -69,7 +73,18 @@ func OTelFilter(service string, opts ...Option) restful.FilterFunction {
...
+		if err != nil {
+			span.SetStatus(codes.Error, err.Error())
+			span.SetAttributes(otelsemconv.ErrorType(err))
+		} else {
+			stCode, stMsg := semconvServer.Status(status)
+			span.SetStatus(stCode, stMsg)
+			if status >= 500 && status < 600 {
```
dashpole: Span status can be `Error` with no `error.type` (out-of-range status codes). Derive the condition from the single source of truth instead of duplicating the range: `if stCode == codes.Error { span.SetAttributes(otelsemconv.ErrorTypeKey.String(strconv.Itoa(status))) }`.

instrumentation/github.com/emicklei/go-restful/otelrestful/restful.go
```diff
@@ -69,7 +73,18 @@ func OTelFilter(service string, opts ...Option) restful.FilterFunction {
 		chain.ProcessFilter(req, resp)
```
dashpole: Recovered panics still produce a span with `Unset` status, no `error.type`, and no status code. If you want it in scope, a recover-and-repanic in the filter is the standard shape (note `forbidigo` in `.golangci.yml:70-72` bans `Span.RecordError`, so set attributes, not an exception event):.

instrumentation/github.com/emicklei/go-restful/otelrestful/restfultest_test.go
```diff
@@ -188,10 +192,124 @@ func TestSpanStatus(t *testing.T) {
...
+	require.Eventually(t, func() bool {
+		return len(sr.Ended()) == 1
+	}, 1*time.Second, 10*time.Millisecond)
+
+	spans := sr.Ended()
+	require.Len(t, spans, 1)
+	span := spans[0]
+	assert.Equal(t, codes.Error, span.Status().Code)
+	assert.Contains(t, span.Attributes(), attribute.Int("http.response.status_code", http.StatusInternalServerError))
+	assert.Contains(t, span.Attributes(), otelsemconv.ErrorType(context.Canceled))
```
dashpole: `TestClientDisconnect`'s `error.type` assertion is tautological. Assert the literal string you intend to ship, e.g.

CHANGELOG.md
```diff
@@ -18,6 +18,7 @@ The next release will require at least [Go 1.26].
```
dashpole: No documentation of the `error.type` values emitted. Two or three lines in `doc.go` listing what `error.type` can be (`"<5xx status code>"`, plus whatever F2 settles on for cancellation/deadline).

dashpole: Minor nits

XSAM (approved):

PR #9388, opened 2026-07-29
detectors/consul: add Consul resource detector

### LINKED ISSUE #8957: [detectors/consul] Contribute consul detector from resourcedetectionprocessor
## Problem Statement

`collector-contrib/processor/resourcedetectionprocessor/internal/consul` implements a resource detector that collects resource attributes from the local Consul agent. There is currently no equivalent in `go-contrib/detectors/`, which means Consul-based detection is only available to collector-contrib pipelines and not to any Go SDK user.

## Proposed Solution

Add a new `detectors/consul` package implementing the Consul detector as a `resource.Detector`, ported from the collector-contrib implementation, so that `collector-contrib/processor/resourcedetectionprocessor` can delegate to it and any other Go SDK user can adopt it.

## Prior Art

- Collector-contrib implementation to port from: [`processor/resourcedetectionprocessor/internal/consul`](https://github.com/open-telemetry/opentelemetry-collector-contrib/tree/main/processor/resourcedetectionprocessor/internal/consul)

## Additional Context

- Queries the local Consul agent's self endpoint.
- Emits: `host.name` (node name), `cloud.region` (datacenter), `host.id` (node ID), and any user-configured Consul metadata keys as resource attributes.

paulojmdias:
Ports the Consul Cloud resource detector from `processor/resourcedetectionprocessor/internal/consul` in
`opentelemetry-collector-contrib` to a new standalone Go module `go.opentelemetry.io/contrib/detectors/consul`.

Detected attributes:
- `cloud.region`
- `host.id`
- `host.name`

Also registers the detector in `detectors/autodetect` as `IDConsul` (`"consul"`).

Fixes #8957
Relates #8940


detectors/consul/consul.go
```diff
@@ -0,0 +1,270 @@
...
+// only cancels the wait. The client timeout bounds that goroutine.
+func self(ctx context.Context, client *api.Client) (map[string]map[string]any, error) {
+	type result struct {
+		self map[string]map[string]any
+		err  error
+	}
+
+	// Buffered so the goroutine never blocks once ctx is done.
+	ch := make(chan result, 1)
+	go func() {
```
ps-mir, line 194: Need a Additional note in `WithClient`, as client created without timeout could result in go routine leak inside `self`.

detectors/consul/consul.go
```diff
@@ -0,0 +1,270 @@
...
+		s, err := client.Agent().Self()
```
ps-mir, line 195: Endpoint connectivity seems to be detection mechanism. 

Can we separate out dial failure (tcp not accepted or address didn't resolve) as detection specific failures and return nil.

That would reduce a lot of error noise for non-detection. Error past dial can be returned as it is.

ps-mir (commented):
Overall change looks ok. Have left some comment on separating specific errors to reduce detection based errors.

detectors/consul/consul.go
paulojmdias: Good catch, added.

detectors/consul/consul.go
paulojmdias: Agreed, I added it. I would like to have @dashpole input here

detectors/consul/consul.go
```diff
@@ -0,0 +1,284 @@
...
+	agent, err := self(ctx, client)
+	if err != nil {
+		if isDialFailure(err) {
+			// Nothing answered: no Consul agent here.
+			return resource.Empty(), nil
+		}
+		return nil, fmt.Errorf("failed to get local agent information: %w", err)
+	}
+
+	agentCfg, ok := agent["Config"]
```
dashpole, line 239: Can agent["Config"] is be null in the Consul response? If so, we should check that it isn't nil even if it is in the map.

detectors/consul/consul.go
dashpole: I agree with being as specific as we can with non-detection

detectors/consul/consul.go
```diff
@@ -0,0 +1,284 @@
...
+}
+
+type optionFunc func(*config)
+
+func (f optionFunc) apply(c *config) { f(c) }
+
+// WithAddress sets the address of the Consul agent, for example
+// "127.0.0.1:8500". By default the address resolved by the Consul client is
+// used, which honors the CONSUL_HTTP_ADDR environment variable.
+func WithAddress(address string) Option {
```
dashpole, line 53: Thats a lot of options!

I assume these are needed to be able to support the current collector configuration?

dashpole (approved):
just nits

detectors/consul/go.mod
```diff
@@ -0,0 +1,39 @@
+module go.opentelemetry.io/contrib/detectors/consul
+
+go 1.25.0
+
+require (
+	github.com/hashicorp/consul/api v1.32.1
+	github.com/stretchr/testify v1.11.1
+	go.opentelemetry.io/otel v1.44.1-0.20260723093731-251b96b24897
+	go.opentelemetry.io/otel/sdk v1.44.1-0.20260625150014-c84013202f01
```
ps-mir: We also need a test for composition regression, which ensures that this detector can merge without any issues (schema conflict) with core detectors.

This test forces detector to use SDK's current main revision.

[Example](https://github.com/ps-mir/opentelemetry-go-contrib/blob/4e7137a5da3e6cfa24771abb846fa9f483495d88/detectors/docker/schema_regression_test.go)

paulojmdias: Added it, and it caught a real conflict: the detector was on schema 1.43.0 while the pinned SDK's default resource was still 1.42.0, so merging failed. Bumping otel/sdk to the current main revision fixes it. Other detectors likely have the same issue.

detectors/consul/consul.go
paulojmdias: Most of them mirror the collector's config one-for-one, so users porting over get the same knobs. Only WithTimeout and WithClient are extra. I can trim it if you'd prefer a smaller surface.

detectors/consul/consul.go
paulojmdias: Good catch, it can be null. Added the nil check plus a test for it.

ps-mir (approved):

paulojmdias:
@dashpole, to be coherent with what we want in https://github.com/open-telemetry/opentelemetry-collector-contrib/issues/49988, I pushed a [commit](https://github.com/open-telemetry/opentelemetry-go-contrib/pull/9388/commits/f2ad7542cb835a144d5c0e08a23734ca0cbafbf5) here with the prefix to the metadata. Can be removed later in the collector when the feature gate is disabled, but the desired outcome can already be updated on this side.

detectors/consul/consul.go
dashpole: WithClient seems fine.

For WithTimeout, it seems like we should expect users to pass in their own deadline via the context provided to Detect?  I'm not sure it makes sense to have a different timeout

paulojmdias: Makes sense, PTAL https://github.com/open-telemetry/opentelemetry-go-contrib/pull/9388/commits/6d6e83b02b9e87e6a094434a4be213d8fcfbebf1

paulojmdias:
@dashpole, can we get this in?

dashpole:
Needs a second approval @open-telemetry/go-approvers

PR #9394, opened 2026-07-30
fix(otelgin): set error.type on client disconnect instead of a bare 500

### LINKED ISSUE #9374: otelgin: client-initiated request termination is recorded as a 500 with span status Error and no error.type
Split out from #9358 as requested in https://github.com/open-telemetry/opentelemetry-go-contrib/issues/9358#issuecomment-5108871781. Same defect class as the `otelhttp` report, confirmed independently against `otelgin`.

### Symptom

When a client disconnects mid-request, the handler observes the cancelled request context and typically returns an error, which surfaces as a 500. `otelgin` then records `http.response.status_code=500` and sets the span status to `Error`, and sets no `error.type` at all, so a client disconnect is indistinguishable from a genuine server fault.

`Middleware` reads the status from Gin's own `c.Writer.Status()` after `c.Next()` and never inspects `c.Request.Context().Err()`:

https://github.com/open-telemetry/opentelemetry-go-contrib/blob/5f30960897bf324a81b44247dec93e51f902cfd1/instrumentation/github.com/gin-gonic/gin/otelgin/gin.go#L113-L129

`otelgin` does set `error.type`, but only from `c.Errors`. A client disconnect does not populate `c.Errors`, so that path does not fire. Because the middleware uses Gin's `ResponseWriter` rather than a wrapper, there is also no observed response-write error, and the request body is not wrapped, so there is no observed body-read error either. The metric path (`sc.RecordMetrics`) carries `StatusCode` only and no `error.type`.

### Reproduction

Confirmed on `main` at 5f30960 with a real HTTP/1.1 disconnect. The handler waits on `c.Request.Context().Done()` and then writes 500:

```go
router.Use(otelgin.Middleware("srv", otelgin.WithTracerProvider(tp)))
router.GET("/hello", func(c *gin.Context) {
    <-c.Request.Context().Done()
    c.String(http.StatusInternalServerError, "cancelled")
})
```

```
span name   = "GET /hello"
span status = Error ""
```

### Suggested fix

Mirroring the triage in https://github.com/open-telemetry/opentelemetry-go-contrib/issues/9358#issuecomment-5095882060, adapted to Gin:

1. Prefer an observed failure cause over the bare status code — request-context error, and any response-write or request-body-read error if `otelgin` starts observing those.
3. Preserve the observed/selected HTTP status code; for an error status with no more specific cause, use the status-code string as `error.type`. Keep the existing `c.Errors` classification working, and decide precedence between a `c.Errors` entry and a request-context error.
4. Add `error.type` to the metric attributes recorded by `sc.RecordMetrics`, per the [HTTP metric conventions](https://opentelemetry.io/docs/specs/semconv/http/http-metrics/#metric-httpserverrequestduration).
5. Add a real disconnect regression test so cancellation is detected before the middleware returns.

### Note on shared code

`otelgin/internal/semconv/server.go` is generated from `internal/shared/semconv/server.go.tmpl` and is byte-identical to the copy vendored into `otelhttp` (verified by `diff`). The same is true for `otelmux`, `otelecho` and `otelrestful`. So whatever `error.type` classification lands in the shared template is picked up by all of them, but each middleware still needs its own wiring to detect and pass the failure cause, plus its own regression test.

pujitha24:
This is the same client-disconnect problem in `otelgin`.

Today `otelgin` gets `error.type` from errors explicitly recorded through `c.Error(...)`. If the client disconnects and the request context is cancelled, there may be no Gin error recorded, so the request can show up as a 500/error span without any indication that cancellation was the cause.

This keeps the existing `c.Errors` behavior first. When there isn't a recorded Gin error and the response is an error, it falls back to the request context error and records that as `error.type`. The attribute is also added to the HTTP server metrics.

I added a real-server regression test that cancels the request while the handler is running. It reproduces the missing attribute before this change and passes afterward, including repeated race runs.

Fixes #9374


CHANGELOG.md
```diff
@@ -53,6 +53,7 @@ This project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0.htm
...
 - Strip connection number suffix from connection ID in `go.opentelemetry.io/contrib/instrumentation/go.mongodb.org/mongo-driver/v2/mongo/otelmongo` to prevent unbounded metric cardinality. (#9352)
 - Strip connection number suffix from connection ID in `go.opentelemetry.io/contrib/instrumentation/go.mongodb.org/mongo-driver/mongo/otelmongo` so the `network.peer.address` and `net.peer.name` span attributes reflect the actual host instead of the raw connection ID. (#9352)
+- Set `error.type` on the span and on the request-duration, request-body-size, and response-body-size metrics when a client disconnects mid-request in `go.opentelemetry.io/contrib/instrumentation/github.com/gin-gonic/gin/otelgin`, using the request context's cancellation error as the classification source when the handler has not already recorded an error via `c.Error`. Previously a disconnect was recorded with span status `Error` and no `error.type`, indistinguishable from a genuine server fault.
```
```suggestion
- Set `error.type` on the span and on the request-duration, request-body-size, and response-body-size metrics when a client disconnects mid-request in `go.opentelemetry.io/contrib/instrumentation/github.com/gin-gonic/gin/otelgin`, using the request context's cancellation error as the classification source when the handler has not already recorded an error via `c.Error`. Previously a disconnect was recorded with span status `Error` and no `error.type`, indistinguishable from a genuine server fault. (#9394)
```

instrumentation/github.com/gin-gonic/gin/otelgin/gin.go
```diff
@@ -136,6 +148,9 @@ func Middleware(service string, opts ...Option) gin.HandlerFunc {
 		if cfg.GinMetricAttributeFn != nil {
 			additionalAttributes = append(additionalAttributes, cfg.GinMetricAttributeFn(c)...)
 		}
+		if errorTypeAttr != (attribute.KeyValue{}) {
```
dashpole: Using errorTypeAttr.Valid() is standard for checking if a KeyValue attribute is initialized.
```suggestion
		if errorTypeAttr.Valid() {
```

instrumentation/github.com/gin-gonic/gin/otelgin/gin_test.go
```diff
@@ -616,3 +618,66 @@ func TestMetrics(t *testing.T) {
...
+		return len(sr.Ended()) == 1
+	}, time.Second, 10*time.Millisecond, "handler should finish and end the span after the client disconnects")
+
+	span := sr.Ended()[0]
+	assert.Equal(t, codes.Error, span.Status().Code)
+	assert.Contains(t, span.Attributes(), attribute.Int("http.response.status_code", http.StatusInternalServerError))
+	assert.Contains(t, span.Attributes(), semconv.ErrorType(context.Canceled))
+
+	rm := metricdata.ResourceMetrics{}
+	require.NoError(t, reader.Collect(t.Context(), &rm))
```
dashpole, line 689: Indexing Metrics[2] directly can be brittle if metric registration order changes. Consider iterating through ScopeMetrics[0].Metrics to find the metric by name.

instrumentation/github.com/gin-gonic/gin/otelgin/gin_test.go
```diff
@@ -616,3 +618,66 @@ func TestMetrics(t *testing.T) {
...
+	require.Len(t, rm.ScopeMetrics, 1)
+	durationMetric, ok := rm.ScopeMetrics[0].Metrics[2].Data.(metricdata.Histogram[float64])
+	require.True(t, ok)
```
dashpole, line 710: Consider asserting that the error.type attribute value matches context.Canceled to ensure the attribute value is correct.

pujitha24:
Thanks for the review! Addressed all four points:

- Added `(#9394)` to the CHANGELOG entry.
- Switched to `errorTypeAttr.Valid()` instead of comparing against the zero value.
- `TestClientDisconnect` now looks up the request-duration metric by name (`http.server.request.duration`) instead of indexing `Metrics[2]`.
- The test now also asserts the `error.type` attribute's value matches `context.Canceled` (via the same `semconv.ErrorType()` helper used for the span assertion), not just that the key is present.

instrumentation/github.com/gin-gonic/gin/otelgin/gin.go
```diff
@@ -113,18 +113,30 @@ func Middleware(service string, opts ...Option) gin.HandlerFunc {
...
+				errorTypeAttr = otelsemconv.ErrorTypeOther
+			}
+			span.SetAttributes(errorTypeAttr)
+		case spanCode == codes.Error:
+			// No explicit c.Errors entry, but the response is a server
+			// error. If the client disconnected mid-request, the request
+			// context carries the real cause instead of a generic 5xx.
+			if reqErr := c.Request.Context().Err(); reqErr != nil {
+				errorTypeAttr = otelsemconv.ErrorType(reqErr)
+				span.SetAttributes(errorTypeAttr)
```
dashpole, line 151: Should we set the span status here?

instrumentation/github.com/gin-gonic/gin/otelgin/gin_test.go
```diff
@@ -616,3 +618,77 @@ func TestMetrics(t *testing.T) {
...
+	assert.Contains(t, span.Attributes(), attribute.Int("http.response.status_code", http.StatusInternalServerError))
+	assert.Contains(t, span.Attributes(), semconv.ErrorType(context.Canceled))
+
+	rm := metricdata.ResourceMetrics{}
+	require.NoError(t, reader.Collect(t.Context(), &rm))
+	require.Len(t, rm.ScopeMetrics, 1)
+
+	var durationMetric *metricdata.Metrics
+	for i, m := range rm.ScopeMetrics[0].Metrics {
+		if m.Name == "http.server.request.duration" {
```
dashpole, line 784: Since the PR description and changelog state that error.type is recorded on request-duration, request-body-size, and response-body-size metrics upon a client disconnect, consider verifying that the attribute is present on the body size metrics here as well, not just request duration.

pujitha24:
Thanks for the follow-up review! Addressed both points:

- Set an explicit span status message (`span.SetStatus(codes.Error, reqErr.Error())`) in the client-disconnect branch, mirroring the pattern the `c.Errors` branch already uses just above it.
- `TestClientDisconnect` now asserts the `error.type` attribute value on `http.server.request.body.size` and `http.server.response.body.size` as well, not just `http.server.request.duration`.

On the `protect-released-changelog` check that's currently red: it looks unrelated to this diff — main cut release v1.45.0 on 2026-08-03, which moved the entries this branch still has under `[Unreleased]` into the new released section, and that's what the check is flagging. Happy to rebase once this round of review settles, or let me know if you'd rather I do it now.

instrumentation/github.com/gin-gonic/gin/otelgin/gin.go
```diff
@@ -113,18 +113,31 @@ func Middleware(service string, opts ...Option) gin.HandlerFunc {
...
 			span.SetStatus(codes.Error, c.Errors.String())
 			if len(c.Errors) == 1 {
-				span.SetAttributes(otelsemconv.ErrorType(c.Errors[0].Err))
+				errorTypeAttr = otelsemconv.ErrorType(c.Errors[0].Err)
 			} else {
-				span.SetAttributes(otelsemconv.ErrorTypeOther)
+				errorTypeAttr = otelsemconv.ErrorTypeOther
+			}
+			span.SetAttributes(errorTypeAttr)
+		case spanCode == codes.Error:
```
MrAlias: This gate makes the selected HTTP status a prerequisite for recognizing the disconnect. If the handler waits for `Context().Done()` and returns without writing an error response, Gin keeps status 200, so the span remains `Unset` and both the span and metrics omit `error.type`.

Could we classify a non-nil request-context error independently of `spanCode` while preserving the selected status? A bounded live-disconnect test where the handler only waits for `Context().Done()` and returns would protect this path.

MrAlias (commented):
The current change covers the reproduced 500 response, but it still leaves part of the client-disconnect lifecycle unclassified.

pujitha24:
Thanks for catching that — you're right that gating on `spanCode == codes.Error` made the disconnect classification depend on the handler happening to write an error status. I changed that branch from `case spanCode == codes.Error` to a bare `default`, so `error.type` is now classified from the request context's error whenever there's no explicit `c.Errors` entry, independent of the response status. The span status message override (`span.SetStatus(codes.Error, reqErr.Error())`) still only fires when `spanCode == codes.Error`, so a non-error status like the default 200 is preserved as you described.

Added `TestClientDisconnectWithoutErrorStatus`, a live-disconnect test where the handler only waits on `<-c.Request.Context().Done()` and returns without writing a status. It asserts the span keeps its non-error status, `http.response.status_code` stays 200, and `error.type=context.Canceled` is set on both the span and the `http.server.request.duration` metric. Both this test and the existing `TestClientDisconnect` pass, along with the rest of the package suite.

dashpole (approved):

instrumentation/github.com/gin-gonic/gin/otelgin/gin.go
```diff
@@ -113,18 +113,38 @@ func Middleware(service string, opts ...Option) gin.HandlerFunc {
...
+			// when the handler never wrote a response that would surface
+			// it as a 5xx (e.g. the handler returned after observing
+			// cancellation without writing a status). Classify it
+			// independent of the response status, but only override the
+			// span status message when the status code already selected
+			// codes.Error so a successful-looking status is preserved.
+			if reqErr := c.Request.Context().Err(); reqErr != nil {
+				if spanCode == codes.Error {
+					span.SetStatus(codes.Error, reqErr.Error())
+				}
```
MrAlias: My earlier request to preserve the selected status was imprecise: we should preserve `http.response.status_code=200`, but not leave the span status unset. The [HTTP conventions](https://github.com/open-telemetry/semantic-conventions/blob/v1.43.0/docs/http/http-spans.md#status) require status `Error` when a 1xx–3xx request has another detected error, and here `reqErr` identifies the client disconnect.

Could we set the span status to `Error` whenever `reqErr != nil` and update `TestClientDisconnectWithoutErrorStatus` accordingly?

```suggestion
				span.SetStatus(codes.Error, reqErr.Error())
```

MrAlias (commented):
The latest change addresses the missing classification path, but it leaves one semantic-convention mismatch in the new default-200 case.

pujitha24:
Thanks for the precision — you're right, an unset span status alongside a detected error doesn't match the HTTP semantic conventions. I removed the `spanCode == codes.Error` gate so `span.SetStatus(codes.Error, reqErr.Error())` now runs unconditionally whenever `reqErr != nil`, while `http.response.status_code` still reflects the real response status (e.g. `200`) since that's set independently earlier in the handler.

Updated `TestClientDisconnectWithoutErrorStatus` to assert `codes.Error` (previously it asserted the opposite) while keeping the `http.response.status_code == 200` assertion. Both this test and `TestClientDisconnect` pass, and the rest of the package suite is unaffected.

pujitha24:
Rebased onto current main to resolve the CHANGELOG conflict and squashed the fixup commits into one. The code change is identical (`gin.go`/`gin_test.go` untouched upstream since the branch point — verified); the changelog entry moved under the current Unreleased section. Module tests pass. @dashpole PTAL — apologies if the force-push dismissed your approval.

instrumentation/github.com/gin-gonic/gin/otelgin/gin.go
```diff
@@ -113,18 +113,37 @@ func Middleware(service string, opts ...Option) gin.HandlerFunc {
...
+			// No explicit c.Errors entry. If the client disconnected
+			// mid-request, the request context carries the real cause even
+			// when the handler never wrote a response that would surface
+			// it as a 5xx (e.g. the handler returned after observing
+			// cancellation without writing a status). Classify it
+			// independent of the response status: per the HTTP semantic
+			// conventions, span status is Error whenever a detected error
+			// exists, even if the response status itself is not an error
+			// (e.g. http.response.status_code stays 200).
+			if reqErr := c.Request.Context().Err(); reqErr != nil {
```
dashpole: If a downstream handler attaches a child context to `c.Request` (e.g. via `c.Request = c.Request.WithContext(...)` with a deferred `cancel()`), `c.Request.Context().Err()` returns `context.Canceled` when the handler finishes even if the request was successful. Should we check `savedCtx.Err()` instead so we only detect true client disconnects?

instrumentation/github.com/gin-gonic/gin/otelgin/gin.go
```diff
@@ -136,6 +155,9 @@ func Middleware(service string, opts ...Option) gin.HandlerFunc {
 		if cfg.GinMetricAttributeFn != nil {
 			additionalAttributes = append(additionalAttributes, cfg.GinMetricAttributeFn(c)...)
 		}
+		if errorTypeAttr.Valid() {
```
dashpole: If a caller configures `MetricAttributeFn` or `GinMetricAttributeFn` with a custom `error.type`, appending `errorTypeAttr` afterwards will overwrite it in the attribute set. Maybe append `errorTypeAttr` before user-provided attribute functions (i.e. put this block above `cfg.GinMetricAttributeFn != nil`) so custom attributes take precedence?

pujitha24:
Pushed a fix for both of your latest points (classifying from the saved context, and ordering error.type before the custom attribute funcs) — it's rebased and green now, let me know if anything else would help move this along.

instrumentation/github.com/gin-gonic/gin/otelgin/gin_test.go
```diff
@@ -616,3 +618,274 @@ func TestMetrics(t *testing.T) {
...
+
+	<-handlerStarted
+	cancel()
+	<-requestDone
+
+	var rm metricdata.ResourceMetrics
+	require.Eventually(t, func() bool {
+		rm = metricdata.ResourceMetrics{}
+		require.NoError(t, reader.Collect(t.Context(), &rm))
+		return len(rm.ScopeMetrics) == 1
```
pellared: Please wait for the duration metric, not merely the metric scope.

This predicate can succeed while `RecordMetrics` has recorded only the body-size histograms; `http.server.request.duration` is recorded last. The subsequent lookup can therefore return nil.

I reproduced this on head `d0538134` from the `otelgin` module directory with:

```sh
go test -run '^TestClientDisconnectCustomErrorTypeAttribute$' -count=5000 -timeout=5m .
```

That run failed twice at line 883 with:

```text
Expected value not to be nil
expected to find the http.server.request.duration metric
```

Please wait until the collected snapshot contains the named duration metric and its data point, or synchronize on middleware completion.

pellared (commented):
I found one intermittent test reliability issue; reproduction details are included inline.

instrumentation/github.com/gin-gonic/gin/otelgin/gin.go
pellared: Rechecked on current head `409122d6`. The functional behavior is addressed: disconnect classification now checks `savedCtx.Err()` independently of the selected HTTP status, and `TestClientDisconnectWithoutErrorStatus` covers the default-200 case. I am leaving this thread open because the requested bounded test is not fully addressed: the test still uses unbounded `<-handlerStarted` and `<-requestDone` receives, so a setup or handler failure can hang until the package test timeout.

instrumentation/github.com/gin-gonic/gin/otelgin/gin.go
pellared: Verified on current head `409122d6`: `span.SetStatus(codes.Error, reqErr.Error())` now runs whenever `savedCtx.Err()` is non-nil, while `http.response.status_code` still preserves the selected HTTP status. `TestClientDisconnectWithoutErrorStatus` asserts span status `Error` together with HTTP status 200. This feedback is addressed; resolving the thread.

instrumentation/github.com/gin-gonic/gin/otelgin/gin_test.go
pellared: Rechecked on current head `409122d6`; this is still not addressed. The `Eventually` predicate still returns when only `len(rm.ScopeMetrics) == 1`, and the duration-metric lookup remains outside the predicate. A 20,000-run stress test reproduced five failures at `gin_test.go:883` with `expected to find the http.server.request.duration metric`. Leaving this thread open.

pellared:
@pujitha24 please make sure to answer to the comments and request review when the PR is ready

pujitha24:
Pushed a fix for both open points already (2550c6174):

- The three disconnect tests now use a `waitOrFail` helper (5s timeout) instead of the unbounded `<-handlerStarted`/`<-requestDone` receives, so a setup or handler failure fails fast instead of hanging to the package timeout.
- `TestClientDisconnectCustomErrorTypeAttribute`'s `Eventually` predicate now polls until the `http.server.request.duration` metric itself is found, not just until a scope is present — `RecordMetrics` writes the body-size histograms before the duration histogram, so the old check could observe a partial snapshot.

Reran your repro locally (`go test -run '^TestClientDisconnectCustomErrorTypeAttribute$' -count=2000 -timeout=5m .`) with no failures, and the full package suite with `-race` is green. @pellared PTAL when you get a chance.

MrAlias (approved):

XSAM (approved):

PR #9395, opened 2026-07-30
fix(propagators/ot): return baggage.New's aggregate limit error

### LINKED ISSUE #9047: ot: Baggage extraction drops size-limit errors from baggage.New
### Description

The OT propagator can ignore aggregate errors returned by `baggage.New` when individual baggage members parse successfully. Current-tree validation showed `extractBags` can return baggage with nil error even if aggregate baggage limits fail during final construction.

### Steps To Reproduce

[10-db3dc95a227c8191bf4ffc1f342ac41b-baggage-extraction-drops-size-limit-errors-from-baggage-new.zip](https://github.com/user-attachments/files/28384860/10-db3dc95a227c8191bf4ffc1f342ac41b-baggage-extraction-drops-size-limit-errors-from-baggage-new.zip)

1. Use `go.opentelemetry.io/contrib/propagators/ot`.
2. Send valid OT trace headers plus many or large valid `ot-baggage-*` headers.
3. Ensure individual `baggage.NewMember` calls succeed but aggregate `baggage.New` validation fails.
4. Run the PoC test from the artifact zip or inspect the extracted context.
5. Observe that baggage can still be attached instead of being rejected.

### Expected behavior

Errors returned by `baggage.New` should be returned or joined unconditionally so aggregate baggage limit failures prevent invalid baggage from being attached to the context.

om7057:
## Summary

`extractBags` in `go.opentelemetry.io/contrib/propagators/ot` only returns an error when an individual `ot-baggage-*` header fails `baggage.NewMember`. Once every member parses on its own, the function still calls `baggage.New(members...)`, which enforces its own aggregate limits (max 64 members, max 8192 bytes total). If that aggregate call fails while the per-member loop produced no errors, the code takes the `err == nil` branch and returns the accumulated `nil` instead of the error `baggage.New` just returned, so the failure is silently dropped.

Fixes #9047

## Changes

- `propagators/ot/ot_propagator.go`: collapsed the two-branch return in `extractBags` into a single `multierr.Append(err, e)` call. `multierr.Append` already handles all four nil/non-nil combinations of the per-member and aggregate errors correctly, so the branch was both the bug and unnecessary.

## Test plan

- Added `TestExtractBagsAggregateLimitError`, which builds a carrier with 65 individually-valid `ot-baggage-*` headers (one over `baggage.New`'s 64-member cap) and asserts `extractBags` returns an error. This test fails against the pre-fix code, since the aggregate error was discarded whenever every member parsed individually.
- `go build`, `go vet`, `go test -race`, and `golangci-lint run` all pass clean for `propagators/ot`.


propagators/ot/ot_propagator_test.go
```diff
@@ -134,6 +135,21 @@ func TestOT_Extract(t *testing.T) {
...
+	// surfaced by baggage.New itself.
+	carrier := propagation.MapCarrier{}
+	for i := range 65 {
+		carrier.Set(fmt.Sprintf("%sk%d", baggageHeaderPrefix, i), "v")
+	}
+
+	bags, err := extractBags(carrier)
+
+	require.Error(t, err)
+	assert.LessOrEqual(t, len(bags.Members()), 64)
```
dashpole: Since baggage.New returns an empty Baggage struct on limit validation error, consider asserting that members is empty instead:
```suggestion
	assert.Empty(t, bags.Members())
```

propagators/ot/ot_propagator_test.go
```diff
@@ -134,6 +135,21 @@ func TestOT_Extract(t *testing.T) {
 	}
 }
 
+func TestExtractBagsAggregateLimitError(t *testing.T) {
+	// baggage.New enforces a 64-member limit. Each member below parses
```
ps-mir:
```suggestion
	// Test depends on 64 member limit enforced by `baggage.maxMembers`. Each member below parses
```

ps-mir (approved):
A nit comment.

propagators/ot/ot_propagator_test.go
om7057: Resolved.

om7057:
> A nit comment.

Thanks for approving.

om7057:
@pellared PR is approved, could you please check once?

propagators/ot/ot_propagator_test.go
om7057: Tightened the assertion to assert.Len(t, bags.Members(), 64).
Thank you for the suggestion @dashpole!

om7057:
@dashpole just a small reminder regarding this!

MrAlias (approved):

propagators/ot/ot_propagator.go
```diff
@@ -113,10 +113,7 @@ func extractBags(carrier propagation.TextMapCarrier) (baggage.Baggage, error) {
 		members = append(members, member)
 	}
 	bags, e := baggage.New(members...)
-	if err != nil {
-		return bags, multierr.Append(err, e)
-	}
-	return bags, err
+	return bags, multierr.Append(err, e)
```
dashpole, line 119: By returning the aggregate limit error here, `OT.Extract` (line 86) will now completely discard all baggage from the context whenever `baggage.New` exceeds limits, rather than attaching the up to 64 valid members as it did previously. Given this module is stable, we likely want to preserve the previous behavior of attaching partial baggage. We should consider adjusting `OT.Extract` to attach `bags` when non-empty even on validation errors, and pass any errors to `otel.GetErrorHandler().Handle(err)` instead of swallowing them.

See how baggage error handling is done in the core Baggage propagator: https://github.com/open-telemetry/opentelemetry-go/blob/main/propagation/baggage.go#L62-L78

om7057: Fixed. OT.Extract now attaches the successfully parsed baggage members even when baggage.New's limit validation fails, and reports the error via otel.Handle instead of swallowing it, matching propagation.Baggage's behavior in core. Added a test covering this end-to-end.

dashpole (approved):

om7057:
@dashpole @MrAlias all the checks are green, just a small reminder for the merge
Thank you!

PR #9398, opened 2026-07-31
bridges/logutil: fix recursive log conversion crash by adding cycle detection

### LINKED ISSUE #9044: otellogr: Recursive log conversion can crash on cyclic data

tharunn0:
## Description

When converting logged map, slice, array, pointer, and interface values into OpenTelemetry `attribute.Value` instances, the shared converter (`internal/shared/logutil/convert.go.tmpl`) recursively walked composite structures without cycle detection. Passing a cyclic value (such as `m := map[string]any{}; m["self"] = m`) resulted in unbounded recursion and process stack overflow crashes.

### Key Changes
- **Path-Local Cycle Detection (`visited []uintptr`)**: Added active ancestor pointer tracking during recursive traversal of `reflect.Slice`, `reflect.Map`, and `reflect.Pointer`.
- **DAG Compatibility**: Active pointers unwind automatically when recursing back up the stack via Go slice value semantics, allowing non-cyclic shared subtrees in Directed Acyclic Graphs (DAGs) to convert cleanly without false positives.
- **Safe Fallback**: Converts detected cyclic references to `attribute.StringValue("<cycle>")`.
- **Zero-Allocation for Primitives & Empty Collections**:
  - Primitive log attributes return immediately without reflect or pointer tracking.
  - Early-return optimization for empty collections (`val.Len() == 0`) bypasses `val.Pointer()` lookups and avoids any interaction with Go runtime `zerobase` addresses.
- **Generated Modules Updated**: Regenerated code and added comprehensive unit tests for `otellogr`, `otellogrus`, `otelslog`, and `otelzap`.

---

## Type of Change
- [x] Bug fix (non-breaking change which fixes an issue)

---

## Testing
Added unit tests in `convert_test.go` for all bridge modules covering:
- Direct cyclic map (`m["self"] = m`)
- Direct cyclic slice (`s[0] = s`)
- Direct cyclic pointers (`p1 -> p2 -> p1`)
- Indirect cyclic maps & slices (`m1 -> m2 -> m1`)
- Non-cyclic shared DAG maps (`left: shared, right: shared`)

#### Fixes: #9044

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -38,6 +38,7 @@ This project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0.htm
 
 ### Fixed
 
+- Fix recursive log value conversion crash on cyclic data by adding path-local cycle detection in `go.opentelemetry.io/contrib/bridges/otellogr`, `go.opentelemetry.io/contrib/bridges/otellogrus`, `go.opentelemetry.io/contrib/bridges/otelslog`, and `go.opentelemetry.io/contrib/bridges/otelzap`.
 - Fix Prometheus reader resource label filter configuration in `go.opentelemetry.io/contrib/otelconf/v0.2.0`. (#9045)
 - Apply `resource.detection/development.attributes.included` and `excluded` filtering to resource detector attributes in `go.opentelemetry.io/contrib/otelconf/x`. (#9131)
 - Honor the context configured with `WithContext` when constructing resources in `go.opentelemetry.io/contrib/otelconf` and `go.opentelemetry.io/contrib/otelconf/x`. (#9160)
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -38,7 +38,7 @@ This project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0.htm
 
 ### Fixed
 
-- Fix recursive log value conversion crash on cyclic data by adding path-local cycle detection in `go.opentelemetry.io/contrib/bridges/otellogr`, `go.opentelemetry.io/contrib/bridges/otellogrus`, `go.opentelemetry.io/contrib/bridges/otelslog`, and `go.opentelemetry.io/contrib/bridges/otelzap`.
+- Fix recursive log value conversion crash on cyclic data by adding path-local cycle detection in `go.opentelemetry.io/contrib/bridges/otellogr`, `go.opentelemetry.io/contrib/bridges/otellogrus`, `go.opentelemetry.io/contrib/bridges/otelslog`, and `go.opentelemetry.io/contrib/bridges/otelzap`.(#9398)
 - Fix Prometheus reader resource label filter configuration in `go.opentelemetry.io/contrib/otelconf/v0.2.0`. (#9045)
 - Apply `resource.detection/development.attributes.included` and `excluded` filtering to resource detector attributes in `go.opentelemetry.io/contrib/otelconf/x`. (#9131)
 - Honor the context configured with `WithContext` when constructing resources in `go.opentelemetry.io/contrib/otelconf` and `go.opentelemetry.io/contrib/otelconf/x`. (#9160)
--- a/bridges/otelslog/convert.go
+++ b/bridges/otelslog/convert.go
@@ -11,67 +11,34 @@ import (
 	"math"
 	"reflect"
 	"strconv"
+	"sync"
 	"time"
 
 	"go.opentelemetry.io/otel/attribute"
 )
 
-// convertValue converts various types to attribute.Value.
-func convertValue(v any) attribute.Value {
-	return convertValueVisited(v, nil)
+var visitedMapPool = sync.Pool{
+	New: func() any {
+		return make(map[uintptr]bool)
+	},
 }
 
-func convertValueVisited(v any, visited []uintptr) attribute.Value {
-	// Handling the most common types without reflect is a small perf win.
-	switch val := v.(type) {
-	case bool:
-		return attribute.BoolValue(val)
-	case string:
-		return attribute.StringValue(val)
-	case int:
-		return attribute.Int64Value(int64(val))
-	case int8:
-		return attribute.Int64Value(int64(val))
-	case int16:
-		return attribute.Int64Value(int64(val))
-	case int32:
-		return attribute.Int64Value(int64(val))
-	case int64:
-		return attribute.Int64Value(val)
-	case uint:
-		return convertUintValue(uint64(val))
-	case uint8:
-		return attribute.Int64Value(int64(val))
-	case uint16:
-		return attribute.Int64Value(int64(val))
-	case uint32:
-		return attribute.Int64Value(int64(val))
-	case uint64:
-		return convertUintValue(val)
-	case uintptr:
-		return convertUintValue(uint64(val))
-	case float32:
-		return attribute.Float64Value(float64(val))
-	case float64:
-		return attribute.Float64Value(val)
-	case time.Duration:
-		return attribute.Int64Value(val.Nanoseconds())
-	case complex64:
-		r := attribute.Float64("r", real(complex128(val)))
-		i := attribute.Float64("i", imag(complex128(val)))
-		return attribute.MapValue(r, i)
-	case complex128:
-		r := attribute.Float64("r", real(val))
-		i := attribute.Float64("i", imag(val))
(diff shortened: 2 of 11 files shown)


bridges/otelslog/convert.go
```diff
@@ -79,12 +83,44 @@ func convertValue(v any) attribute.Value {
...
+		if val.Len() == 0 {
+			return attribute.SliceValue()
+		}
+		if t.Kind() == reflect.Slice {
+			if val.IsNil() {
+				return attribute.Value{}
+			}
+			ptr := val.Pointer()
+			if ptr != 0 {
+				for _, p := range visited {
```
pellared: Iterating over a slice is very inefficient. Why not using `map[uintptr]bool`? We can have a `sync.Pool` of `map[uintptr]bool` to avoid heap allocations.

internal/shared/logutil/convert.go.tmpl
```diff
@@ -79,12 +83,44 @@ func convertValue(v any) attribute.Value {
 	case reflect.Struct:
 		return attribute.StringValue(fmt.Sprintf("%+v", v))
 	case reflect.Slice, reflect.Array:
+		if val.Len() == 0 {
+			return attribute.SliceValue()
+		}
+		if t.Kind() == reflect.Slice {
+			if val.IsNil() {
```
khushijain21: this code is unreachable. `val.Len()==0` will return true for both empty and nil slice and hence we will return early.

tharunn0:
Added fastConvertValue() to handle primitive types without reflection, keeping the common cases fast by avoiding pool overhead

internal/shared/logutil/convert.go.tmpl
```diff
@@ -79,12 +50,36 @@ func convertValue(v any) attribute.Value {
 	case reflect.Struct:
 		return attribute.StringValue(fmt.Sprintf("%+v", v))
 	case reflect.Slice, reflect.Array:
+		if t.Kind() == reflect.Slice {
+			if val.IsNil() {
```
MrAlias: Removing the zero-length early return changed existing behavior: typed nil slices and maps previously converted to empty collection values, but now return an invalid `attribute.Value`. Tracking zero-length slices also treats an empty subslice sharing its parent’s backing array as a cycle (`parent[0] = parent[:0]` produces `"<cycle>"`), even though it cannot recurse.

Could we return the appropriate empty collection when `val.Len() == 0` before pointer tracking? That preserves existing behavior and keeps terminal collections out of cycle detection.

MrAlias (commented):
The cycle detection direction looks sound, but the collection handling still has a conversion regression that should be addressed before merge.

internal/shared/logutil/convert.go.tmpl
```diff
@@ -11,63 +11,34 @@ import (
...
 	"strconv"
+	"sync"
 	"time"
 
 	"go.opentelemetry.io/otel/attribute"
 )
 
+var visitedMapPool = sync.Pool{
+	New: func() any {
+		return make(map[uintptr]bool)
```
MrAlias: This uses a raw address as the visit key, so overlapping slices and differently typed pointers at the same address are mistaken for cycles. Both `parent := []any{42, nil}; parent[1] = parent[:1]` and `parent := &[2]any{42, nil}; parent[1] = &parent[0]` convert the second element to `"<cycle>"`, even though each traversal terminates at `42`.

Could we key visits by type and address, include the length for slices, and add these cases to the shared test template?

MrAlias (commented):
The empty-collection handling now preserves the previous behavior, but the cycle detector still has a correctness gap in how it identifies traversed values that should be addressed before merge.

pellared:
@tharunn0, are you planning to continue working on this PR?

internal/shared/logutil/convert.go.tmpl
```diff
@@ -79,12 +56,38 @@ func convertValue(v any) attribute.Value {
 	case reflect.Struct:
 		return attribute.StringValue(fmt.Sprintf("%+v", v))
```
pellared, line 57: **Blocking correctness:** This exits the guarded traversal and starts a separate, unbounded recursive traversal inside `fmt`. For example:

```go
m := map[string]any{}
m["wrapped"] = struct{ M map[string]any }{M: m}
convertValue(m)
```

This produces a process-fatal stack overflow in `fmt.(*pp).printValue`; the outer map is registered in `visited`, but `fmt.Sprintf` cannot see that state. The logged root is a cyclic map, so this is within the stated scope of the fix. Please use a cycle-safe representation here and add a map → struct → original map regression test to the shared test template.

internal/shared/logutil/convert.go.tmpl
```diff
@@ -11,63 +11,40 @@ import (
...
-		return attribute.MapValue(r, i)
-	case time.Time:
-		return attribute.Int64Value(val.UnixNano())
-	case []byte:
-		return attribute.ByteSliceValue(val)
-	case error:
-		return attribute.StringValue(val.Error())
-	case attribute.Value:
-		return val
+	if res, ok := fastConvertValue(v); ok {
```
pellared, line 34: **Blocking performance regression:** Extracting the primitive switch into `fastConvertValue` adds another non-inlined function call to every primitive conversion. On supported Go 1.25, paired benchmarks of the exact merge-base and PR converters measured:

- `int64`: 19.05 → 38.74 ns/op (+103%)
- `[]int{1,2,3}`: 559 → 1,160 ns/op (+108%)
- one-entry map: 625.6 → 1,187 ns/op (+90%)
- pointer depth 64: 2,922 → 18,006 ns/op (+516%)

`fastConvertValue` has compiler cost 897 against an inlining budget of 80, and composites call it again from `convertValueVisited`. Please keep the common primitive switch directly in `convertValue`, enter the cycle-aware slow path only after it misses, and add direct conversion benchmarks to the shared template.

internal/shared/logutil/convert.go.tmpl
```diff
@@ -11,63 +11,40 @@ import (
...
+		return res
+	}
+
+	visited := visitedMapPool.Get().(map[visitKey]bool)
+	res := convertValueVisited(v, visited)
+	clear(visited)
```
pellared, line 40: **Performance/memory:** `clear` removes entries but does not reduce the map’s bucket capacity. A single 65,536-level pointer chain left approximately 6.3 MiB retained by this pooled map through the following GC. Multiple processor Ps can retain separate oversized maps.

The pool is also acquired for `nil`, structs, and empty collections before their early returns, adding cold and post-GC allocations to terminal values. Please handle terminal values before `Get` and avoid returning unusually large maps to the pool.

CHANGELOG.md
```diff
@@ -38,6 +38,7 @@ This project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0.htm
 
 ### Fixed
 
+- Fix recursive log value conversion crash on cyclic data by adding path-local cycle detection in `go.opentelemetry.io/contrib/bridges/otellogr`, `go.opentelemetry.io/contrib/bridges/otellogrus`, `go.opentelemetry.io/contrib/bridges/otelslog`, and `go.opentelemetry.io/contrib/bridges/otelzap`.(#9398)
```
pellared, line 41: The four bridge names are correct, but this should describe the observable behavior rather than “path-local cycle detection.” Detected references become the ordinary string `"<cycle>"`, which is indistinguishable downstream from a user-provided string with the same value. There is also a missing space before `(#9398)`.

After the struct-boundary crash is fixed, suggested wording:

```suggestion
- Prevent stack-overflow crashes while recursively converting cyclic map, slice, and pointer log values in `otellogr`, `otellogrus`, `otelslog`, and `otelzap`; detected recursive references are emitted as the string `"<cycle>"`. (#9398)
```

pellared (changes requested):
I validated this at head `4cffc074`. Direct and branching map/slice/pointer cycle detection works, and the shared template correctly generates the implementation for otellogr, otellogrus, otelslog, and otelzap.

I am requesting changes because a cyclic value passing through a struct can still cause process-fatal stack overflow, and paired benchmarks on supported Go 1.25 show material hot-path regressions. The generated tests pass, including under the race detector, but do not cover these failures. Details and reproductions are inline.

pellared:
I created https://github.com/open-telemetry/opentelemetry-go-contrib/pull/9542 instead.

PR #9400, opened 2026-08-01
fix(otelmux): set error.type on client disconnect instead of a bare 500

### LINKED ISSUE #9375: otelmux: client-initiated request termination is recorded as a 500 with span status Error and no error.type
Split out from #9358 as requested in https://github.com/open-telemetry/opentelemetry-go-contrib/issues/9358#issuecomment-5108871781. Same defect class as the `otelhttp` report, confirmed independently against `otelmux`.

### Symptom

When a client disconnects mid-request, the handler observes the cancelled request context and typically returns an error, which surfaces as a 500. `otelmux` then records `http.response.status_code=500` and sets the span status to `Error`, and sets no `error.type` at all, so a client disconnect is indistinguishable from a genuine server fault.

`otelmux` is the closest of the framework instrumentations to `otelhttp`: it uses the same `internal/request.RespWriterWrapper` (defaulting to 200, recording only what the handler wrote) and reads the status after `tw.handler.ServeHTTP` without inspecting `r.Context().Err()`:

https://github.com/open-telemetry/opentelemetry-go-contrib/blob/5f30960897bf324a81b44247dec93e51f902cfd1/instrumentation/github.com/gorilla/mux/otelmux/mux.go#L181-L207

Note that `otelmux` already plumbs `ReadError` and `WriteError` into `ResponseTraceAttrs` from the body and response-writer wrappers, but the shared semconv implementation never turns those into `error.type` — so the observed failure cause is collected and then discarded. `RecordMetrics` likewise carries `StatusCode` only, with no `error.type`.

### Reproduction

Confirmed on `main` at 5f30960 with a real HTTP/1.1 disconnect. The handler waits on `r.Context().Done()` and then writes 500:

```go
router.Use(otelmux.Middleware("srv", otelmux.WithTracerProvider(tp)))
router.HandleFunc("/hello", func(w http.ResponseWriter, r *http.Request) {
    <-r.Context().Done()
    w.WriteHeader(http.StatusInternalServerError)
    _, _ = w.Write([]byte("cancelled"))
})
```

```
span name   = "GET /hello"
span status = Error ""
```

### Suggested fix

Mirroring the triage in https://github.com/open-telemetry/opentelemetry-go-contrib/issues/9358#issuecomment-5095882060:

1. Prefer an observed response-write error, request-body read error, or request-context error as the failure cause — the first two are already available via `rww.Error()` and `bw.Error()`.
4. Add `error.type` to the metric attributes recorded by `RecordMetrics`, per the [HTTP metric conventions](https://opentelemetry.io/docs/specs/semconv/http/http-metrics/#metric-httpserverrequestduration).
5. Add a real disconnect regression test so cancellation is detected before the outer `ServeHTTP` returns.

Because `otelmux` follows `otelhttp`'s structure so closely, the fix here should be close to a direct port of whatever lands in #9358.

### Note on shared code

`otelmux/internal/semconv/server.go` is generated from `internal/shared/semconv/server.go.tmpl` and is byte-identical to the copy vendored into `otelhttp` (verified by `diff`). The same is true for `otelgin`, `otelecho` and `otelrestful`. So whatever `error.type` classification lands in the shared template is picked up by all of them, but each middleware still needs its own wiring to detect and pass the failure cause, plus its own regression test.

pujitha24:
A client disconnect in `otelmux` can currently look just like a real server failure in telemetry.

If the request context is cancelled and the handler ends up returning a 500, the span is marked as an error but there is no `error.type` explaining what actually happened.

This change sets `error.type` using the underlying failure when we have one: response write error first, then request body read error, then the request context error. The same attribute is also included on the HTTP server metrics.

I kept this scoped to `otelmux`; it doesn't change the shared generated semconv code.

I added a test using a real HTTP server and cancelled request. It reproduces the missing `error.type` before the fix and passes afterward, including under repeated race runs.

Fixes #9375


instrumentation/github.com/gorilla/mux/otelmux/mux.go
```diff
@@ -197,11 +200,36 @@ func (tw traceware) ServeHTTP(w http.ResponseWriter, r *http.Request) {
 		WriteError: rww.Error(),
 	})...)
 
+	var errorTypeAttr attribute.KeyValue
+	if spanCode == codes.Error {
```
MrAlias, line 204: This gate makes the selected status the prerequisite for detecting I/O errors. If a handler sends 2xx and the later body write fails, the span remains `Unset` and both the span and metrics omit `error.type`. Conversely, a 5xx with no wrapper or context error still has no status-code fallback, so the bare server-error case remains unclassified.

Could we route the full priority and fallback through the shared server semconv change tracked in #9404? The classifier needs to set `Error` for observed I/O failures regardless of status and use the status-code string when no more specific cause exists.

instrumentation/github.com/gorilla/mux/otelmux/muxtest_test.go
```diff
@@ -495,6 +498,99 @@ func TestHandlerWithMetricAttributesFn(t *testing.T) {
...
+		defer close(requestDone)
+		resp, doErr := srv.Client().Do(req)
+		if doErr == nil {
+			_ = resp.Body.Close()
+		}
+	}()
+
+	<-handlerStarted
+	cancel()
+	<-requestDone
```
MrAlias: Non-blocking: these receives have no timeout, and the goroutine drops `doErr`. If the request never reaches the handler or cancellation fails to complete it, this test hangs until the global test timeout. Could we bound both waits and surface the client error?

CHANGELOG.md
```diff
@@ -54,6 +54,7 @@ This project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0.htm
...
 - Strip connection number suffix from connection ID in `go.opentelemetry.io/contrib/instrumentation/go.mongodb.org/mongo-driver/mongo/otelmongo` so the `network.peer.address` and `net.peer.name` span attributes reflect the actual host instead of the raw connection ID. (#9352)
 - Record `network.protocol.version` (and `network.protocol.name`) from the response's negotiated protocol instead of the request's stamped `Proto` in `go.opentelemetry.io/contrib/instrumentation/net/http/otelhttp`, so an HTTP/1.1-shaped request that gets upgraded to HTTP/2 via ALPN is no longer misreported as HTTP/1.1 on the client span and request metrics. (#9371)
+- Set `error.type` on the span and on the request-duration, request-body-size, and response-body-size metrics when a client disconnects mid-request in `go.opentelemetry.io/contrib/instrumentation/github.com/gorilla/mux/otelmux`, using an observed response-write error, request-body read error, or the request context's cancellation error (in that order) as the classification source. Previously a disconnect was recorded with span status `Error` and no `error.type`, indistinguishable from a genuine server fault.
```
MrAlias: Non-blocking: the PR number is now known, so this entry should end with `(#9400)`.

```suggestion
- Set `error.type` on the span and on the request-duration, request-body-size, and response-body-size metrics when a client disconnects mid-request in `go.opentelemetry.io/contrib/instrumentation/github.com/gorilla/mux/otelmux`, using an observed response-write error, request-body read error, or the request context's cancellation error (in that order) as the classification source. Previously a disconnect was recorded with span status `Error` and no `error.type`, indistinguishable from a genuine server fault. (#9400)
```

MrAlias (dismissed):
This handles the reported 5xx cancellation case, but it does not yet implement the complete server error-classification behavior tracked by #9375.

pujitha24:
Thanks for the detailed look. You're right that the status-gate has real holes: a 2xx response with a later write failure leaves the span `Unset` with no `error.type`, and a bare 5xx with no observed I/O/context error still falls through unclassified. I'd rather not patch those in isolation here, since #9404 (the shared `otelhttp` classifier this template feeds via `internal/shared/semconv`) hasn't merged yet and would let `otelgin`/`otelecho`/`otelrestful`/`otelmux` share one implementation instead of diverging. Once #9404 lands I'll port its priority/fallback logic into this PR rather than keep the local gate. Let me know if you'd rather I hold this PR open until then, or land the current fix as an interim step and follow up separately.

Separately: the `TestClientDisconnect` timeout/error-surfacing nit and the CHANGELOG `(#9400)` reference were already addressed in the last commit. The `protect-released-changelog` CI failure looks unrelated to this diff — this branch predates the v1.45.0 release cut on `main` (#9413), so the locked section comparison is comparing against a differently-shaped file; I'd expect that to clear once the branch is rebased onto current `main` rather than needing a code change here.

MrAlias:
Looks good overall. I think we should wait for #9404 to land prior to moving forward here though.

pujitha24:
Thanks for the detailed feedback.

The two non-blocking nits (bounding the `TestClientDisconnect` waits and surfacing the client error, and adding `(#9400)` to the changelog entry) are already addressed in the latest commits.

On the classification gap: you're right that gating on `spanCode == codes.Error` misses a 2xx that fails during a later body write, and that a bare 5xx with no I/O/context cause still gets no `error.type`. This PR only covers the disconnect case from the #9375 reproduction (I/O-error-or-context-cancellation priority, gated on error status) — not the full priority/fallback (I/O errors regardless of status, status-code string fallback) that #9404 is still working out for the shared `internal/semconv` path.

Since #9404 hasn't merged yet and touches the generated semconv code otelmux shares, I don't want to hand-port a design that's still moving upstream. Once #9404 lands I can port the finalized classifier here to close the gap — or if you'd rather I implement the fuller behavior locally now, independent of #9404's timeline, let me know and I'll take another pass.

Issue #9376, opened 2026-07-28
otelecho: client-initiated request termination is recorded as a 500 with span status Error and no error.type


danielloader:
Split out from #9358 as requested in https://github.com/open-telemetry/opentelemetry-go-contrib/issues/9358#issuecomment-5108871781. Same defect class as the `otelhttp` report, confirmed independently against `otelecho`.

### Symptom

When a client disconnects mid-request, the handler observes the cancelled request context and typically returns an error, which surfaces as a 500. `otelecho` then records `http.response.status_code=500` and sets the span status to `Error`, and sets no `error.type` at all, so a client disconnect is indistinguishable from a genuine server fault.

`Middleware` reads the status from `c.Response().Status` after `next(c)` and never inspects the request context:

https://github.com/open-telemetry/opentelemetry-go-contrib/blob/5f30960897bf324a81b44247dec93e51f902cfd1/instrumentation/github.com/labstack/echo/otelecho/echo.go#L96-L131

`otelecho` has an extra wrinkle worth fixing in the same pass. When the handler returns an error it records the full error string as a non-semconv `echo.error` attribute:

```go
if err != nil {
    span.SetAttributes(attribute.String("echo.error", err.Error()))
    cfg.OnError(c, err)
}
```

That is unbounded-cardinality free text on a span attribute, and it is the only place the failure cause is captured — `error.type` is never set. Idiomatic Echo handlers return the error rather than writing a status themselves, so this is the common path, not an edge case. `RecordMetrics` carries `StatusCode` only, with no `error.type`.

### Reproduction

Confirmed on `main` at 5f30960 with a real HTTP/1.1 disconnect, in both handler shapes. The client issues the request with a cancellable context and cancels it after 100ms.

**Handler returns the context error (idiomatic Echo):**

```go
e.Use(otelecho.Middleware("srv", otelecho.WithTracerProvider(tp)))
e.GET("/hello", func(c echo.Context) error {
    <-c.Request().Context().Done()
    return c.Request().Context().Err()
})
```

```
span name   = "GET /hello"
span status = Error ""
attr http.route                 = /hello
attr echo.error                 = context canceled
attr http.response.body.size    = 36
attr http.response.status_code  = 500
```

**Handler writes the status itself (matching the #9358 repro):**

```go
e.GET("/hello", func(c echo.Context) error {
    <-c.Request().Context().Done()
    return c.String(http.StatusInternalServerError, "cancelled")
})
```

```
[...]

PR #9404, opened 2026-08-01
fix(otelhttp): set error.type on server spans for disconnects and errors

### LINKED ISSUE #9358: otelhttp: client-initiated request termination is recorded as a 500 with span status Error
When a client disconnects mid-request, the server handler observes the cancelled request context and typically returns an error, which our handlers surface as a 500. otelhttp then records `http.response.status_code=500` and sets the span status to Error, because `serveHTTP` reads the status from `RespWriterWrapper` (which defaults to 200 and only records what the handler wrote) and never inspects `r.Context().Err()`. 

No response reached the client, so per the HTTP semantic conventions `http.response.status_code` should arguably be absent, since it is required only if a status was received or sent, and `error.type` should carry a low cardinality identifier instead. The client side already does this in `transport.go` via `semconv.ErrorType`, but the server path never sets `error.type` at all. Is the current server behaviour intentional, and would a check on the request context after `next.ServeHTTP` to distinguish a client disconnect from a genuine 500 be in scope?

DuttaNeel07:
Fixes #9358

Adds `error.type` to server-side spans and metrics in `otelhttp` when a
request ends without a more specific outcome, per the triage in #9358:

1. Prefers an observed response write error, request-body read error,
   or request-context error (in that order) as the failure cause.
2. Sets a low-cardinality `error.type` for that cause (`context_canceled`,
   `context_deadline_exceeded`, or the error's Go type name).
3. Preserves the existing HTTP status code behavior, nothing about
   status recording changes; for an error status with no more specific
   cause, the status code itself is used as `error.type`.
4. Adds a regression test covering the disconnect scenario.

**On the regression test approach:** `TestServeHTTPClientDisconnect`
starts a live `httptest.NewServer`, lets the handler begin, cancels the
client's request context mid-flight, and waits for the span to end via
`require.Eventually`. That covers the in-flight disconnect boundary.
Additional handler tests inject write, non-EOF read, and combined
causes to lock in write > read > context priority, and they assert the
literal `error.type` values `context_canceled` and
`context_deadline_exceeded`.

This PR covers `otelhttp` only. The same fix is tracked for otelgin
(#9374), otelmux (#9375), otelecho (#9376), and otelrestful (#9377)
as follow up PRs.


MrAlias (commented):
The current branch does not build, and the server failure classification is not yet wired consistently through spans and metrics.

internal/shared/semconv/server.go.tmpl
```diff
@@ -20,7 +21,10 @@ import (
 	"go.opentelemetry.io/otel/attribute"
 	"go.opentelemetry.io/otel/codes"
 	"go.opentelemetry.io/otel/metric"
+<<<<<<< HEAD
```
MrAlias: These conflict markers are copied into generated Go files; the current lint job fails with `missing import path`. `httpconvtest_test.go.tmpl` has another marker block, and this template is also consumed by otelhttptrace, otelgin, otelmux, otelecho, and otelrestful while this PR only updates the otelhttp outputs.

Please resolve both template conflicts and regenerate every consumer, or keep this behavior local to otelhttp so the checked-in generated files remain consistent with their source.

internal/shared/semconv/server.go.tmpl
```diff
@@ -331,6 +360,10 @@ func (HTTPServer) ResponseTraceAttrs(resp ResponseTelemetry) []attribute.KeyValu
 	if resp.StatusCode > 0 {
 		count++
 	}
+	errAttr, hasErr := n.errorTypeAttr([]error{resp.WriteError, resp.ReadError, resp.RequestError}, resp.StatusCode)
```
MrAlias: `ResponseTraceAttrs` has an unnamed `HTTPServer` receiver, so this reference to `n` is undefined and the generated packages will not compile after the conflict markers are removed.

Could we name the receiver in the template and regenerate the outputs?

instrumentation/net/http/otelhttp/handler.go
```diff
@@ -187,17 +187,17 @@ func (h *middleware) serveHTTP(w http.ResponseWriter, r *http.Request, next http
...
-		ReadBytes:  bytesRead,
-		ReadError:  bw.Error(),
-		WriteBytes: bytesWritten,
-		WriteError: rww.Error(),
+		StatusCode:   statusCode,
+		ReadBytes:    bytesRead,
+		ReadError:    bw.Error(),
+		WriteBytes:   bytesWritten,
+		WriteError:   rww.Error(),
+		RequestError: r.Context().Err(),
```
MrAlias: The new context error can set `error.type`, but span status still comes only from `Status(statusCode)`. If the handler selected 2xx and a later write, read, or context failure prevents the response from completing, this leaves the span `Unset` even though the operation failed.

Could we set status `Error` whenever the selected write/read/context cause is non-nil, regardless of the HTTP status? The HTTP semantic conventions call for detected incomplete sends or receives to be recorded as errors.

instrumentation/net/http/otelhttp/handler.go
```diff
@@ -187,17 +187,17 @@ func (h *middleware) serveHTTP(w http.ResponseWriter, r *http.Request, next http
...
 	})...)
 
 	h.semconv.RecordMetrics(ctx, semconv.ServerMetricData{
```
MrAlias, line 250: `MetricAttributes.Err` is read by `RecordMetrics`, but this handler never sets it. A cancelled request that selected 500 is reduced to `error.type="500"`, and a failed 2xx request records no metric `error.type` at all.

Could we select the write/read/context error once, in the documented priority, and pass it into this metric data? An integration assertion on `http.server.request.duration` would protect the wiring.

instrumentation/net/http/otelhttp/handler_test.go
```diff
@@ -791,3 +792,41 @@ func BenchmarkHandlerServeHTTP(b *testing.B) {
...
+			}
+			w.WriteHeader(http.StatusOK)
+		}), "test_handler",
+		WithTracerProvider(tp),
+	)
+
+	ctx, cancel := context.WithCancel(context.Background())
+	cancel() // already canceled before the request is even served
+
+	r := httptest.NewRequestWithContext(ctx, http.MethodGet, "/", http.NoBody)
```
MrAlias: This only proves that otelhttp reads a context that was cancelled before `ServeHTTP` began. It does not cover the reported boundary where `net/http` cancels an in-flight server request after the client connection closes.

Could we retain a bounded live-disconnect test, or otherwise exercise that boundary without relying on an already-cancelled request?

CHANGELOG.md
```diff
@@ -48,6 +48,10 @@ This project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0.htm
 - `TextMapPropagator` in `go.opentelemetry.io/contrib/propagators/autoprop` returns the no-op propagator for empty input, matching the behavior of `none`. An unknown `OTEL_PROPAGATORS` value still returns an error with a nil propagator so `NewTextMapPropagator` falls back to the default TraceContext and Baggage propagators instead of disabling propagation. (#9163)
 - Preserve error-valued attributes nested in a group as grouped attributes instead of silently dropping them in `go.opentelemetry.io/contrib/bridges/otelslog`. (#9238)
 - Fix a data race in `go.opentelemetry.io/contrib/bridges/otelslog` where concurrent `Handle` calls could corrupt each other's log attributes because `kvBuffer.KeyValues` returned a slice aliasing a shared buffer. (#9229)
+- Add `error.type` to server spans and metrics in `otelhttp` when a request
+  ends without a more specific status, preferring an observed response
+  write error, request-body read error, or request-context error as the
+  cause, and falling back to the status code otherwise. (#9358)
```
MrAlias: Non-blocking: this entry should name `go.opentelemetry.io/contrib/instrumentation/net/http/otelhttp` and end with `(#9404)` rather than the issue number. The branch also predates the latest release, so the entry needs to remain in the current `Unreleased` section after rebasing.

MrAlias:
@DuttaNeel07 are you still able to work on this? This is blocking #9400.

DuttaNeel07:
> @DuttaNeel07 are you still able to work on this? This is blocking #9400.

I got caught up with some other work and couldn’t get to this sooner. I’ll prioritize it and try to get it finished by today

CHANGELOG.md
DuttaNeel07: Fixed. The entry now names go.opentelemetry.io/contrib/instrumentation/net/http/otelhttp and ends with #9404, and it sits in the current Unreleased section.

instrumentation/net/http/otelhttp/handler_test.go
DuttaNeel07: Replaced it with TestServeHTTPClientDisconnect. It starts a real httptest.NewServer, lets the handler begin, cancels the client's request context mid flight, and waits for the span to end via require.Eventually.

instrumentation/net/http/otelhttp/handler.go
DuttaNeel07: Fixed. The span status is now forced to codes.Error whenever we detect a response-write, body-read, or context-cancellation cause, independent of the HTTP status the handler wrote.

internal/shared/semconv/server.go.tmpl
DuttaNeel07: Reworked this so internal/shared/semconv/server.go.tmpl is left untouched. The cause or error.type classification now lives entirely in handler.go, using go.opentelemetry.io/otel/semconv/v1.43.0 directly. That keeps the PR scoped to otelhttp

internal/shared/semconv/server.go.tmpl
DuttaNeel07: No longer applicable

instrumentation/net/http/otelhttp/handler.go
```diff
@@ -187,26 +190,62 @@ func (h *middleware) serveHTTP(w http.ResponseWriter, r *http.Request, next http
...
+	spanCode, spanMsg := h.semconv.Status(statusCode)
+
+	// Select the failure cause once, in priority order: an observed response
+	// write error, a request-body read error, or the request context's error
+	// (e.g. the client disconnected mid-request). A detected failure is
+	// recorded as an error regardless of the HTTP status the handler set,
+	// since the response did not complete as the handler intended.
+	cause := rww.Error()
+	if cause == nil {
+		cause = bw.Error()
```
dashpole: Reading a request body to completion sets `bw.Error()` to `io.EOF`, which causes successful 200 requests that consume `r.Body` to be marked with status `Error` and `error.type`. We should ignore `io.EOF` when checking for body read errors:
```suggestion
		if err := bw.Error(); err != nil && !errors.Is(err, io.EOF) {
			cause = err
		}
```
=

instrumentation/net/http/otelhttp/handler_test.go
```diff
@@ -791,3 +794,139 @@ func BenchmarkHandlerServeHTTP(b *testing.B) {
...
+	var foundSpanErrorType bool
+	for _, attr := range span.Attributes() {
+		if attr.Key == otelsemconv.ErrorTypeKey {
+			spanErrorTypeAttr = attr
+			foundSpanErrorType = true
+			break
+		}
+	}
+	require.True(t, foundSpanErrorType, "expected an error.type attribute on the span")
+	assert.NotEmpty(t, spanErrorTypeAttr.Value.AsString())
```
dashpole: Because the handler writes 500, `assert.NotEmpty` would pass even if context cancellation wasn't detected and it fell back to status code `error.type="500"`. Should we assert `assert.Equal(t, "context.canceledError", spanErrorTypeAttr.Value.AsString())` (or avoid `WriteHeader(500)`) to verify context error classification?

pellared:
@DuttaNeel07, sorry for reminding again, but can you address the feedback? I am asking as we are planning to make a new release soon.

DuttaNeel07:
> @DuttaNeel07, sorry for reminding again, but can you address the feedback? I am asking as we are planning to make a new release soon.

Sorry for the delay! I'll address the feedback and push the changes shortly. I'll make sure the changes are done before the release.

danielloader:
@DuttaNeel07 if you need me to finish this off just say because I'm keen to have this out the door for the pending release.

DuttaNeel07:
> @DuttaNeel07 if you need me to finish this off just say because I'm keen to have this out the door for the pending release.

@danielloader thanks, but no need, the updated branch is pushed with @dashpole's feedback addressed and rebased on latest main. Please take another look when you can.

instrumentation/net/http/otelhttp/handler.go
```diff
@@ -187,26 +192,68 @@ func (h *middleware) serveHTTP(w http.ResponseWriter, r *http.Request, next http
...
+
+	if cause != nil {
+		spanCode = codes.Error
+	}
+	span.SetStatus(spanCode, spanMsg)
+
+	var errorType attribute.KeyValue
+	hasErrorType := cause != nil
+	if cause != nil {
+		errorType = otelsemconv.ErrorType(cause)
```
MrAlias, line 227: `otelsemconv.ErrorType` does not produce the `context_canceled` and `context_deadline_exceeded` values promised by the PR description: `context.Canceled` becomes `*errors.errorString`, while `context.DeadlineExceeded` becomes `context.deadlineExceededError`. The new test compares this helper to itself, so it cannot detect that mismatch.

Could we normalize these two sentinel errors and assert their literal values, or update the documented contract if type names are intended?

instrumentation/net/http/otelhttp/handler.go
```diff
@@ -187,26 +192,68 @@ func (h *middleware) serveHTTP(w http.ResponseWriter, r *http.Request, next http
...
-	span.SetAttributes(h.semconv.ResponseTraceAttrs(semconv.ResponseTelemetry{
+
+	spanCode, spanMsg := h.semconv.Status(statusCode)
+
+	// Select the failure cause once, in priority order: an observed response
+	// write error, a request-body read error, or the request context's error
+	// (e.g. the client disconnected mid-request). A detected failure is
+	// recorded as an error regardless of the HTTP status the handler set,
+	// since the response did not complete as the handler intended.
+	cause := rww.Error()
```
MrAlias, line 206: `RespWriterWrapper.Error` and `BodyWrapper.Error` each acquire a mutex. The new success path reads both here and then reads both again for `ResponseTelemetry`, adding two avoidable locks to every successful instrumented request.

Could we capture `writeErr` and `readErr` once and reuse them for cause selection and telemetry?

instrumentation/net/http/otelhttp/handler_test.go
```diff
@@ -791,3 +794,192 @@ func BenchmarkHandlerServeHTTP(b *testing.B) {
...
+			<-r.Context().Done()
+			w.WriteHeader(http.StatusInternalServerError)
+			_, _ = w.Write([]byte("cancelled"))
+		}),
+		"test_handler",
+		WithTracerProvider(provider),
+		WithMeterProvider(meterProvider),
+	)
+
+	srv := httptest.NewServer(h)
```
MrAlias, line 819: This is now a bounded live-server cancellation test, but the PR description still says the live approach was discarded in favor of an already-cancelled context. Could we update that section so it describes the regression coverage on the current head?

instrumentation/net/http/otelhttp/handler_test.go
```diff
@@ -791,3 +794,192 @@ func BenchmarkHandlerServeHTTP(b *testing.B) {
...
+
+	histogram, ok := durationMetric.Data.(metricdata.Histogram[float64])
+	require.True(t, ok)
+	require.Len(t, histogram.DataPoints, 1)
+	metricErrorType, ok := histogram.DataPoints[0].Attributes.Value(otelsemconv.ErrorTypeKey)
+	require.True(t, ok, "expected error.type attribute on the request duration metric")
+	assert.Equal(t, spanErrorTypeAttr.Value.AsString(), metricErrorType.AsString())
+}
+
+func TestServeHTTPServerErrorSetsErrorType(t *testing.T) {
```
MrAlias, line 896: The new coverage exercises context cancellation, 5xx fallback, and EOF exclusion, but not the response-write or non-EOF body-read branches, or the documented write > read > context priority. A regression in either branch would leave the advertised behavior unprotected, and the non-EOF assignment is currently the uncovered production line.

Could we add focused handler cases that inject each error and a combined-cause case, asserting the selected `error.type` on the span and metrics?

MrAlias (commented):
The server failure classification is now wired consistently through spans and metrics, but its documented cancellation behavior, cause-precedence coverage, and hot-path wrapper access still need tightening.

DuttaNeel07:
> The server failure classification is now wired consistently through spans and metrics, but its documented cancellation behavior, cause-precedence coverage, and hot-path wrapper access still need tightening.

Thanks for the detailed review. I’m tied up with some work at the moment, so I’ll need a little time to work through these changes properly. I’ll follow up once I’ve had a chance to address the comments.

instrumentation/net/http/otelhttp/handler.go
DuttaNeel07: Fixed. `otelsemconv.ErrorType` reports Go type names, thus `context.Canceled` is mapped to `errors.errorString` and `context.DeadlineExceeded` is mapped to `context.deadlineExceededError`. `errorTypeAttr` normalizes the two sentinels with `errors.Is` to `context_canceled` and `context_deadline_exceeded`, and leaves all other errors reported as `otelsemconv.ErrorType`. The disconnect test asserts on the literal `context_canceled` value, not comparing the helper to itself.

instrumentation/net/http/otelhttp/handler_test.go
DuttaNeel07: Updated the description

instrumentation/net/http/otelhttp/handler.go
DuttaNeel07: Fixed. `writeErr` and `readErr` are captured once after the handler returns and reused for cause selection and `ResponseTelemetry`, so the success path no longer takes the wrapper locks twice.

instrumentation/net/http/otelhttp/handler.go
```diff
@@ -187,26 +193,70 @@ func (h *middleware) serveHTTP(w http.ResponseWriter, r *http.Request, next http
...
+		WriteError: writeErr,
+	})
+	if hasErrorType {
+		attrs = append(attrs, errorType)
+	}
+	span.SetAttributes(attrs...)
+
+	additionalAttributes := append(labeler.Get(), h.metricAttributesFromRequest(r)...)
+	if hasErrorType {
+		additionalAttributes = append(additionalAttributes, errorType)
```
dashpole, line 250: Appending `errorType` after `labeler.Get()` and `metricAttributesFromRequest` causes it to overwrite any caller-supplied `error.type` under `attribute.NewSet` last-value-wins rules. Consider appending `errorType` before them so user overrides take precedence, similar to `otelgin`.

DuttaNeel07: I moved the detected error.type in front of the labeler and WithMetricAttributesFn attrs

pellared (commented):
Thanks, this now covers the main cancellation and cause-priority cases. I left three inline comments. The current lint failure is already reported by CI, so I did not duplicate it here.

instrumentation/net/http/otelhttp/handler.go
```diff
@@ -187,26 +193,73 @@ func (h *middleware) serveHTTP(w http.ResponseWriter, r *http.Request, next http
...
-	span.SetAttributes(h.semconv.ResponseTraceAttrs(semconv.ResponseTelemetry{
+
+	spanCode, spanMsg := h.semconv.Status(statusCode)
+
+	// Select the failure cause once, in priority order: an observed response
+	// write error, a request-body read error, or the request context's error
+	// (e.g. the client disconnected mid-request). A detected failure is
+	// recorded as an error regardless of the HTTP status the handler set,
+	// since the response did not complete as the handler intended.
+	writeErr := rww.Error()
```
pellared, line 207: `httpsnoop.Wrap` preserves the underlying writer’s `io.ReaderFrom`, but this middleware does not hook that method, so calls such as `io.Copy(w, r.Body)` bypass `rww.Write` and leave `writeErr` nil. With a real `httptest` server and a reset client connection, `io.Copy` returned a `*net.OpError` for `broken pipe` while the span recorded `error.type="context_canceled"`, so the documented write > read > context priority was not applied. Please track `ReadFrom` through the response wrapper and add an `io.Copy` regression test.

instrumentation/net/http/otelhttp/handler.go
```diff
@@ -187,26 +193,73 @@ func (h *middleware) serveHTTP(w http.ResponseWriter, r *http.Request, next http
...
+	if cause == nil {
+		// Reading a request body to completion reports io.EOF as the last read
+		// error; that is the success path and must not be classified as a
+		// failure.
+		if readErr != nil && !errors.Is(readErr, io.EOF) {
+			cause = readErr
+		}
+	}
+	if cause == nil {
+		cause = r.Context().Err()
```
pellared, line 219: `r` has already been handed to downstream middleware, which may replace its context in place and defer-cancel a handler-scoped child before returning. In that composition, a successful 2xx is reclassified as `Error` and all three metrics gain `error.type=context_canceled`, even though the original server request is still live. Please inspect the middleware-owned `ctx.Err()` instead; the context returned by `Tracer.Start` remains linked to the original request cancellation but cannot be replaced downstream.

instrumentation/net/http/otelhttp/handler.go
```diff
@@ -187,26 +193,73 @@ func (h *middleware) serveHTTP(w http.ResponseWriter, r *http.Request, next http
...
-		WriteError: rww.Error(),
-	})...)
+		WriteError: writeErr,
+	})
+	if hasErrorType {
+		attrs = append(attrs, errorType)
+	}
+	span.SetAttributes(attrs...)
+
+	var additionalAttributes []attribute.KeyValue
```
pellared, line 248: On successful requests with Labeler attributes, starting from a nil slice copies the slice already allocated by `labeler.Get()`. A one-label benchmark regressed from 32 to 33 allocations per request and from 2850 to 2914 B/op. Could this start with `additionalAttributes := labeler.Get()` and prepend `errorType` only on the failure path? That preserves caller precedence without adding an allocation to every labeled request.

PR #9419, opened 2026-08-05
Add timeout on Kubernetes API calls for EKS detector

### LINKED ISSUE #9073: aws: EKS detector lacks timeout on Kubernetes API calls
### Description

The EKS detector relies entirely on caller-provided context cancellation for Kubernetes API calls. Current-tree validation showed `Detect` passes the caller context through to `isEKS` and `getClusterName`, and `getConfigMap` passes that context directly to client-go `ConfigMaps().Get(...)` without adding a timeout or deadline.

### Steps To Reproduce

[13-3bb8797755288191ba7ed4d8311136ce-eks-detector-lacks-timeout-on-kubernetes-api-calls.zip](https://github.com/user-attachments/files/28464767/13-3bb8797755288191ba7ed4d8311136ce-eks-detector-lacks-timeout-on-kubernetes-api-calls.zip)

1. Use `go.opentelemetry.io/contrib/detectors/aws/eks`.
2. Run the detector with a caller context that has no deadline, such as `context.Background()`.
3. In an EKS-like environment, make the Kubernetes API request hang or remain unreachable while ConfigMap retrieval is attempted.
4. Run the PoC from the artifact zip or inspect `detectors/aws/eks/detector.go`.
5. Observe that `Detect` can remain blocked until the caller context is canceled.

### Expected behavior

The EKS detector should apply a bounded timeout or require an explicit deadline for Kubernetes API calls so resource detection cannot hang indefinitely when callers provide a context without cancellation.

iblancasa:
Fixes #9073

### DIFF
--- a/detectors/aws/eks/detector.go
+++ b/detectors/aws/eks/detector.go
@@ -14,6 +14,7 @@ import (
 	"os"
 	"regexp"
 	"strings"
+	"time"
 
 	"go.opentelemetry.io/otel/attribute"
 	"go.opentelemetry.io/otel/sdk/resource"
@@ -30,6 +31,7 @@ const (
 	cwConfigmapName   = "cluster-info"
 	defaultCgroupPath = "/proc/self/cgroup"
 	containerIDLength = 64
+	configMapTimeout  = 10 * time.Second
 )
 
 // detectorUtils is used for testing the resourceDetector by abstracting functions that rely on external systems.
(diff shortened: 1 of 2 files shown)


dashpole (commented):
Should we apply an overall deadline to the entire Detect call, rather than just to the k8s GET requests?

MrAlias (commented):
The per-request timeout is applied in the right lifecycle; the test suite should pin this contract.

detectors/aws/eks/detector.go
```diff
@@ -163,6 +165,9 @@ func (eksDetectorUtils) fileExists(filename string) bool {
 
 // getConfigMap retrieves the configuration map from the k8s API.
 func (eksUtils eksDetectorUtils) getConfigMap(ctx context.Context, namespace, name string) (map[string]string, error) {
+	ctx, cancel := context.WithTimeout(ctx, configMapTimeout)
```
MrAlias, line 168: The current tests exercise this line only through requests that complete immediately, so they would still pass if the timeout were removed. Could we add a regression test with a custom `RoundTripper` that inspects the request context deadline? That would pin the new behavior without waiting ten seconds.

MrAlias (approved):

ps-mir (commented):
Detect can take upto ~20s in worst case. Consider clarifying this in the changelog.

A single Detect() level deadline can be a followup if needed.

PR #9429, opened 2026-08-06
fix(otelgrpc): record error.type on call duration metrics for failed RPCs

### LINKED ISSUE #9426: [Bug]: otelgrpc never sets error.type on rpc.server.call.duration / rpc.client.call.duration despite it being Conditionally Required on failure
### Component

Instrumentation: otelgrpc

Per the [RPC metrics semantic conventions](https://opentelemetry.io/docs/specs/semconv/rpc/rpc-metrics/), error.type is Conditionally Required on rpc.server.call.duration / rpc.client.call.duration — specifically "if and only if the operation failed". otelgrpc-go never sets this attribute in any case, even when the RPC clearly failed (non-OK status).

Confirmed live on v0.69.0: triggering a server-side failure and inspecting the emitted rpc.server.call.duration metric keyset shows rpc.response.status_code present (correctly reflecting the failure) but error.type absent entirely.

Additional Context
Related to the OTEL_SEMCONV_STABILITY_OPT_IN work in #8683/#8726 — that PR restores the legacy metric name and rpc.grpc.status_code attribute for rpc/old, but doesn't address error.type being unset in any mode.

### Expected behavior

When the RPC status code is not OK, error.type should be set on the recorded metric (e.g. to the gRPC status code name, per semconv guidance for RPC error types).

### Steps to Reproduce

Instrument a gRPC server
Make an RPC call that returns a non-OK status (e.g. codes.Internal).
Inspect the resulting rpc.server.call.duration metric's attributes — error.type is absent.
Component Version
v0.69.0 (also confirmed present in rpc/old / rpc/dup compat modes added in #8726 — those only revert the metric name and status-code attribute key, not the missing error.type)

### Operating System

Not relevant (reproduces on any OS — confirmed on macOS host + Linux container in minikube)

### Device Architecture

x86_64

### Go Version

1.25.0

### Component Version

v0.69.0

theycallmeloki:
fix(otelgrpc): record error.type on call duration metrics for failed RPCs

Fixes #9426

Per the [RPC metric semantic conventions](https://opentelemetry.io/docs/specs/semconv/rpc/rpc-metrics/),
`error.type` is Conditionally Required on `rpc.client.call.duration` and
`rpc.server.call.duration` — if and only if the operation failed — and SHOULD
be set to the returned status code. otelgrpc never set it.

The `stats.End` handler now appends `error.type` (e.g. `INTERNAL`) to the
duration metric attributes when the RPC ends with a non-OK gRPC status, and
continues to omit it on successful calls. The value uses the same canonical
status-code name already emitted for `rpc.response.status_code`.

Tests:
- `TestStatsHandlerHandleRPCServerErrors` (shared table, all 18 gRPC codes
  incl. the out-of-range `CODE(9999)` case) now asserts `error.type` on every
  failed code and its absence on `OK`.
- New `TestStatsHandlerErrorType` drives a real failing `UnaryCall`
  (`ResponseStatus` = `codes.Internal`) through both client and server
  handlers and asserts `error.type=INTERNAL` on the failing data point and
  its absence on a successful call, for both duration metrics.

This applies in all `OTEL_SEMCONV_STABILITY_OPT_IN` modes (rpc, rpc/old,
rpc/dup) since the attribute is added to the shared metric attributes.


instrumentation/google.golang.org/grpc/otelgrpc/grpc_stats_handler_test.go
```diff
@@ -266,6 +266,103 @@ func checkServerSpans(t *testing.T, sr *tracetest.SpanRecorder, addr string) {
...
+			for _, m := range sm.Metrics {
+				if m.Name != metricName {
+					continue
+				}
+				hist, ok := m.Data.(metricdata.Histogram[float64])
+				if !ok {
+					continue
+				}
+				for _, dp := range hist.DataPoints {
+					if v, ok := dp.Attributes.Value(attribute.Key("rpc.method")); ok && v.AsString() == method {
```
dashpole: Since both a successful and a failed call are executed, this check returns true as soon as the first data point arrives. Consider waiting until both data points are present (or extracting `okDP` and `errDP` directly inside `Eventually`) to avoid flakiness.

instrumentation/google.golang.org/grpc/otelgrpc/grpc_stats_handler_test.go
```diff
@@ -266,6 +266,103 @@ func checkServerSpans(t *testing.T, sr *tracetest.SpanRecorder, addr string) {
...
+					errDP = dp
+				} else {
+					okDP = dp
+				}
+			}
+		}
+	}
+
+	require.NotNil(t, errDP, "%s: expected a failed-call data point with error.type=INTERNAL", metricName)
+	require.NotNil(t, okDP, "%s: expected a successful-call data point without error.type", metricName)
```
dashpole: Consider also asserting that `okDP` does not have `error.type` to verify it is omitted on successful calls:
```suggestion
	require.NotNil(t, okDP, "%s: expected a successful-call data point without error.type", metricName)
	_, hasErrType := okDP.Attributes.Value(attribute.Key("error.type"))
	assert.False(t, hasErrType, "%s: successful call data point must not contain error.type", metricName)
```

instrumentation/google.golang.org/grpc/otelgrpc/stats_handler.go
```diff
@@ -407,10 +407,17 @@ func (*config) handleRPC(
...
-				metricAttrs = make([]attribute.KeyValue, 0, len(gctx.metricAttrs)+1)
+				metricAttrs = make([]attribute.KeyValue, 0, len(gctx.metricAttrs)+2)
 				metricAttrs = append(metricAttrs, gctx.metricAttrs...)
 			}
 			metricAttrs = append(metricAttrs, rpcStatusAttr)
+			// Per the RPC semantic conventions, error.type is Conditionally
+			// Required on the call duration metrics if and only if the
+			// operation failed, and SHOULD be set to the returned status code.
+			// Do not set it on successful calls.
+			if s != nil && s.Code() != grpc_codes.OK {
```
MrAlias: This applies the client rule to server metrics. The [gRPC conventions](https://github.com/open-telemetry/semantic-conventions/blob/v1.43.0/docs/rpc/grpc.md#server) treat every non-`OK` client status as an error, but only `UNKNOWN`, `DEADLINE_EXCEEDED`, `UNIMPLEMENTED`, `INTERNAL`, `UNAVAILABLE`, and `DATA_LOSS` are server errors; a cancelled client is therefore reported as `error.type=CANCELLED` on `rpc.server.call.duration`.

Could we derive `error.type` from the role-specific `recordStatus` result and update the server table expectations? `serverStatus` already implements the server classification.

MrAlias (changes requested):
The metric attribute is added in the right lifecycle, but the shared failure check does not preserve the client/server classification required by the gRPC conventions.

instrumentation/google.golang.org/grpc/otelgrpc/stats_handler.go
theycallmeloki: Ah, the shared check didn't preserve the client/server classification: I missed this. 

Mainly I'm thinking a cancelled client would be reported as `error.type=CANCELLED` on `rpc.server.call.duration`

I have addressed deriving `error.type` from the role-specific `recordStatus` that's already passed into handleRPC: `serverStatus` already implemented the server classification, so I used that directly. 

client's `recordStatus` treats every non-OK status as an error. 

I also split the code-table expectations: server metrics now carry `error.type` only for the six server-error codes, and I added a client mirror test asserting every non-OK code gets `error.type` on the client side

instrumentation/google.golang.org/grpc/otelgrpc/stats_handler.go
```diff
@@ -407,10 +407,24 @@ func (*config) handleRPC(
...
+			// operation failed, and SHOULD be set to the returned status code.
+			// The failure classification is role-specific: clients treat every
+			// non-OK status as an error, while servers only classify Unknown,
+			// DeadlineExceeded, Unimplemented, Internal, Unavailable and
+			// DataLoss as errors. Use recordStatus so the client and server
+			// rules are applied independently (e.g. a client cancellation is
+			// not a server error).
+			if s != nil {
+				if c, _ := recordStatus(s); c == codes.Error {
+					metricAttrs = append(metricAttrs, semconv.ErrorTypeKey.String(canonicalString(s.Code())))
```
MrAlias: This appends `error.type` to the attribute set shared by `duration` and `oldDuration`, which changes the legacy metrics in `rpc/old` and `rpc/dup`. The [v1.37 `rpc.client.duration` and `rpc.server.duration` conventions](https://github.com/open-telemetry/semantic-conventions/blob/v1.37.0/docs/rpc/rpc-metrics.md) did not include `error.type`; compatibility mode needs to preserve that telemetry shape so existing series and queries remain stable.

Please use separate record options: include `error.type` only when recording the stable `rpc.*.call.duration` instruments, and omit it from the legacy instruments. In `rpc/dup`, record the stable and legacy histograms with their respective attribute sets. Please also add failing-call coverage for `rpc/old` and `rpc/dup` to lock this down.

MrAlias (commented):
The role-specific error classification now matches the gRPC conventions. The remaining issue is that the stable error attribute is also applied to the legacy compatibility metrics.

instrumentation/google.golang.org/grpc/otelgrpc/stats_handler.go
theycallmeloki: Yes, I have added seperate attribute sets for `error.type`, now it only lands on `rpc.*.call.duration` the older modes keep their existing shape. Also added failing-call coverage for rpc/old and rpc/dup to lock it down. Thanks for the catch.

instrumentation/google.golang.org/grpc/otelgrpc/stats_handler.go
```diff
@@ -407,22 +407,43 @@ func (*config) handleRPC(
...
+				if c, _ := recordStatus(s); c == codes.Error {
+					stableMetricAttrs = make([]attribute.KeyValue, 0, len(metricAttrs)+1)
+					stableMetricAttrs = append(stableMetricAttrs, metricAttrs...)
+					stableMetricAttrs = append(stableMetricAttrs, semconv.ErrorTypeKey.String(canonicalString(s.Code())))
+				}
+			}
+
+			// Allocate vararg slices once.
+			stableRecordOpts := []metric.RecordOption{metric.WithAttributeSet(attribute.NewSet(stableMetricAttrs...))}
+			legacyRecordOpts := []metric.RecordOption{metric.WithAttributeSet(attribute.NewSet(metricAttrs...))}
```
MrAlias: Both attribute sets are built whenever either histogram is enabled, even though `rpc` and `rpc/old` enable only one. This puts an unused `attribute.NewSet` on every normal RPC completion, and the error path also copies `metricAttrs` for a stable set that `rpc/old` never records.

I benchmarked this commit against `12d218e` with a `ManualReader` so the histogram is enabled: successful calls went from 4 to 7 allocs (488 B to 720 B, ~42% slower), and failed calls from 4 to 8 allocs (552 B to 1,040 B, ~54% slower). Could we construct each set only in its enabled branch and copy the attributes only when `rpc/dup` actually needs both shapes? The existing End benchmark uses a provider without a reader, so both instruments are disabled and it cannot catch this regression.

MrAlias (commented):
The compatibility behavior is now correct, but the split recording path introduces a measurable hot-path regression in the common single-mode configurations.

instrumentation/google.golang.org/grpc/otelgrpc/stats_handler.go
theycallmeloki: Let me try to reproduce the same behaviour and come back with a fix. I'll try to work out setting up the benchmarking on my end so I can be thorough in testing that it doesnt regress performance.

theycallmeloki: Thank you for catching this, I wasn't looking into the performance and the mallocs when I was doing it, at this layer I definetly see the merit in hindsight. Did a sweep and caught it in this push, each record option is now built only in its enabled branch.

`error.type` gets appended to the stable set only after the legacy instruments record so only single-mode configs allocate, and they allocate to a single set, without copies.

Benched against the `12d218e` baseline: 
success is back to 4 allocs / 488 B and failed to 4 allocs / 552 B (down from 7/8)

Also added reader-enabled end benchmarks with a `ManualReader` for the success, error, and rpc/old error paths, this should help catch this class of regression.

MrAlias (approved):

instrumentation/google.golang.org/grpc/otelgrpc/grpc_stats_handler_test.go
theycallmeloki: Good catch, fixed. The data-point extraction now happens inside Eventually and it only returns once both the ok and err data points are present, so the outer assertions can't read a partial snapshot. The sibling checkLegacyNoErrorType had the same early-return race, so I applied the same fix there. Full otelgrpc package passes under -race across all three stability modes, pls recommend if there's any other edges to test I may have missed.

instrumentation/google.golang.org/grpc/otelgrpc/grpc_stats_handler_test.go
theycallmeloki: Applied exactly as suggested: the successful-call data point now asserts `error.type` is absent.

instrumentation/google.golang.org/grpc/otelgrpc/grpc_stats_handler_test.go
```diff
@@ -266,6 +267,184 @@ func checkServerSpans(t *testing.T, sr *tracetest.SpanRecorder, addr string) {
...
+				hist, ok := m.Data.(metricdata.Histogram[float64])
+				if !ok {
+					continue
+				}
+				for i := range hist.DataPoints {
+					dp := &hist.DataPoints[i]
+					if v, ok := dp.Attributes.Value(attribute.Key("rpc.method")); !ok || v.AsString() != method {
+						continue
+					}
+					if v, ok := dp.Attributes.Value(attribute.Key("error.type")); ok && v.AsString() == "INTERNAL" {
```
dashpole: Selecting `errDP` by checking `error.type` in the loop and then asserting `rpc.response.status_code` below inverts the test logic (if `error.type` is missing on failure, `errDP` stays nil and `okDP` is overwritten). Identifying `errDP` by `rpc.response.status_code == "INTERNAL"` and asserting `error.type == "INTERNAL"` on `errDP` directly tests the new attribute.

dashpole (approved):

instrumentation/google.golang.org/grpc/otelgrpc/grpc_stats_handler_test.go
theycallmeloki: a missing `error.type` would route the failed data point into `okDP` and fail on a nil `errDP` instead of on the attribute itself. Now `errDP` is selected by `rpc.response.status_code=INTERNAL` (independent of the new attribute) and `error.type=INTERNAL` is asserted on it directly, matching how the legacy check already selects its failed data point. Tests pass under -race.

PR #9434, opened 2026-08-07
fix(otelecho): set error.type on client disconnect, retain echo.error

### LINKED ISSUE #9376: otelecho: client-initiated request termination is recorded as a 500 with span status Error and no error.type

theycallmeloki:
fix(otelecho): set error.type on client disconnect, retain echo.error

Fixes #9376

When a client disconnects mid-request, `otelecho` recorded the handler's
cancellation as a bare 500 (or no status at all) with no `error.type`, and
captured the failure cause only as an unbounded-cardinality `echo.error`
free-text attribute — so a client disconnect was indistinguishable from a
genuine server fault.

## Changes

- **`error.type` classification** (low-cardinality, per the HTTP semantic conventions):
  - `context.Canceled` → `error.type=context_canceled`
  - `context.DeadlineExceeded` → `error.type=context_deadline_exceeded`
  - other handler errors → classified when they surface as a server error
    (5xx); Echo handlers idiomatically return `*echo.HTTPError` for 4xx/3xx
    responses, which are not operation failures and keep the status-derived
    span status
  - when the handler returns without an error, the request context error is
    consulted (covers handlers that observe the cancellation and return nil)
  - a 5xx response committed without a handler error (e.g.
    `c.NoContent(http.StatusInternalServerError)`) falls back to the status
    code as the error type (e.g. `error.type=500`)
- `error.type` is set on the span **and** on the request-duration,
  request-body-size, and response-body-size metrics
- **`echo.error` is retained** alongside the low-cardinality `error.type`
  attribute for backward compatibility (called out in #9376 as unbounded
  cardinality); its deprecation and removal are tracked in #9485

## Tests

- `TestClientDisconnect` — real HTTP/1.1 disconnect, idiomatic handler
  returning the context error: `error.type=context_canceled` on span +
  metrics, span status Error
- `TestClientDisconnectWithoutErrorStatus` — handler observes the
  cancellation and returns nil: span still Error with `error.type`, even
  though no 5xx response is recorded
- `TestStatusError` — includes a `NoContentServerError` case covering the
  status-code fallback: handler commits a 500 without returning an error,
  `error.type=500`
- Updated `TestServerErr`/`TestStatusError` to assert `error.type` instead
  of `echo.error` (including the 4xx case asserting `error.type` is absent)


instrumentation/github.com/labstack/echo/otelecho/echo_test.go
```diff
@@ -438,3 +444,131 @@ func TestWithOnError(t *testing.T) {
...
+	assert.Contains(t, span.Attributes(), otelsemconv.ErrorTypeKey.String("context_canceled"))
+
+	assertErrorTypeOnMetrics := func(rm metricdata.ResourceMetrics, metricName string) {
+		t.Helper()
+		for _, sm := range rm.ScopeMetrics {
+			for _, m := range sm.Metrics {
+				if m.Name != metricName {
+					continue
+				}
+				hist, ok := m.Data.(metricdata.Histogram[float64])
```
dashpole: Since `http.server.request.body.size` and `http.server.response.body.size` are `metricdata.Histogram[int64]`, this `float64` type assertion fails and silently skips evaluating data points for those metrics. Consider switching on both `metricdata.Histogram[float64]` and `metricdata.Histogram[int64]` (and asserting that data points were evaluated).

instrumentation/github.com/labstack/echo/otelecho/echo.go
```diff
@@ -94,18 +98,46 @@ func Middleware(serverName string, opts ...Option) echo.MiddlewareFunc {
...
+			// Classify the failure cause with a low-cardinality error.type
+			// instead of recording the full error string as an unbounded
+			// "echo.error" attribute. A cancelled request context (client
+			// disconnect or deadline) is always an error, independent of the
+			// response status code; other handler errors are only classified
+			// when they surface as a server error, since Echo handlers
+			// idiomatically return *echo.HTTPError for 4xx/3xx responses.
+			var errorTypeAttr attribute.KeyValue
+			switch {
+			case err != nil:
```
dashpole: Should we still fall back to `c.Request().Context().Err()` if the `errors.Is(err, context.Canceled) || errors.Is(err, context.DeadlineExceeded) || status >= 500` condition below isn't met?

instrumentation/github.com/labstack/echo/otelecho/echo.go
```diff
@@ -153,3 +188,18 @@ func spanNameFormatter(c echo.Context) string {
...
+
+// errorTypeAttrFrom returns a low-cardinality error.type attribute for err.
+// Context cancellation and deadline errors are classified with their
+// canonical values so a client disconnect is distinguishable from a genuine
+// server fault; other errors fall back to their concrete type.
+func errorTypeAttrFrom(err error) attribute.KeyValue {
+	switch {
+	case errors.Is(err, context.Canceled):
+		return otelsemconv.ErrorTypeKey.String("context_canceled")
+	case errors.Is(err, context.DeadlineExceeded):
```
dashpole, line 215: Looks like the `context_deadline_exceeded` branch is not covered by tests in `echo_test.go` or `echotest_test.go`. Mind adding a test case verifying `context.DeadlineExceeded` sets `error.type=context_deadline_exceeded` on spans and metrics?

instrumentation/github.com/labstack/echo/otelecho/echo_test.go
theycallmeloki: Fixed: the helper now switches on both `metricdata.Histogram[float64]` (duration) and `metricdata.Histogram[int64]` (body sizes), and `assert.Positive` fails the test if no data point was evaluated, so silent skips are impossible. The test passing confirms the body-size metrics genuinely carry       
 `error.type now`.

instrumentation/github.com/labstack/echo/otelecho/echo.go
theycallmeloki: Yes, added the same, when a handler returns an error that doesn't wrap `context.Canceled/DeadlineExceeded` and isn't a `5xx` or a `4xx *echo.HTTPError`, the request context may still be the real failure cause: a disconnect the handler observed but didn't propagate. Implemented as the else if fallback, also added a test for it.

instrumentation/github.com/labstack/echo/otelecho/echo.go
theycallmeloki: Done: added `TestClientDisconnectWithDeadline` asserting `error.type=context_deadline_exceeded` on both the span and all three metrics.

instrumentation/github.com/labstack/echo/otelecho/echo.go
```diff
@@ -94,18 +98,54 @@ func Middleware(serverName string, opts ...Option) echo.MiddlewareFunc {
...
+			// instead of recording the full error string as an unbounded
+			// "echo.error" attribute. A cancelled request context (client
+			// disconnect or deadline) is always an error, independent of the
+			// response status code; other handler errors are only classified
+			// when they surface as a server error, since Echo handlers
+			// idiomatically return *echo.HTTPError for 4xx/3xx responses.
+			var errorTypeAttr attribute.KeyValue
+			switch {
+			case err != nil:
+				if errors.Is(err, context.Canceled) || errors.Is(err, context.DeadlineExceeded) || status >= 500 {
```
MrAlias: This still skips the request-context error whenever `status >= 500`. With a canceled request and a handler that returns an ordinary error, the default error hook selects 500 and `error.type` is reported as `*errors.errorString` rather than `context_canceled`.

Could we evaluate both signals and prefer request-context cancellation here as well?

instrumentation/github.com/labstack/echo/otelecho/echo.go
```diff
@@ -94,18 +98,54 @@ func Middleware(serverName string, opts ...Option) echo.MiddlewareFunc {
 
 			// serve the request to the next middleware
 			err := next(c)
+
+			// Call the error hook before capturing the response status: the
+			// default OnError writes the error response via c.Error, which is
+			// what makes c.Response().Status reflect the error status code.
 			if err != nil {
-				span.SetAttributes(attribute.String("echo.error", err.Error()))
```
MrAlias, line 98: Removing `echo.error` here changes the telemetry schema for existing users. The issue scoped that compatibility change to a separate follow-up because consumers may depend on the attribute; could we retain it in this fix and handle deprecation or removal separately?

MrAlias (commented):
The cancellation classification still has a gap when a canceled request also produces a 5xx handler error, and this patch also changes an existing telemetry contract.

instrumentation/github.com/labstack/echo/otelecho/echo.go
theycallmeloki: Fixed, restructured so the request-context error is evaluated before the 5xx classification, so a disconnect or deadline observed while the handler is running wins over an ordinary handler error that surfaces as a 5xx, and `error.type` reports `context_canceled` instead of the concrete error type. 

Added `TestClientDisconnectWithServerError` for exactly that case; `TestStatusError` still verifies non-canceled 5xx classify as the concrete error type.

instrumentation/github.com/labstack/echo/otelecho/echo.go
theycallmeloki: Ok I wasn't sure of the same, for now `echo.error` is retained: it's set whenever the handler returns an error, exactly as before, with deprecation/removal left to a separate follow-up.

instrumentation/github.com/labstack/echo/otelecho/echotest_test.go
```diff
@@ -120,40 +121,39 @@ func TestError(t *testing.T) {
 	attrs := span.Attributes()
 	assert.Contains(t, attrs, attribute.String("server.address", "foobar"))
 	assert.Contains(t, attrs, attribute.Int("http.response.status_code", http.StatusInternalServerError))
-	assert.Contains(t, attrs, attribute.String("echo.error", "oh no"))
+	assert.Contains(t, attrs, otelsemconv.ErrorType(wantErr))
```
dashpole, line 125: Since `echo.error` was retained in `echo.go` for backward compatibility, should we keep the `assert.Contains(t, attrs, attribute.String("echo.error", wantErr.Error()))` assertion here to ensure it is covered by regression tests?

theycallmeloki: Good catch, apologies I missed it: have restored it. `TestError` now asserts both `error.type` (the preferred classification) and `echo.error` (the retained attribute):                                                                                                                                                                                                                       
                                                                                                                                                                                                                                                                                                                                                              
 ```go                                                                                                                                                                                                                                                                                                                                                        
   assert.Contains(t, attrs, otelsemconv.ErrorType(wantErr))                                                                                                                                                                                                                                                                                                  
   assert.Contains(t, attrs, attribute.String("echo.error", wantErr.Error()))                                                                                                                                                                                                                                                                                 
 ```                                                                                                                                                                                                                                                                                                                                                          
                                                                                                                                                                                                                                                                                                                                                              
So the backward-compat attribute stays regression-covered even though `error.type` is the primary signal.

instrumentation/github.com/labstack/echo/otelecho/echo.go
```diff
@@ -94,18 +98,56 @@ func Middleware(serverName string, opts ...Option) echo.MiddlewareFunc {
...
+			switch {
+			case errors.Is(err, context.Canceled), errors.Is(err, context.DeadlineExceeded):
+				span.SetStatus(codes.Error, err.Error())
+				errorTypeAttr = errorTypeAttrFrom(err)
+			default:
+				// Prefer the request-context error over the handler error: a
+				// client disconnect or deadline observed while the handler is
+				// running is the actual failure cause, even when the handler
+				// returns an ordinary error that surfaces as a 5xx response.
+				if reqErr := c.Request().Context().Err(); reqErr != nil {
```
dashpole: If downstream middleware creates a child context with `defer cancel()` on `c.Request()`, `c.Request().Context().Err()` will be `context.Canceled` even on successful 2xx responses. Consider checking `savedCtx.Err()` (or `ctx.Err()`) instead so scoped downstream contexts don't mark successful requests as canceled errors:
```suggestion
				if reqErr := savedCtx.Err(); reqErr != nil {
```

instrumentation/github.com/labstack/echo/otelecho/echo.go
```diff
@@ -94,18 +98,56 @@ func Middleware(serverName string, opts ...Option) echo.MiddlewareFunc {
...
+					span.SetStatus(codes.Error, reqErr.Error())
+					errorTypeAttr = errorTypeAttrFrom(reqErr)
+				} else if err != nil && status >= 500 {
```
MrAlias, line 145: When a handler commits a 5xx response and returns `nil` while the request context is still active—for example, `return c.NoContent(http.StatusInternalServerError)`—neither branch sets `errorTypeAttr`. The span is marked Error from the response status, but the span and server metrics omit `error.type`; this also misses #9376's status-code fallback when no more specific cause exists.

Could we fall back to the response status string here and add coverage for this path?

instrumentation/github.com/labstack/echo/otelecho/echo.go
```diff
@@ -94,18 +98,56 @@ func Middleware(serverName string, opts ...Option) echo.MiddlewareFunc {
...
 
+			// Retain the historical echo.error attribute for existing consumers;
+			// deprecation and removal are tracked separately.
+			if err != nil {
+				span.SetAttributes(attribute.String("echo.error", err.Error()))
+			}
+
+			// Classify the failure cause with a low-cardinality error.type
+			// instead of recording the full error string as an unbounded
+			// "echo.error" attribute. A cancelled request context (client
```
MrAlias: `echo.error` is retained immediately above, so this comment still describes the superseded removal behavior. The PR title and description make the same replacement claim, while the changelog now says removal is deferred.

Could we update the PR metadata and make this comment match the compatibility behavior?

```suggestion
			// Classify the failure cause with a low-cardinality error.type
			// alongside the retained unbounded-cardinality "echo.error"
			// attribute. A cancelled request context (client
```

MrAlias (commented):
The earlier feedback is resolved, but the fallback behavior and replacement wording still need correction.

instrumentation/github.com/labstack/echo/otelecho/echo.go
theycallmeloki: Done, now when a handler commits a 5xx and returns nil with the request context still active (like `c.NoContent(http.StatusInternalServerError)` ) `error.type` now falls back to the status code string (`error.type=500`) Added test to cover this `NoContentServerError` case in `TestStatusError`

instrumentation/github.com/labstack/echo/otelecho/echo.go
theycallmeloki: Applied the same. now describes `error.type` as classified "alongside retained unbounded-cardinality `echo.error` attribute" and updated both the PR title/description to better reflect the retention

instrumentation/github.com/labstack/echo/otelecho/echo.go
theycallmeloki: Subtle catch, thank you, appreciate it! Done and added a regression test for it as well. `savedCtx` is already captured at middleware entry (before `next(c)`), so the classification now reads `savedCtx.Err()` :: a downstream middleware's scoped child context can no longer be mistaken for a client disconnect.                                  
                                                                                                                                                                              
Added `TestDownstreamContextCancellationIgnored` (downstream `context.WithCancel + defer cancel()` , handler returns 200): it reproduced the false positive before the fix (span marked Error, `error.type=context_canceled` on a successful request) and passes after.

instrumentation/github.com/labstack/echo/otelecho/echo.go
```diff
@@ -94,18 +99,60 @@ func Middleware(serverName string, opts ...Option) echo.MiddlewareFunc {
...
-			span.SetStatus(semconvSrv.Status(status))
+			spanCode, spanMsg := semconvSrv.Status(status)
+			span.SetStatus(spanCode, spanMsg)
 			span.SetAttributes(semconvSrv.ResponseTraceAttrs(semconv.ResponseTelemetry{
 				StatusCode: status,
 				WriteBytes: c.Response().Size,
 			})...)
 
+			// Retain the historical echo.error attribute for existing consumers;
+			// deprecation and removal are tracked separately.
```
dashpole: Please provide a link to the tracking

dashpole (approved):

instrumentation/github.com/labstack/echo/otelecho/echo.go
theycallmeloki: Added: #9485 which tracks the `echo.error` deprecation and removal, added as comment inline

.github/workflows/ossf-scorecard.yml
```diff
@@ -16,4 +16,4 @@ jobs:
       contents: read # for actions/checkout
       id-token: write # for Scorecard to publish results
       security-events: write # for the SARIF upload to code scanning
-    uses: open-telemetry/shared-workflows/.github/workflows/scorecard.yml@71c77e7624e3dccefec108b135186c22e5d9d69b # v0.9.0
+    uses: open-telemetry/shared-workflows/.github/workflows/scorecard.yml@8079110d7e793bd38b1a5219ef8f8280dab9139e # v0.7.0
```
dashpole: Reverting this file on the branch leaves a diff that will downgrade `shared-workflows` from `v0.9.0` back to `v0.7.0` on `main` when merged. Rebasing the branch onto `main` and dropping the revert commit should keep `main`'s versions intact without triggering PAT workflow scope issues:
```suggestion
    uses: open-telemetry/shared-workflows/.github/workflows/scorecard.yml@71c77e7624e3dccefec108b135186c22e5d9d69b # v0.9.0
```

.github/workflows/zizmor.yml
```diff
@@ -15,4 +15,4 @@ jobs:
     permissions:
       contents: read # for actions/checkout
       security-events: write # for zizmor to upload SARIF results
-    uses: open-telemetry/shared-workflows/.github/workflows/zizmor.yml@71c77e7624e3dccefec108b135186c22e5d9d69b # v0.9.0
+    uses: open-telemetry/shared-workflows/.github/workflows/zizmor.yml@8079110d7e793bd38b1a5219ef8f8280dab9139e # v0.7.0
```
dashpole: Similar to `ossf-scorecard.yml`, this will downgrade `zizmor.yml` from `v0.9.0` back to `v0.7.0` when merged:
```suggestion
    uses: open-telemetry/shared-workflows/.github/workflows/zizmor.yml@71c77e7624e3dccefec108b135186c22e5d9d69b # v0.9.0
```

pellared:
Closing per https://github.com/open-telemetry/opentelemetry-go-contrib/pull/9613

PR #9453, opened 2026-08-11
fix(zpages): use non-deprecated attribute.Value.String for formatting

### LINKED ISSUE #9008: zpages: use non-OTLP string formatting
Change this to follow non-OTLP string formatting per spec.

Reference: https://github.com/open-telemetry/opentelemetry-specification/blob/main/specification/common/README.md#anyvalue-representation-for-non-otlp-protocols

_Originally posted by @pellared in https://github.com/open-telemetry/opentelemetry-go-contrib/pull/9006#discussion_r3286339377_

om7057:
## Summary

Closes #9008.

`zpages` formats span attributes using the deprecated `attribute.Value.Emit`. `attribute.Value.String` is the documented replacement and follows the OpenTelemetry AnyValue representation for non-OTLP protocols.

Beyond the deprecation, `Emit` can produce outright invalid output. For example, a `float64` slice containing `NaN` renders as `invalid: [NaN 1.5]` via `Emit`, while `String` renders the spec-compliant `["NaN",1.5]`.

## Test plan

- [x] Added a regression test that renders a span with a `NaN`-containing float64 slice attribute through the real `/tracez` HTTP handler and asserts the spec-compliant formatting, replacing the previous `invalid:` output
- [x] `go build ./...`, `go vet ./...`, `go test ./... -race` pass in `zpages`
- [x] `golangci-lint run` passes with 0 issues


om7057:
@dashpole a small reminder on this!

zpages/tracez_test.go
```diff
@@ -174,6 +177,52 @@ func TestTracezHandler_ServeHTTP(t *testing.T) {
 	}
 }
 
+func TestTracezHandler_FormatsAttributesUsingSpecString(t *testing.T) {
```
ps-mir, line 180: Can we make this a table driven test This fix also changes the output for 2 other cases.

- BOOLSLICE
- FLOAT64

ps-mir (commented):
Code change looks ok, a comment on covering regression for other cases.

zpages/tracez_test.go
om7057: Made it table driven. FLOAT64 (scalar `+Inf`/`-Inf` → `Infinity`/`-Infinity`) and BOOLSLICE (space- vs comma-separated) both regress differently, added cases for both alongside the original NaN case.

om7057:
@ps-mir I've resolved the comment, please check the changes made.

zpages/tracez_test.go
```diff
@@ -174,6 +177,105 @@ func TestTracezHandler_ServeHTTP(t *testing.T) {
...
+
+			resp := w.Result()
+			defer resp.Body.Close()
+
+			require.Equal(t, http.StatusOK, resp.StatusCode)
+
+			body := w.Body.String()
+			assert.Contains(t, body, tc.want)
+			assert.NotContains(t, body, "invalid:")
+			if tc.wantNot != "" {
```
ps-mir: The positive assertion already takes care of this. I think we can remove `wantNot` handling here, and from the type definition above as well.

ps-mir (commented):
I think `wantNot` can be dropped, it adds conditional path only for 1 specific test which is already covered by positive assertion.

zpages/tracez_test.go
om7057: Sure, dropped wantNot and the field from the struct.

om7057:
@ps-mir done!

> I think `wantNot` can be dropped, it adds conditional path only for 1 specific test which is already covered by positive assertion.

ps-mir (approved):
Requested changes to improve regression coverage are done.

Thanks for fixing this.

dashpole (approved):

om7057:
@dashpole small reminder on this one!

om7057:
@dashpole just a reminder on this one, ready for merge!

dashpole:
PRs need two approvals (green checks).

@open-telemetry/go-approvers

pellared (approved):

Issue #8946, opened 2026-05-08
[detectors/gcp] Audit and reconcile differences with resourcedetectionprocessor gcp detector


paulojmdias:
## Problem Statement

`go-contrib/detectors/gcp` and `collector-contrib/processor/resourcedetectionprocessor/internal/gcp` both implement a GCP resource detector (both delegating to the same underlying `GoogleCloudPlatform/opentelemetry-operations-go/detectors/gcp` library), but collector-contrib adds extra capabilities on top. Two parallel implementations in the same binary can produce inconsistent resource attributes — one seen on the collector's self-telemetry, another on user pipeline telemetry.

## Proposed Solution

Audit all attribute keys, values, semantics, and failure modes across both implementations across all GCP platforms (GCE, GKE, Cloud Run, Cloud Run Jobs, Cloud Functions, GAE, Bare Metal Solutions). Reconcile any gaps in this repo so that `go-contrib/detectors/gcp` becomes the canonical implementation, and `collector-contrib/processor/resourcedetectionprocessor` can delegate to it without loss of functionality.

## Prior Art

- Existing go-contrib detector: [`detectors/gcp`](https://github.com/open-telemetry/opentelemetry-go-contrib/tree/main/detectors/gcp)
- Collector-contrib detector: [`processor/resourcedetectionprocessor/internal/gcp`](https://github.com/open-telemetry/opentelemetry-collector-contrib/tree/main/processor/resourcedetectionprocessor/internal/gcp)
- Tracking issue: open-telemetry/opentelemetry-collector-contrib#48159

## Additional Context

Known differences to investigate:

- **GCE instance labels:** collector-contrib supports fetching GCE instance labels via the Compute Engine REST API (regex-filtered via `gce_metadata_labels` config) and emits them as `gcp.gce.instance.labels.*` attributes. go-contrib has no equivalent.
- **Bare Metal Solutions (BMS):** collector-contrib handles the BMS platform. Verify go-contrib coverage.
- **Attribute coverage:** verify all GKE, GCE, Cloud Run, Cloud Run Jobs, Cloud Functions, GAE attributes and semconv versions are aligned between both detectors.

ps-mir:
@dashpole If there are no concerns, can i take this one.

dashpole:
Happy to have your help @ps-mir 

[...]

PR #9462, opened 2026-08-12
feat(detectors/gcp): add parity for BMS, Cloud Run Worker Pool, GCE MIG, and GKE hostname

dashpole:
Part of https://github.com/open-telemetry/opentelemetry-go-contrib/issues/8946

## Description

Brings feature parity to `detectors/gcp` to match the detector capabilities present in `opentelemetry-collector-contrib`'s resourcedetectionprocessor:

1. **Bare Metal Solution (BMS)**: Check for Bare Metal Solution before the GCE metadata check, and emit BMS resource attributes (`cloud.platform = "gcp_bare_metal_solution"`, `cloud.account.id`, `host.name`, `cloud.region`).
2. **Cloud Run Worker Pool**: Handle `CloudRunWorkerPool` identically to `CloudRun`.
3. **GCE Managed Instance Group (MIG)**: Detect MIG attributes (`gcp.gce.instance_group_manager.name`, `zone`, `region`) on GCE.
4. **Best-effort GKE Hostname**: Attempt to collect `host.name` on GKE nodes when available.

This is required for the collector to migrate to using this detector: https://github.com/open-telemetry/opentelemetry-collector-contrib/pull/50211.


detectors/gcp/detector.go
```diff
@@ -147,6 +161,22 @@ func (r *resourceBuilder) addZoneOrRegion(detect func() (string, gcp.LocationTyp
...
+func (r *resourceBuilder) addManagedInstanceGroup(detect func() (gcp.ManagedInstanceGroup, error)) {
+	if mig, err := detect(); err == nil {
+		if mig.Name != "" {
+			r.attrs = append(r.attrs, attribute.String("gcp.gce.instance_group_manager.name", mig.Name))
+			switch mig.Type {
+			case gcp.Zone:
+				r.attrs = append(r.attrs, attribute.String("gcp.gce.instance_group_manager.zone", mig.Location))
+			case gcp.Region:
+				r.attrs = append(r.attrs, attribute.String("gcp.gce.instance_group_manager.region", mig.Location))
+			}
```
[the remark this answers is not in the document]
dashpole: addressed in eb5d8dbf0

ps-mir:
@dashpole Does the scope also includes changes required for detector migration ?

#8946 

If so, you can take the ownership of the task as well.

dashpole:
Ah, I forgot that issue existed.  Sorry for taking it without asking.

dashpole:
I just need to add gcp.gce.instance.labels.* support, and move the library from opentelemetry-operations-go here.

ps-mir:
> Ah, I forgot that issue existed. Sorry for taking it without asking.

No issues :)

dashpole:
> I just need to add gcp.gce.instance.labels.* support, and move the library from opentelemetry-operations-go here.

I'll do those as follow-ups

detectors/gcp/README.md
```diff
@@ -8,6 +8,7 @@ The GCP resource detector supports detecting resources on:
 * Cloud Run
 * Cloud Run jobs
 * Cloud Functions
+* Bare Metal Solution (BMS)
```
ps-mir, line 12: Do we need to specify that setting [ops agent](https://docs.cloud.google.com/bare-metal/docs/set-up-ops-agent-for-bms) is required for BMS detection to work ?

detectors/gcp/detector.go
```diff
@@ -46,7 +56,10 @@ func (d *detector) Detect(context.Context) (*resource.Resource, error) {
 		b.addZoneOrRegion(d.detector.GKEAvailabilityZoneOrRegion)
 		b.add(semconv.K8SClusterNameKey, d.detector.GKEClusterName)
 		b.add(semconv.HostIDKey, d.detector.GKEHostID)
-	case gcp.CloudRun:
+		if v, err := d.detector.GCEHostName(); err == nil && v != "" {
```
ps-mir, line 61: Err is being dropped, should be accumulated in `b.errs`.

detectors/gcp/detector_test.go
```diff
@@ -222,6 +225,147 @@ func TestDetect(t *testing.T) {
...
+				semconv.CloudAccountID("my-project"),
+				semconv.CloudPlatformGCPComputeEngine,
+				semconv.HostID("1472385723456792345"),
+				semconv.HostName("my-gke-node-1234"),
+				semconv.GCPGCEInstanceNameKey.String("my-gke-node-1234"),
+				semconv.GCPGCEInstanceHostnameKey.String("hostname"),
+				semconv.HostType("n1-standard1"),
+				semconv.CloudRegion("us-central1"),
+				semconv.CloudAvailabilityZone("us-central1-c"),
+				attribute.String("gcp.gce.instance_group_manager.name", "my-mig"),
```
ps-mir: Need semconv helper functions. Ex; `GCPGCEInstanceGroupManagerName`

detectors/gcp/detector_test.go
```diff
@@ -222,6 +225,147 @@ func TestDetect(t *testing.T) {
...
```
ps-mir: Semconv helpers (same as above)

detectors/gcp/detector_test.go
```diff
@@ -222,6 +225,147 @@ func TestDetect(t *testing.T) {
...
```
ps-mir: Semconv helpers.

detectors/gcp/detector_test.go
ps-mir:
```suggestion
			desc: "GCE without MIG",
```

ps-mir (commented):
Comments on error handling and semconv helpers in tests.

detectors/gcp/README.md
dashpole: You shouldn't need to use the ops-agent. The ops-agent is actually built on the OTel collector, so it will eventually use this code (once the collector migrates to it) to make that documentation work!

detectors/gcp/detector.go
dashpole: addressed in d6d539fb6

detectors/gcp/detector_test.go
dashpole: addressed in 74bfeb04b

detectors/gcp/detector_test.go
dashpole: addressed in a2b14f6e2

detectors/gcp/detector_test.go
dashpole: addressed in 309fa637e

detectors/gcp/detector_test.go
dashpole: addressed in e3989f31a

detectors/gcp/detector.go
```diff
@@ -31,8 +32,17 @@ type detector struct {
 }
 
 // Detect detects associated resources when running on GCE, GKE, GAE,
-// Cloud Run, and Cloud functions.
+// Cloud Run, Cloud Functions, and Bare Metal Solution.
 func (d *detector) Detect(context.Context) (*resource.Resource, error) {
+	if d.detector.CloudPlatform() == gcp.BareMetalSolution {
```
ps-mir: Detection depends on `detector.CloudPlatform()`. Would be ideal to call it once and reuse the result.

ps-mir (commented):
Small concern on CloudPlatform() being called twice for detection.

detectors/gcp/detector.go
dashpole: addressed in 147f2d663

ps-mir (approved):
LGTM

detectors/gcp/detector.go
```diff
@@ -31,22 +32,33 @@ type detector struct {
 }
 
 // Detect detects associated resources when running on GCE, GKE, GAE,
-// Cloud Run, and Cloud functions.
+// Cloud Run, Cloud Functions, and Bare Metal Solution.
 func (d *detector) Detect(context.Context) (*resource.Resource, error) {
+	cp := d.detector.CloudPlatform()
```
dashpole: `CloudPlatform()` before `metadata.OnGCE()` adds seconds of blocking startup I/O on every non-GCP host. BMS is detected from environment variables only (`opentelemetry-operations-go/detectors/gcp/bms.go` — `onBareMetalSolution()` does three `LookupEnv` calls and no I/O), so gate the expensive probe on a cheap env check and keep the fast exit first:.

detectors/gcp/detector.go
```diff
@@ -31,22 +32,33 @@ type detector struct {
...
 
-	switch d.detector.CloudPlatform() {
+	switch cp {
 	case gcp.GKE:
 		b.attrs = append(b.attrs, semconv.CloudPlatformGCPKubernetesEngine)
 		b.addZoneOrRegion(d.detector.GKEAvailabilityZoneOrRegion)
 		b.add(semconv.K8SClusterNameKey, d.detector.GKEClusterName)
 		b.add(semconv.HostIDKey, d.detector.GKEHostID)
-	case gcp.CloudRun:
+		b.add(semconv.HostNameKey, d.detector.GCEHostName)
```
dashpole: GKE `host.name` now turns a normal Workload Identity node into `ErrPartialResource`. restore the fallible handling (matching the collector), or, if the reviewer's preference stands, at minimum (a) add a test case pinning `GKE + GCEHostName error → ErrPartialResource and no host.name`, and (b) call the new error out in `CHANGELOG.md`/`README.md` so GKE users who treat `resource.New`'s error as fatal are warned.

detectors/gcp/detector_test.go
```diff
@@ -29,6 +29,7 @@ func TestDetect(t *testing.T) {
 				projectID:           "my-project",
```
dashpole: The BMS-before-`OnGCE` invariant is not pinned by any test. the `metadata.OnGCE()` result is cached by a package-level `sync.Once`, so a second test that simply omits the env var is order-dependent and fragile.

detectors/gcp/detector.go
```diff
@@ -147,6 +160,24 @@ func (r *resourceBuilder) addZoneOrRegion(detect func() (string, gcp.LocationTyp
...
+			switch mig.Type {
+			case gcp.Zone:
+				r.attrs = append(r.attrs, semconv.GCPGCEInstanceGroupManagerZone(mig.Location))
+			case gcp.Region:
+				r.attrs = append(r.attrs, semconv.GCPGCEInstanceGroupManagerRegion(mig.Location))
+			default:
+				r.errs = append(r.errs, fmt.Errorf("managed instance group location must be zone or region. Got %v", mig.Type))
+			}
+		}
+	} else {
```
dashpole: MIG detection-error path is dead code in the tests. add a `gcp.GCE` case with a fake that returns an error only from `GCEManagedInstanceGroup` (give `fakeGCPDetector` a dedicated `migErr` field, since the shared `err` field short-circuits every other getter too), asserting `expectErr: true` and the full non-MIG GCE attribute set.

detectors/gcp/detector.go
```diff
@@ -31,22 +32,33 @@ type detector struct {
...
 // Detect detects associated resources when running on GCE, GKE, GAE,
-// Cloud Run, and Cloud functions.
+// Cloud Run, Cloud Functions, and Bare Metal Solution.
 func (d *detector) Detect(context.Context) (*resource.Resource, error) {
+	cp := d.detector.CloudPlatform()
+	if cp == gcp.BareMetalSolution {
+		b := &resourceBuilder{}
+		b.attrs = append(b.attrs, semconv.CloudProviderGCP, semconv.CloudPlatformGCPBareMetalSolution)
+		b.add(semconv.CloudAccountIDKey, d.detector.BareMetalSolutionProjectID)
+		b.add(semconv.HostNameKey, d.detector.BareMetalSolutionInstanceID)
```
dashpole: BMS instance ID lands on `host.name`, but GCP's own monitored-resource mapping reads `host.id`. `b.add(semconv.HostNameKey, d.detector.BareMetalSolutionInstanceID)`.

detectors/gcp/README.md
dashpole: Docs cover one of the four new features. add ` Cloud Run worker pools` to the README list and to both doc comments; extend the changelog line to name the attribute keys (`cloud.platform=gcpbaremetalsolution`, `gcp.gce.instancegroup_manager.{name,zone,region}`, `host.name` on GKE).

detectors/gcp/detector.go
dashpole: GKE `host.name` silently collides with `resource.WithHost()`. documentation only — a sentence in `README.md` stating that on GKE `host.name` is the node name and will conflict with `resource.WithHost()`.

detectors/gcp/detector.go
```diff
@@ -31,22 +32,33 @@ type detector struct {
...
 	}
 	b := &resourceBuilder{}
 	b.attrs = append(b.attrs, semconv.CloudProviderGCP)
 	b.add(semconv.CloudAccountIDKey, d.detector.ProjectID)
 
-	switch d.detector.CloudPlatform() {
+	switch cp {
 	case gcp.GKE:
 		b.attrs = append(b.attrs, semconv.CloudPlatformGCPKubernetesEngine)
 		b.addZoneOrRegion(d.detector.GKEAvailabilityZoneOrRegion)
```
dashpole: "Parity" is still incomplete, and the two implementations now diverge in three new places. After this PR the contrib and collector mappings still differ:.

ps-mir (commented):
Need a composition test with core detectors to ensure no merge regression. Forces detector on stay on SDK main version.

[Example](https://github.com/paulojmdias/opentelemetry-go-contrib/blob/c31823c61c080c58453d4bb341c702ec8a49e6cd/detectors/consul/composition_test.go)

detectors/gcp/detector.go
```diff
@@ -31,8 +34,17 @@ type detector struct {
 }
 
 // Detect detects associated resources when running on GCE, GKE, GAE,
-// Cloud Run, and Cloud functions.
+// Cloud Run, Cloud Run jobs, Cloud Run worker pools, Cloud Functions, and Bare Metal Solution.
 func (d *detector) Detect(context.Context) (*resource.Resource, error) {
+	if _, err := d.detector.BareMetalSolutionProjectID(); err == nil && d.detector.CloudPlatform() == gcp.BareMetalSolution {
```
dashpole: This gate only avoids the non-GCP startup probe because `&&` short-circuits with the env lookup on the left. Nothing pins that. I re-applied the `cp := d.detector.CloudPlatform()` caching from `e9851c66` (the form `6f727737` reverted to fix the startup-I/O issue) and the whole suite stays green while the unconditional pre-`OnGCE()` metadata probe comes back. Worth a small spy-based test asserting `BareMetalSolutionProjectID` is called first and `CloudPlatform` exactly once, plus a comment on line 39 naming the invariant so the next person who suggests caching sees why it was rejected.

dashpole:
@psx95

detectors/gcp/composition_test.go
```diff
@@ -0,0 +1,58 @@
...
+	require.NoError(t, err)
+
+	merged, err := resource.Merge(resource.Default(), detected)
+	require.NoError(t, err)
+	assert.NotErrorIs(t, err, resource.ErrSchemaURLConflict)
+	assert.Equal(t, resource.Default().SchemaURL(), merged.SchemaURL())
+}
+
+// TestComposition_WithCoreDetectors asserts this detector composes with the
+// detectors built into the SDK.
```
psx95, line 45: Could you clarify this? 
I don't see how the assertion is related, I'm maybe misunderstanding the intent of this test.

ps-mir: Regarding the test

Each detector pins its own semconv independently of SDK's resource.Default() schema. If those drift apart, `resource.Merge` returns ErrSchemaURLConflict. This breaks the composition with SDK's core resource detectors.


Each detectors needs its own copy of this test to prevent the schema drift which is per module.

dashpole: This test checks that the gcp detector can be composed with the SDK's opt-in built-in detectors (WithHost(), WithOS(), etc.) without a semconv schema version mismatch.

When resource.New merges detectors with conflicting SchemaURLs (e.g. if contrib uses v1.43.0 while WithHost() in the pinned SDK emits v1.42.0), resource.Merge returns ErrSchemaURLConflict and drops the schema URL on the resulting resource (res.SchemaURL() becomes "").

Asserting res.SchemaURL() == resource.Default().SchemaURL() confirms that all composed detectors agreed on the schema version and that the schema URL wasn't cleared during merge (mirroring detectors/docker/composition_test.go from [#9001](https://github.com/open-telemetry/opentelemetry-go-contrib/pull/9001#discussion_r3677186896)).

detectors/gcp/detector.go
pellared: Google’s [documented BMS setup](https://docs.cloud.google.com/bare-metal/docs/set-up-ops-agent-for-bms#authorize_the_ops_agent) and the [Ops Agent detector](https://github.com/GoogleCloudPlatform/ops-agent/blob/b0fc29f48d87fef5b454d9d98e7d934cf8472884/confgenerator/resourcedetector/bms_detector.go#L23-L26) use `BMS_PROJECT_ID`, `BMS_LOCATION`, and `BMS_INSTANCE_ID`, but the pinned operations-go detector [expects `BMS_REGION`](https://github.com/GoogleCloudPlatform/opentelemetry-operations-go/blob/e024e63ca63d8aec968076f5aa25de2f15210680/detectors/gcp/bms.go#L18-L29). A host configured according to Google’s documentation therefore misses this branch, probes a GCE metadata service that BMS does not provide, and returns no resource. Please support `BMS_LOCATION` (optionally retaining `BMS_REGION` as an alias) and regression-test the documented variable set.

This guard should also read and cache all three values and require each to be non-empty before any platform probe. The upstream getter treats a present-but-empty project variable as success, while `CloudPlatform()` falls through to an uncancellable metadata probe for a partial BMS environment; line 48 then probes metadata again.

detectors/gcp/detector.go
```diff
@@ -31,8 +34,17 @@ type detector struct {
...
+		b := &resourceBuilder{}
+		b.attrs = append(b.attrs, semconv.CloudProviderGCP, semconv.CloudPlatformGCPBareMetalSolution)
+		b.add(semconv.CloudAccountIDKey, d.detector.BareMetalSolutionProjectID)
+		b.add(semconv.HostIDKey, d.detector.BareMetalSolutionInstanceID)
```
pellared: The current Collector GCP detector maps `BMS_INSTANCE_ID` to `host.name`, and the [migration issue](https://github.com/open-telemetry/opentelemetry-go-contrib/issues/8946) calls for delegation without loss of functionality. Emitting `host.id` is semantically appropriate and matches the Ops Agent, but emitting only that attribute would remove `host.name` when the Collector migrates, potentially breaking existing selectors, dashboards, and grouping. Please emit both during a compatibility period, or add an explicit compatibility shim and regression test in the Collector migration.

detectors/gcp/detector.go
```diff
@@ -86,6 +101,7 @@ func (d *detector) Detect(context.Context) (*resource.Resource, error) {
 		b.add(semconv.HostNameKey, d.detector.GCEHostName)
 		b.add(semconv.GCPGCEInstanceNameKey, d.detector.GCEInstanceName)
 		b.add(semconv.GCPGCEInstanceHostnameKey, d.detector.GCEInstanceHostname)
+		b.addManagedInstanceGroup(d.detector.GCEManagedInstanceGroup)
```
pellared, line 145: This makes an optional MIG metadata lookup part of every existing GCE detection path, and `addManagedInstanceGroup` turns any lookup failure into `ErrPartialResource`. A GCE host that previously detected successfully can now return an error solely because `instance/attributes/created-by` failed, and callers commonly treat an error from `resource.New` as fatal. Please make lookup failures best-effort (or opt-in), while retaining validation errors when a successfully returned MIG value is malformed.

detectors/gcp/README.md
```diff
@@ -7,7 +7,9 @@ The GCP resource detector supports detecting resources on:
 * Google App Engine (GAE)
 * Cloud Run
 * Cloud Run jobs
+* Cloud Run worker pools
 * Cloud Functions
+* Bare Metal Solution (BMS)
```
pellared, line 12: Please document that BMS detection is entirely environment-driven and list the three required non-empty variables after the `BMS_LOCATION`/`BMS_REGION` contract is resolved. It would also help to state their resource-attribute mappings. Without those variables, a real BMS host has no GCE metadata fallback and produces no BMS resource; the Ops Agent itself is not required, but the environment configuration is.

detectors/gcp/README.md
```diff
@@ -75,3 +77,6 @@ res, err := resource.New(ctx,
 )
 ...
 ```
+
+> [!NOTE]
+> On GKE, `host.name` is collected on a best-effort basis from the underlying node name. If `resource.WithHost()` is also configured, `host.name` will reflect the pod hostname or the node name depending on detector option order.
```
pellared: Could this spell out both the value sources and precedence? `GCEHostName()` supplies the backing Compute Engine instance name, whereas `resource.WithHost()` supplies the OS/container hostname, and later resource options win. For example: “On GKE, the detector best-effort reads `host.name` as the backing Compute Engine instance name. If `WithHost` is also used, put the GCP detector later to prefer the instance name, or `WithHost` later to prefer the OS/container hostname.”

detectors/gcp/composition_test.go
```diff
@@ -0,0 +1,58 @@
...
+// detector and the SDK. [resource.Merge] reports [resource.ErrSchemaURLConflict]
+// and drops the schema URL when the two disagree, so this fails as soon as the
+// semconv version here and the one behind [resource.Default] diverge.
+func TestComposition_MergeWithDefault(t *testing.T) {
+	detected, err := testDetector(t).Detect(t.Context())
+	require.NoError(t, err)
+
+	merged, err := resource.Merge(resource.Default(), detected)
+	require.NoError(t, err)
+	assert.NotErrorIs(t, err, resource.ErrSchemaURLConflict)
```
pellared: `require.NoError(t, err)` immediately above aborts the test for every non-nil error, so this assertion can never fail. Please remove it, or place the specific `NotErrorIs` check before the general assertion if both are desired. The following schema URL assertion already guards against a conflict clearing the URL.

detectors/gcp/detector.go
```diff
@@ -21,7 +21,10 @@ import (
 // * Google Kubernetes Engine (GKE).
 // * Google App Engine (GAE).
 // * Cloud Run.
+// * Cloud Run jobs.
```
pellared: This list is collapsed into one paragraph by `go doc`. Go doc-comment lists need a blank comment line before the list and indented `-` markers. Could this use the documented list syntax so the expanded public API comment renders as intended?

pellared (commented):
The GKE hostname, Cloud Run worker pool, MIG, and composition coverage looks solid. I found several items worth addressing, primarily around BMS environment compatibility and Collector migration behavior.

In particular, Google’s documented BMS setup uses `BMS_LOCATION`, while the pinned detector expects `BMS_REGION`. There are also potential regressions involving the existing BMS `host.name` attribute and error propagation from optional MIG enrichment.

detectors/gcp/detector.go
dashpole: addressed in d24396e4c (supported `BMS_LOCATION` with `BMS_REGION` fallback in `internal/bms.go`, and guarded BMS detection in `detector.go` by reading and requiring all 3 non-empty values before calling `CloudPlatform()` or `metadata.OnGCE()`)

detectors/gcp/detector.go
dashpole: We will keep only `host.id` here in `opentelemetry-go-contrib` to stay strictly aligned with OpenTelemetry semantic conventions and the Ops Agent, and we will add a `host.name` compatibility shim in `opentelemetry-collector-contrib` during the Collector migration (#8946).

detectors/gcp/detector.go
dashpole: `GCEManagedInstanceGroup()` in `internal/gce.go` already explicitly catches HTTP 404 (`metadata.NotDefinedError` when `instance/attributes/created-by` is absent on standalone VMs) as well as non-MIG `created-by` values, returning a `nil` error in both cases. It only returns a non-nil error if the metadata server request itself fails with a transport/HTTP error, which is identical to all other GCE metadata lookups (`GCEHostType`, `GCEHostID`, `GCEInstanceName`, etc.) that report `ErrPartialResource`.

detectors/gcp/README.md
dashpole: addressed in 29862632c

detectors/gcp/README.md
dashpole: addressed in b4f2948ae — Upon further consideration, I decided to remove `GCEHostName()` (`host.name`) from GKE detection altogether. In Kubernetes, the GCE VM instance name represents `k8s.node.name` rather than the container/pod's `host.name` (which `resource.WithHost()` / `os.Hostname()` reports), conflicts with `resource.WithHost()`, and `instance/name` is blocked under GKE Workload Identity anyway.

cc @iblancasa

detectors/gcp/composition_test.go
dashpole: addressed in 9718245bc

detectors/gcp/detector.go
dashpole: addressed in fd4649aa5

detectors/gcp/README.md
iblancasa: Thanks for the ping. Dropping `host.name` on GKE makes sense to me for the reasons you listed.

Would you be open to exposing it as an opt-in? Something like `NewDetector(opts ...Option)` with a `WithGKEHostType()` option, off by default given the extra API call and the `compute.instances.get` permission it needs.

dashpole: I'm slightly confused. Maybe I pinged you by accident. The host name and host type are different things, right?

iblancasa: Yes. Those are different things. I thought you pinged me because of the work I did regarding `host.type` and this change might affect that, maybe?

Since we are here anyway: the port in internal already has `GKEHostType()`, but the public `Detect()` does not call it in the GKE branch and the collector migration PR drops the collector side call too. As a consequence GKE `host.type` disappears after the migration

dashpole: Got it -- I did confuse your work on host.type with host.name.  I think we should add an option for host.type.

dashpole: addressed in 04b752f5a

iblancasa (approved):

detectors/gcp/detector.go
```diff
@@ -47,7 +99,10 @@ func (d *detector) Detect(context.Context) (*resource.Resource, error) {
 		b.addZoneOrRegion(d.detector.GKEAvailabilityZoneOrRegion)
 		b.add(semconv.K8SClusterNameKey, d.detector.GKEClusterName)
 		b.add(semconv.HostIDKey, d.detector.GKEHostID)
-	case internal.CloudRun:
+		if d.cfg.gkeHostType {
+			b.add(semconv.HostTypeKey, d.detector.GKEHostType)
```
MrAlias: `Detect` drops the context passed by `resource.New`, and `GKEHostType` creates both its credential lookup and `instances.get` request with `context.TODO()`. If either call stalls, cancelling resource detection cannot stop it, so startup can remain blocked after the caller's deadline. Could we thread the detector context through this lookup and cover cancellation in a test?

MrAlias (commented):
The parity additions and follow-up fixes look sound overall. The new opt-in GKE API lookup still needs to preserve the detector's cancellation boundary.

detectors/gcp/detector.go
dashpole: addressed in 03889a57c

MrAlias (approved):

PR #9474, opened 2026-08-14
fix(otelaws): don't record 304 Not Modified as a span error

### LINKED ISSUE #6331: otelaws s3 error responses should not always be recorded
### Description

Instrumentation for s3 records 3xx http responses as errors.

https://github.com/open-telemetry/opentelemetry-go-contrib/blob/11cf0df7388a/instrumentation/github.com/aws/aws-sdk-go-v2/otelaws/aws.go#L75-L79

There is already `http.status_code` that covers the result of the call, so recording the error does not add any useful information. 

When getting objects from s3 conditionally a 3xx http response is expected if the condition is not satisfied. For example, you might get an object if is has been modified since the last time it was accessed.

### Environment

- OS: Linux
- Architecture: x86
- Go Version: 1.23.2
- `otelaws` version:  v0.52.0

### Steps To Reproduce

```
s3Client.GetObject(ctx, &s3.GetObjectInput{
		Bucket:          aws.String(bucket),
		Key:             aws.String(object),
		IfModifiedSince: lastModified,
	})
```

### Expected behavior

Don't record an error for 3xx responses.

### Potential fix

The error could be checked using something like:

```
import (
	"net/http"
	awshttp "github.com/aws/aws-sdk-go-v2/aws/transport/http"

...

	var respErr *awshttp.ResponseError
	if errors.As(err, &respErr) && respErr.HTTPStatusCode() == http.StatusNotModified {
```

It seems reasonable to do this for all services, or there could be an "IsTraceableError" check for each service which would not record these "error" responses.

I can create a PR for this if useful.

pujitha24:
I noticed that `otelaws` treats any non-nil AWS SDK error as a span error, which also catches HTTP 304 responses.

For conditional requests, a 304 `Not Modified` can be completely expected behavior. For example, an S3 request using `IfModifiedSince` can return 304 without anything actually failing.

This change detects the AWS SDK's wrapped HTTP 304 response and avoids setting span status Error / `error.type` for that case. Other redirects and actual errors are unchanged.

I added coverage for both sides: 304 is no longer reported as an error, while a 301 still is. The package tests, vet and lint pass.

Fixes #6331


instrumentation/github.com/aws/aws-sdk-go-v2/otelaws/aws.go
```diff
@@ -131,6 +133,21 @@ func (m otelMiddlewares) buildAttributes(ctx context.Context, in middleware.Init
...
 }
 
+// isNotModifiedResponseError reports whether err wraps an HTTP response
+// error with a 304 Not Modified status code. This is returned, for example,
+// by a conditional S3 GetObject request whose precondition (such as
+// IfModifiedSince) was not satisfied. It indicates the precondition was not
+// met, not a failure of the request, and should not be recorded as a span
+// error. Other 3xx status codes (e.g. a 301 signaling a misconfigured
+// bucket region) can indicate a real problem and are still recorded.
+func isNotModifiedResponseError(err error) bool {
```
ps-mir: I think this is packing two independent questions
1) Can status code be extracted from the error (Does a response exist)
2) Is that status code the one of interest (304)

Keeping it as functional is useful for testing, would recommend something like this

```go
func responseStatusCode(err error) (code int, ok bool) {
        var respErr *smithyhttp.ResponseError
        if !errors.As(err, &respErr) {
                return 0, false
        }
        return respErr.HTTPStatusCode(), true
}

func isNotModifiedStatus(err error) bool {
        code, ok := responseStatusCode(err)
        return ok && code == http.StatusNotModified
}
```

ps-mir (commented):
A comment regarding refactoring. Spec wise, change seems ok, as instrumentation can have specific error handling.

CHANGELOG.md
```diff
@@ -14,6 +14,7 @@ This project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0.htm
 
 ### Fixed
 
+- Don't mark spans as errors in `go.opentelemetry.io/contrib/instrumentation/github.com/aws/aws-sdk-go-v2/otelaws` when the AWS SDK returns an error wrapping an HTTP `304 Not Modified` response, such as from a conditional S3 `GetObject` request whose precondition (e.g. `IfModifiedSince`) was not satisfied.
```
ps-mir: Need to be moved to bottom. Also please attach the PR number.

pujitha24:
Split `isNotModifiedResponseError` into `responseStatusCode` and `isNotModifiedStatus` as suggested, and added a direct test for `responseStatusCode` alongside the existing 304/301 cases. Also moved the changelog entry to the bottom of `### Fixed` and attached `(#9474)`.

instrumentation/github.com/aws/aws-sdk-go-v2/otelaws/aws_test.go
```diff
@@ -172,6 +173,87 @@ func Test_otelMiddlewares_presignedRequests(t *testing.T) {
 	assert.NotContains(t, input.Header[key], value)
 }
 
+func Test_responseStatusCode(t *testing.T) {
```
ps-mir: This can dropped as `Test_isNotModifiedStatus` provides equivalent coverage for `responseStatusCode`.

ps-mir (commented):
Test for responseStatusCode can be dropped as its covered by tests of calling function. Rest looks ok.

ps-mir (approved):

CHANGELOG.md
```diff
@@ -17,6 +17,7 @@ This project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0.htm
 - Report `ot-baggage-*` extraction errors from `go.opentelemetry.io/contrib/propagators/ot` to `otel.Handle` instead of silently discarding them, while still attaching the successfully parsed baggage members to the context. (#9395)
 - Set `error.type` on the `rpc.client.call.duration` and `rpc.server.call.duration` metrics in `go.opentelemetry.io/contrib/instrumentation/google.golang.org/grpc/otelgrpc` when the RPC fails with a non-OK status, per the RPC semantic conventions. (#9429)
 - Reject OTLP exporter headers with an empty `name` in `go.opentelemetry.io/contrib/otelconf`, `go.opentelemetry.io/contrib/otelconf/x`, and `go.opentelemetry.io/contrib/otelconf/v0.3.0`, instead of forwarding invalid header names to OTLP exporters. (#9102)
+- Don't mark spans as errors in `go.opentelemetry.io/contrib/instrumentation/github.com/aws/aws-sdk-go-v2/otelaws` when the AWS SDK returns an error wrapping an HTTP `304 Not Modified` response, such as from a conditional S3 `GetObject` request whose precondition (e.g. `IfModifiedSince`) was not satisfied. (#9474)
```
iblancasa: Should this be marked as a breaking change?

instrumentation/github.com/aws/aws-sdk-go-v2/otelaws/aws.go
```diff
@@ -131,6 +133,28 @@ func (m otelMiddlewares) buildAttributes(ctx context.Context, in middleware.Init
...
 }
 
+// responseStatusCode reports the HTTP status code carried by err, if err
+// wraps a *smithyhttp.ResponseError.
+func responseStatusCode(err error) (code int, ok bool) {
+	var respErr *smithyhttp.ResponseError
+	if !errors.As(err, &respErr) {
+		return 0, false
+	}
+	return respErr.HTTPStatusCode(), true
```
dashpole, line 143: `respErr.HTTPStatusCode()` internally evaluates `e.Response.StatusCode`. If `respErr` is a typed nil pointer or if `respErr.Response` / `respErr.Response.Response` is nil (e.g. in test mocks or custom middleware wrapping `*smithyhttp.ResponseError`), calling `respErr.HTTPStatusCode()` will panic with a nil pointer dereference on the request path.

Consider adding defensive nil checks:
```go
func responseStatusCode(err error) (code int, ok bool) {
	var respErr *smithyhttp.ResponseError
	if !errors.As(err, &respErr) || respErr == nil || respErr.Response == nil || respErr.Response.Response == nil {
		return 0, false
	}
	return respErr.HTTPStatusCode(), true
}
```

dashpole (approved):

PR #9476, opened 2026-08-14
fix(otelhttp): avoid pinning the connection read buffer via standardizeHTTPMethod

### LINKED ISSUE #6315: Otel Memory Leak
### Description

Memory Leak in otel library code.

### Environment

- OS: Linux
- Architecture: x86
- Go Version: 1.23.2
        - go.opentelemetry.io/contrib/instrumentation/net/http/otelhttp v0.56.0
	- go.opentelemetry.io/otel v1.31.0
	- go.opentelemetry.io/otel/exporters/otlp/otlpmetric/otlpmetricgrpc v1.31.0
	- go.opentelemetry.io/otel/exporters/otlp/otlptrace v1.31.0
	- go.opentelemetry.io/otel/exporters/otlp/otlptrace/otlptracegrpc v1.31.0
	- go.opentelemetry.io/otel/sdk v1.31.0
	- go.opentelemetry.io/otel/trace v1.31.0

### Steps To Reproduce

See Comment here: https://github.com/open-telemetry/opentelemetry-go-contrib/issues/5190#issuecomment-2463638063

### Expected behavior

Memory does not continuously increase over time.

pujitha24:
While looking into #6315, I found one concrete retention path in `standardizeHTTPMethod`.

`net/http` can derive `Request.Method` as a substring of its connection read buffer. For an already-uppercase method like `GET`, `strings.ToUpper` can return that same string, and we were then storing it in the `http.request.method` attribute. That can keep the larger connection buffer reachable longer than needed.

This changes `standardizeHTTPMethod` to return the canonical `http.Method*` constants for known methods, or `_OTHER`, instead of returning a string derived from the input.

I added a regression test that checks the returned method does not share the caller's backing buffer. The test fails with the previous implementation and passes with this change across the affected instrumentations.

I don't want to overstate this as the full explanation for the memory growth in #6315. It fixes a specific retention path I could reproduce; the other possibilities discussed in that issue are outside this PR.

Fixes #6315


internal/shared/semconv/util.go.tmpl
```diff
@@ -117,14 +117,34 @@ func handleErr(err error) {
 	}
 }
 
+// standardizeHTTPMethod returns the canonical form of method, or "_OTHER" if
+// it does not match a known HTTP method. The returned string is always a
+// package-level constant or literal, never a derivative of method, so it
+// does not retain a reference to the buffer method may have been sliced
+// from (e.g. net/http's per-connection read buffer).
 func standardizeHTTPMethod(method string) string {
```
dashpole, line 125: Looks like `HTTPServer.SpanName` in `server.go.tmpl` has the same buffer retention issue when `r.Pattern` is empty, so we might want to use `standardizeHTTPMethod` there too.

instrumentation/github.com/emicklei/go-restful/otelrestful/internal/semconv/util.go
```diff
@@ -117,14 +117,34 @@ func handleErr(err error) {
...
+		return http.MethodPatch
+	case http.MethodPost:
+		return http.MethodPost
+	case http.MethodPut:
+		return http.MethodPut
+	case http.MethodTrace:
+		return http.MethodTrace
 	default:
-		method = "_OTHER"
+		return "_OTHER"
```
iblancasa, line 146: Why is this change needed?

CHANGELOG.md
```diff
@@ -14,6 +14,7 @@ This project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0.htm
 
 ### Fixed
 
+- Return canonical `http.Method*` constants from `standardizeHTTPMethod` in `go.opentelemetry.io/contrib/instrumentation/net/http/otelhttp`, `go.opentelemetry.io/contrib/instrumentation/net/http/httptrace/otelhttptrace`, `go.opentelemetry.io/contrib/instrumentation/github.com/gorilla/mux/otelmux`, `go.opentelemetry.io/contrib/instrumentation/github.com/gin-gonic/gin/otelgin`, `go.opentelemetry.io/contrib/instrumentation/github.com/labstack/echo/otelecho`, and `go.opentelemetry.io/contrib/instrumentation/github.com/emicklei/go-restful/otelrestful`, instead of the upper-cased input string. `net/http` derives `Request.Method` as a substring of the connection's read buffer via `strings.Cut`, and `strings.ToUpper` returns its input unchanged when no case conversion is needed, so recorded `http.request.method` attribute values could keep that entire read buffer reachable for as long as the span or metric data point referencing them was retained.
```
iblancasa: I think this is too much verbose

iblancasa (commented):
Please: can you rewrite the PR description on your own words? It is difficult to follow.

iblancasa (changes requested):

instrumentation/github.com/gin-gonic/gin/otelgin/internal/semconv/util_test.go
```diff
@@ -73,3 +74,16 @@ func TestStandardizeHTTPMethod(t *testing.T) {
...
+	// method is a substring of a larger buffer, similar to how
+	// net/http derives Request.Method from the underlying connection's
+	// read buffer via strings.Cut. The returned, standardized method
+	// must not keep that buffer alive by sharing its backing array.
+	buf := "GET / HTTP/1.1\r\nHost: example.com\r\n\r\n"
+	method := buf[:3]
+
+	got := standardizeHTTPMethod(method)
+	assert.Equal(t, "GET", got)
+	assert.NotSame(t, unsafe.StringData(buf), unsafe.StringData(got))
```
sonalgaud12, line 88: could we add a second case to this test covering the _OTHER path?

sonalgaud12 (changes requested):

PR #9477, opened 2026-08-14
Reduce the semconv v1.37.0 import scope

skitt:
In ParseFullMethodOld, the only use of semconv/v1.37.0 is to retrieve three attribute definitions. Since they are referenced for backward compatibility only, and won't change in the future, they can be defined in situ, which eliminates the only use of the full semconv/v1.37.0 package. semconv/v1.37.0/rpcconv is still required for otelgrpc but is much smaller.

This reduces the impact of importing otelgrpc in downstream projects using newer semconv versions.

### DIFF
--- a/instrumentation/google.golang.org/grpc/otelgrpc/internal/parse.go
+++ b/instrumentation/google.golang.org/grpc/otelgrpc/internal/parse.go
@@ -8,7 +8,6 @@ import (
 	"strings"
 
 	"go.opentelemetry.io/otel/attribute"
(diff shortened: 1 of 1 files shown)


dashpole:
I don't think we should do this. Its probably close to time to remove the backwards-compatibility modes anyways, but I need to dig into what is left to do there, and if we've waited long enough.

dashpole:
Looks like we have to wait until Oct 7th, so another 2 months to wait

dashpole:
Opened https://github.com/open-telemetry/opentelemetry-go-contrib/issues/9482 to track it

skitt:
Fair enough, thanks for keeping track!

PR #9489, opened 2026-08-16
detectors/scaleway: add Scaleway resource detector

### LINKED ISSUE #8967: [detectors/scaleway] Contribute scaleway detector from resourcedetectionprocessor
## Problem Statement

`collector-contrib/processor/resourcedetectionprocessor/internal/scaleway` implements a resource detector for Scaleway Instances. There is currently no equivalent in `go-contrib/detectors/`, which means Scaleway detection is only available to collector-contrib pipelines and not to any Go SDK user.

## Proposed Solution

Add a new `detectors/scaleway` package implementing the Scaleway detector as a `resource.Detector`, ported from the collector-contrib implementation, so that `collector-contrib/processor/resourcedetectionprocessor` can delegate to it and any other Go SDK user can adopt it.

## Prior Art

- Collector-contrib implementation to port from: [`processor/resourcedetectionprocessor/internal/scaleway`](https://github.com/open-telemetry/opentelemetry-collector-contrib/tree/main/processor/resourcedetectionprocessor/internal/scaleway)

## Additional Context

- Queries the Scaleway Instance Metadata API.
- Emits: `cloud.provider` (scaleway), `cloud.platform` (scaleway_baremetal or scaleway_instance), `cloud.region`, `cloud.availability_zone`, `host.id`, `host.name`, `host.type`.

paulojmdias:
Ports the Scaleway resource detector from `processor/resourcedetectionprocessor/internal/scaleway` in
`opentelemetry-collector-contrib` to a new standalone Go module `go.opentelemetry.io/contrib/detectors/scaleway`.

Also registers the detector in `detectors/autodetect` as `IDScaleway` (`"scaleway"`).

Fixes #8967
Relates #8940


CHANGELOG.md
```diff
@@ -11,6 +11,7 @@ This project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0.htm
 ### Added
 
 - Add support for the `aws.ec2` resource detector in `go.opentelemetry.io/contrib/otelconf/x`. (#9139)
+- Add `go.opentelemetry.io/contrib/detectors/scaleway`, a new resource detector for Scaleway Instances, ported from `processor/resourcedetectionprocessor/internal/scaleway` in `opentelemetry-collector-contrib`. Detects `cloud.provider`, `cloud.platform`, `cloud.account.id`, `cloud.availability_zone`, `cloud.region`, `host.id`, `host.image.id`, `host.image.name`, `host.name`, and `host.type`. (#9487)
```
paulojmdias:
```suggestion
- Add `go.opentelemetry.io/contrib/detectors/scaleway`, a new resource detector for Scaleway Instances, ported from `processor/resourcedetectionprocessor/internal/scaleway` in `opentelemetry-collector-contrib`. Detects `cloud.provider`, `cloud.platform`, `cloud.account.id`, `cloud.availability_zone`, `cloud.region`, `host.id`, `host.image.id`, `host.image.name`, `host.name`, and `host.type`. (#9489)
```

detectors/scaleway/scaleway.go
```diff
@@ -0,0 +1,218 @@
...
+	api := newMetadataAPI()
+
+	// A client that already knows where to look is used as it is.
+	if api.MetadataURL != nil {
+		return getMetadata(ctx, api)
+	}
+
+	var firstErr error
+	for _, url := range metadataURLs {
+		api.MetadataURL = &url
```
dashpole: Since `getMetadata` spawns a background goroutine that might outlive the timeout, mutating `api.MetadataURL` on the shared `api` instance across loop iterations could cause a data race. Consider instantiating a fresh `api` inside the loop:
```suggestion
		api := newMetadataAPI()
		api.MetadataURL = &url
```

detectors/scaleway/scaleway.go
```diff
@@ -0,0 +1,218 @@
...
+	for _, opt := range opts {
+		opt.apply(&cfg)
+	}
+	return &ResourceDetector{cfg: cfg}
+}
+
+// getMetadata queries the metadata service. The client does not accept a
+// [context.Context], so the request runs in its own goroutine and ctx only
+// cancels the wait. The client also offers no way to bound its own request, so
+// that goroutine can outlive this call.
```
ps-mir: Created scaleway/scaleway-sdk-go#3290 for followup, can be included in comments.

detectors/scaleway/scaleway_test.go
```diff
@@ -0,0 +1,286 @@
...
+	} {
+		_, ok := res.Set().Value(k)
+		assert.False(t, ok, "expected %s to be absent", k)
+	}
+}
+
+func TestDetect_ContextCanceled(t *testing.T) {
+	// The handler blocks until the test is done, so only the canceled context
+	// can end the wait.
+	release := make(chan struct{})
```
ps-mir: This seems to work for most cases, but I see a faint possibility of race.

srv.Close() in `serveMetadata` will be called first (LIFO push on t.Cleanup), and it succeeds when no request is in progress.

Most of the time the `ctx.Done` state is achieved faster than API handler being invoked and blocked on `<-release`.

Moving `t.Cleanup(func() { close(release) })` after detect seems like one way to avoid,but would prefer `defer close(release)` since it doesn't depend on `t.Cleanup` registration order.

ps-mir (commented):
Minor comment and test related changes. Also please add a composition test with one of the core detectors (Same as consul).

detectors/scaleway/scaleway.go
```diff
@@ -0,0 +1,218 @@
...
+
+	type result struct {
+		md  *instance.Metadata
+		err error
+	}
+
+	// Buffered so the goroutine never blocks once ctx is done.
+	ch := make(chan result, 1)
+	go func() {
+		md, err := api.GetMetadata()
```
MrAlias: The upstream SDK issue covers the uncancellable request in general, but the workaround here creates two current-PR failures. With two metadata endpoints that accept a request and never respond, I measured `Detect(context.Background())` returning after about four seconds while both SDK calls, goroutines, and connections remained active; repeated detection accumulates two of each per call. The per-call timeout only bounds our wait and is restarted for the second address.

When the default transport selects `HTTP_PROXY` and `NO_PROXY` does not exclude these addresses, each attempted link-local metadata URL is sent through the proxy. If both proxied attempts fail while the caller context remains active, the terminal error is converted to `resource.Empty(), nil`, so detection silently misses the Scaleway instance. Could we own this metadata lookup with a dedicated client that uses one context for the entire operation, bypasses proxies, and validates HTTP status? This also removes the Scaleway SDK module dependency used only for this lookup.

MrAlias (changes requested):
The current SDK workaround leaves metadata requests running after the detector returns and can route link-local lookups through the process proxy. That request boundary needs to be owned by the detector.

detectors/scaleway/scaleway.go
paulojmdias: Makes sense, thanks for that! PTAL when you get a chance 🙏

detectors/autodetect/go.mod
```diff
@@ -15,6 +15,7 @@ require (
 	go.opentelemetry.io/contrib/detectors/hetzner v0.17.0
 	go.opentelemetry.io/contrib/detectors/ibmcloud/vpc v0.17.0
 	go.opentelemetry.io/contrib/detectors/k8sapi v0.17.0
+	go.opentelemetry.io/contrib/detectors/scaleway v0.17.0
 	go.opentelemetry.io/contrib/detectors/vultr v0.17.0
```
ps-mir: version is bumped to `v0.18.0` for experimental detectors post last release.

detectors/scaleway/go.mod
```diff
@@ -0,0 +1,23 @@
+module go.opentelemetry.io/contrib/detectors/scaleway
+
+go 1.25.0
+
+require (
+	github.com/stretchr/testify v1.11.1
+	go.opentelemetry.io/otel v1.45.0
+	go.opentelemetry.io/otel/sdk v1.45.0
```
ps-mir: Need a composition test to ensure dep stay on sdk's main.

[Consul ref](https://github.com/paulojmdias/opentelemetry-go-contrib/blob/c31823c61c080c58453d4bb341c702ec8a49e6cd/detectors/consul/composition_test.go)

ps-mir (commented):
Context plumbing is much better now. Need a composition test with core detector.

[Consul ref](https://github.com/paulojmdias/opentelemetry-go-contrib/blob/c31823c61c080c58453d4bb341c702ec8a49e6cd/detectors/consul/composition_test.go)

ps-mir (approved):

PR #9499, opened 2026-08-17
detectors/openshift: add Openshift resource detector

### LINKED ISSUE #8955: [detectors/openshift] Contribute openshift detector from resourcedetectionprocessor
## Problem Statement

`collector-contrib/processor/resourcedetectionprocessor/internal/openshift` implements a resource detector for OpenShift clusters. There is currently no equivalent in `go-contrib/detectors/`, which means OpenShift detection is only available to collector-contrib pipelines and not to any Go SDK user.

## Proposed Solution

Add a new `detectors/openshift` package implementing the OpenShift detector as a `resource.Detector`, ported from the collector-contrib implementation, so that `collector-contrib/processor/resourcedetectionprocessor` can delegate to it and any other Go SDK user can adopt it.

## Prior Art

- Collector-contrib implementation to port from: [`processor/resourcedetectionprocessor/internal/openshift`](https://github.com/open-telemetry/opentelemetry-collector-contrib/tree/main/processor/resourcedetectionprocessor/internal/openshift)

## Additional Context

- Queries the OpenShift API (Infrastructure resource) to determine cluster infrastructure details.
- Emits: `cloud.provider`, `cloud.platform`, `cloud.region`, `k8s.cluster.name`.

paulojmdias:
Ports the Openshift resource detector from `processor/resourcedetectionprocessor/internal/openshift` in
`opentelemetry-collector-contrib` to a new standalone Go module `go.opentelemetry.io/contrib/detectors/openshift`.

Also registers the detector in `detectors/autodetect` as `IDOpenShift` (`"openshift"`).

Fixes #8955
Relates #8940


detectors/openshift/openshift.go
```diff
@@ -0,0 +1,361 @@
...
+func (d *ResourceDetector) address() (string, bool) {
+	if d.cfg.address != "" {
+		return d.cfg.address, true
+	}
+
+	host, port := os.Getenv(hostEnvVar), os.Getenv(portEnvVar)
+	if host == "" || port == "" {
+		return "", false
+	}
+	return "https://" + host + ":" + port, true
```
dashpole: If `KUBERNETES_SERVICE_HOST` is an IPv6 address, this produces an invalid URL without square brackets. Consider using `net.JoinHostPort(host, port)` to format IPv6 addresses correctly.

detectors/openshift/openshift.go
```diff
@@ -0,0 +1,361 @@
...
+	if !ok {
+		return nil, false, nil
+	}
+
+	client, err := d.client(address)
+	if err != nil {
+		return nil, true, err
+	}
+
+	req, err := http.NewRequestWithContext(ctx, http.MethodGet, address+infrastructurePath, http.NoBody)
```
dashpole, line 250: If `WithAddress` is configured with a trailing slash, this creates a double slash in the request path. Consider trimming trailing slashes:
```suggestion
	req, err := http.NewRequestWithContext(ctx, http.MethodGet, strings.TrimRight(address, "/")+infrastructurePath, http.NoBody)
```

dashpole (commented):
@iblancasa to review as a new codeowner

paulojmdias:
Thanks @dashpole! Please take a look again when you have a chance 👍

detectors/openshift/openshift.go
```diff
@@ -0,0 +1,366 @@
...
+		}
+		tlsConfig = &tls.Config{RootCAs: pool, MinVersion: tls.VersionTLS12}
+	}
+
+	// Use a transport with Proxy explicitly disabled. The API server is reached
+	// over a cluster-internal address that must never be contacted through an
+	// HTTP(S) proxy: doing so would hand the service account bearer token to
+	// the proxy in environments where users set HTTP_PROXY/HTTPS_PROXY for
+	// outbound traffic.
+	return &http.Client{
```
ps-mir, line 226: Need http timeout to cover transport issues. Atm detect will block until caller's ctx gets cancelled.

detectors/openshift/openshift.go
```diff
@@ -0,0 +1,366 @@
...
+		return nil, true, err
+	}
+
+	req, err := http.NewRequestWithContext(ctx, http.MethodGet, address+infrastructurePath, http.NoBody)
+	if err != nil {
+		return nil, true, err
+	}
+	req.Header.Set("Authorization", "Bearer "+token)
+
+	resp, err := client.Do(req)
```
ps-mir, line 256: The underlying allocation for connection is not released due to idle-conn keep alive. This repeats per detect call.

Something like `CloseIdleConnection` call or DisableKeepAlive setting can be used.

detectors/openshift/openshift_test.go
```diff
@@ -0,0 +1,455 @@
+// Copyright The OpenTelemetry Authors
+// SPDX-License-Identifier: Apache-2.0
```
ps-mir, line 2: Need a composition test with core resource detector.

ps-mir (commented):
Couple of concerns:
* The per `Detect()` http.Client is never closed. 
* Need timeout for http client.
* [Minor] Add composition test with core detector.

The new connection per `detect()` itself is a bit of overhead. It would be much better to hold connection in Detector config so that it can be reused per `Detect()`.

If error surfacing is a concern, I would be ok to collect them in a error field inside detector (or detector config) during `NewResourceDetector` construction.

SA tokens can rotate, so `token()` fetch needs to be per Detect()

> No need to handle idle conn if client construction moves to NewReourceDetector.

Would welcome maintainers` view.

detectors/openshift/openshift.go
```diff
@@ -0,0 +1,366 @@
...
+// API server. By default the service account token projected into the pod is
+// used.
+func WithToken(token string) Option {
+	return optionFunc(func(c *config) { c.token = token })
+}
+
+// WithTLSConfig sets the TLS configuration used to reach the OpenShift API
+// server. By default the certificate authority projected into the pod is used
+// as the sole root of trust.
+func WithTLSConfig(tlsConfig *tls.Config) Option {
```
dashpole, line 127: Looks like `WithTLSConfig` and the certificate pool loading path in `client()` aren't covered by unit tests. Adding a test with a custom `tls.Config` or `httptest.NewTLSServer` would verify this and resolve the codecov patch warning.

detectors/openshift/openshift.go
```diff
@@ -0,0 +1,366 @@
...
+	if d.cfg.token != "" {
+		return d.cfg.token, true
+	}
+
+	token, err := os.ReadFile(d.tokenPath)
+	if err != nil {
+		return "", false
+	}
+	// Trailing whitespace would make the Authorization header value invalid.
+	return strings.TrimSpace(string(token)), true
```
dashpole, line 200: If the token file is empty or only whitespace, `token()` still returns `true`, which causes `infrastructure()` to send an empty `Bearer ` header. Returning `false` when `strings.TrimSpace` yields an empty string would avoid that unnecessary request:
```suggestion
	t := strings.TrimSpace(string(token))
	if t == "" {
		return "", false
	}
	return t, true
```

detectors/openshift/openshift.go
```diff
@@ -0,0 +1,366 @@
...
+				filtered = append(filtered, kv)
+			}
+		}
+		attrs = filtered
+	}
+
+	res := resource.NewWithAttributes(semconv.SchemaURL, attrs...)
+
+	if len(errs) > 0 {
+		return res, fmt.Errorf("%w: %v", resource.ErrPartialResource, errs)
```
dashpole, line 363: Consider using `errors.Join` with `%w: %w` so callers can unwrap and inspect the underlying errors via `errors.Is` or `errors.As`:
```suggestion
		return res, fmt.Errorf("%w: %w", resource.ErrPartialResource, errors.Join(errs...))
```

detectors/openshift/doc.go
```diff
@@ -0,0 +1,43 @@
...
+/*
+Package openshift provides a [resource.Detector] which supports detecting
+attributes specific to OpenShift 4 clusters.
+
+The detector reads the cluster-scoped Infrastructure config object from the
+OpenShift API server, authenticating with the service account token projected
+into the pod. It requires the following RBAC:
+
+  - apiGroups: ["config.openshift.io"]
+    resources: ["infrastructures"]
```
ps-mir, line 13: RBAC for `infrastructure/status` is missing. Its needed for GET /status subresource.

detectors/openshift/openshift.go
```diff
@@ -0,0 +1,366 @@
...
+			return nil, true, err
+		}
+		// The API server is unreachable: not running on OpenShift.
+		return nil, false, err
+	}
+	defer resp.Body.Close()
+
+	switch {
+	case resp.StatusCode >= 200 && resp.StatusCode < 300:
+	case resp.StatusCode >= 400 && resp.StatusCode < 500:
```
ps-mir, line 270: HTTP 404 for non-detection is correct.

Detection logic also treats 401, 403 as non-detection, but they are possible when running in Openshift. Like:
* Bearer token missing/malformed/expired
* token authenticated by SA account missing the RBAC.

They should surface appropriate error about misconfigured detector.

ps-mir (commented):
Need a fix in README for rbac permission. Also 401/403 need to be treated separately for openshift based errors due to possible issue with malformed tokens and absent rbac permissions.

detectors/openshift/go.mod
```diff
@@ -0,0 +1,23 @@
+module go.opentelemetry.io/contrib/detectors/openshift
+
+go 1.25.0
+
+require (
+	github.com/stretchr/testify v1.11.1
+	go.opentelemetry.io/otel v1.45.0
+	go.opentelemetry.io/otel/sdk v1.45.0
```
ps-mir: otel dep now at `1.46.0`. Go at `1.26.0`. Please update both.

ps-mir (commented):
go.mod needs changes to avoid drift with other detectors and autodetect.

detectors/openshift/openshift.go
```diff
@@ -0,0 +1,389 @@
...
+	}
+
+	platform := infra.Status.PlatformStatus
+	switch strings.ToLower(platform.Type) {
+	case "aws":
+		attrs = append(attrs, semconv.CloudProviderAWS, semconv.CloudPlatformAWSOpenShift)
+		region(platform.AWS.Region, "aws.region")
+	case "azure":
+		attrs = append(attrs, semconv.CloudProviderAzure, semconv.CloudPlatformAzureOpenShift)
+		region(platform.Azure.CloudName, "azure.cloudName")
```
iblancasa: `cloud.region` ends up with values that are not regions here. For Azure, `platformStatus.azure.cloudName` is the cloud environment name (AzurePublicCloud, AzureUSGovernmentCloud, and so on) and the Infrastructure object does not expose an Azure region at all. For OpenStack, cloudName is the entry name from `clouds.yaml`. So every Azure cluster will report `cloud.region=azurepubliccloud` after lowercasing.

The collector detector has the same behaviour, but porting it into a new public module cements the bug. I would prefer to omit `cloud.region` for Azure and OpenStack (keeping cloud.provider and `cloud.platform` for Azure)

detectors/openshift/openshift.go
iblancasa: Every transport error except context cancellation is mapped to "not running on OpenShift", so Detect returns an empty resource and nil error. That covers a wrong root CA passed via `WithTLSConfig`, a CA file that does not match the API server, a TLS handshake failure, a DNS failure, and the 5 second client timeout. A slow or briefly unavailable API server silently produces no attributes, with no way for the user to notice. It is especially surprising when the address was set explicitly through WithAddress

detectors/openshift/openshift.go
```diff
@@ -0,0 +1,389 @@
...
+}
+
+// NewResourceDetector returns a [resource.Detector] that detects resource
+// attributes of the OpenShift 4 cluster the process is running in.
+//
+// The detector reads the cluster-scoped Infrastructure config object and
+// therefore requires the following RBAC:
+//
+//   - apiGroups: ["config.openshift.io"]
+//     resources: ["infrastructures", "infrastructures/status"]
```
iblancasa: I guess we only need `infrastructures/status`

detectors/openshift/openshift.go
paulojmdias: Agreed, dropped cloud.region for Azure and OpenStack. Azure keeps cloud.provider/cloud.platform. On OpenStack now reports no cloud attributes. The collector detector has the same issue, I'll follow up there to ensure that when we start using the SDK, we will report it as a breaking change or have a feature gate for the transition.

detectors/openshift/openshift.go
paulojmdias: Agreed. Once an address and token are resolved we're either in a pod or were pointed at a server explicitly, so transport failures (TLS, DNS, timeout) are now returned as errors. Only a 4xx other than 401/403 (e.g. vanilla Kubernetes answering 404 for the config API group) is treated as not running on OpenShift.

detectors/openshift/openshift.go
paulojmdias: Thanks, updated.

ps-mir (approved):

iblancasa (approved):

detectors/openshift/version.go
```diff
@@ -0,0 +1,7 @@
+// Copyright The OpenTelemetry Authors
+// SPDX-License-Identifier: Apache-2.0
+
+package openshift
+
+// Version is the current release version of the OpenShift resource detector.
+const Version = "0.17.0"
```
dashpole: Looks like this stayed at `0.17.0` after rebasing—it should be `0.18.0` to match the `experimental-detectors` module set in `versions.yaml` and `detectors/autodetect/go.mod`:
```suggestion
const Version = "0.18.0"
```

PR #9542, opened 2026-08-21
[WIP] logutil: detect cycles during log value conversion

### LINKED ISSUE #9044: otellogr: Recursive log conversion can crash on cyclic data

pellared:
Fixes #9044.

## Summary

- Detect path-local cycles while converting maps, slices, arrays, pointers, interfaces, and values traversed by `fmt`.
- Replace proven recursive edges with `"<cycle>"`, over-depth edges with `"<depth-limit>"`, and edges that exhaust the total conversion-work budget with `"<work-limit>"`.
- Enforce a package-internal 1,000-edge path-depth bound and a root-owned 10,000-unit work bound across normal conversion, nested formatting preflights, conditional snapshots, reflection copies, variable-size payloads, and raw map-key hashing.
- Keep identities path-local, so shared acyclic values are preserved instead of mistaken for cycles; keep the work counter root-owned and monotonic, so shared suffixes cannot amplify conversion work.
- Close the formatter-mutation TOCTOU found during review: when a formatting method and mutation-sensitive state coexist, format an exact-type, bounded entry snapshot so the method cannot invalidate the completed safety proof before `fmt` traverses the rest of the value.
- Derive multi-entry string-key accounting and result capacity from the snapshotted map key set, not a live map length that a formatter can shrink.
- Apply the shared-template changes and generated regressions to `otellogr`, `otellogrus`, `otelslog`, and `otelzap`.

## Review map

The template and design are the primary review surface. The eight generated `convert*.go` files are mechanical copies and are byte-identical after package-name normalization.

| Review concern | Primary location | Focused regression |
| --- | --- | --- |
| Invariants, bounds, and compatibility policy | [`DESIGN.md`](https://github.com/pellared/opentelemetry-go-contrib/blob/f71c7b5d379f0e0577fd8cc9e9c5e120ed9553a8/internal/shared/logutil/DESIGN.md) | Design sections “Traversal safety bounds” and “Formatting boundary” |
| Path-local cycle/depth/work accounting | [`convert.go.tmpl`](https://github.com/pellared/opentelemetry-go-contrib/blob/f71c7b5d379f0e0577fd8cc9e9c5e120ed9553a8/internal/shared/logutil/convert.go.tmpl) | Existing cycle, boundary, shared-DAG, and amplification suites |
| Formatter-mutation isolation | `formattingMutationState`, `prepareFormattingValue`, and `detachFormattingValue` in the template | `TestConvertValueFormattingMethodMutation`, `TestConvertValuePreservesSafeFormatting` |
| Map-key snapshot cardinality | The four map conversion loops in the template | `TestConvertValueTraversalWorkLimit/MapStringKeySnapshotSortWork` |
| Snapshot copy/hash bounds | `snapshotFormattingValue`, `reserveValueCopyWork`, `reserveMapKeyHashWork` | `FormattingSnapshotMapKeyHash`, depth/work boundary tests |
| Hot-path impact | [`bridges/otellogr/bench_test.go`](https://github.com/pellared/opentelemetry-go-contrib/blob/f71c7b5d379f0e0577fd8cc9e9c5e120ed9553a8/bridges/otellogr/bench_test.go) | `BenchmarkConvertValue`, including the two conditional-isolation cases |

Suggested order: review `DESIGN.md`, then the implementation and test templates, then treat the generated bridge files as a parity check.

## Design

See [`internal/shared/logutil/DESIGN.md`](https://github.com/pellared/opentelemetry-go-contrib/blob/f71c7b5d379f0e0577fd8cc9e9c5e120ed9553a8/internal/shared/logutil/DESIGN.md) for identity rules, marker precedence, work accounting, formatting behavior, hot-path constraints, and rejected alternatives.

### Why this still uses `fmt`

Structs and non-string map keys already use `%+v`, so replacing `fmt` here would be a representation change rather than a local safety fix. An exact replacement through public Go APIs would need to reproduce aggregate and pointer formatting, `reflect.Value` handling, formatter-method dispatch, and the private `internal/fmtsort` ordering across supported Go versions. Sanitized proxy values are smaller, but change concrete types, method selection, ordering, and output.

This PR therefore keeps `fmt` only after a bounded preflight has proved the entry graph safe. If a callable method could mutate state that `fmt` will subsequently traverse, the converter first builds a bounded detached snapshot and passes that snapshot to `fmt`. A deterministic project-owned representation remains a possible follow-up, but needs an explicit compatibility decision; it would not, by itself, bound arbitrary work performed inside user formatting methods.

### Trust boundary and conservative behavior

- Converter-owned traversal, copying, snapshot allocation, output payload accounting, and raw map-key hashing share the 10,000-unit root budget and the 1,000-edge path bound.
- `fmt.Formatter`, `fmt.Stringer`, and `error` method bodies remain trusted for their own runtime, receiver ABI work, panic behavior, and returned output. Their side effects are *not* trusted to preserve the enclosing graph.
- If a required detached map/slice lies behind an unexported non-settable field, safe reflection cannot replace it. The converter returns `"<work-limit>"` without invoking the method instead of using `unsafe`; inaccessible branches without shared mutable storage retain their normal output.
- Concurrent mutation from another goroutine is a caller data race and is outside this contract.
- Public reflection exposes a map’s logical length, but not retained bucket capacity. The bound caps logical entries and repeated expansion, but cannot strictly price one scan of a very sparse retained map without depending on private runtime layout.

## Tests and verification

The generated suite covers direct, mutual, mixed, branching, and deep cycles across maps, slices, arrays, pointers, interfaces, structs, map keys, and `reflect.Value`; shared DAGs; overlapping slices; pointer/array snapshots; formatting-method precedence and mutation; tracker growth/cleanup; depth/work/cycle precedence; wide and zero-sized collections; reflection-copy and payload accounting; and bounded raw map-key hashing.

Review-finding regressions additionally cover all four map dispatch paths, all four non-string map-key formatting sites, dynamic interface and `reflect.Value` keys, root pointers/addressable values, nil/empty header replacement, scalar-interface-to-cycle and short-to-large-string replacement, an empty map at the exact depth boundary, map shrinkage after key snapshot, detached-map raw-key hashing, preservation of safe formatting, and the reflection-inaccessible fallback. Fatal-recursion cases run in a subprocess with a 1 MiB maximum stack.

Verification for pushed head `f71c7b5d379f0e0577fd8cc9e9c5e120ed9553a8`:

- `make precommit` passed after the final change, including repository-wide generation, module tidying, lint, and race-enabled tests.
- The focused mutation, preservation, and work-limit regressions passed 10 repeated runs in each of the four bridge modules; the final mutation/depth-boundary safety pass also passed 20 repeated runs in `otellogr`.
- Full `GOARCH=386` tests passed in all four bridge modules.
- Running the four canonical bridge generation targets after the final edit left the working tree unchanged.
- Generated implementation hashes normalize to `9d726a23096c74ab9fd42b274f1fe2032330f9063e39c612a69989aca7f4eba4`; generated test hashes normalize to `371b2f1d077e6f9b480571a5b90a05f444bf720274ae0895f2aa28222bb896d3`.
- `git diff --check` passed.

The fatal-safety regressions execute first in low-stack child processes. After that crash guard succeeds, the same fresh cases run in the parent at its normal stack limit so coverage instrumentation observes the safe implementation; `otellogr` package coverage is 89.9% locally.

### Review-finding delta (`1fb7c4d8` → `349042dd4`; `f71c7b5d3` is test-only)

Environment: Go 1.27.0, Linux/amd64, AMD EPYC 7571; prebuilt test binaries pinned to one logical CPU, `GOMAXPROCS=1`, `-cpu=1`, `-benchtime=300ms`, `-benchmem`. The same benchmark harness, including the two new isolation cases, is compiled into both binaries. Runs are alternated base/head to reduce drift.

Eight balanced, interleaved pairs produced:

| Case | Before | After | Time change | Before → after B/op | Before → after allocs/op |
| --- | ---: | ---: | ---: | ---: | ---: |
| `Bool` | 16.09 ns | 16.09 ns | no difference | 0 → 0 | 0 → 0 |
| `Nil` | 14.78 ns | 14.78 ns | no difference | 0 → 0 | 0 → 0 |
| `Map` | 1.781 µs | 1.779 µs | no significant difference | 536 → 536 | 9 → 9 |
| `Struct` | 552.2 ns | 545.9 ns | -1.15% | 16 → 16 | 1 → 1 |
| `StructMap` | 1.788 µs | 1.872 µs | +4.73% | 112 → 112 | 5 → 5 |
| `StructRecursiveSlice` | 1.346 µs | 1.385 µs | +2.86% | 72 → 72 | 3 → 3 |
| `IntMapKey` | 1.159 µs | 1.212 µs | +4.62% | 176 → 176 | 6 → 6 |
| `NonStringMapKey` | 1.845 µs | 1.874 µs | +1.57% | 192 → 192 | 6 → 6 |
| `FormattingSnapshotMap` | 2.086 µs | 4.429 µs | +112.35% | 136 → 448 | 5 → 11 |
| `FormattingSnapshotSlice` | 2.176 µs | 3.635 µs | +67.03% | 48 → 184 | 1 → 5 |

Allocation counts and bytes are unchanged for all 30 ordinary cases. Their timing geomean is 228.6 ns → 226.8 ns (-0.78%); individual ordinary-path changes are flat to low-single-digit in this run. Including the two newly exercised isolation paths produces a +3.27% all-case timing geomean.

The snapshot cost is paid only when one formatted value contains both a callable formatting method and state the method could mutate before `fmt` traverses it. Aggregate allocations, copies, recursive descent, map reinsertion, and key hashing on that path remain hard-capped by the shared work/depth bounds.

<details>
<summary>Historical original-implementation measurements against main</summary>

These measurements predate the review-finding fix and are retained to show the cost of the original cycle/work-bounding implementation.

- Base: `a9b86ba93ccec95c6d809ef09820d68462820f2a`
- Previous PR head: `65ef372fc1d81dd0bfc70482918e791e67810a0a`
- Benchmarked implementation: `1ca1bf73d1b11d84b1a60601f2c5c6f5eeb98008`
- Go 1.27.0, Linux/amd64, AMD EPYC 7571
- Prebuilt test binaries pinned to logical CPU 4; `GOMAXPROCS=1`, `-cpu=1`, `-benchtime=300ms`, `-count=10`, `-benchmem`

| Case | Base | Head | Time change | Head B/op | Head allocs/op |
| --- | ---: | ---: | ---: | ---: | ---: |
| `Bool` | 15.99 ns | 16.02 ns | +0.2% | 0 | 0 |
| `Nil` | 15.99 ns | 14.77 ns | -7.6% | 0 | 0 |
| `PointerInt` | 90.13 ns | 99.89 ns | +10.8% | 8 | 1 |
| `PointerDepth8` | 341.6 ns | 551.6 ns | +61.5% | 0 | 0 |
| `EmptySlice` | 38.47 ns | 28.59 ns | -25.7% | 0 | 0 |
| `Slice` | 551.5 ns | 639.9 ns | +16.0% | 312 | 5 |
| `SliceAny` | 431.0 ns | 545.5 ns | +26.6% | 288 | 2 |
| `OneEntryMap` | 583.6 ns | 770.8 ns | +32.1% | 176 | 5 |
| `Map` | 1.466 µs | 1.815 µs | +23.8% | 536 | 9 |
| `Struct` | 400.5 ns | 570.9 ns | +42.5% | 16 | 1 |
| `StructMap` | 1.198 µs | 1.793 µs | +49.6% | 112 | 5 |
| `StructRecursiveSlice` | 832.8 ns | 1.371 µs | +64.6% | 72 | 3 |
| `IntMapKey` | 825.6 ns | 1.150 µs | +39.2% | 176 | 6 |
| `NonStringMapKey` | 1.256 µs | 1.892 µs | +50.6% | 192 | 6 |
| **27-case geomean** | **169.0 ns** | **202.9 ns** | **+20.1%** | | |

The original accounting added 12.8% geomean relative to the prior PR head and 20.1% relative to base. Exact scalar and empty paths remained allocation-free. Allocation counts matched base in 26 of 27 cases; `StructMap` added one 16-byte reflected-key snapshot so string-key payload work could be bounded through public reflection.

Head-only adversarial measurements showed compact amplification plateauing at the shared work bound:

| Case | Median | B/op | allocs/op |
| --- | ---: | ---: | ---: |
| `BranchingCycleDepth18` | 2.07 ms | 630 KiB | 6,667 |
| `BranchingCycleDepth64` | 2.12 ms | 630 KiB | 6,667 |
| `FormattingSharedDAGDepth24` | 620 µs | 4.75 KiB | 1 |
| `FormattingSharedDAGDepth32` | 621 µs | 4.75 KiB | 1 |
| `RejectedPointerArray16384` | 142 ns | 0 | 0 |
| `RejectedNestedPointerArray16384` | 160 ns | 0 | 0 |
| `ZeroSizedSlice100000` | 51 ns | 0 | 0 |
| `WideFormattingStruct10000` | 81 ns | 0 | 0 |

The base converter does not terminate on the cyclic adversarial inputs, so those cases are head-only.

</details>


bridges/otellogr/bench_test.go
```diff
@@ -8,8 +8,89 @@ import (
...
+		{name: "SharedDAG", value: []any{shared, shared}},
+		{name: "Struct", value: struct{ Value int }{Value: 42}},
+		{name: "StructInterface", value: struct{ Value any }{Value: 42}},
+		{name: "StructMap", value: struct{ Value map[string]int }{Value: shared}},
+	} {
+		b.Run(tt.name, func(b *testing.B) {
+			b.ReportAllocs()
+			for range b.N {
+				benchmarkValue = convertValue(tt.value)
+			}
```
pellared: Use `b.Loop` instead of `b.N`.
Note that then we do not need `benchmarkValue`.
Reference: https://pkg.go.dev/testing#B.Loop


```suggestion
			for b.Loop() {
				convertValue(tt.value)
			}
```

bridges/otellogr/bench_test.go
```diff
@@ -8,8 +8,89 @@ import (
...
+	}{
+		{name: "Cycle", value: cyclicMap},
+		{name: "BranchingCycle", value: branchingCycle},
+		{name: "StructCycle", value: wrapper{Value: structCycle}},
+	} {
+		b.Run(tt.name, func(b *testing.B) {
+			b.ReportAllocs()
+			for range b.N {
+				benchmarkValue = convertValue(tt.value)
+			}
```
pellared: Use `b.Loop` instead of `b.N`.
Note that then we do not need `benchmarkValue`.
Reference: https://pkg.go.dev/testing#B.Loop

```suggestion
			for b.Loop() {
				convertValue(tt.value)
			}
```

bridges/otellogr/bench_test.go
```diff
@@ -8,8 +8,89 @@ import (
 	"testing"
 
 	"github.com/go-logr/logr"
+
+	"go.opentelemetry.io/otel/attribute"
 )
 
+var benchmarkValue attribute.Value
+
```
pellared: This should not be needed per other comments

```suggestion
```

bridges/otellogr/bench_test.go
pellared: Updated this benchmark to use `b.Loop` in fec676f7. The refreshed paired benchmark results are in the PR description.

bridges/otellogr/bench_test.go
pellared: Updated this benchmark to use `b.Loop` in fec676f7. The refreshed paired benchmark results are in the PR description.

bridges/otellogr/bench_test.go
pellared: Removed `benchmarkValue` in fec676f7 now that both converter benchmarks use `b.Loop`.

pellared:
@khushijain21, @dmathieu, @MrAlias, I am asking you for review as you are code owners of different log bridges.

internal/shared/logutil/convert.go.tmpl
```diff
@@ -69,52 +268,526 @@ func convertValue(v any) attribute.Value {
...
+func convertPointerValue(current reflect.Value, visited *visitTracker) attribute.Value {
+	anchor := visitKey{typ: current.Type(), ptr: current.UnsafePointer()}
+	power, distance := 1, 0
+	checkError := false
+	var terminal any
+
+	for {
+		// The fast path checked the initial pointer type for error. Only a type
+		// transition can expose a new implementation.
+		if checkError {
```
dmathieu, line 525: Shouldn't this be after checking for `current.IsNil`?

dmathieu (commented):
I'm not feeling super confident about this.

internal/shared/logutil/convert.go.tmpl
pellared: Not in this case. The ordering is intentional: the previous recursive implementation re-entered `convertValue`, whose error fast path runs before pointer nil handling. Moving the IsNil check first would change indirect typed-nil errors while direct typed-nil errors would still call `Error()`. I added a clarifying comment and a regression test with a nil-safe error receiver behind three pointer levels in 41170d13c.

pellared:
> I'm not feeling super confident about this.

Any particular thing? I know that this code is not simple. Personally, I could not find any issue. It mitigates the issue when someone adds a log field with a cycle with minimal overhead.

I told Codex 5.6 Sol Ultra to find a bug in this PR. Let's give him a chance.

EDIT: First iteration with Codex found no issue. Trying again more aggressively. One perf issue (not really bug) found https://github.com/open-telemetry/opentelemetry-go-contrib/pull/9542#discussion_r3969904141 (IMO overkill for now).

Note that we can also add some option like `WithoutCycleDetection` if the behavior causes any issues for users.

internal/shared/logutil/convert.go.tmpl
```diff
@@ -69,52 +268,528 @@ func convertValue(v any) attribute.Value {
...
+			if formatHasCycle(iter.Key(), depth+1, visited) || formatHasCycle(iter.Value(), depth+1, visited) {
+				visited.pop(key)
+				return true
+			}
+		}
+		visited.pop(key)
+		return false
+	case reflect.Array:
+		for i := range v.Len() {
+			if formatHasCycle(v.Index(i), depth+1, visited) {
```
pellared, line 778: `formatHasCycle` consumes a Go stack frame for every aggregate edge, so it can exhaust the default 1 GB goroutine stack before reaching a repeated identity. This is reproducible with safe, method-free input: arrange 1,000 one-element slice views in a ring and wrap each edge in nested `[1]` arrays created with `reflect.ArrayOf`. At array depth 580 the converter returns `"<cycle>"`; at 581 it terminates the process with `fatal error: stack overflow`. Calling `formattingCycle` directly reproduces the failure, so neither `fmt` nor attribute construction is responsible.

This cyclic input is covered by the termination goal in `DESIGN.md`; it is not the explicitly excluded hardening for deeply acyclic values. An explicit iterative DFS would remove the dependency on the goroutine stack without introducing a depth limit. A subprocess regression using `debug.SetMaxStack(1<<20)` and a smaller `64 × 64` graph catches the current failure in about 20 ms.

pellared: > arrange 1,000 one-element slice views in a ring and wrap each edge in nested `[1]` arrays created with reflect.ArrayOf

This is not something very common and  I do not think this should be a blocker.  :wink: 

The iterative DFS code would be even larger and make the code review even harder. I would only address it when somebody reports performance issues with the current approach. 

CC @dmathieu

dmathieu (approved):

khushijain21 (approved):

internal/shared/logutil/convert.go.tmpl
pellared: Decision documented in https://github.com/open-telemetry/opentelemetry-go-contrib/pull/9542/commits/ca321970ca928e64282b85f2529dda1600119601

internal/shared/logutil/convert.go.tmpl
```diff
@@ -69,52 +268,528 @@ func convertValue(v any) attribute.Value {
...
+				added = visited.pushSlow(key)
+			}
+			if !added {
+				return attribute.StringValue(cycleMarker)
+			}
+		}
 		items := make([]attribute.Value, 0, val.Len())
 		for i := range val.Len() {
-			items = append(items, convertValue(val.Index(i).Interface()))
+			items = append(items, convertValueWithTracker(val.Index(i).Interface(), visited))
```
MrAlias, line 411: `convertValueWithTracker` still adds stack frames for every slice edge before the tracker reaches the repeated identity. At the current head, a 32,768-node ring of one-element slices with an 8 MiB limit set through `debug.SetMaxStack` terminates the process on this recursive call instead of returning `\"<cycle>\"`.

Should we make normal container traversal iterative, or explicitly narrow the termination guarantee and keep #9044 open for this remaining crash path? A subprocess regression would make the boundary concrete. This is separate from the documented `formatHasCycle` limitation: it occurs in the normal slice-conversion path covered by this PR's termination goal

pellared: Similarly to https://github.com/open-telemetry/opentelemetry-go-contrib/pull/9542#discussion_r3970376993 this rather is not something that people would log and I do not think this should be a blocker. Also note that the initial stack for 64-bit machines is 1 GiB :wink: 

For now, I have documented the limitation here https://github.com/open-telemetry/opentelemetry-go-contrib/pull/9542/commits/2e2aac59a8f8ef81a81ea8bbf1e238db96616237

In the meantime, I am trying to change the implementation from recursive to iterative or/and also give some big hard-coded limits (like 1000). Depending on the complexity, I will propose if this should be addressed here (in this PR) or in a separate PR.

pellared: After thinking it thought I think we need hard-coded limits anyway as a safety-net.

pellared: Addressed by bounding to 1000 https://github.com/open-telemetry/opentelemetry-go-contrib/pull/9542/commits/b446701745262cc5918602b90cce0a346cd49241

internal/shared/logutil/convert.go.tmpl
pellared: Addressed by bounding to 1000 https://github.com/open-telemetry/opentelemetry-go-contrib/pull/9542/commits/b446701745262cc5918602b90cce0a346cd49241

pellared:
I do not like where it is going. I will try to focus on finding some simpler/easier fix as otherwise I think we will be going circles. I also do not think it needs to be a more robust implementation than things implemented in Go standard library itself.

pellared:
Closing per https://github.com/open-telemetry/opentelemetry-go-contrib/pull/9542

PR #9543, opened 2026-08-21
fix(bridges/prometheus): keep explicitly supplied epoch sample timestamps

### LINKED ISSUE #9539: [Bug]: Prometheus bridge replaces explicit epoch timestamps with production time
### Component

`go.opentelemetry.io/contrib/bridges/prometheus`

### Describe the issue

All Prometheus-to-OTLP metric converters determine whether an explicit sample timestamp is present by checking whether `m.GetTimestampMs() != 0`.

`TimestampMs` is an optional `*int64`. Its generated getter returns `0` both when the field is absent and when it is explicitly set to Unix epoch (`0`). As a result, an explicitly supplied epoch timestamp is treated as absent and replaced with the time of `Produce`.

The affected converters are:

- `convertGauge`
- `convertCounter`
- `convertExponentialHistogram`
- `convertSummary`
- `convertHistogram`
- `convertUntyped`, introduced in #9099

This was identified in https://github.com/open-telemetry/opentelemetry-go-contrib/pull/9099#discussion_r3833239014.

The OpenTelemetry Prometheus compatibility specification requires a present Prometheus sample timestamp to be converted to the OTLP data point timestamp: https://github.com/open-telemetry/opentelemetry-specification/blob/main/specification/compatibility/prometheus_and_openmetrics.md#timestamps.

### Expected behavior

Converters should test field presence with `m.TimestampMs != nil`. If the field is present, including when its value is `0`, the OTLP data point time should be set with `time.UnixMilli(m.GetTimestampMs())`. Only an absent timestamp should use the time of `Produce`.

### Steps to reproduce

1. Create a Prometheus metric with an explicit epoch timestamp, for example with `prometheus.NewMetricWithTimestamp(time.Unix(0, 0), metric)` or with `TimestampMs: proto.Int64(0)` on a `dto.Metric`.
2. Gather and convert the metric using the Prometheus bridge.
3. Observe that the OTLP data point timestamp is the current production time instead of `1970-01-01T00:00:00Z`.

### Suggested regression coverage

Add table-driven coverage for every converter verifying:

- absent `TimestampMs` uses the production time;
- present `TimestampMs` equal to `0` produces Unix epoch;
- present non-zero `TimestampMs` preserves that timestamp.

### Environment

Observed on the current `main` branch and in the implementation proposed by #9099. This behavior is independent of operating system and architecture.

yzxcj797:
Fixes #9539.

## Summary

Every Prometheus-to-OTLP converter in `bridges/prometheus` decided whether a sample carries an explicit timestamp with `m.GetTimestampMs() != 0`. The generated getter returns `0` **both** when the field is absent and when it is explicitly set to Unix epoch, so an explicitly supplied epoch timestamp was treated as absent and silently replaced with the time of `Produce` — contrary to the [Prometheus compatibility specification](https://github.com/open-telemetry/opentelemetry-specification/blob/main/specification/compatibility/prometheus_and_openmetrics.md#timestamps), which requires a present sample timestamp to become the OTLP data point timestamp.

All six converters now check the field itself (`m.TimestampMs != nil`):

- `convertGauge`, `convertUntyped`, `convertCounter` (simple form)
- `convertExponentialHistogram`, `convertSummary` (the `if t := ...; t != 0` form)
- plus `convertHistogram`, which used the same idiom

Absent timestamps still fall back to the production time, unchanged.

## Testing

Added `TestConvertersExplicitEpochTimestamp` — builds `dto.Metric` values with `TimestampMs` explicitly set to `0` and asserts every converter keeps `time.UnixMilli(0)` instead of the (deliberately non-epoch) production time — and `TestConvertersAbsentTimestampUsesProductionTime`, pinning the absent-field behavior.

- With the fix: `go test . -run TestConverters` → 8/8 pass, full module suite `go test .` → ok, `gofmt` clean.
- Differential on unpatched `producer.go`: the epoch test fails on all converters.

### DIFF
--- a/bridges/prometheus/producer.go
+++ b/bridges/prometheus/producer.go
@@ -132,7 +132,9 @@ func convertGauge(metrics []*dto.Metric, now time.Time) metricdata.Gauge[float64
 			Time:       now,
 			Value:      m.GetGauge().GetValue(),
 		}
-		if m.GetTimestampMs() != 0 {
+		// The getter conflates an absent timestamp with an explicit Unix
+		// epoch (0), so presence must be checked on the field itself (#9539).
+		if m.TimestampMs != nil {
 			dp.Time = time.UnixMilli(m.GetTimestampMs())
 		}
 		otelGauge.DataPoints[i] = dp
@@ -150,7 +152,9 @@ func convertUntyped(metrics []*dto.Metric, now time.Time) metricdata.Gauge[float
 			Time:       now,
 			Value:      m.GetUntyped().GetValue(),
 		}
-		if m.GetTimestampMs() != 0 {
+		// The getter conflates an absent timestamp with an explicit Unix
+		// epoch (0), so presence must be checked on the field itself (#9539).
+		if m.TimestampMs != nil {
 			dp.Time = time.UnixMilli(m.GetTimestampMs())
 		}
 		otelGauge.DataPoints[i] = dp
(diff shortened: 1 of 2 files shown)


dashpole:
Please sign the CLA when you are able to.  Then we can review this

PR #9581, opened 2026-08-26
detectors/gcp: move gcp detector logic to internal package

dashpole:
Contribute GCP detector source files from opentelemetry-operations-go/detectors/gcp into an internal package under detectors/gcp/internal, preserving license headers. Remove the
github.com/GoogleCloudPlatform/opentelemetry-operations-go/detectors/gcp module dependency.

This change includes no behavioral changes, but fixes documentation and test issues discovered by copilot.

The plan is to deprecate opentelemetry-operations-go/detectors/gcp, and later archive the repository.  This makes the resource detector self-contained, and should make future changes much simpler.


detectors/gcp/internal/detector_print_test.go
```diff
@@ -0,0 +1,99 @@
...
+	log("FaaSCloudRegion", v, err)
+	v, err = d.FaaSID()
+	log("FaaSID", v, err)
+	v, err = d.FaaSName()
+	log("FaaSName", v, err)
+	v, err = d.FaaSVersion()
+	log("FaaSVersion", v, err)
+	v1, v2, err = d.GCEAvailabilityZoneAndRegion()
+	log("GCEAvailabilityZoneAndRegion.Zone", v1, err)
+	log("GCEAvailabilityZoneAndRegion.Region", v1, err)
```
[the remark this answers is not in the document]
dashpole: addressed in fc4b01b46

detectors/gcp/internal/detector.go
```diff
@@ -0,0 +1,112 @@
...
+}
+
+// ProjectID returns the ID of the project in which this program is running.
+func (d *Detector) ProjectID() (string, error) {
+	// N.B. d.metadata.ProjectIDWithContext(context.TODO()) is cached globally, so if we use it here it's untestable.
+	s, err := d.metadata.GetWithContext(context.TODO(), "project/project-id")
+	return strings.TrimSpace(s), err
+}
+
+// instanceID returns the ID of the project in which this program is running.
```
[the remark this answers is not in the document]
dashpole: addressed in 5398fa1ee

detectors/gcp/internal/detector.go
```diff
@@ -0,0 +1,112 @@
...
+
+// Detector collects resource information for all GCP platforms.
+type Detector struct {
+	metadata       *metadata.Client
+	os             osProvider
+	httpClient     *http.Client
+	computeBaseURL string
+}
+
+// osProvider contains the subset of the os package functions used by.
```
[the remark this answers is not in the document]
dashpole: addressed in 5fea4b8e9

detectors/gcp/internal/detector.go
```diff
@@ -0,0 +1,112 @@
...
+
+// realOSProvider uses the os package to lookup env vars.
+type realOSProvider struct{}
+
+func (realOSProvider) LookupEnv(env string) (string, bool) {
+	return os.LookupEnv(env)
+}
+
+// NewDetector returns a *Detector which can get detect the platform,
+// and fetch attributes of the platform on which it is running.
```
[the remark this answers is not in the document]
dashpole: addressed in fa95e386b

detectors/gcp/internal/utils_test.go
```diff
@@ -0,0 +1,125 @@
...
+		"project/numeric-project-id": fakeProjectNumber,
+		"instance/id":                fakeInstanceID,
+		"instance/name":              fakeInstanceName,
+		"instance/hostname":          fakeInstanceHostname,
+		"instance/machine-type":      fakeInstanceMachineType,
+		"instance/zone":              fmt.Sprintf("projects/%s/zones/%s", fakeProjectNumber, fakeZone),
+	}
+	for i := 0; i < len(keyValues); i += 2 {
+		out[keyValues[i]] = keyValues[i+1]
+	}
```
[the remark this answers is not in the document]
dashpole: addressed in cf8755b3d

detectors/gcp/internal/gke.go
```diff
@@ -0,0 +1,168 @@
...
+	_, err := d.metadata.InstanceAttributeValueWithContext(context.TODO(), clusterLocationMetadataAttr)
+	return err == nil
+}
+
+// GKEHostID returns the instance ID of the instance on which this program is running.
+func (d *Detector) GKEHostID() (string, error) {
+	return d.GCEHostID()
+}
+
+// GKEClusterName returns the name if the GKE cluster in which this program is running.
```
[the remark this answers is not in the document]
dashpole: addressed in 3872aa006

detectors/gcp/detector.go
```diff
@@ -41,44 +42,44 @@ func (d *detector) Detect(context.Context) (*resource.Resource, error) {
 	b.add(semconv.CloudAccountIDKey, d.detector.ProjectID)
 
 	switch d.detector.CloudPlatform() {
-	case gcp.GKE:
+	case internal.GKE:
```
ps-mir, line 45: The metadata lookup is sequence of independent HTTP calls each with its own retry/backoff. This is invisible on healthy server, but on a slow server, each call pays its full retry budget. All running serially, this could result in longer delay for Detect().

Context plumbing will allow caller to cap entire thing to a single deadline.

detectors/gcp/internal/detector.go
```diff
@@ -0,0 +1,112 @@
...
+	case d.onGCE():
+		return GCE
+	}
+	return UnknownPlatform
+}
+
+// ProjectID returns the ID of the project in which this program is running.
+func (d *Detector) ProjectID() (string, error) {
+	// N.B. d.metadata.ProjectIDWithContext(context.TODO()) is cached globally, so if we use it here it's untestable.
+	s, err := d.metadata.GetWithContext(context.TODO(), "project/project-id")
```
ps-mir, line 103: Metadata client honors context cancellation. Reusing context will allow for caller supplied deadlines.

Edit: The caching argument is not load bearing for `context.TODO`. The value gets cached irrespective of context used.

detectors/gcp/internal/gke.go
```diff
@@ -0,0 +1,168 @@
...
+	zone, err := d.metadata.ZoneWithContext(ctx)
+	if err != nil {
+		return "", err
+	}
+	instanceName, err := d.metadata.InstanceNameWithContext(ctx)
+	if err != nil {
+		return "", err
+	}
+
+	machineType, err := d.computeInstanceMachineType(ctx, projectID, lastPathSegment(zone), instanceName)
```
ps-mir, line 83: This skips metadata client, so its timeout/retry handling doesn't apply. The eventually used `google.DefaultClient` returns client with no timeout. 

If server accepts connection, and then stalls, this could block the client indefinitely.

Context plumbing will be useful, more important is to ensure that resulting client is created with timeout (Ideally with context, or by other means)

ps-mir (commented):
Lots of `context.TODO`s. Enabling context reuse would help in applying timeouts effectively.

dashpole:
@ps-mir I completely agree with doing context plumbing.  I was intending this to be just a code transfer without behavioral changes.  Do you think we could save improvements for a follow-up?

detectors/gcp/internal/gke.go
```diff
@@ -0,0 +1,168 @@
...
+
+func (d *Detector) onGKE() bool {
+	// Check if we are on k8s first
+	_, found := d.os.LookupEnv(k8sServiceHostEnv)
+	if !found {
+		return false
+	}
+	// If we are on k8s, make sure that we are actually on GKE, and not a
+	// different managed k8s platform.
+	_, err := d.metadata.InstanceAttributeValueWithContext(context.TODO(), clusterLocationMetadataAttr)
```
ps-mir, line 53: `cluster-location` is fetched twice, here only to check if read works, later for actual value at line 153 in `GKEAvailabilityZoneOrRegion`.

Can be simplified to a single call.

ps-mir:
> @ps-mir I completely agree with doing context plumbing. I was intending this to be just a code transfer without behavioral changes. Do you think we could save improvements for a follow-up?

I'm ok doing it as followup. We need a tracking issue for identified issues.

detectors/gcp/internal/gke.go
psx95: I think we follow a similar pattern in other environments too (e.g., GCE). Reducing it to a single call would involve storing/caching the state from the metadata server, which is what we do in Java's resource detector. 

Perhaps we could do something similar for Go in a follow-up PR.

@dashpole your thoughts ?

psx95 (approved):

MrAlias (approved):

detectors/gcp/detector.go
```diff
@@ -41,44 +42,44 @@ func (d *detector) Detect(context.Context) (*resource.Resource, error) {
...
 		b.attrs = append(b.attrs, semconv.CloudPlatformGCPKubernetesEngine)
 		b.addZoneOrRegion(d.detector.GKEAvailabilityZoneOrRegion)
 		b.add(semconv.K8SClusterNameKey, d.detector.GKEClusterName)
 		b.add(semconv.HostIDKey, d.detector.GKEHostID)
-	case gcp.CloudRun:
+	case internal.CloudRun:
```
dashpole: `git merge-tree --write-tree` against `pr-9462` at both current heads conflicts here and in `detector_test.go`, and the conflict straddles both the GKE `host.name` block and the `CloudRunWorkerPool` case — taking this PR's side of that hunk drops both silently, and the auto-merged regions leave `gcp.`-qualified identifiers behind so the result doesn't compile until they're re-qualified. Note also that `TestDetect` forces `OnGCE()` true via `GCE_METADATA_HOST`, so a resolution that moves the BMS block below the `OnGCE()` gate stays green. The `internal` package has every API #9462 needs, so nothing is lost by ordering; suggest landing #9462 first and re-running the rename over it.

detectors/gcp/internal/doc.go
```diff
@@ -0,0 +1,5 @@
+// Copyright The OpenTelemetry Authors
+// SPDX-License-Identifier: Apache-2.0
+
+// Package internal provides detection logic for GCP environments.
```
dashpole: Nothing in the tree records that this is `.../opentelemetry-operations-go/detectors/gcp@v1.36.0` — the only in-tree pointer was the `go.mod` pin this PR removes, so recovering it means digging through git history. Two lines in `internal/doc.go` naming the upstream module and version would let the next person re-verify the "no behavioral changes" claim or re-sync a late upstream fix without that. (`v1.37.0` is code-identical to `v1.36.0` apart from `Deprecated:` comments, so the copy is current.)

detectors/gcp/internal/gke.go
dashpole: Yep, lets do it in a follow-up.

detectors/gcp/internal/detector_print_test.go
```diff
@@ -0,0 +1,99 @@
...
+// distributed under the License is distributed on an "AS IS" BASIS,
+// WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+// See the License for the specific language governing permissions and
+// limitations under the License.
+
+// This provides an always-passing test that simply logs the detected environment in which it is run.
+// To run it on an arbitrary machine, build a test binary:
+//
+// go test -c -tags print
+//
```
pellared, line 22: After this file moved under `internal`, running `go test -c -tags print` from `detectors/gcp` builds the public package, which does not contain `TestDetectorPrint`; running it from this directory produces `internal.test`, not the documented `gcp.test`. Please make the command unambiguous, for example:

```sh
go test -c -tags print -o gcp.test ./internal
```

detectors/gcp/internal/detector_print_test.go
```diff
@@ -0,0 +1,99 @@
...
+//go:build print
+
+package internal
+
+import "testing"
+
+func TestDetectorPrint(t *testing.T) {
+	d := NewDetector()
+	t.Logf("CloudPlatform: %+v", d.CloudPlatform())
+	log := func(label, v interface{}, err error) {
```
pellared: This print-tagged file is excluded from default linting, but `golangci-lint run --build-tags=print` reports `revive/use-any` here. Please replace `interface{}` with `any`.

detectors/gcp/internal/detector_print_test.go
dashpole: addressed in dac48b8ad

detectors/gcp/internal/detector_print_test.go
dashpole: addressed in 9602c16a2

ps-mir (approved):

pellared (approved):

PR #9630, opened 2026-09-03
feat(otelhttp/filters): add Methods filter to match multiple HTTP methods

### LINKED ISSUE #9014: [Feature]: add Methods filter to match multiple HTTP methods
### Component

Instrumentation: otelhttp

### Problem Statement

The filters package provides a Method filter for matching a single HTTP  method, but there is no way to match multiple methods without chaining with Any(). This leads to verbose code for a common use case.

### Proposed Solution

Add a Methods filter that accepts multiple HTTP methods:

func Methods(ms ...string) otelhttp.Filter {
    return func(r *http.Request) bool {
        return slices.Contains(ms, r.Method)
    }
}

Usage:
otelhttp.WithFilter(filters.Methods("GET", "HEAD"))

Instead of:
otelhttp.WithFilter(filters.Any(filters.Method("GET"), filters.Method("HEAD")))

### Alternatives

Users can currently achieve this by combining Any() with multiple Method() calls, but this is verbose for a common use case.

### Prior Art

Similar multi-value filter patterns exist in other HTTP middleware libraries. The existing filters package already follows this pattern with Query/QueryContains and Header/HeaderContains pairs.

### Additional Context

Happy to submit a PR implementing this if the approach is accepted.

SurbhiAgarwal1:
## Description
Adds a `Methods(ms ...string)` filter to `instrumentation/net/http/otelhttp/filters` to allow matching multiple HTTP methods cleanly without needing to chain `Any(Method("GET"), Method("HEAD"))`.

Fixes #9014

## Type of Change
- [x] New feature (non-breaking change which adds functionality)

## Checklist
- [x] Added godoc comments for exported function `Methods`.
- [x] Added unit tests in `filters_test.go`.
- [x] Verified unit tests pass locally (`PASS`).

### DIFF
--- a/instrumentation/net/http/otelhttp/filters/filters.go
+++ b/instrumentation/net/http/otelhttp/filters/filters.go
@@ -111,3 +111,12 @@ func Method(m string) otelhttp.Filter {
 		return m == r.Method
 	}
 }
+
+// Methods returns a Filter that returns true if the request
+// method matches any of the provided values ms.
+func Methods(ms ...string) otelhttp.Filter {
+	return func(r *http.Request) bool {
+		return slices.Contains(ms, r.Method)
+	}
+}
+
--- a/instrumentation/net/http/otelhttp/filters/filters_test.go
+++ b/instrumentation/net/http/otelhttp/filters/filters_test.go
@@ -172,6 +172,41 @@ func TestMethod(t *testing.T) {
 	}
 }
 
+func TestMethods(t *testing.T) {
+	for _, s := range []scenario{
+		{
+			name:   "non-matching method",
+			filter: Methods(http.MethodGet, http.MethodPost),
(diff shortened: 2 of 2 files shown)


instrumentation/net/http/otelhttp/filters/filters.go
```diff
@@ -111,3 +111,12 @@ func Method(m string) otelhttp.Filter {
 		return m == r.Method
 	}
 }
+
+// Methods returns a Filter that returns true if the request
+// method matches any of the provided values ms.
+func Methods(ms ...string) otelhttp.Filter {
```
dashpole: Could you please add an entry for `filters.Methods` under the `### Added` section in `CHANGELOG.md`?

dashpole (commented):
This needs a changelog entry

SurbhiAgarwal1:
Hi @dashpole,

Thanks for the review! I have added the changelog entry in `CHANGELOG.md` under `## [Unreleased]` and updated the formatting.

PR #9636, opened 2026-09-04
exporters/autoexport: treat empty Prometheus env values as unset

cuishuang:
Fix `go.opentelemetry.io/contrib/exporters/autoexport` so empty `OTEL_EXPORTER_PROMETHEUS_HOST` and `OTEL_EXPORTER_PROMETHEUS_PORT` values are treated the same as unset values.

Previously, `os.LookupEnv` distinguished an empty environment variable from an unset variable. This could produce an invalid Prometheus bind address such as:

```text
localhost:
```

The implementation now uses the documented defaults when either value is empty:

- host: `localhost`
- port: `9464`

### DIFF
--- a/exporters/autoexport/metrics_test.go
+++ b/exporters/autoexport/metrics_test.go
@@ -7,6 +7,7 @@ import (
 	"context"
 	"fmt"
 	"io"
+	"net"
 	"net/http"
 	"net/http/httptest"
 	"reflect"
@@ -125,6 +126,7 @@ func TestMetricExporterPrometheus(t *testing.T) {
 	assertNoOtelHandleErrors(t)
 
 	t.Setenv("OTEL_METRICS_EXPORTER", "prometheus")
+	t.Setenv("OTEL_EXPORTER_PROMETHEUS_HOST", "")
 	t.Setenv("OTEL_EXPORTER_PROMETHEUS_PORT", "0")
 
 	r, err := NewMetricReader(t.Context())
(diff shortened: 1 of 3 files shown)


exporters/autoexport/metrics_test.go
```diff
@@ -26,6 +27,37 @@ import (
 	prometheusbridge "go.opentelemetry.io/contrib/bridges/prometheus"
 )
 
+func TestGetenv(t *testing.T) {
```
dmathieu: This is testing a deeply internal method. I would make this test higher-level.
Set the env var as empty, and ensure the behavior is correct.

cuishuang: Thanks for the feedback. I updated the tests to exercise the Prometheus exporter behavior directly instead of testing the internal getenv helper.

The tests now verify that empty Prometheus environment variables result in:
- localhost for the empty host
- port 9464 for the empty port

Please review it again.

dmathieu (approved):

dashpole (approved):

cuishuang:
Thanks for pointing this out. The changelog structure is valid under the repository's .markdownlint.yaml, which configures MD024 with siblings_only: true.  

The fail-fast workflow was not passing that configuration to markdownlint, so it used the default rule and incorrectly treated ### Fixed headings from different release sections as duplicates. I’ve updated the workflow to load .markdownlint.yaml, squashed the change into the existing commit, and verified the changelog locally.  


Please let me know if you would prefer the workflow fix to be submitted as a separate PR.

PR #9643, opened 2026-09-08
feat(otelhttp): add WithRedactedQueryParams to Transport

### LINKED ISSUE #6987: Add option to redact (mask) sensitive URL parameters like ?password=... or ?token=...
### Problem Statement

Some of the external APIs use authentication via URL parameters e.g.:
`https://myapi.com/query?token=abcd`
When using otelhttp Transport transaction contains sensitive information in the attributes like http.url.
```
	otelHttpClient := http.Client{
		Transport: otelhttp.NewTransport(http.DefaultTransport),
	}
```

### Proposed Solution

Have an Option to pass, something like WithParameterRedact(keysToSanitize []string{}),  to the function:
`func NewTransport(base http.RoundTripper, opts ...Option) *Transport `
Which would replace actual token with the masked string like `?token=xxxxx`

#### Alternatives

- Create own implementation (code duplication, additional code lines)
- https://betterstack.com/community/guides/observability/redacting-sensitive-data-opentelemetry/ (not always possible/flexible)

#### Prior Art

See alternatives

### Additional Context

I just need this option to protect sensitive data leakage.

om7057:
## Summary

Fixes #6987.

`otelhttp.Transport` records the full outbound URL, including the query string, in the `url.full` span attribute. There's no way to keep sensitive values (API tokens, etc.) that are passed as query parameters out of that attribute without dropping the whole thing.

This adds `WithRedactedQueryParams(keys ...string)`, which replaces the value of the given query parameters with `"REDACTED"` in the recorded `url.full` attribute. It only affects what's recorded on the span; the request sent over the wire, and the original `*http.Request` passed to `RoundTrip`, are both untouched.

As discussed on the issue, the existing [`redact`](https://github.com/MrAlias/redact) plugin can't do this: it works by matching whole attribute keys or dropping spans by name, so there's no way to use it to mask just one query parameter's value while keeping the rest of the URL visible.

## Implementation notes

- Scoped to the `Transport` (client) side only, distinct from the server-side `url.query` capture work in #8869/#8867.
- Redaction is applied by briefly mutating the request's `URL.RawQuery` around the call that builds the trace attributes, then restoring it before the request is sent — a regression test (`TestTransportOriginRequestNotModify`-style assertion) confirms the original request is untouched and the real query string reaches the server unchanged.
- Query parameters not being redacted are left as-is in value, but since the query string is rebuilt via `url.Values.Encode()`, they may be reordered/canonically re-encoded. Documented on the option.

## Test plan

- [x] Added `TestTransportRedactsQueryParams` (end-to-end: real request unaffected, span attribute redacted)
- [x] Added `TestTransportNoRedactedQueryParamsLeavesURLUnchanged` (no-op when unconfigured)
- [x] Added `TestRedactQueryParams` (unit tests for the redaction helper, including invalid query strings and repeated keys)
- [x] `go build ./...`, `go vet ./...`, `go test ./... -race` pass
- [x] `golangci-lint run` passes with 0 issues

### DIFF
--- a/instrumentation/net/http/otelhttp/config.go
+++ b/instrumentation/net/http/otelhttp/config.go
@@ -36,6 +36,8 @@ type config struct {
 	TracerProvider     trace.TracerProvider
 	MeterProvider      metric.MeterProvider
 	MetricAttributesFn func(*http.Request) []attribute.KeyValue
+
+	RedactedQueryParams []string
 }
 
 // Option interface used for setting optional config properties.
@@ -182,6 +184,24 @@ func WithSpanNameFormatter(f func(operation string, r *http.Request) string) Opt
 func WithClientTrace(f func(context.Context) *httptrace.ClientTrace) Option {
 	return optionFunc(func(c *config) {
 		c.ClientTrace = f
+	})
+}
+
+// WithRedactedQueryParams returns an Option that replaces the value of the
+// given query parameters with "REDACTED" in the url.full attribute recorded
+// by the otelhttp Transport, for parameters present on the outbound request.
+// It has no effect on the request that is actually sent; only the recorded
+// span attribute is changed.
+//
+// Query parameters not listed are left as-is, but the query string as a
+// whole is re-encoded (parameters may be reordered, and their values
+// percent-encoded canonically) as a side effect of the redaction.
+//
+// Use this for URLs that carry sensitive values in query parameters, such as
+// API tokens, that would otherwise be captured verbatim in span data.
+func WithRedactedQueryParams(keys ...string) Option {
+	return optionFunc(func(c *config) {
+		c.RedactedQueryParams = keys
 	})
 }
 
--- a/instrumentation/net/http/otelhttp/config.go
+++ b/instrumentation/net/http/otelhttp/config.go
@@ -195,7 +195,11 @@ func WithClientTrace(f func(context.Context) *httptrace.ClientTrace) Option {
 //
 // Query parameters not listed are left as-is, but the query string as a
 // whole is re-encoded (parameters may be reordered, and their values
-// percent-encoded canonically) as a side effect of the redaction.
+// percent-encoded canonically) as a side effect of the redaction. Any
+// parameter pair that fails to parse (a malformed percent-escape, for
+// example) is dropped rather than passed through raw, so a malformed pair
+// elsewhere in the query can never cause a value that should be redacted to
+// leak unredacted.
 //
 // Use this for URLs that carry sensitive values in query parameters, such as
 // API tokens, that would otherwise be captured verbatim in span data.
--- a/instrumentation/net/http/otelhttp/config.go
+++ b/instrumentation/net/http/otelhttp/config.go
@@ -203,9 +203,13 @@ func WithClientTrace(f func(context.Context) *httptrace.ClientTrace) Option {
 //
 // Use this for URLs that carry sensitive values in query parameters, such as
 // API tokens, that would otherwise be captured verbatim in span data.
+//
+// Like [WithFilter], keys accumulate across multiple calls to
+// WithRedactedQueryParams rather than the last call replacing the previous
+// ones.
 func WithRedactedQueryParams(keys ...string) Option {
 	return optionFunc(func(c *config) {
-		c.RedactedQueryParams = keys
+		c.RedactedQueryParams = append(c.RedactedQueryParams, keys...)
 	})
 }
 
--- a/instrumentation/net/http/otelhttp/transport.go
+++ b/instrumentation/net/http/otelhttp/transport.go
@@ -106,28 +106,40 @@ func (t *Transport) requestTraceAttrs(r *http.Request) []attribute.KeyValue {
 }
 
 // redactQueryParams returns rawQuery with the values of any parameter in
-// keys replaced with "REDACTED". If rawQuery cannot be parsed, or none of
-// keys are present, rawQuery is returned unchanged. Otherwise, the returned
-// query string is re-encoded as a whole: parameters may be reordered and
-// their values percent-encoded canonically.
+// keys replaced with "REDACTED". If none of keys are present, rawQuery is
+// returned unchanged. Otherwise, the returned query string is re-encoded as
+// a whole: parameters may be reordered and their values percent-encoded
+// canonically.
+//
+// url.ParseQuery returns the pairs it was able to parse alongside an error
+// for any it wasn't (e.g. a malformed percent-escape in an unrelated
+// parameter, or a "token=secret;x=1" query rejected for its semicolon
+// separator). A pair that failed to parse is dropped rather than passed
+// through raw: returning rawQuery unchanged on any parse error would risk
+// leaking a value we were asked to redact, if it happened to sit next to a
+// malformed pair.
 func redactQueryParams(rawQuery string, keys []string) string {
 	values, err := url.ParseQuery(rawQuery)
-	if err != nil {
+	if err == nil && !hasAny(values, keys) {
 		return rawQuery
 	}
 
-	var redacted bool
 	for _, key := range keys {
 		if _, ok := values[key]; ok {
 			values.Set(key, "REDACTED")
-			redacted = true
 		}
-	}
-	if !redacted {
-		return rawQuery
 	}
 
 	return values.Encode()
+}
+
+func hasAny(values url.Values, keys []string) bool {
+	for _, key := range keys {
+		if _, ok := values[key]; ok {
+			return true
+		}
+	}
+	return false
 }
 
 // RoundTrip creates a Span and propagates its context via the provided request's headers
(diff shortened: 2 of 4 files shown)


instrumentation/net/http/otelhttp/transport.go
```diff
@@ -76,12 +79,57 @@ func (t *Transport) applyConfig(c *config) {
...
+
+// redactQueryParams returns rawQuery with the values of any parameter in
+// keys replaced with "REDACTED". If rawQuery cannot be parsed, or none of
+// keys are present, rawQuery is returned unchanged. Otherwise, the returned
+// query string is re-encoded as a whole: parameters may be reordered and
+// their values percent-encoded canonically.
+func redactQueryParams(rawQuery string, keys []string) string {
+	values, err := url.ParseQuery(rawQuery)
+	if err != nil {
+		return rawQuery
```
sierikov, line 124: Do we want to return the query here? For example `url.full = "http://127.0.0.1:63118?token=supersecret;x=1"` will go through here. `ParseQuery` will return pairs, that it was able to pass, we can leave them there and go through the clean up. What do you think? Here some examples to illustrate what I mean:

`token=secret;x=1` - dropped.
`a=%zz&token=secret` = `token=REDACTED`.
`token=sec%zzret&a=1` = `a=1`.

instrumentation/net/http/otelhttp/config.go
```diff
@@ -185,6 +187,24 @@ func WithClientTrace(f func(context.Context) *httptrace.ClientTrace) Option {
...
+//
+// Query parameters not listed are left as-is, but the query string as a
+// whole is re-encoded (parameters may be reordered, and their values
+// percent-encoded canonically) as a side effect of the redaction.
+//
+// Use this for URLs that carry sensitive values in query parameters, such as
+// API tokens, that would otherwise be captured verbatim in span data.
+func WithRedactedQueryParams(keys ...string) Option {
+	return optionFunc(func(c *config) {
+		c.RedactedQueryParams = keys
```
sierikov: For example: `NewTransport(rt, WithRedactedQueryParams("a"), WithRedactedQueryParams("b"))` will take only `b`. Which can be unxepected. There is also `WithFilter` [here](https://github.com/open-telemetry/opentelemetry-go-contrib/pull/9643/changes#diff-2c7e041c055b5f16b11842955600bb09c9613fd21713ca02da521c27cf413c5cL129) which accumulates values. Let's accumulate as filter does.

instrumentation/net/http/otelhttp/transport.go
om7057: Thanks! fixed. Now rebuilds from whatever `url.ParseQuery` recovers and drops unparsable pairs instead of returning rawQuery raw. Matches your three examples.

instrumentation/net/http/otelhttp/config.go
om7057: Fixed, now accumulates like WithFilter.

sierikov (approved):
Looks good to me, but there is also `otelhttptrace` package, should it be included there as well? I also found some http tracing in templating files in `internal`, but didn't look for what they are used. Let's see what maintainers will say. Thank you for your work 👍🏻

dmathieu:
How is this better than the redact plugin, which allows redacting data globally, not just within otelhttp.
With your current setup, we would need to setup redaction rules in every module, and configure them independently every time.
With the possibility that modules that aren't in this repository work differently, or have different method signatures, making this much harder to use.

om7057:
Fair concern. Two reasons I still think this belongs at the source rather than as a downstream processor:

1. `redact` operates on whatever attributes are already on the span by the time it runs as a `SpanProcessor`. If any other processor or exporter is registered ahead of it (or a user misconfigures the order), the raw token is already exported before `redact` gets a chance to touch it. Stripping it at the point the attribute is built in `Transport` means it never enters the pipeline unredacted in the first place.
2. `redact` matches on whole attribute values or drops whole spans by name. It has no way to reach into `url.full` and touch just the `token=` substring while leaving host/path/other params intact for diagnostics, that needs URL-aware parsing this package doesn't have and, I'd guess, doesn't want to take on generically.

Agreed that per-module options are a worse experience if this pattern proliferates with inconsistent shapes across otelgin/otelmux/otelrestful/etc. Happy to keep this scoped to otelhttp per the original issue for now, or take a different steer if you'd rather see a shared helper other instrumentation packages could reuse consistently, let me know which direction you want.

cc: @dmathieu

sierikov:
A redact-style processor rewrites the attribute in `OnEnd`, using the slice that `recordingSpan.snapshot()` aliases rather than copying. However, `RegisterSpanProcessor` appends, thus with a provider constructed using `WithBatcher`, the batch processor is already ahead of you, having enqueued that exact snapshot before your `OnEnd` runs. The write is then done concurrently with the exporter goroutine reading it (`--race` flags it) and `BatchSpanProcessor` exports instantly whenever a batch reaches `MaxExportBatchSize` (512), so the window is hit deterministically rather than hypothetically. There is no ordering fix for outside provider construction, which is usually held by a shared framework rather than the service code.

So we'd use `WithRedactedQueryParams` as written, and constructing it where `url.full` is formed is the only place it's safe.  

Regarding your last question: shared helper can solve it. Putting it in `internal/shared/semconv` so that other HTTP instrumentations inherit it. This will avoid the inconsistent issue. Per-module rule flexibility would be useful, but I don't see it as a blocker. 

What do you think?

om7057:
The race point is a good one, that's a stronger version of my first argument. Agreed a shared helper in `internal/shared/semconv` is the right home if we go broader; worth noting most of the other consumers of that template are server middleware without an outbound `Transport`, so in practice it's otelhttp (and maybe otelhttptrace) that gains the feature, with the shared location mainly buying consistency. I'd rather get @dmathieu's read on whether the source-level approach is acceptable at all before restructuring, but if he's on board I'm happy to move it into the shared template.

dmathieu:
> If any other processor or exporter is registered ahead of it (or a user misconfigures the order), the raw token is already exported before redact gets a chance to touch it

Well, that's configuration issues. They can be a problem to users, but they shouldn't decide the architecture of software.

> redact matches on whole attribute values or drops whole spans by name. It has no way to reach into url.full and touch just the token= substring while leaving host/path/other params intact for diagnostics, that needs URL-aware parsing this package doesn't have and, I'd guess, doesn't want to take on generically.

It could though. redact, or any other processor could do regexp matching (no need to parse the URL, which is time consuming) and replace partial values that seem to be URL arguments.

> A redact-style processor rewrites the attribute in OnEnd, using the slice that recordingSpan.snapshot() aliases rather than copying. However, RegisterSpanProcessor appends, thus with a provider constructed using WithBatcher, the batch processor is already ahead of you

This is because `OnEnd` is the wrong step in a processor to do this. `OnEnding` is the right step to do that (it's unfortunately not yet supported by the SDK).
https://github.com/open-telemetry/opentelemetry-specification/blob/main/specification/trace/sdk.md#onending

sierikov:
Thank you for expanation, I missed `OnEnding`. I looked at the SDK and saw that there was no OnEnding in v1.46.0, open-telemetry/opentelemetry-go#6593 was still open, and your PR open-telemetry/opentelemetry-go#5756 was closed as stale. Are you or someone else planning to pick open-telemetry/opentelemetry-go#5756 back up? If not, I would be pleased to take over, either by starting a new implementation along the same path or continuing your branch, with solution you and @MrAlias find the right one. If you'd be up for guiding it, that would help a lot 🙂.  This hook will allow basically any logic we want, not limiting to redaction.

dmathieu:
I've taken a look at it today. https://github.com/open-telemetry/opentelemetry-go/pull/8957

om7057:
Nice, glad `OnEnding` is moving forward on that PR. Given that's still in progress upstream and unlikely to land + get released soon, do you want this PR to go ahead as a scoped `otelhttp`-only fix in the meantime, or would you rather hold off until `OnEnding` ships and revisit this as a `redact`-based processor instead? Happy to go either way, just want to make sure we're not duplicating effort once that hook lands.

dmathieu:
I wouldn't ship this. Removing a public API method is always tricky.

om7057:
Thanks for taking a look. Just to double check, this PR only adds a new `WithRedactedQueryParams` option, it doesn't remove or change any existing public API. Could you let me know which method you had in mind? It's possible you're thinking of a different PR, perhaps something related to the `OnEnding` discussion. I'd be glad to address the concern once I understand it better.

cc: @dmathieu

dmathieu:
That option would be part of the public API.

om7057:
Thanks. Would you prefer we hold off on this PR entirely until the OnEnding hook lands and revisit it as a redact-based processor at that point, or is there something else you'd suggest in the meantime?

cc: @dmathieu

dmathieu:
My opinion is the same as 2 comments above. `WithRedactedQueryParams` would be part of the public API. So I don't think we should ship this if we have a better solution in mind.

om7057:
Thanks for the clarity, I'll go ahead and close this PR. Let me know if you'd rather I keep it open for reference instead.

cc: @dmathieu

PR #9653, opened 2026-09-09
fix(otelconf): report Prometheus HTTP server errors in v0.2.0

### LINKED ISSUE #9056: otelconf: Prometheus metrics server errors are silently ignored

KR-Ravindra:
## Problem

`go.opentelemetry.io/contrib/otelconf/v0.2.0` starts an HTTP server for the Prometheus pull exporter in a goroutine. The error check on `server.Serve(lis)` is inverted: a clean shutdown (`http.ErrServerClosed`) is reported through `otel.Handle` as "the Prometheus HTTP server exited unexpectedly", while every genuine `Serve` failure (accept errors, TLS/listener problems, etc.) is silently dropped. Operators therefore get a spurious error on every shutdown and no signal at all when the metrics endpoint actually stops serving.

## Root cause

`otelconf/v0.2.0/metric.go:369`:

```go
if err := server.Serve(lis); err != nil && errors.Is(err, http.ErrServerClosed) {
```

The condition is missing the negation. The sibling implementations in `otelconf/v0.3.0/metric.go` and `otelconf/x/metric.go` already use `!errors.Is(err, http.ErrServerClosed)`.

## Fix

- Negate the predicate (`err != nil && !errors.Is(err, http.ErrServerClosed)`), so the goroutine in `prometheusReader` is now identical to the one in `v0.3.0` and `x`.
- Add a `CHANGELOG.md` entry under Unreleased / Fixed.

No public API changes.

The first revision extracted the check into a `handleServeErr` helper with a table test; it was inlined again in e769464 at review request so the three implementations stay the same.

## How tested

The handler is the process-global `otel.Handle`, so this branch has no isolated unit test (see the discussion on #9097); the fix is the same code that `v0.3.0` and `x` already run.

From `otelconf/`: `go build ./...`, `go vet ./v0.2.0/...`, `go test -race ./v0.2.0/...` (ok), `golangci-lint run ./v0.2.0/...` (0 issues).

## Links

- Fixes https://github.com/open-telemetry/opentelemetry-go-contrib/issues/9056
- Supersedes https://github.com/open-telemetry/opentelemetry-go-contrib/pull/9097 (closed by its author).

This change was prepared with an AI agent operated by KR-Ravindra, who reviewed and tested it.


dmathieu (approved):

otelconf/v0.2.0/metric.go
```diff
@@ -366,14 +366,20 @@ func prometheusReader(ctx context.Context, prometheusConfig *Prometheus) (sdkmet
...
-		}
+		handleServeErr(server.Serve(lis), otel.Handle)
 	}()
 
 	return readerWithServer{reader, &server}, nil
 }
 
+// handleServeErr reports the error returned by [http.Server.Serve] via handle
+// unless the server was closed intentionally.
+func handleServeErr(err error, handle func(error)) {
```
pellared: Can we inline this and just fix the condition?
I prefer having the same code as in https://github.com/open-telemetry/opentelemetry-go-contrib/blob/main/otelconf/v0.3.0/metric.go

KR-Ravindra: @pellared done in e769464: inlined the condition in `prometheusReader` so the goroutine is identical to `v0.3.0/metric.go`, and dropped the `handleServeErr` helper and its test. The diff is now the one-character `!` plus the changelog entry. Could you take another look?

dashpole (approved):

pellared (approved):

PR #9685, opened 2026-09-14
fix(otelhttp): preserve multipart form for request cleanup

### LINKED ISSUE #9684: [Bug]: otelhttp leaks multipart temporary files when the request is copied with WithContext
### Component

Instrumentation: otelhttp

otelhttp can prevent net/http from cleaning up temporary files created by ParseMultipartForm.

The handler replaces the incoming request with a shallow copy using Request.WithContext(). If downstream code calls ParseMultipartForm(), the resulting MultipartForm is attached to the copied request rather than the original request held by net/http.

When the request finishes, net/http checks the original request's MultipartForm to clean up temporary files. Because it remains nil, the temporary files are not removed.

This can cause temporary files to accumulate for applications using multipart uploads through otelhttp.

### Expected behavior

Temporary files created by ParseMultipartForm should be cleaned up when the request finishes, the same way they are when using net/http without otelhttp.

### Steps to Reproduce

1. Create an HTTP server using net/http and wrap its handler with otelhttp.NewHandler.
2. Send a multipart/form-data request large enough for ParseMultipartForm to create temporary files.
3. Call ParseMultipartForm() in the downstream handler.
4. Allow the request to complete.
5. Check the temporary directory for files created by the multipart parser.

With plain net/http, the temporary files are removed after the request completes.

With the current otelhttp handler, the temporary files remain because MultipartForm is attached to the request copy created by WithContext(), while net/http performs cleanup using the original request.

The behavior is reproducible using an httptest.Server with a real HTTP server lifecycle.

### Operating System

Windows 11 Home Single Language 25H2

### Device Architecture

x86_64

### Go Version

1.26.4

### Component Version

otelhttp v1.46.0, 1395f632c

fabledruns:
## Summary

This PR fixes a resource leak in `otelhttp` where temporary files created by `ParseMultipartForm` may not be cleaned up after the request completes.

**Multipart form cleanup:**
- Copies `MultipartForm` from the context-derived request back to the original request after the handler returns.
- This allows `net/http` to find the multipart form during request cleanup and remove any temporary files.

**Testing:**
- Adds a regression test using a real `httptest.Server`.
- Forces multipart parsing to create a disk-backed temporary file and verifies that it is removed after the request completes.

Fixes #9684


ps-mir (commented):
Need a changelog entry explicitly stating that cleanup is not covered in handler panic scenarios.

fabledruns:
> Need a changelog entry explicitly stating that cleanup is not covered in handler panic scenarios.

Added the requested changelog entry, including the note that cleanup is not covered when the handler panics.

ps-mir (approved):

instrumentation/net/http/otelhttp/handler.go
```diff
@@ -181,8 +181,14 @@ func (h *middleware) serveHTTP(w http.ResponseWriter, r *http.Request, next http
 		ctx = ContextWithLabeler(ctx, labeler)
 	}
 
+	origReq := r
 	r = r.WithContext(ctx)
 	next.ServeHTTP(w, r)
+	// Copy MultipartForm back to the original request so net/http can find
+	// and remove any temp files ParseMultipartForm created on the copy.
+	if r.MultipartForm != nil {
```
dashpole: If `next.ServeHTTP` panics after calling `ParseMultipartForm`, this copy-back is skipped and temp files will still leak when using HTTP/2 or an outer panic recovery middleware. Could we wrap this assignment in a `defer` before `next.ServeHTTP(w, r)` (similar to `origReq.Body` above) so panic unwinding is covered too?

CHANGELOG.md
```diff
@@ -31,6 +31,7 @@ This project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0.htm
 - Format span attributes in `go.opentelemetry.io/contrib/zpages` using `attribute.Value.String` instead of the deprecated `attribute.Value.Emit`, following the OpenTelemetry AnyValue representation for non-OTLP protocols. (#9453)
 - Format span attributes in `go.opentelemetry.io/contrib/zpages` using `attribute.Value.String` instead of the deprecated `attribute.Value.Emit`, following the OpenTelemetry AnyValue representation for non-OTLP protocols. (#9453)
 - Set `error.type` on the span and on the request-duration, request-body-size, and response-body-size metrics when a client disconnects mid-request in `go.opentelemetry.io/contrib/instrumentation/github.com/gin-gonic/gin/otelgin`, using the request context's cancellation error as the classification source when the handler has not already recorded an error via `c.Error`. Previously a disconnect was recorded with span status `Error` and no `error.type`, indistinguishable from a genuine server fault. (#9394)
+- Fix temporary file cleanup for multipart requests in `go.opentelemetry.io/contrib/instrumentation/net/http/otelhttp` when the handler returns normally. Cleanup is not covered when the handler panics. (#9685)
```
dashpole: If we switch to `defer` in `handler.go`, we can drop the panic caveat here as well:
```suggestion
- Fix temporary file cleanup for multipart requests in `go.opentelemetry.io/contrib/instrumentation/net/http/otelhttp`. (#9685)
```

instrumentation/net/http/otelhttp/handler.go
fabledruns: Fixed this by moving the `MultipartForm` copy-back into a defer so cleanup is also covered during panic unwinding

CHANGELOG.md
dashpole: Since the copy-back in `handler.go` was moved into a `defer` to cover panic unwinding, this note about panics is outdated. Could we simplify the changelog entry?
```suggestion
- Fix temporary file cleanup for multipart requests in `go.opentelemetry.io/contrib/instrumentation/net/http/otelhttp`. (#9685)
```

fabledruns:
@dashpole Could you take a look when you get a chance? Thanks

instrumentation/net/http/otelhttp/handler.go
```diff
@@ -181,8 +181,18 @@ func (h *middleware) serveHTTP(w http.ResponseWriter, r *http.Request, next http
...
+
+	// Copy MultipartForm back to the original request so net/http can find
+	// and remove any temp files ParseMultipartForm created on the copy.
+ defer func() {
+	 if r.MultipartForm != nil {
+	  origReq.MultipartForm = r.MultipartForm
+	 }
+ }()
+
+ next.ServeHTTP(w, r)
```
MrAlias: `handler.go` is not `gofmt`-clean at this head, so the repository formatting check will fail when the full CI suite runs.

```suggestion
	defer func() {
		if r.MultipartForm != nil {
			origReq.MultipartForm = r.MultipartForm
		}
	}()

	next.ServeHTTP(w, r)
```

instrumentation/net/http/otelhttp/handler_multipart_test.go
```diff
@@ -0,0 +1,84 @@
...
+
+	"github.com/stretchr/testify/require"
+)
+
+// TestHandlerMultipartTempFileCleanup verifies that a disk-backed multipart
+// upload parsed downstream is cleaned up by net/http after the request
+// completes. The middleware passes a context-derived request copy downstream,
+// so the parsed MultipartForm must be copied back onto the original request
+// for net/http's finishRequest to find and remove its temp files.
+func TestHandlerMultipartTempFileCleanup(t *testing.T) {
```
MrAlias, line 60: The defer is what preserves cleanup under HTTP/2 panic handling or an outer recovery middleware, but this test only exercises normal return. Moving the copy-back after `next.ServeHTTP` would leave this test green and reintroduce the panic leak. Could we add a panic-path case that verifies the temp file is removed?

CHANGELOG.md
MrAlias: The new defer copies the form back during panic unwinding. `net/http` cleans it up on HTTP/2 panics, and an outer recovery middleware lets the HTTP/1 cleanup path run, so this release note now describes covered cases as unsupported. Could we narrow the caveat to unrecovered HTTP/1 handler panics?

MrAlias (changes requested):
The copy-back is on the right lifecycle boundary, but the panic behavior is not yet consistently tested or documented, and the current head is not format-clean. The branch also conflicts with `main`, so only the CLA and code-owner routing checks have run.

MrAlias (approved):

fabledruns:
Fixed the lint issues reported by CI in `handler_multipart_test.go`.

pellared:
@sonalgaud12, @dmathieu, PTAL as code owners

instrumentation/net/http/otelhttp/handler.go
```diff
@@ -181,7 +181,18 @@ func (h *middleware) serveHTTP(w http.ResponseWriter, r *http.Request, next http
 		ctx = ContextWithLabeler(ctx, labeler)
 	}
 
+	origReq := r
 	r = r.WithContext(ctx)
```
dmathieu: Nit, but I think it would be clearer to keep `r` as the origin request, and create a new `rCtx` variable for the new request with the context.

Similar to what's already done in otelmux.
https://github.com/open-telemetry/opentelemetry-go-contrib/pull/9361/changes

fabledruns: Done, kept `r` as original request and added `rCtx`

dmathieu (approved):

instrumentation/net/http/otelhttp/handler.go
```diff
@@ -181,8 +181,19 @@ func (h *middleware) serveHTTP(w http.ResponseWriter, r *http.Request, next http
 		ctx = ContextWithLabeler(ctx, labeler)
 	}
 
-	r = r.WithContext(ctx)
-	next.ServeHTTP(w, r)
+	origReq := r
```
dmathieu: Sorry for the wrong approval. The whole point of my previous review was to remove this variable definition.

fabledruns: Fixed it. I’m very sorry for wasting both of our time on this. Thanks for bearing with me

dmathieu (approved):

sonalgaud12 (approved):

fabledruns:
Fixed the formatting issue reported by CI

instrumentation/net/http/otelhttp/handler.go
ps-mir: I think **`http.route`** will be dropped as its still reading the original request via `Req` and `metricAttributesFromRequest` also gets original request

These were not caught likely due to missing coverage. As a verification you can add test cases to confirm the failure first.
* routes GET /foo/123 through a real ServeMux and asserts all three server metrics carry http.route="/foo/{id}".
*  asserts the WithMetricAttributesFn callback sees the routed request, meaning its Pattern, PathValue("id"), and span context are all set.

ps-mir (commented):
I think ctx needs to be used further to pass `http.route`.

PR #9691, opened 2026-09-15
logutil: bound recursive log value conversion

pellared:
Towards #9044.

Scope note: this change bounds only converter-owned recursive traversal. It still permits a fatal cycle through the existing `fmt` boundary; for example, when a cycle crosses a struct or a non-string map key. This residual risk is documented in the [trust boundary and residual risks section of `DESIGN.md`](https://github.com/open-telemetry/opentelemetry-go-contrib/blob/7ca370847126bcea59c3cb9e6dcb9095796e131a/internal/shared/logutil/DESIGN.md#trust-boundary-and-residual-risks).

Simpler alternative to #9542.

## Summary

Bound converter-owned recursive traversal in `otellogr`, `otellogrus`, `otelslog`, and `otelzap` to 100 non-empty map/slice/array and non-nil pointer levels. Non-nil interface wrappers are depth-transparent: unwrapping one does not consume a level, while its concrete value is counted according to its own kind. If conversion would enter counted level 101, it immediately unwinds and replaces that whole source field with `"<max-depth-exceeded>"`. Other fields and the log record are still emitted.

The conversion is atomic per source field: it does not retain a partial tree or continue visiting siblings after failure. This makes a branching cycle reach the bound and unwind instead of expanding every branch to the bound.

The implementation deliberately has no identity table, custom hash, total-work accounting, snapshots, `sync.Pool`, global state, public option, or dependency change. The detailed contract and rejected alternatives are in [`internal/shared/logutil/DESIGN.md`](https://github.com/open-telemetry/opentelemetry-go-contrib/blob/7ca370847126bcea59c3cb9e6dcb9095796e131a/internal/shared/logutil/DESIGN.md).

## Rationale

- **Correctness:** a `(value, ok)` helper propagates the first depth failure to the field root, so no partial aggregate escapes.
- **Go precedent:** [`log/slog.Value.Resolve`](https://go.dev/src/log/slog/value.go) uses a fixed 100-call safeguard, and [`encoding/json`](https://go.dev/src/encoding/json/encode.go) rejects a whole cyclic value rather than inserting edge-local markers.
- **OpenTelemetry:** this private bridge guard is intentionally separate from the configurable SDK [`AttributeValueDepthLimit`](https://opentelemetry.io/docs/specs/otel/common/#attribute-limits) implemented by [open-telemetry/opentelemetry-go#8938](https://github.com/open-telemetry/opentelemetry-go/pull/8938). The bridge must first produce a finite `attribute.Value`, so a later SDK policy cannot prevent this conversion-time recursion.
- **Performance:** the existing exact-type switch remains first. The first reflective level also stays in `convertValue`; only descendants enter the status-returning helper. This preserves the ordinary root path and introduces no visit-tracking allocation.

## Compatibility and scope

There is no public API change. Acyclic values requiring at most 100 counted levels retain their existing conversion. Values requiring another recursive descent now become the diagnostic string instead of continuing without a bound.

This narrowly bounds reflection recursion owned by the converter. Existing `fmt` calls for structs, non-string map keys, and unhandled values remain unchanged. `fmt.Formatter`, `fmt.Stringer`, `error.Error`, their side effects or pathological recursion, concurrent mutation, huge-width inputs, and compact acyclic graph amplification remain explicit trust boundaries/out of scope.

## Performance

- Base production commit: `ab9739e67112e66015ee07f5e7ddae9bdaff46d5`
- Head commit: `7ca370847126bcea59c3cb9e6dcb9095796e131a`
- Benchmarked implementation commit: `8eda7adce5d30c1dc694f9b09338a8e7427a6453`
- Go 1.27.0, Linux/amd64 (`6.17.0-1017-aws`), AMD EPYC 7571
- `GOMAXPROCS=1`, `-cpu=1`, pinned to logical CPU 2 with `taskset`
- 10 alternating base/head samples; each test binary built once from an identical benchmark file; `-benchtime=300ms -benchmem`

Raw `benchstat` output:

```text
goos: linux
goarch: amd64
pkg: go.opentelemetry.io/contrib/bridges/otellogr
cpu: AMD EPYC 7571
                             │ base-r2.txt │             head-r2.txt              │
                             │   sec/op    │    sec/op     vs base                │
ConvertValue/Bool              20.05n ± 0%    20.07n ± 0%   +0.10% (p=0.007 n=10)
ConvertValue/String            21.23n ± 1%    21.20n ± 0%   -0.16% (p=0.000 n=10)
ConvertValue/Int64             19.99n ± 0%    20.00n ± 0%        ~ (p=0.324 n=10)
ConvertValue/Nil               19.98n ± 0%    19.99n ± 0%        ~ (p=0.293 n=10)
ConvertValue/AttributeValue    20.38n ± 0%    20.38n ± 0%        ~ (p=0.910 n=10)
ConvertValue/PointerScalar     91.86n ± 0%   104.00n ± 1%  +13.22% (p=0.000 n=10)
ConvertValue/EmptySlice        39.64n ± 6%    35.81n ± 0%   -9.67% (p=0.000 n=10)
ConvertValue/SliceInt3         543.6n ± 3%    543.6n ± 3%        ~ (p=0.684 n=10)
ConvertValue/SliceAny3         464.4n ± 5%    507.3n ± 2%   +9.24% (p=0.000 n=10)
ConvertValue/MapOneEntry       576.2n ± 2%    587.9n ± 1%   +2.03% (p=0.007 n=10)
ConvertValue/MapThreeEntries   1.403µ ± 1%    1.416µ ± 3%   +0.93% (p=0.000 n=10)
ConvertValue/StructScalar      409.2n ± 1%    418.2n ± 1%   +2.20% (p=0.000 n=10)
ConvertValue/StructMap         1.145µ ± 1%    1.187µ ± 5%   +3.67% (p=0.027 n=10)
ConvertKVs                     364.9n ± 4%    364.5n ± 3%        ~ (p=0.796 n=10)
geomean                        131.0n         132.9n        +1.42%

                             │ base-r2.txt  │             head-r2.txt             │
                             │     B/op     │    B/op     vs base                 │
ConvertValue/Bool              0.000 ± 0%     0.000 ± 0%       ~ (p=1.000 n=10) ¹
ConvertValue/String            0.000 ± 0%     0.000 ± 0%       ~ (p=1.000 n=10) ¹
ConvertValue/Int64             0.000 ± 0%     0.000 ± 0%       ~ (p=1.000 n=10) ¹
ConvertValue/Nil               0.000 ± 0%     0.000 ± 0%       ~ (p=1.000 n=10) ¹
ConvertValue/AttributeValue    0.000 ± 0%     0.000 ± 0%       ~ (p=1.000 n=10) ¹
ConvertValue/PointerScalar     8.000 ± 0%     8.000 ± 0%       ~ (p=1.000 n=10) ¹
ConvertValue/EmptySlice        0.000 ± 0%     0.000 ± 0%       ~ (p=1.000 n=10) ¹
ConvertValue/SliceInt3         312.0 ± 0%     312.0 ± 0%       ~ (p=1.000 n=10) ¹
ConvertValue/SliceAny3         288.0 ± 0%     288.0 ± 0%       ~ (p=1.000 n=10) ¹
ConvertValue/MapOneEntry       176.0 ± 0%     176.0 ± 0%       ~ (p=1.000 n=10) ¹
ConvertValue/MapThreeEntries   536.0 ± 0%     536.0 ± 0%       ~ (p=1.000 n=10) ¹
ConvertValue/StructScalar      16.00 ± 0%     16.00 ± 0%       ~ (p=1.000 n=10) ¹
ConvertValue/StructMap         96.00 ± 0%     96.00 ± 0%       ~ (p=1.000 n=10) ¹
ConvertKVs                     256.0 ± 0%     256.0 ± 0%       ~ (p=1.000 n=10) ¹
geomean                                   ²               +0.00%                ²
¹ all samples are equal
² summaries must be >0 to compute geomean

                             │ base-r2.txt  │             head-r2.txt             │
                             │  allocs/op   │ allocs/op   vs base                 │
ConvertValue/Bool              0.000 ± 0%     0.000 ± 0%       ~ (p=1.000 n=10) ¹
ConvertValue/String            0.000 ± 0%     0.000 ± 0%       ~ (p=1.000 n=10) ¹
ConvertValue/Int64             0.000 ± 0%     0.000 ± 0%       ~ (p=1.000 n=10) ¹
ConvertValue/Nil               0.000 ± 0%     0.000 ± 0%       ~ (p=1.000 n=10) ¹
ConvertValue/AttributeValue    0.000 ± 0%     0.000 ± 0%       ~ (p=1.000 n=10) ¹
ConvertValue/PointerScalar     1.000 ± 0%     1.000 ± 0%       ~ (p=1.000 n=10) ¹
ConvertValue/EmptySlice        0.000 ± 0%     0.000 ± 0%       ~ (p=1.000 n=10) ¹
ConvertValue/SliceInt3         5.000 ± 0%     5.000 ± 0%       ~ (p=1.000 n=10) ¹
ConvertValue/SliceAny3         2.000 ± 0%     2.000 ± 0%       ~ (p=1.000 n=10) ¹
ConvertValue/MapOneEntry       5.000 ± 0%     5.000 ± 0%       ~ (p=1.000 n=10) ¹
ConvertValue/MapThreeEntries   9.000 ± 0%     9.000 ± 0%       ~ (p=1.000 n=10) ¹
ConvertValue/StructScalar      1.000 ± 0%     1.000 ± 0%       ~ (p=1.000 n=10) ¹
ConvertValue/StructMap         4.000 ± 0%     4.000 ± 0%       ~ (p=1.000 n=10) ¹
ConvertKVs                     1.000 ± 0%     1.000 ± 0%       ~ (p=1.000 n=10) ¹
geomean                                   ²               +0.00%                ²
¹ all samples are equal
² summaries must be >0 to compute geomean
```

The all-case timing geomean is 131.0 ns → 132.9 ns (**+1.42%**). For the guarded shallow pointer/slice/map group it is 301.0 ns → 308.1 ns (**+2.36%**). Bytes and allocations are unchanged in every case.

The exact-value paths remain zero-allocation and within 3%, and the guarded shallow composite group remains below the 5% geomean threshold. The largest per-case change is `PointerScalar` at +13.22%; this isolated result is reported as measured rather than adding a special-case fast path to this narrowly scoped fix.

For code-size context, the built `otellogr` package archive changes from 384,742 to 423,814 bytes (+10.2%). This is the cost of keeping the direct root path and a separate bounded descendant path; primitives bypass the latter, and a failure unwinds with only depth/status state rather than identity or work tracking.


pellared:
@khushijain21, @dmathieu, @MrAlias, can you please review this PR given it is a simplified version of the one you have already reviewed?

dmathieu (approved):

khushijain21 (approved):

MrAlias (approved):

PR #9699, opened 2026-09-17
fix(ci): CODEOWNERS path matching in ping_codeowners_issues

### LINKED ISSUE #9050: Owner ping workflow has non-security path matching gaps
### Description

The CODEOWNERS ping workflow can miss owners for labels whose path matching differs from the script assumptions.

Current-tree check: relevant paths are present in this checkout: tools/ping_codeowners_issues.sh, .github/workflows/ping-codeowners-issues.yml, CODEOWNERS.

Validation note: Validated the finding as a functional correctness bug in the new issue-label automation. The commit c17e09fe67f429ee048e6c9cf46e3c32e0758e0b adds .github/workflows/ping-codeowners-issues.yml and tools/ping_codeowners_issues.sh. The workflow passes the GitHub issue label name directly as COMPONENT at .github/workflows/ping-codeowners-issues.yml:17-22. The script strips any prefix before ':' into COMPONENT_NAME at tools/ping_codeowners_issues.sh:20, then resolves CODEOWNERS with grep -F "/${COMPONENT_NAME}/" at tools/ping_codeowners_issues.sh:21, and silently exits when COMPONENT_PATH is empty at lines 23-24. This exact slash-surrounded substring requirement skips valid CODEOWNERS entries such as CODEOWNERS:42 'exporters/autoexport' and CODEOWNERS:75 'zpages/'.

### Steps To Reproduce

[16-27679bee01d88191ae0aa4e339e3e3d2-owner-ping-workflow-has-non-security-path-matching-gaps.zip](https://github.com/user-attachments/files/28385119/16-27679bee01d88191ae0aa4e339e3e3d2-owner-ping-workflow-has-non-security-path-matching-gaps.zip)

1. Run the PoC script from the artifact zip against the repository CODEOWNERS data.
2. Use labels that map to paths covered by CODEOWNERS entries with patterns the script does not resolve correctly.
3. Observe that legitimate owners are not selected for ping comments.

### Expected behavior

The workflow should resolve owners for valid labels consistently with CODEOWNERS semantics and ping the relevant maintainers.

sonalgaud12:
Fixes #9050

grep -F "/${COMPONENT_NAME}/" requires slashes on both sides of the component name, so it misses last-segment paths like exporters/autoexport and single-segment paths like zpages/. When no match is found the script silently exits without pinging an owner.


dmathieu:
Have you tested this against multiple components? Could you maybe run the check against all the ones in CODEOWNERS and paste the validation here?

tools/ping_codeowners_issues.sh
```diff
@@ -18,7 +18,17 @@ CUR_DIRECTORY=$(dirname "$0")
 # Extract the component name (the part after ': ') and resolve the full path
 # from CODEOWNERS, since get-codeowners.sh matches by path prefix.
 COMPONENT_NAME=$(echo "${COMPONENT}" | sed 's/^[^:]*:[[:space:]]*//')
-COMPONENT_PATH=$(grep -F "/${COMPONENT_NAME}/" "${CUR_DIRECTORY}/../CODEOWNERS" | awk '{ print $1 }' | sed 's|/$||' | head -1 || true)
+COMPONENT_PATH=$(awk -v component_name="${COMPONENT_NAME}" '
```
MrAlias, line 21: The issue-labeler adds component labels with `${{ github.token }}`, but this workflow only runs on a later `issues:labeled` event. [GitHub suppresses workflow runs for events created with `GITHUB_TOKEN`](https://docs.github.com/en/actions/concepts/security/github_token#when-github_token-triggers-workflow-runs). Issue #9684 demonstrates the result: its user-applied `bug` label started run 34806763124 with `COMPONENT: bug`, while the bot-applied `instrumentation: otelhttp` label produced no owner-ping run. The automatic path therefore never reaches this matcher.

Could we run the owner lookup in the labeler job, or explicitly call or dispatch the ping workflow after adding the labels?

tools/ping_codeowners_issues.sh
```diff
@@ -18,7 +18,17 @@ CUR_DIRECTORY=$(dirname "$0")
...
+    $1 !~ /^#/ {
+        path = $1
+        sub(/\/$/, "", path)
+        component_count = split(path, components, "/")
+        if (path == component_name || components[component_count] == component_name) {
```
MrAlias, line 26: `COMPONENT_NAME` still cannot be compared directly to either the whole path or its last segment for several labels emitted by the issue-labeler workflow. For example, `detector: aws:ec2` becomes `aws:ec2`, while CODEOWNERS contains `detectors/aws/ec2/v2`; this condition fails, `COMPONENT_PATH` remains empty, and the script exits without pinging `@iblancasa`. The same failure is reachable for the generated ECS/EKS/Lambda detector labels and `sampler: probability:consistent`. The supported `AWS X-Ray` form values fail one step earlier: they sanitize to `aws:x-ray`, while the existing labels are `aws:xray`, so the labeler rejects them before this matcher runs.

Could we make the component label-to-path mapping exhaustive and add a check that every supported form value produces an existing label and resolves to a CODEOWNERS entry? The linked issue suggests renaming tags or using an explicit mapping table.

MrAlias (changes requested):
The owner-ping path is still silent for supported issue components because the automatic label flow never reaches this matcher and the label vocabulary remains only partially mapped.

tools/ping_codeowners_issues.sh
dashpole: Since line 20 strips any prefix before a colon, matching `zpages` here also causes `area: zpages` (which is applied alongside `zpages`) to resolve to `zpages` and ping the code owner twice. Also worth noting if you switch to an explicit mapping: repo labels like `bridge: slog`, `bridge: zap`, `bridge: logrus`, `bridge: logr`, and `processor: baggage` don't match their `bridges/otelslog` (etc.) and `processors/baggagecopy` directory names in `CODEOWNERS`.

sonalgaud12: Thanks for catching both of these.

Added a CI test that checks that every form value resolves, every CODEOWNERS path is either mapped or explicitly allowlisted, and the allowlist itself can't go stale. Ran it against the full CODEOWNERS file: 48 paths, 26 covered, 22 explicitly excluded (no form option exists for them yet; happy to file a follow-up for those).

tools/ping_codeowners_issues_test.sh
```diff
@@ -0,0 +1,146 @@
...
+    'propagator: Autoprop|propagators/autoprop'
+    'propagator: AWS X-Ray|propagators/aws'
+    'propagator: B3|propagators/b3'
+    'propagator: Jaeger|propagators/jaeger'
+    'propagator: OpenCensus|propagators/opencensus'
+    'propagator: OT|propagators/ot'
+    'sampler: JaegerRemote|samplers/jaegerremote'
+    'sampler: probability:consistent|samplers/probability/consistent'
+    'exporter: exporters/autoexport|exporters/autoexport'
+    'zpages: zpages|zpages'
```
MrAlias, line 38: These cases validate only namespace-to-CODEOWNERS resolution, while the workflow also depends on every emitted repository label existing. It emits `area: sampler`, `exporter: exporters/autoexport`, and `zpages: zpages`, none of which exist. The current labels are `sampler: jaegerremote`, `sampler: probability:consistent`, `exporter: autoexport`, and `zpages`, with no `area: sampler`. For either Sampler choice, the missing area label makes the combined existence check fail even though its namespace exists; AutoExporter and zPages fail on their namespace. In every case no component output is set, so the owner-ping step is skipped while this test stays green.

Could we align every emitted label with the existing repository labels, or provision the intended replacements, and validate this invariant in CI?

tools/ping_codeowners_issues_test.sh
```diff
@@ -0,0 +1,146 @@
...
+if [[ -n "${stale_exclusions}" ]]; then
+    echo "Non-issue component exclusions are missing from CODEOWNERS:" >&2
+    printf '%s\n' "${stale_exclusions}" >&2
+    exit 1
+fi
+
+normalize_component_label() {
+    case "$1" in
+        AutoExporter) echo "exporter: exporters/autoexport" ;;
+        Config) return 1 ;;
```
MrAlias, line 111: `Config` is a supported value in both issue forms, and CODEOWNERS assigns `otelconf/`, but this branch explicitly skips it while `non_issue_components` classifies `otelconf` as not selectable. In the workflow, Config produces only `area: file-config`; no component output is set, so the direct ping step never runs, and the resolver has no file-config mapping. Every Config issue therefore still misses its owners even though this test passes.

Could we include Config in the mapping and assert that it resolves to `otelconf` instead of excluding it?

MrAlias (commented):
The follow-up adds the direct lookup and mapping coverage, but several supported form selections still fail the label-to-owner path on the current repository state.

PR #9708, opened 2026-09-18
otelhttp: record error.type on server span

### LINKED ISSUE #9358: otelhttp: client-initiated request termination is recorded as a 500 with span status Error

dmathieu:
Fixes https://github.com/open-telemetry/opentelemetry-go-contrib/issues/9358


dmathieu:
This is a simpler solution to https://github.com/open-telemetry/opentelemetry-go-contrib/pull/9404, and closer to what we already do with otelrestful (#9387).

instrumentation/net/http/otelhttp/handler.go
```diff
@@ -190,8 +193,18 @@ func (h *middleware) serveHTTP(w http.ResponseWriter, r *http.Request, next http
 
 	statusCode := rww.StatusCode()
 	bytesWritten := rww.BytesWritten()
-	span.SetStatus(h.semconv.Status(statusCode))
 	bytesRead := bw.BytesRead()
+
+	if err := ctx.Err(); err != nil {
```
pellared, line 198: The new classification checks only `ctx.Err()`, even though `rww.Error()` and non-EOF `bw.Error()` are captured below. A failed response write or request-body read with status 200 therefore remains `Unset` with no `error.type`. Please select the cause in response-write → non-EOF request-read → context order, force `codes.Error`, and reuse the selected `error.type` for spans and metrics.

instrumentation/net/http/otelhttp/handler.go
```diff
@@ -190,8 +193,18 @@ func (h *middleware) serveHTTP(w http.ResponseWriter, r *http.Request, next http
...
+		span.SetStatus(codes.Error, err.Error())
+		span.SetAttributes(otelsemconv.ErrorType(err))
```
pellared, line 200: The selected `error.type` is attached only to the span. `RecordMetrics` receives no corresponding attribute, so failed request-duration and body-size measurements omit the conditionally required `error.type`. Please append the computed attribute to `MetricAttributes.AdditionalAttributes` and add metric assertions for cancellation, 5xx, read, and write failures.

pellared (commented):
Minor edge case: statuses 600–999 are classified as errors by `HTTPServer.Status`, but the `error.type` fallback only covers 500–599.

dmathieu:
@pellared with your reviews, let's close this PR and go back to #9404. What you're asking for is what that PR does.

pellared:
> @pellared with your reviews, let's close this PR and go back to #9404. What you're asking for is what that PR does.

#9404 got stale and has comment that are not addressed.

PR #9717, opened 2026-09-21
fix(otelconf): honor standalone without_type_suffix/without_units Prometheus options

### LINKED ISSUE #9051: otelconf: Prometheus config ignores single without_* options
### Description

Prometheus configuration translation ignores standalone `without_*` options when no other options are set.

### Steps To Reproduce

[18-536fb72f9f2481918e2215e550bfd747-prometheus-config-ignores-single-without-options.zip](https://github.com/user-attachments/files/28385257/18-536fb72f9f2481918e2215e550bfd747-prometheus-config-ignores-single-without-options.zip)

1. Create a Prometheus config using only a `without_*` option covered by the PoC.
2. Run the translation PoC from the artifact zip.
3. Observe that the translated configuration omits the intended option.

### Expected behavior

A standalone `without_*` option should be translated and applied even when no other Prometheus options are present.

om7057:
## Summary

- `prometheusReaderOpts` in `go.opentelemetry.io/contrib/otelconf/v0.3.0` only applied a translation option when both `without_type_suffix` and `without_units` were set together, so a standalone occurrence of either was silently ignored.
- When only one of the two flags is set, the fix now also applies the corresponding `otelprom.WithoutCounterSuffixes()`/`otelprom.WithoutUnits()` option alongside the default translation strategy, so the flag takes effect on its own.

Fixes #9051

## Test plan

- Added table-test cases in `TestPrometheusReaderOpts` covering a standalone `without_type_suffix` and `without_units`.
- Added `TestPrometheusReaderOptsStandaloneWithoutOptions`, which builds an actual Prometheus exporter with the returned options, records a counter with a unit, and asserts the produced metric name has the expected suffix dropped.
- `go test ./v0.3.0/...` and `go vet ./v0.3.0/...` pass.


om7057:
cc: @pellared @dashpole please go through it once you get the time to!

dmathieu:
Could you fix the lint issues?

om7057:
Fixed, the new test now uses t.Context() instead of context.Background() as flagged by usetesting.

cc: @dmathieu

otelconf/v0.3.0/metric.go
```diff
@@ -386,14 +386,26 @@ func prometheusReaderOpts(prometheusConfig *Prometheus) ([]otelprom.Option, erro
 		opts = append(opts, otelprom.WithoutScopeInfo())
 	}
 
-	if prometheusConfig.WithoutTypeSuffix != nil && *prometheusConfig.WithoutTypeSuffix && prometheusConfig.WithoutUnits != nil && *prometheusConfig.WithoutUnits {
+	withoutTypeSuffix := prometheusConfig.WithoutTypeSuffix != nil && *prometheusConfig.WithoutTypeSuffix
+	withoutUnits := prometheusConfig.WithoutUnits != nil && *prometheusConfig.WithoutUnits
+
+	if withoutTypeSuffix && withoutUnits {
```
dashpole: It looks like `otelconf/v0.2.0/metric.go` has the exact same regression from #8595 where standalone `without_type_suffix` and `without_units` options are ignored. Could we apply this fix and test coverage to `otelconf/v0.2.0` as well?

